PR #2, opened 2015-03-20
dummy test

tansaku:
just testing


spec/feature/grand_finale_spec.rb
```diff
@@ -10,6 +10,9 @@
 
 feature 'Grand Finale' do
 
-  xscenario 'all planes can land and all planes can take off'
+  scenario 'all planes can land and all planes can take off' do
```
tansaku: No I don't like double quoted!

PR #3, opened 2015-03-20
Pull request

sanjsanj:


lib/airport.rb
```diff
@@ -1,2 +1,44 @@
+require_relative 'plane'
+
 class Airport
+  def initialize
+    @capacity = 6
+  end
+
+  def locate_plane plane
```
stephenlloyd: can you make the name of this method express whats its doing, def location_of plane

lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+
+  def landed_planes
+    @landed_planes ||= []
+  end
+
+  def airborne_planes
+    @airborne_planes ||= []
+  end
+
+  def weather set_weather = %w(sunny stormy).sample
```
stephenlloyd: this is very imagintive. I would say theres too much logic in the defaulted value.  essentailly you want this to either return whats its entered or a random [sunny, stormy]. It took me a couple of reads to see this when you could easily express this in the method body. I would encourage you to simplify this as it's a huge code smell.

lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+    set_weather
+  end
+
+  def weather_good
+    weather == 'sunny'
+  end
+
+  def can_land
```
stephenlloyd: this is a boolean method so should be can_land?

lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+
+  def airborne_planes
+    @airborne_planes ||= []
+  end
+
+  def weather set_weather = %w(sunny stormy).sample
+    set_weather
+  end
+
+  def weather_good
```
stephenlloyd: this is a boolean method so weather_good?

lib/plane.rb
```diff
@@ -1,2 +1,28 @@
...
+  attr_reader :status, :location
+  def initialize
+    @status = 'flying'
+  end
+
+  def request_land airport
+    airport.landing_permission self
+  end
+
+  def land(*)
```
stephenlloyd: this splat is hugely worrying, is it even needed?

lib/plane.rb
```diff
@@ -1,2 +1,28 @@
...
+    fail 'already landed' if @status == 'landed'
+    @status = 'landed'
+  end
+
+  def request_takeoff airport
+    airport.order_plane_takeoff
+  end
+
+  def takeoff
+    fail 'already flying' if @status == 'flying'
```
stephenlloyd: I would change @status == 'flying' to flying?

spec/airport_spec.rb
```diff
@@ -1,42 +1,50 @@
...
 
-  context 'taking off and landing' do
+  it 'knows when a plane is in the airport' do
+    plane = double :plane, location: 'airport'
+    expect(subject.locate_plane plane).to eq('airport')
+  end
 
-    xit 'a plane can land'
+  context 'can order a plane to take off' do
+    it { is_expected.to respond_to :order_plane_takeoff }
```
stephenlloyd: I would always have the longer format of testing and never this it {..} format. it "descriptive name" do .. end is far far better

spec/airport_spec.rb
```diff
@@ -1,42 +1,50 @@
...
-
 describe Airport do
+  it 'knows when a plane is in the air' do
+    plane = double :plane, location: 'air'
+    expect(subject.locate_plane plane).to eq('air')
+  end
 
-  context 'taking off and landing' do
+  it 'knows when a plane is in the airport' do
+    plane = double :plane, location: 'airport'
```
stephenlloyd: great use of doubles, I would put this as a let(:plane){double :plane ...} at the top though as you're using it a few times.

spec/airport_spec.rb
```diff
@@ -1,42 +1,50 @@
...
+  end
 
-    xit 'a plane can take off'
+  context 'can respond to a plane wanting to land' do
+    it { is_expected.to respond_to :landing_permission }
   end
 
   context 'traffic control' do
+    it 'a plane cannot land if the airport is full' do
+      Airport.any_instance.stub(:can_land).and_return(false)
```
stephenlloyd: this is old syntax. if airport was a let at the top like so let(:airport){Airport.new} then you can simple say allow(airport).to receive(:can_land).and_return true
this is the most recent syntax. But otherwise really good :)

spec/plane_spec.rb
```diff
@@ -1,25 +1,48 @@
...
-
-  xit 'has a flying status when created'
-
-  xit 'has a flying status when in the air'
-
-  xit 'can take off'
-
-  xit 'changes its status to flying after taking off'
-
+  plane = Plane.new
```
stephenlloyd: this is a massibve code smell. Make this a let so that its a different plane for each test. this will force you to write better tests. Right now they only work in sequence which will only cause complications later on. You always always always want every test to be able to run independently.

spec/plane_spec.rb
```diff
@@ -1,25 +1,48 @@
...
+    expect(plane.location).to eq 'air'
+  end
+
+  it 'can \'request to land\' at airport' do
+    airport = double :airport, landing_permission: true
+    expect(plane.request_land airport).to eq true
+  end
+
+  it 'responds to \'land\'' do
+    expect(plane).to respond_to :land
```
stephenlloyd: all this tests is that there is a method called land. Whilst this is ok you can get too wins by testing the return value or expected change. Passing this test give you no functionality.

spec/plane_spec.rb
```diff
@@ -1,25 +1,48 @@
...
+    plane.land :airport
+    expect(plane.status).to eq 'landed'
+  end
+
+  it 'does not respond to \'land\' after landing' do
+    expect { (plane.land :airport) }.to raise_error 'already landed'
+  end
+
+  it 'responds to \'takeoff\'' do
+    expect(plane).to respond_to :takeoff
```
stephenlloyd: again all of these respond_to tests are a waste. They're easy to pass and test nothing. You're better of testing that this method is actually doing.

stephenlloyd:
Hey, Sorry to be quite harsh. Despite what it sounds like this is really good. You just need to neaten it up a bit and it would be perfect. I'm impressed with some of your code but just try to make sure that everything is consitant and has purpose. Really good :checkered_flag:

PR #5, opened 2015-03-21
Airport challenge - Pull request

armi1189:


lib/airport.rb
```diff
@@ -1,2 +1,36 @@
...
+
 class Airport
+  attr_reader :control_tower, :planes, :available_room, :capacity
+
+  DEFAULT_CAPACITY = 30
+  def initialize(options = {})
+    @control_tower = ControlTower.new
+    @planes = []
+    @capacity = options.fetch(:capacity, DEFAULT_CAPACITY)
+    @available_room = @capacity
```
stephenlloyd: is this needed. Couldn't you just calculate this when needed from the capacity and plane count.

lib/airport.rb
```diff
@@ -1,2 +1,36 @@
+require_relative 'control_tower'
+
 class Airport
+  attr_reader :control_tower, :planes, :available_room, :capacity
+
+  DEFAULT_CAPACITY = 30
+  def initialize(options = {})
+    @control_tower = ControlTower.new
```
stephenlloyd: this is breaking dependency inversion principle

lib/control_tower.rb
```diff
@@ -0,0 +1,35 @@
+class ControlTower
+  def land_request plane, airport
```
stephenlloyd: this would have been so cool if it were a module.

lib/control_tower.rb
```diff
@@ -0,0 +1,35 @@
...
+    fail 'CONTROL TOWER: Impossible to land because of a storm'
+  end
+
+  def error_take_off_stormy
+    `say -v Kathy -r 190 "Impossible to take off because of a storm"`
+    fail 'CONTROL TOWER: Impossible to take off because of a storm'
+  end
+
+  def stormy?
+    [true, false, false, false, false].sample
```
stephenlloyd: is there an easier way to do this? maybe 
```ruby
rand(5) == 1
```

lib/plane.rb
```diff
@@ -1,2 +1,20 @@
 class Plane
+  attr_reader :flying
+  alias_method :flying?, :flying
+
+  def initialize
+    @flying = false
+  end
+
+  def take_off_from airport
+    puts 'PILOT: Airplane to Control Tower, ready for take off'
```
stephenlloyd: try not to have puts in methods. In real life they'll just be putting out to a server somewhere for noone to see.

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
...
+  end
+
+  def land_to airport
+    puts 'PILOT: Airplane to Control Tower, requests for landing'
+    `say -v Alex -r 200 "Airplane to Control Tower, requests for landing"`
+    airport.control_tower.land_request self, airport
+    @flying = false
+  end
+
+  def travel from, to
```
stephenlloyd: this is cool. You could play with the ruby syntax a little and do trave 
```ruby
def from: airport, to: airport 
    take_off_from from
    and_to to
end
```

spec/control_tower_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'control_tower'
+
+describe ControlTower do
+  before do
+    allow($stdout).to receive(:puts)
+    allow(subject).to receive(:`)
```
stephenlloyd: I'm impressed you found this. using the tick ` command is quite dangerous and normally frowned upon as is puts but at least you stubbed it out correctly :)

spec/control_tower_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+  context 'basic operations' do
+    before do
+      allow(subject).to receive(:stormy?).and_return(false)
```
stephenlloyd: nice

stephenlloyd:
this is really good, Probably one of the best I've seen :) take my notes on bored but you definitely get a sticker and i'm really impressed :) 
You should definitely try to make the control tower a module which you could include in your airport.

stephenlloyd:
:checkered_flag:

PR #6, opened 2015-03-22
As much as I could do

SammyAbukmeil:
I wasn't able to complete the full challenge, I'm still trying to wrap my head around BDD.


SammyAbukmeil:
Also, RuboCop says it prefers single quotes whereas hound prefers double. I decided to leave single quotes.

SammyAbukmeil:
The run-through you did today was really beneficial, I'm going to run through that now to make sure I understand how to approach this challenge.

spec/airport_spec.rb
```diff
@@ -11,32 +11,9 @@
 #
 # If the airport is full then no planes can land
 
-describe Airport do
-
-  context 'taking off and landing' do
-
-    xit 'a plane can land'
```
stephenlloyd: This doesn't have a test or code for any of this. Ideally you'd want an array of planes in the airport and you should be able to land and take of planes into this array providing the weather is good. If you go through the video from yesterday you should have a better idea of how to do this.

lib/plane.rb
```diff
@@ -1,2 +1,13 @@
 class Plane
+  def initialize
+    @landed = true
+  end
+
+  def landed?
+    @landed = true
```
stephenlloyd: this is a boolean method which either returns true or false. As it stands right now it will only ever return true.  You should have a land method which changes this method to return false. Draw out what you want the plane to do and then code it.

lib/airport.rb
```diff
@@ -1,2 +1,5 @@
 class Airport
+  def ok_to_land
+    true
```
stephenlloyd: with all tests you should have the easy path which you can just slime to pass like you've done here and then do something to change the state. Can it ever not be ok to land, if so what state do you need to do to make this happen. This is kind of a pointless method and it's the only one.

stephenlloyd:
This test isn't complete. Go through the video from yesterday and do this again, I would expect far more tests and even the ability not to land the same plane twice and take of a plane that doesnt exist. 
Let me know if you have any trouble with this :construction:

PR #8, opened 2015-03-22
Airport Challenge Pull

GJMcGowan:
Like everyone else I'm leaving single quotes as opposed to doubles so that the rubocop all passes without being edited.


lib/airport.rb
```diff
@@ -1,2 +1,31 @@
...
+  def initialize
+    @capacity = 20
+    @planes = []
+  end
+
+  def land_plane(plane)
+    fail 'Airport Full' if @planes.count >= @capacity
+    fail 'Stormy Weather' if stormy_weather?
+    plane.land!
+    @planes << plane
```
tansaku: it's good practice to refer to instance variables like @planes with the attr_reader you have created, so this line of code could be planes << plane

lib/airport.rb
```diff
@@ -1,2 +1,31 @@
...
+  end
+
+  def plane_take_off(plane)
+    fail 'Airport Empty' if @planes.empty?
+    fail 'Stormy Weather' if stormy_weather?
+    plane1 = @planes.delete(plane)
```
tansaku: I'm not so sure about defining plane1 here - when the plane is in the array it will just be the same as plane.  Ideally we should be failing with an error of our own when the plane is not in the array rather than leaving it to through a no method exception for nil when it is not present

lib/airport.rb
```diff
@@ -1,2 +1,31 @@
...
+  end
+
+  def land_plane(plane)
+    fail 'Airport Full' if @planes.count >= @capacity
+    fail 'Stormy Weather' if stormy_weather?
+    plane.land!
+    @planes << plane
+  end
+
+  def plane_take_off(plane)
```
tansaku: I'd be tempted to just call this take_off

lib/plane.rb
```diff
@@ -1,2 +1,18 @@
 class Plane
+  attr_reader :flying
+  alias_method :flying?, :flying
+
+  def initialize
+    @flying = true
+  end
+
+  def land!
+    fail 'Already Landed' unless @flying
```
tansaku: use your attr_reader here rather than the instance variable

lib/plane.rb
```diff
@@ -1,2 +1,18 @@
...
+    @flying = false
+  end
+
+  def take_off!
+    fail 'Already Flying' if @flying
```
tansaku: use your attr_reader here rather than the instance variable

spec/airport_spec.rb
```diff
@@ -1,42 +1,50 @@
...
+    it 'a plane can take off in normal conditions' do
+      subject.land_plane(plane)
+      expect(subject.plane_take_off(plane)).to eq plane
+    end
   end
 
   context 'traffic control' do
 
-    xit 'a plane cannot land if the airport is full'
+    it 'can return the landed planes' do
```
tansaku: I don't think that this test does what it says it does

spec/airport_spec.rb
```diff
@@ -1,42 +1,50 @@
...
+      expect { subject.plane_take_off(plane) }.to raise_error 'Airport Empty'
+    end
 
     context 'weather conditions' do
-      xit 'a plane cannot take off when there is a storm brewing'
-
-      xit 'a plane cannot land in the middle of a storm'
+      let(:airport) { Airport.new }
+      it 'a plane cannot take off when there is a storm brewing' do
+        allow(subject).to receive(:stormy_weather?).and_return(false, true)
```
tansaku: I don't think you want to pass two arguments into and_return here

spec/feature/grand_finale_spec.rb
```diff
@@ -8,8 +9,61 @@
 # Check when all the planes have landed that they have status "landed"
 # Once all planes are in the air again, check that they have status "flying!"
 
-feature 'Grand Finale' do
+feature 'Grand Finale: Good Weather' do
+  airport = Airport.new
```
tansaku: you shouldn't really be declaring variables here - they should be inside a before block

spec/plane_spec.rb
```diff
@@ -1,25 +1,28 @@
...
-  xit 'has a flying status when created'
-
-  xit 'has a flying status when in the air'
-
-  xit 'can take off'
-
-  xit 'changes its status to flying after taking off'
-
+  it 'has a flying status when created' do
+    expect(subject.flying?).to be true
```
tansaku: this could use a predicate, e.g. expect(subject).to be_flying

tansaku:
Overall good work here - see the embedded comments for some minor tweaks but you are basically there.

:checkered_flag:

PR #10, opened 2015-03-22
Airport challenge solution submission - James

james-miller:


james-miller:
Hound is asking for single-quoted strings whilst RuboCop asks for double-quoted. Given that if I disobey RuboCop I am unable to run any tests locally I have stuck with single quotes.

PR #15, opened 2015-03-22
Challenge Submission

augustinas:


lib/airport.rb
```diff
@@ -1,2 +1,43 @@
 class Airport
+  DEFAULT_CAPACITY = 20
```
sjmog: Great use of constants for static numbers that rarely change

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+  end
+
+  def plane_count
+    planes.length
+  end
+
+  private
+
+  attr_accessor :planes
+  attr_reader :weather
```
sjmog: You **can** do this but it's super-weird to see `attr_...` midway through a class. If you *really* want these to be private, see http://heartmindcode.com/2013/04/25/private_accessors_in_ruby/ and get back to me. Otherwise, just make them public and be slightly upset :tongue:

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
 class Airport
+  DEFAULT_CAPACITY = 20
+  attr_reader :capacity
+
+  def initialize options = {}
+    @planes = []
+    @capacity = options.fetch(:capacity, DEFAULT_CAPACITY)
```
sjmog: Nice use of `fetch`. See Rob's submission for using `merge` with a `defaults` hash, too.

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+    @weather = %w(sunny sunny sunny sunny stormy).shuffle.pop
```
sjmog: It seems weird that `Airport` is responsible for its own weather. This probably belongs in a separate `Weather` class (or, since `Weather` should never be instantiated, as a `Module`).

Also, I think there are more efficient and natural ways of setting this sort of thing up. If we want to extend weather, we have to modify this `%w` array, which is not nice. We can't really test weather neatly, in isolation. This is also not very DRY.

I would suggest investigating `rand` to generate a random weather, with parameters. See Jo's submission for an awesome example of how to randomise weather (he even has a modifier for `climate`), and Rob's submission for splitting `Weather` out as a module.

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+  end
+
+  def land plane
+    fail 'Sorry, Airport Full' if full?
+    fail 'Impossible, Stormy Weather' if stormy_weather?
```
sjmog: These error messages could be more descriptive. As a user trying to land a plane, I'll get

`Runtime Error: Sorry, Airport Full` and `Runtime Error: Impossible, Stormy Weather`

which kind of tell me what the problem *is*, but not how to avoid them. Better would be

`Runtime Error: Airport Full. Try taking off some planes from the Airport and trying again`

This is in line with Nielsen's advice for providing useful error messages, which always give the user the resources to **avoid the error in future**.

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+    planes << plane
+  end
+
+  def take_off
+    fail 'Airport is Empty' if empty?
+    fail 'Impossible, Stormy Weather' if stormy_weather?
+    planes.pop
```
sjmog: This doesn't allow us to take off a specific plane - only the last plane to land :(

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+  end
+
+  def plane_count
+    planes.length
+  end
```
sjmog: Does this method add anything above just using `planes.count`?

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+
+  attr_accessor :planes
+  attr_reader :weather
+
+  def empty?
+    planes.empty?
+  end
+
+  def full?
+    planes.length >= capacity
```
sjmog: Awesome: you've handled the edge case where Airport could be initialised with too many planes.

lib/plane.rb
```diff
@@ -1,2 +1,5 @@
 class Plane
+  def status
+    'airborne'
+  end
```
sjmog: This seems unfinished.

spec/airport_spec.rb
```diff
@@ -15,14 +15,36 @@
 
   context 'taking off and landing' do
 
-    xit 'a plane can land'
+    it 'a plane can land' do
+      allow(subject).to receive(:weather) { 'sunny' }
+      expect(subject.plane_count).to eq(0)
```
sjmog: You can use `be_zero` here

spec/airport_spec.rb
```diff
@@ -15,14 +15,36 @@
...
+      subject.land double :plane
```
sjmog: Nice use of inline `double`s

spec/airport_spec.rb
```diff
@@ -15,14 +15,36 @@
...
+      expect(subject.plane_count).to eq(1)
+    end
 
-    xit 'a plane can take off'
+    it 'a plane can take off' do
```
sjmog: Maybe refactor this description so it reads more naturally, like `it 'can take off a plane'`

spec/airport_spec.rb
```diff
@@ -15,14 +15,36 @@
...
+      subject.land :plane
+      expect(subject.take_off).to eq :plane
+    end
   end
 
   context 'traffic control' do
 
-    xit 'a plane cannot land if the airport is full'
+    it 'has a fixed capacity' do
+      expect(subject.capacity).to eq(20)
```
sjmog: Good use of `subject` and knowing that `Airport.new` defaults to a `DEFAULT_CAPACITY`

spec/airport_spec.rb
```diff
@@ -15,14 +15,36 @@
...
+    end
+
+    it 'cannot take off a plane if the airport is empty' do
+      allow(subject).to receive(:weather) { 'sunny' }
+      expect { subject.take_off }.to raise_error 'Airport is Empty'
+    end
```
sjmog: Nice test

spec/feature/grand_finale_spec.rb
```diff
@@ -9,7 +10,29 @@
...
 
 feature 'Grand Finale' do
+  let(:airport) { Airport.new }
+  let(:plane) { Plane.new }
 
-  xscenario 'all planes can land and all planes can take off'
+  scenario 'planes can land and take off' do
+    allow(airport).to receive(:weather) { 'sunny' }
+    expect(plane.status).to eq 'airborne'
+    airport.land plane
```
sjmog: Shouldn't you `expect(plane.status).to eq 'grounded'`?

spec/feature/grand_finale_spec.rb
```diff
@@ -9,7 +10,29 @@
...
+    airborne_plane = airport.take_off
+    expect(airborne_plane.status).to eq 'airborne'
```
sjmog: Because you've only implemented a test for one half of this feature, you haven't implemented a variable `status` on `Plane`!

spec/feature/grand_finale_spec.rb
```diff
@@ -9,7 +10,29 @@
...
+  end
 
+  scenario 'airport has a limited capacity' do
+    allow(airport).to receive(:weather) { 'sunny' }
+    expect { airport.take_off }.to raise_error 'Airport is Empty'
+    airport.capacity.times { airport.land plane }
+    expect { airport.land plane }.to raise_error 'Sorry, Airport Full'
+  end
```
sjmog: Very nice :)

spec/feature/grand_finale_spec.rb
```diff
@@ -9,7 +10,29 @@
...
+
+  scenario 'the planes cannot fly during stormy weather' do
+    allow(airport).to receive(:weather) { 'sunny' }
+    airport.land plane
+    allow(airport).to receive(:weather) { 'stormy' }
+    expect { airport.land plane }.to raise_error 'Impossible, Stormy Weather'
+    expect { airport.take_off }.to raise_error 'Impossible, Stormy Weather'
+  end
```
sjmog: :+1:

spec/plane_spec.rb
```diff
@@ -14,7 +14,9 @@
 
 describe Plane do
 
-  xit 'has a flying status when created'
+  it 'has a airborne status when created' do
+    expect(subject.status).to eq 'airborne'
+  end
 
   xit 'has a flying status when in the air'
```
sjmog: We're missing quite a bit of this test. What happened?

sjmog:
A lot of great stuff here, Gus: use of constants, understanding of feature vs. unit tests, neat implementation, good understanding of encapsulation. However, there are still three pending tests and some tests for Plane have not been written. Your feature tests fall short of the user stories, and your plane implementation is as a result currently failing the client. You need to rectify this!

You might want to consider retracing your steps through this and making sure you've met all the requirements before moving on. There's a lot of learning to do here (like creating and including a `Weather` module, and then trying to isolate it for testing, or using `merge` to create a `defaults` hash). Once you've re-submitted a fully-working version that truly demonstrates your prowess, I'll pass it :+1:

sjmog:
:construction:

PR #17, opened 2015-03-23
Plane and AirportSlot classes, 2 methods, 2 Unit and 1 Feature Test work...

maciejk77:
I have managed to define basics: Plane and Airport (AirportSlot) classes. I also created working Feature and two unit tests.

I am not sure on the choice of methods where, I suspect 'available?' and 'accepted_plane' possibly refer to the same class AirportSlot - not to the the separate ones. 

Apologies, I was not able to do anymore, which I am really gutted about

PS
So far I am struggling with course material and 'scaffolding video' sessions (confident only in 1,5 / 4 of those)

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,2 +1,5 @@
-class Airport
+class AirportSlot
+    def available?
+    true
+    end
 end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,2 +1,7 @@
+require_relative 'airport_slot'
+
 class Plane
+  def accepted_plane
+  AirportSlot.new
+  end
 end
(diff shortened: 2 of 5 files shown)


tansaku:
Hi Maciej,

No apologies necessary.  Sad to hear that you feel gutted :-( One of the key things I hope you can take from this is that it is okay not to make the progress you hoped.  Try to put those feelings to one side and let's focus on what we can learn from the process.

I'm seeing a lot of indenting issues that we really need to get sorted.  Maybe we can set you up with RuboCop in your text editor to help with that.

It looks to me like the key thing here was that you strayed from the naming conventions creating an Airport Slot rather than an Airport.  Will be very happy to discuss further - you do have a start here that could be turned into a complete solution.

:construction:

PR #20, opened 2015-03-23
weekend challenge - changes

M1lena:
I found this to be a very fun challenge! Up until the "stub!" with the weather at least.

My answers are pretty much only for the tests given in the original repo files (and hopefully those were correct).

I hope to come back and add/update this as the stub method is cleared up.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,2 +1,34 @@
 class Airport
+
+  DEFAULT_CAPACITY = 100
+
+  attr_reader :planes, :weather
+
+  def initialize(options = {})
+    @capacity = options.fetch(:capacity, DEFAULT_CAPACITY)
+    @planes = []
+    @weather = rand(5)
+  end
+
+  def capacity
+    @capacity
+  end
+
+  def ok_landing(plane)
+    raise "Airport is full" if full?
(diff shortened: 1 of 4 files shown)


tansaku:
Hi Milena,

Well done completing most of the challenge.  It sounds like you are aware of the remaining issues.  But just in case, here are my main concerns

* lots of hound/rubocop issues
* no grand finale spec
* airport unit test includes Plane.new --> Airport unit test should be independent of Plane
* lots of commented code hanging around
* plane landing at airport does not update plane's status, similarly for takeoff

```sh
2.1.5 :001 > plane = Plane.new
 => #<Plane:0x007fdcc2affe00 @flying=true> 
2.1.5 :002 > airport = Airport.new
 => #<Airport:0x007fdcc2afc570 @capacity=100, @planes=[], @weather=0> 
2.1.5 :003 > airport.ok_landing plane
 => [#<Plane:0x007fdcc2affe00 @flying=true>] 
2.1.5 :004 > airport.ok_takeoff plane
 => #<Plane:0x007fdcc2affe00 @flying=true> 
```

but given that we've discussed the weather and there are now some code-alongs available, I imagine you'll be able to finish this up without too much effort

:construction:

PR #21, opened 2015-03-23
First Weekend Challenge

Munded:
Hopefully it's up to scratch


tansaku:
Some great work here Ed!  Marred slightly by lots of indentation issues :-)

Rubocop/Hound also identifies some other important issues:

* lib/plane.rb:9:7: C: Rename has_landed to landed?. def has_landed <-- actually what we should be doing is replacing this with 'land', since you are setting instance variable via this method - not querying one
* lib/airport.rb:27:3: C: Use attr_reader to define trivial reader methods. --> def plane_check

One pro-tip is that: airport spec could refer to DEFAULT_CAPACITY to avoid magic number

All your tests pass, which is great :-)

There is some cleanup needed but you are basically there.  Get the indenting fixed and I'll be happy to give you a badge.

:checkered_flag:

PR #22, opened 2015-03-23
Weekend Challenge but not yet complete

massud:
I gave it a good go, but still needs lots of work to get it being a fully functioning Airport.


tansaku:
Hi Massud,

You're off to a good start here.  Very pleased to see that you have good indenting and that Rubocop is happy.

Couple of minor points.  

* The Plane unit test could use 'subject' for RSpec to save a line or two of code
* 'plane.flight' <-- prefer status? --> I wonder if plane.flight as a method is a little ambiguous

You could certainly get a badge spending a bit more time on this.  How long did this take you?  Please do chat with me if you need pointers on how to continue :-)

:construction:

PR #23, opened 2015-03-23
The sorry efforts of IF

ilyafaybisovich:


README.md
```diff
@@ -1,62 +1,3 @@
...
-* Tests coverage is good
-* The code is elegant: every class has a clear responsibility, methods are short etc.
-
-Note that is a practice 'Tech Test' of the kinds that employers use to screen developer applicants.  More detailed submission requirements/guidelines are in [CONTRIBUTING.md](CONTRIBUTING.md)
-
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first. 
-
-* Finally submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am
+Not much to explain, I am afraid. 
+Lost almost all Friday for personal reasons, spent too much time reading and redoing Boris Bikes, did not plan enough time to code in non-panic mode.
```
sjmog: This is OK! Hopefully today was helpful?

README.md
```diff
@@ -1,62 +1,3 @@
...
-* The code is elegant: every class has a clear responsibility, methods are short etc.
-
-Note that is a practice 'Tech Test' of the kinds that employers use to screen developer applicants.  More detailed submission requirements/guidelines are in [CONTRIBUTING.md](CONTRIBUTING.md)
-
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first. 
-
-* Finally submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am
+Not much to explain, I am afraid. 
+Lost almost all Friday for personal reasons, spent too much time reading and redoing Boris Bikes, did not plan enough time to code in non-panic mode. 
+Need to rethink the whole exercise and my approach.
```
sjmog: Take another stab along with Tansaku's video :)

sjmog: And thanks for putting this in the README! Makes it nice and clear :+1:

lib/airport.rb
```diff
@@ -1,2 +1,11 @@
+require_relative 'plane'
+
 class Airport
+  attr_accessor :capacity, :planes, :weather
+
+  def initialize(capacity = 10)
```
sjmog: Great idea to put a default for capacity :+1:

lib/airport.rb
```diff
@@ -1,2 +1,11 @@
...
+    @capacity = capacity
```
sjmog: I've mostly seen this as a constant, I like your way better

lib/airport.rb
```diff
@@ -1,2 +1,11 @@
...
+    @planes = []
+    @weather = :good
```
sjmog: Should the airport's *state* include the weather?

lib/airport.rb
```diff
@@ -1,2 +1,11 @@
+require_relative 'plane'
+
 class Airport
+  attr_accessor :capacity, :planes, :weather
```
sjmog: When using `attr_...`, try to use `reader` first and only grant extra privileges like `writer` and `accessor` as and when they are necessary.

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
...
+  end
+
+  def take_off(airport)
+    airport.planes.reject! { |plane| self == plane }
+    @status = :flying
+    'Took off!'
+  end
+
+  def land(airport)
+    fail 'cannot land in storm' if airport.weather == :stormy
```
sjmog: Nice use of `fail`

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
...
+
 class Plane
+  attr_reader :status
+
+  def initialize
+    @status = :flying
+  end
+
+  def take_off(airport)
+    airport.planes.reject! { |plane| self == plane }
```
sjmog: This seems like a very odd call for a plane to be making. The plane shouldn't really be asking the airport to do anything, in my opinion.

Simpler objects should rarely, if ever, make calls to more complex ones.

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
...
+    @status = :flying
+    'Took off!'
```
sjmog: This would be a nice thing to put in a `puts` statement but returning it is strange (I call `plane.take_off(airport)` and I get a string? seems odd)

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
...
+  end
+
+  def land(airport)
+    fail 'cannot land in storm' if airport.weather == :stormy
+    fail 'all spaces are taken' if airport.planes.length == airport.capacity
+    airport.planes << self
```
sjmog: Again, if you find yourself modifying the state of one object from within another, question if this is the right way round. You might want to use **Dependency Inversion** to sort out the unnecessary dependency.

spec/airport_spec.rb
```diff
@@ -15,28 +15,39 @@
 
   context 'taking off and landing' do
 
-    xit 'a plane can land'
+    it 'a plane can land' do
+      plane = Plane.new
```
sjmog: We should never be talking about other classes in a unit test, or they aren't properly isolated. This means changes to `Plane` could break the tests for `Airport`, which is super-weird.

Instead, use a `double`, with `let(:plane) { double :plane }` and mock up extra behaviour as required.

spec/airport_spec.rb
```diff
@@ -15,28 +15,39 @@
 
   context 'taking off and landing' do
 
-    xit 'a plane can land'
+    it 'a plane can land' do
```
sjmog: If you find yourself using awkward statements like this, that's a good indication you need to reconsider your dependencies. If we changed this to `it 'lands a plane'` you could have used that to refactor all the `land` and `take_off` logic into the `Airport` class, saving yourself a lot of bother

spec/airport_spec.rb
```diff
@@ -15,28 +15,39 @@
...
-
-    # Include a weather condition.
-    # The weather must be random and only have two states "sunny" or "stormy".
-    # Try and take off a plane, but if the weather is stormy,
-    # the plane can not take off and must remain in the airport.
-    #
-    # This will require stubbing to stop the random return of the weather.
-    # If the airport has a weather condition of stormy,
-    # the plane can not land, and must not be in the airport
+    it 'a plane cannot land if the airport is full' do
```
sjmog: Again, `it 'cannot land a plane if the airport is full'` makes more sense to me...and again suggests that `land` should be the airport's responsibility, not the plane's

spec/airport_spec.rb
```diff
@@ -15,28 +15,39 @@
...
+      airport = Airport.new
+      plane = Plane.new
+      allow(airport).to receive(:weather) { :good }
+      airport.capacity.times do
```
sjmog: Nice use of airport capacity :+1:

spec/airport_spec.rb
```diff
@@ -15,28 +15,39 @@
...
+        plane.land(airport)
+      end
+      expect { Plane.new.land(airport) }.to raise_error('all spaces are taken')
```
sjmog: You see that by making `Plane` do all the heavy lifting we are giving it too much power. Airport essentially *does* nothing. If you find yourself needing to use a class, or mock up a really complex double, in a unit test for a *different* class (like here), ask yourself if your dependencies can be inverted, or otherwise changed (such as through **injection**)

spec/airport_spec.rb
```diff
@@ -15,28 +15,39 @@
...
-    context 'weather conditions' do
-      xit 'a plane cannot take off when there is a storm brewing'
+  context 'weather conditions' do
+    it 'a plane cannot take off when there is a storm brewing' do
+    end
 
-      xit 'a plane cannot land in the middle of a storm'
+    it 'a plane cannot land in the middle of a storm' do
+      airport = Airport.new
+      allow(airport).to receive(:weather) { :stormy }
```
sjmog: Nice stubbing here

spec/airport_spec.rb
```diff
@@ -15,28 +15,39 @@
...
+      plane = Plane.new
+      expect { plane.land(airport) }.to raise_error('cannot land in storm')
```
sjmog: See earlier comment

spec/feature/pilot_land_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'capybara/rspec'
+require 'plane'
+require 'airport'
+
+feature 'pilot can' do
```
sjmog: Features are usually a little more descriptive: e.g. 'pilot can land a plane'. Scenario in this case could then be 'land a single plane'. That leaves the 'pilot can land a plane' feature open to considerable extension in future if the client wishes.

spec/feature/pilot_land_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  scenario 'arrive at the desired destination' do
+    plane = Plane.new
+    expect(plane).to respond_to(:land)
+  end
```
sjmog: Lovely feature test :+1:

spec/feature/pilot_land_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  scenario 'take off from the appropriate airport' do
+    plane = Plane.new
+    airport = Airport.new
+    plane.land(airport)
+    expect { plane.take_off(airport) }.to be_equal { 'Took off!' }
+  end
```
sjmog: Super awesome feature test :+1:

spec/plane_spec.rb
```diff
@@ -14,12 +14,39 @@
...
 
-  xit 'has a flying status when created'
-
-  xit 'has a flying status when in the air'
-
-  xit 'can take off'
-
-  xit 'changes its status to flying after taking off'
+  it 'has a flying status when created' do
+    plane = Plane.new
```
sjmog: When in a `describe` block for a class (here `Plane`), you can use `subject`. Here, `plane == subject`.

spec/plane_spec.rb
```diff
@@ -14,12 +14,39 @@
...
+    expect(plane.status).to be(:flying)
```
sjmog: So this could be `expect(subject.status).to be(:flying)`

spec/plane_spec.rb
```diff
@@ -14,12 +14,39 @@
...
+  end
+
+  it 'has a flying status when in the air' do
+    plane = Plane.new
+    expect(plane.take_off(airport)).to be(:flying)
```
sjmog: Since we only have two states of plane (`:flying` and `:landed`), maybe we should think about storing this as a boolean in the plane **state** somehow.

spec/plane_spec.rb
```diff
@@ -14,12 +14,39 @@
...
+  end
+
+  it 'can take off' do
+    plane = Plane.new
+    expect(plane).to respond_to(:take_off)
+  end
```
sjmog: These four lines could be written as:

`it { is_expected.to respond_to :take_off }`

spec/plane_spec.rb
```diff
@@ -14,12 +14,39 @@
...
+
+  it 'can land' do
+    plane = Plane.new
+    expect(plane).to respond_to(:land)
+  end
```
sjmog: Ditto these ^

spec/plane_spec.rb
```diff
@@ -14,12 +14,39 @@
...
+
+  it 'has a landed status after landing' do
+    plane = Plane.new
+    airport = Airport.new('Heathrow')
```
sjmog: I'm guessing this changed somewhere...don't you initialise Airports with a capacity parameter?

spec/plane_spec.rb
```diff
@@ -14,12 +14,39 @@
...
+    allow(airport).to receive(:weather) { :good }
```
sjmog: Nice use of allow

spec/plane_spec.rb
```diff
@@ -14,12 +14,39 @@
...
+    expect(plane.land(airport)).to be(:landed)
```
sjmog: This is quite fragile, as it is dependent on a very particular implementation of the plane's `:land` method (specifically, that it returns a symbol). Maybe think about how to generalise this.

sjmog:
Again, there's a huge amount to love here, Ilya. I think you're being too hard on yourself!

In the main, you have RSpec syntax nailed. You might want to check some places where you could make it more general. On the design front, there are a lot of good ideas - what really threw you was deciding that `Plane` should have the responsibility for landing and take-offing `plane` objects. If you had set off in the other direction - by giving most of the heavy lifting to `Airport` instead - I think you would have found this a lot easier.

My recommendation is *try it again*, this time trying to load more responsibility onto the `Airport` class. `Plane` should be responsible only for keeping track of its own `flying` state, and manipulating it in response to methods from `Airport`.

In general, I'm impressed with a lot here: I think you'll find the process of coding up Battleships to be really handy. :+1:

sjmog:
Don't forget to refactor this so it passes!

:construction:

PR #26, opened 2015-03-23
SOC attempt at the airport_challenge

saramoohead:


lib/airport.rb
```diff
@@ -1,2 +1,21 @@
 class Airport
+  DEFAULT_CAPACITY = 10
+
+  def initialize
+    @planes = []
+  end
+
+  def agree_land plane
```
stephenlloyd: the name of this method doesn't express fully what its doing,  When I call this method i want to know that it will land the plane. maybe just calling it land would be better

lib/airport.rb
```diff
@@ -1,2 +1,21 @@
...
+    fail 'Airport is full' if full?
+    @planes << plane
+    nil
+  end
+
+  private
+
+  attr_reader :planes
```
stephenlloyd: this should be at the top of the class and probably not private

lib/airport.rb
```diff
@@ -1,2 +1,21 @@
...
+
+  def full?
+    @planes.length >= DEFAULT_CAPACITY
```
stephenlloyd: use the planes method you created rather than the instance variable. also I would expose the default capacity as a capacity method and allow it so you can change the capacity should you need to. Look up the open/closed principle. I wouldnt want to fiddle with the class in order to change the capacity if the requirements change.

lib/airport.rb
```diff
@@ -1,2 +1,21 @@
...
+  end
 end
```
stephenlloyd: this class is nearly there. i was surpised not to see something with the weather there. It would eventually be a module but you could start by putting it into the airport. I'll cover this later.

lib/plane.rb
```diff
@@ -1,2 +1,18 @@
 class Plane
+  attr_reader :status
+  alias_method :status?, :status
```
stephenlloyd: a boolean method "status?" implies that it will return true or false, status returns a string so this would never be a boolean method. I would get rid of this line as it's not needed.

spec/airport_spec.rb
```diff
@@ -1,42 +1,49 @@
...
 # A plane currently in the airport can be requested to take off.
-#
-# No more planes can be added to the airport, if it's full.
-# It is up to you how many planes can land in the airport
-# and how that is implemented.
-#
-# If the airport is full then no planes can land
 
 describe Airport do
+  let(:plane) { Plane.new }
```
stephenlloyd: this should be a double and not a 'real' object

spec/airport_spec.rb
```diff
@@ -1,42 +1,49 @@
...
+  end
 
-    xit 'a plane can take off'
+  context 'appropriate plane status checks' do
+    it 'when a plane is landed, it cannot land' do
+      plane.land
+      expect { plane.land }.to raise_error 'Plane already landed.'
+    end
+    it 'when a plane is flying, it cannot take off' do
+      plane.land
```
stephenlloyd: this is testing the plane and not the airport.

spec/airport_spec.rb
```diff
@@ -1,42 +1,49 @@
...
-  context 'taking off and landing' do
-
-    xit 'a plane can land'
+    xit 'when a plane is taking off'
+  end
 
-    xit 'a plane can take off'
+  context 'appropriate plane status checks' do
+    it 'when a plane is landed, it cannot land' do
+      plane.land
```
stephenlloyd: this is testing the plane and not the airport.

spec/plane_spec.rb
```diff
@@ -1,25 +1,33 @@
...
-#
 # When the plane takes of from the airport, the plane's status
 # should become "flying"
 
 describe Plane do
+  let(:plane) { Plane.new }
 
-  xit 'has a flying status when created'
+  it 'has a flying status when created' do
+    expect(plane).to have_attributes(status: 'flying')
```
stephenlloyd: interesting. I've not seen this before but it's good :)

spec/plane_spec.rb
```diff
@@ -1,25 +1,33 @@
...
 
-  xit 'changes its status to flying after taking off'
+  it 'changes its status to landed after landing' do
+    expect { plane.land }.to change(plane, :status).from('flying').to('landed')
+  end
 
+  xit 'must get approval from airport for taking off'
+
+  it 'can take off' do
+    expect(plane).to respond_to :take_off
```
stephenlloyd: I don't like respond_to tests. They're easy to pass without adding functionality. test for what you expect the method to do rather than it's existence and you'll have a better test.

stephenlloyd:
this still needs to be worked on and if you want me to  go through some things with you id be more than happy. You have some good in there but you seem a little confused. Hopefully when I go through it later i'll clear up a few thigns :) :construction:

PR #30, opened 2015-03-23
airport disaster

Stacca:


spec/plane_spec.rb
```diff
@@ -1,25 +1,26 @@
...
-
-  xit 'has a flying status when created'
-
-  xit 'has a flying status when in the air'
-
-  xit 'can take off'
-
-  xit 'changes its status to flying after taking off'
-
+  let(:plane) { Plane.new }
```
stephenlloyd: nice, good use of let.

spec/plane_spec.rb
```diff
@@ -1,25 +1,26 @@
...
-
-# When we create a new plane, it should have a "flying" status,
-# thus planes can not be created in the airport.
-#
-# When we land a plane at the airport, the plane in question should
-# have its status changed to "landed"
-#
-# When the plane takes of from the airport, the plane's status
-# should become "flying"
+require 'airport'
```
stephenlloyd: you don't need this

spec/airport_spec.rb
```diff
@@ -1,42 +1,20 @@
...
-    # If the airport has a weather condition of stormy,
-    # the plane can not land, and must not be in the airport
-
-    context 'weather conditions' do
-      xit 'a plane cannot take off when there is a storm brewing'
-
-      xit 'a plane cannot land in the middle of a storm'
+  describe 'plane must have permission to land' do
+    it 'airport denines landing' do
+      expect(Weather.sunny).to eq false
```
stephenlloyd: I'm not entirely sure what this test is doing

spec/airport_spec.rb
```diff
@@ -1,42 +1,20 @@
...
-# No more planes can be added to the airport, if it's full.
-# It is up to you how many planes can land in the airport
-# and how that is implemented.
-#
-# If the airport is full then no planes can land
+require 'weather'
 
 describe Airport do
+  let(:airport) { Airport.new }
+  let(:plane) { double :plane }
```
stephenlloyd: nice, glad plane is a double. It would be nice to see you using expect(plane).to receive(method_name) to ensure it's landing a plane when they land.

lib/plane.rb
```diff
@@ -1,2 +1,18 @@
 class Plane
+  def initialize
+  end
+  attr_reader :flying
+
+  def status
+    flying? ? :flying : :landed
+    # if flying? :flying  else :landed
```
stephenlloyd: always delete commended code. Although this is a good use of a ternary statement.

lib/plane.rb
```diff
@@ -1,2 +1,18 @@
 class Plane
+  def initialize
```
stephenlloyd: you're not doing anything with this so it may aswell not be there :)

lib/airport.rb
```diff
@@ -1,2 +1,23 @@
...
+  def land plane
+    planes << plane
+  end
+
+  def full?
+    planes_count >= @capacity
+  end
+
+  def runway
+    @runway ||= []
```
stephenlloyd: you don't seem to be using this method anymore.

stephenlloyd:
this is good, definitely some improvements from what I saw yesterday. I would recommend stubbing out the weather call so it can either be sunny or stormy depending on what you want for your tests and then you'll get the badge. :construction:

PR #31, opened 2015-03-23
aiport_challenge

Tomiblanchard:


lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+  MAXIMUM_CAPACITY = 30
+  DEFAULT_CAPACITY = 20
+  include Weather
+
+  def initialize
+    @airport_capacity ||= DEFAULT_CAPACITY
+    @planes ||= []
+  end
+
+  def flying?
```
stephenlloyd: this method isn't being used anywhere. I think it's more for the plane and not the airport. an airport.flying? method doesnt make sense as an airport shouldnt be able to fly.

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+require_relative 'plane'
+
 class Airport
+  attr_reader :airport_capacity, :flying, :planes
+  MAXIMUM_CAPACITY = 30
+  DEFAULT_CAPACITY = 20
+  include Weather
+
+  def initialize
+    @airport_capacity ||= DEFAULT_CAPACITY
```
stephenlloyd: when can this not be the default capacity? It's impossible to change as it is.

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+  end
+
+  def take_off(plane)
+    raise 'plane cannot take off' if @stormy
+    plane.take_off
+    @planes << plane
+    @planes.delete plane
+  end
+
+  def full?
```
stephenlloyd: this method should return true or false as its a boolean method. right now you're returning strings of 'true' and 'false'. You could write it all in one line by doing just
``` def full?
          planes.count >= AIRPORT_CAPACITY 
   end
````
notice i havent used Maximum capacity as you should just check the actual capactiy of the airport.

File: lib/plane.rb [hunk 4]
```diff
@@ -1,2 +1,22 @@
 class Plane
+  attr_reader :flying, :status
+
+  def initialize
+    @flying = true
+    @status = :landed
+  end
+
+  def flying?
+    @flying
```
stephenlloyd: could you figure this out just based on the staus, then you wont need lines 14 and 19

spec/airport_spec.rb
```diff
@@ -1,42 +1,52 @@
...
+    it 'a plane can land' do
+      airport.land(plane)
+      expect(airport.planes).to eq [plane]
+    end
 
-    xit 'a plane can take off'
+    it 'a plane can take off' do
+      airport.land(plane)
+      airport.take_off plane
+      expect(airport.planes).to be_empty
```
stephenlloyd: you're never testing the plane is receiving the method calls, if i remove the code which calls plane.land it doens't break a test which is worrying. you should never be able to remove a line of code without breaking a test. if you can you havent written enough tests.

spec/airport_spec.rb
```diff
@@ -1,42 +1,52 @@
...
+    end
   end
 
   context 'traffic control' do
 
-    xit 'a plane cannot land if the airport is full'
+    it 'a plane cannot land if the airport is full' do
+      Airport.stub(:weather).and_return(:'stormy')
```
stephenlloyd: this is old syntax. you should be using the newer syntax

spec/weather_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'weather'
+
+describe Weather do
```
stephenlloyd: I don't think this test is doing anything :) maybe test the methods in the airport and then extract into a shared_examples test.

stephenlloyd:
This test was pretty good. I think you need to make sure your testing all of you code and not having functionality in your classes which can be removed without a test failing. Also if your using a boolean method make sure youre returning either true or false, this will actually make youre code easier to write/test and read. 
Although you didn't test weather quite right it was still used well in the airport class. I'll give you a sticker for this but feel free to neaten it up and impliment my notes. 
:checkered_flag:

PR #32, opened 2015-03-24
Not finished, but what I could manage

domaro:


tansaku:
hey dias, will be interested to see you take another stab at this at some point

:construction:

PR #34, opened 2015-03-24
Changes

Munded:
everything's sorted


lib/airport.rb
```diff
@@ -1,2 +1,40 @@
+require_relative 'plane'
+
 class Airport
+  DEFAULT_CAPACITY = 10
+
+  attr_reader :planes
+
+  def initialize(airport_size = {})
+    @planes = []
+    @weather_conditions = %(Stormy Sunny)
```
tansaku: this should be %w(Stormy Sunny)

spec/plane_spec.rb
```diff
@@ -13,13 +13,20 @@
...
-
-  xit 'has a flying status when created'
-
-  xit 'has a flying status when in the air'
-
-  xit 'can take off'
-
-  xit 'changes its status to flying after taking off'
-
+  let(:plane) { Plane.new }
```
tansaku: prefer subject here

tansaku:
I want to give you the badge, but you accidentally introduced an error that we should look at first - also would be great if you could sync to master - I'll show you how ...

PR #41, opened 2015-05-01
Created grand finale feature test

DanielJohnston:


silvabox:
Thanks Daniel - that's a pretty good submission.  I would have liked to see the weather responsibility handled outside the airport, but in a more complex domain that refactor would probably have emerged anyway. :checkered_flag:

PR #42, opened 2015-05-01
Plane can land at airport

dwatson62:


silvabox:
Thanks for the submission Daryl - on the whole it's pretty good.  There's a bit of cross over between your feature tests and unit tests and they're a bit confused in terms of which is which, but that will come.  You're also using a slightly unusual `::` syntax which will confuse people!  Use the standard `.` syntax when calling methods.  :checkered_flag:

PR #43, opened 2015-05-01
initial plan added to README.md

timrobertson0122:


silvabox:
Hi Tim, thanks for submitting your airport challenge.  Some of your code is OK, but there are some serious issues with parts of it.  But that's not a bad thing at this stage - it's an opportunity to learn!

My biggest concern is your use of `attr_accessor`.  In `Airport` you have `attr_accessor :hangar`.  While, technically, that fulfils the requirements, it exposes your class to a whole load of untested vulnerabilities.  For example, because you've exposed a `hangar=` setter method, I could say `airport.hangar = nil` at any point and destroy the collection of planes.  Or `airport.hangar = Gerbil.new` or some other completely unrelated thing.  You don't even reference the getter and setters from inside the `Airport` class - you always reference the instance variable `@hangar`.  There are calls to hangar in your tests, which test `subject.hangar.count` but this breaks encapsulation because I am required to know that `Airport` has a `hangar` method and that the `hangar` method returns an object with a `count` method.

You have a similar issue in `Plane` with `attr_accessor :status` which enables me to change the status from outside `Plane` independently of the `land` and `takeoff` methods.

Also in `Airport` you assign `@weather`, (`@weather = weather`) but that instance variable is never used again, so it's superfluous.

Have a look at your unit tests for `Plane` - you test that it has certain methods; but you don't test what happens when you call those methods.

Lot's to take in there!  Give me a shout if you want to go over any of it :construction:

PR #45, opened 2015-05-02
airport-challenge

MollieS:
creating a pull request to check progress


lib/airport.rb
```diff
@@ -1,2 +1,31 @@
...
+    fail "Airport is full" if full?
+    plane.landed = true
+    plane.flying = false
+    planes << plane
+  end
+
+  def take_off plane
+    plane.flying = true
+    plane.landed = false
+    planes.delete(plane)
```
rodcul: @MollieS nice Pull Request!

Just curious, any reason you decided to use delete vs pop (as in boris-bikes)

Thanks!

spec/airport_spec.rb
```diff
@@ -1,42 +1,58 @@
...
-## Feel free to write more tests!!
-
-# A plane currently in the airport can be requested to take off.
-#
-# No more planes can be added to the airport, if it's full.
-# It is up to you how many planes can land in the airport
-# and how that is implemented.
-#
-# If the airport is full then no planes can land
+require 'plane'
```
rodcul: hey @MollieS not 100% sure but if you are using doubles, no need to require 'plane', especially on a unit test

silvabox:
Hi Mollie, thanks for submitting your airport challenge!  What you've done shows promise, but there are a few issues with your tests.  Your code in `Airport` and `Plane` is good, and you've encapsulated weather in a module (note you have redundant ternary expressions e.g. `stormy? ? false : true` which is the same as `!stormy?`

I like that you've added your own tests and given them better descriptions, but the tests are a bit confused.  Most of your `Plane` unit tests are actually testing `Airport` and you're using doubles in feature tests where you shouldn't and not in unit tests where you should.

Have a look at the code in your `Weather` unit tests.  All of these tests are just testing the double of airport; you're testing that the double returns what you just told it to return; none of the tests actually test the weather!

I'm not too worried about that though - I think it's just a minor conceptual issue and once we've fixed that your code will be strong.  :construction:

PR #46, opened 2015-05-02
andygout: airport_challenge

andygout:


spec/feature/traffic_control_outgoing_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    allow(airport).to receive(:weather).and_return('sunny')
+    airport.receive plane
+    airport.launch plane
+    planes = airport.planes
+    expect(planes.include? plane).to be false
+  end
+
+  scenario 'a plane cannot take off when there is a storm brewing' do
+    allow(airport).to receive(:weather).and_return('stormy')
+    expect { airport.launch plane }.to raise_error 'plane cannot take off when storm brewing'
```
ptolemybarnes: Nice!!!

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+
+  attr_accessor :capacity
+
+  def initialize capacity = 10
+    @planes = []
+    @capacity = capacity
+  end
+
+  def capacity=(capacity)
+    fail 'capacity cannot be lower than number of planes' if @planes.count > capacity
```
ptolemybarnes: Consider breaking "@planes.count > capacity" out into another method.

lib/airport.rb
```diff
@@ -1,2 +1,36 @@
...
+    @capacity = capacity
+  end
+
+  def receive plane
+    fail 'plane cannot land when storm brewing' if weather == 'stormy'
```
ptolemybarnes: Likewise with these, something like #stormy?

spec/airport_spec.rb
```diff
@@ -1,42 +1,62 @@
...
+        expect { subject.receive plane }.to raise_error 'plane cannot land when storm brewing'
+      end
+
     end
+
   end
+
+  context 'airport capacity' do
+
+    it 'airports commission instructs amend to airport capacity' do
```
ptolemybarnes: Hmmm "airports commission instructs amend to airport capacity"... what does that mean?

silvabox:
Hi Andy,  thanks for your airport challenge submission.  It's a bit of a mixed bag; there's good and bad!

You've tried to extend the requirements a bit, which is good and you've introduced extra feature tests, which is also good.  But there are some fundamental problems with your code.

In `Plane`, you've defined `attr_accessor :plane` then needlessly redefined the getter `status`.  But by using `attr_accessor`, you've created a `status=` method on the plane's public interface, which means I can set the status independently of `land` and `takeoff`.

Your `Plane` tests are actually incorrect - e.g. you have
```
it 'changes its status to landed after landing' do
  expect(subject.land).to eq 'landed'
end
```
Which is not testing the status of the plane.  It is testing that the `land` method returns `landed`.  The plane's status could be anything.

Beware tricksy code.  It is often hard to read and often wrong.  For example:
```
```
In your airport tests you have `expect(subject.planes.include? plane).to be true` which would read better as `expect(subject.planes).to include plane`.

PR #47, opened 2015-05-02
Airport Challenge - Feature Tests Complete - First Submission

AlexHandy1:


lib/airport.rb
```diff
@@ -1,2 +1,50 @@
+require_relative 'plane'
+
 class Airport
+	DEFAULT_CAPACITY = 10
+	attr_accessor :capacity
+	attr_reader :stormy
+
+	def initialize
+		@capacity ||= DEFAULT_CAPACITY
```
ptolemybarnes: Not sure why you're using the lazy assignment operator there (||=).

lib/airport.rb
```diff
@@ -1,2 +1,50 @@
...
+		@planes = []
+		@stormy = false
+	end
+
+	def rand_weather_change
+		if rand(2) == 0
+			@stormy = true
```
ptolemybarnes: Try to modify the value of an instance variable via a writer method, rather than accessing it directly.

lib/airport.rb
```diff
@@ -1,2 +1,50 @@
...
+		else
+			@stormy = false
+		end
+	end
+
+	def land plane
+		fail "Airport is full" if full?
+		fail "Weather is stormy" if stormy?
+		plane.lands
+		@planes << plane
```
ptolemybarnes: Same again, access the @planes instance variable via a reader method.

lib/airport.rb
```diff
@@ -1,2 +1,50 @@
...
+	end
+
+	def release
+		fail "No planes to take off" if empty?
+		fail "Weather is stormy" if stormy?
+		@planes.pop.fly_again
```
ptolemybarnes: So this will always release the last plane in the @planes array. Is that what you want?

ptolemybarnes:
Nice attempt. You implemented the required functionality while writing the minimum code you could get away with. Two things:
 1 ) Indentation. Please make sure you indent to two spaces, rather than four. Non-conventional code styling is very distracting for other developers when they're trying to read your code.
 2 ) Read / write to instance variables via getter and setter methods, rather than by directly calling the instance variables.
Otherwise, nice job!
:checkered_flag:

PR #48, opened 2015-05-02
first commit

bagolol:
testing first commit


ptolemybarnes:
Nice attempt. You implemented the required functionality while writing the minimum code you could get away with. Two things:
1 ) Indentation. Please make sure you indent to two spaces, rather than four. Non-conventional code styling is very distracting for other developers when they're trying to read your code.
2 ) Read / write to instance variables via getter and setter methods, rather than by directly calling the instance variables.
Otherwise, nice job!
:checkered_flag:

PR #49, opened 2015-05-03
Completed third feature

chidumaga:
Planes cannot land when airport is full


lib/airport.rb
```diff
@@ -1,2 +1,24 @@
 class Airport
+
+  MAX_CAPACITY = 6
+
+  def initialize
+    @planes = []
+  end
+
+  def accept_plane plane
+    fail "Airport is full" if @planes.length == MAX_CAPACITY
```
ptolemybarnes: avoid referring to instance variables directly. Read / write to them via getter and setter methods. Though this may seem trivial now, it will become more apparent why down the line.

lib/airport.rb
```diff
@@ -1,2 +1,24 @@
...
+    fail "Stormy weather" if stormy?
+    @planes << plane
+  end
+
+  def launch_plane plane
+    fail "Stormy weather" if stormy?
+    @plane = plane
+    @plane
```
ptolemybarnes: You needn't use this @plane instance variable as it's not used outside the scope of this method.

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
...
+  def initialize
+    @status = "flying"
+  end
+
+  attr_accessor 'status'
+
+  def land airport
+    @airport = airport
+    @status = "landed"
+    @airport
```
ptolemybarnes: Again, you needn't use the @airport instance variable here.

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
...
+  end
+
+  def take_off airport
+    @status = "flying"
+    @airport = airport
```
ptolemybarnes: @airport... you know what I'm going to say.

spec/feature/grand_finale_spec.rb
```diff
@@ -8,8 +8,48 @@
 # Check when all the planes have landed that they have status "landed"
 # Once all planes are in the air again, check that they have status "flying!"
 
-feature 'Grand Finale' do
+feature "Fully functioning airport" do
+  scenario "Planes land at the appropriate airport" do
```
ptolemybarnes: This is very almost a feature test.. but not quite. It's kind of a mix of feature and unit tests.

ptolemybarnes:
:checkered_flag:

PR #51, opened 2015-05-03
J. Arad-Marks - Airport Challenge

curlygirly:
All the tests are passing  - definitely needs some refactoring and tidying up.


silvabox:
Hi Jennifer, sorry you're only getting this today.  You've done a pretty good job - there are a couple of things though that I suggest you look at:
```
In `land plane` in airport:
 def land(plane)
   fail 'Airport full' if @planes.count >= 6
   @planes << plane
   plane.land
   fail 'Stormy cannot land' if weather == 'stormy'
end
```
You are failing the code for stormy weather after the plane has already landed!

You've used `attr_accessor` for the plane status, which means the status can be set irrespective of the `take_off` and `land` methods.

In airport and plane unit tests, you test that certain methods can be called, but you don't test what the methods actually do.

Well done; good luck with these weekend's challenge!

:checkered_flag:

PR #52, opened 2015-05-03
Test

zanetton:
Latest version. I was unsure how to do random weather so just have stormy as false. Im not really sure how I got here either but seems to work?!?


zanetton:
Latest version.

PR #53, opened 2015-05-04
Partially Completed Challenge

cmanessis:
I did my best to complete the entire challenge.

Had some issues  with the start of the challenge, but as I read through boris-bikes, looked at some examples, everything started to become more clear. 

I will finish the challenge today and re-do it so to ensure I fully understand how to create feature tests, unit tests, and write the respective code.


silvabox:
Hi Chris, thanks for submitting your airport challenge.  My main concern is that your tests are a bit confused.   Your unit tests don't really test the functionality of airport and plane, and your feature tests create a an airport but it's not actually used in the tests.  Hopefully, if you look again at Boris Bikes, you might be able to see what I mean.  :construction:

PR #55, opened 2015-05-04
changes the first adopted approach

bagolol:


lib/airport.rb
```diff
@@ -1,2 +1,39 @@
...
+
+  def land plane
+    fail 'stormy weather, permission denied' if forecast == "stormy"
+    fail 'airport full, permission denied' if full?
+    hangar << plane
+    plane.landed
+  end
+
+  def take_off plane
+    fail 'stormy weather, permission denied' if forecast == "stormy"
```
ptolemybarnes: Consider refactoring "forecast == 'stormy'" out to another method.

lib/airport.rb
```diff
@@ -1,2 +1,39 @@
...
+    hangar.pop
+    plane.flying
+  end
+
+  def forecast
+    Forecast.now
```
ptolemybarnes: Nice use of the #now convention. Was that inspired by the Time class?

lib/plane.rb
```diff
@@ -1,2 +1,16 @@
 class Plane
+
+  attr_accessor :status
+
+  def initialize
+    @status = "flying"
+  end
+
+  def landed
+    @status = "landed"
```
ptolemybarnes: Good. Just be aware of what these methods are returning. Maybe it doesn't matter but if you were to call plane.landed.takeoff, you would get the error: undefined method for string class.

spec/feature/grand_finale_spec.rb
```diff
@@ -11,5 +13,36 @@
 feature 'Grand Finale' do
 
   xscenario 'all planes can land and all planes can take off'
+end
+
+feature 'a pilot flies a plane' do
```
ptolemybarnes: I always try to avoid pushing with lines of code commented out.. it makes it very messy for other developers to read.

ptolemybarnes:
Great solution! Let me know if you want further 1 to 1 feedback.
:checkered_flag:

lib/airport.rb
```diff
@@ -1,2 +1,39 @@
...
+  def land plane
+    fail 'stormy weather, permission denied' if forecast == "stormy"
+    fail 'airport full, permission denied' if full?
+    hangar << plane
+    plane.landed
+  end
+
+  def take_off plane
+    fail 'stormy weather, permission denied' if forecast == "stormy"
+    hangar.pop
```
ptolemybarnes: This is always going to remove the last plane from the hangar. Is that what you want?

PR #58, opened 2015-05-04
latest version but without weather

zanetton:
Trying to submit


ptolemybarnes:
:checkered_flag:

lib/airport.rb
```diff
@@ -1,2 +1,34 @@
...
+   def approves_take_off
+     fail 'No planes at airport' if empty?
+		 planes.pop
+     end
+
+     def stormy?
+       false
+     end
+
+private
```
ptolemybarnes: Have a look at Array#delete, which removes a specific object from an array, not just the last one.

zanetton: Yes meant to amend that! Can I just do array. Delete and it picks the right one!?! Or how to a specify the plane?

+44 774 0933309

lib/plane.rb
```diff
@@ -1,2 +1,21 @@
...
+    def flying?
+      @flying
+    end
+
+    def landed
+	     @flying = false
+    end
+
+    def take_off
+      @flying = true
```
ptolemybarnes: Good. But be careful about what your methods return. This #take_off method will return 'true'. What error would you get if you called plane.take_off.flying?

spec/plane_spec.rb
```diff
@@ -1,25 +1,37 @@
...
-  xit 'has a flying status when in the air'
+   it 'has a flying status when not at an airport'do
+   plane = Plane.new
+   expect(plane).to be_flying
+end
 
-  xit 'can take off'
+	it 'can be landed'do
+	plane =Plane.new
+  airport = Airport.new
```
ptolemybarnes: Careful with your indentation here.. makes it very hard to read.

ptolemybarnes:
:checkered_flag:

PR #59, opened 2015-05-04
Airport Challenge - passing all tests

ashleigh090990:
All tests are passing, but I feel like there's still a lot to be improved on...


ashleigh090990:
Hound is highlighting indentations as incorrect when they're fine in Sublime and on Github

tansaku:
@ashleigh090990 that's probably because of a difference between tabs and spaces ...

lib/airport.rb
```diff
@@ -1,2 +1,59 @@
...
+      fail 'Weather is stormy' 
+    elsif reached_capacity?
+      fail 'Permission denied'
+# tried using "plane.land" to change status of
+# each plane going into "planes" array to
+# 'flying', but method was 'undefined' for plane?
+# will come back to this... Works on irb?
+    else
+      plane.land
+      @planes << plane
```
ptolemybarnes: Great. One thing: pay careful attention to what your methods return.. what does #permission_to_land return? Is that what you want? Maybe it doesn't matter but it could possibly cause problems later on.

lib/airport.rb
```diff
@@ -1,2 +1,59 @@
...
+      @planes.pop
+    end
+  end
+
+  def reached_capacity?
+    planes.length >= capacity
+  end
+
+  def empty?
+    @planes = []
```
ptolemybarnes: Methods ending with a question mark return a boolean by convention. Your #empty? method here seems to empty out the @planes array. Perhaps this is something you want to do but consider a different method for this, such as #empty_planes

lib/airport.rb
```diff
@@ -1,2 +1,59 @@
...
+  end
+
+  def weather
+    %w('stormy' 'sunny').sample
```
ptolemybarnes: This works but I'm looking at it and thinking "Why not ['stormy', 'sunny'].sample?"... If I hadn't seen that way of creating an array out of strings, I would be very confused. As a rule of thumb, assume whoever's reading your codebase is stupid and needs the dumbest version possible.

lib/airport.rb
```diff
@@ -1,2 +1,59 @@
...
+  end
+
+  def stormy?
```
ptolemybarnes: Lines 52 - 58 could be replaced with: weather == "stormy"

spec/airport_spec.rb
```diff
@@ -13,16 +13,39 @@
...
-    xit 'a plane can land'
+    it 'a plane can land' do
+      plane = Plane.new
+      expect(plane).to respond_to :land
+    end
 
-    xit 'a plane can take off'
+    it 'a plane can take off' do
+      plane = Plane.new
+      expect(plane).to respond_to :take_off
```
ptolemybarnes: Here use of the the "subject" variable would dry out these tests a little.

spec/airport_spec.rb
```diff
@@ -13,16 +13,39 @@
...
+    it 'can get permission to land' do
+      expect(airport).to respond_to :permission_to_land
+    end
+
+    it 'a plane cannot land if the airport is full' do
+    #implies the airport has capacity, therefore we
+    #we need to define this
+    #if airport is at capacity
+      plane = Plane.new
+      airport.capacity.times {airport.permission_to_land plane}
```
ptolemybarnes: No magic numbers, nice!!!

spec/feature/grand_finale_spec.rb
```diff
@@ -10,6 +10,9 @@
 
 feature 'Grand Finale' do
 
-  xscenario 'all planes can land and all planes can take off'
+  scenario 'all planes can land and all planes can take off' do
+  	plane = Plane.new
+  	expect(plane).to respond_to :land && :take_off
```
ptolemybarnes: This is cool. Does it work??

ptolemybarnes:
:checkered_flag:

PR #62, opened 2015-06-13
Nicola Hughes Week 1 Challenge submit

DataMinerUK:
Doing this exercise brought up the following:

- How can I setup airport challenge such that a plane can only land and take off from an airport call? Would I directly change the `flying?`/ `landed?`  status of a plane in the `landing`/ `take_off` methods in the airport class and not have those in the plane class. So refactor the plane class and its spec file? Is it ok to delete tests?

- How can I write a single responsibility test for landing a plane and making sure it is landed and releasing a plane and making sure it is flying? Feature tests? I mainly checked it in IRB.

- empty? seems only to be for tests. I tried running the tests to ensure the plane landed is the last plane in the planes array but that meant making attr_reader :planes not private which I found out later was a bad idea.

- Private methods and when to use them. At one point I set current_weather to be private as I didn't want a user to override it. But then testing in IRB showed it needn't be private. Does this mean only an attr_accessor can be set using =?

- Setting a random weather condition means you have to run through tests several times as it may pass randomly when your weather is sunny!!! DANGEROUS

- How come I can still append to an array which is an attr_reader? If it is not private I can do:

`a = Airport.new`
`a.planes`
`=> []`
`a.planes << "p"`
`a.planes`
`=> ["p"]`

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,2 +1,7 @@
 class Airport
+
+  def take_off plane
+    plane.take_off
+  end
+
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -10,6 +10,7 @@ class Airport
   end
 
   def take_off plane
+    fail 'Too stormy to take off' if current_weather == 'stormy'
     fail 'The plane is not at the airport' unless planes.include?(plane)
     planes.delete(plane)
     plane.take_off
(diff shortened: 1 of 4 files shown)


ptolemybarnes:
:checkered_flag:

PR #68, opened 2015-06-14
The Airport Challenge... Sort of complete.

Triffanys101:


silvabox:
Thanks for your submission Lucy.  As discussed, there are quite a few issues, but your thinking is all in the right direction and with a few small changes, you wouldn't have got quite so stuck!  A great learning exercise.

:construction:

PR #75, opened 2015-06-14
airport challenge

zdajani:


lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+  end
+
+
+  private
+  attr_reader :planes
+
+  def weather
+    [:sunny, :sunny, :sunny, :stormy].sample
+  end
+
```
danldb: consider a module

ptolemybarnes:
:checkered_flag:

PR #79, opened 2015-06-14
Airport challenge

natstar93:
So.. I think I've overcomplicated things by setting 'clearing for take off' and 'taking off' as separate methods (same with landing). I will refactor in the morning and implement the weather stuff.

**Update: The commits I made this morning simplify the program and merge the methods I mentioned into one. I would like to know if my first approach is a valid way of doing it though. I merged the two methods because the second one didn't really include any decisioning and was just getting the plane to land. I decided that this wasn't really necessary since if the plane has been deemed fit to land there is no further reason why it should keep flying around waiting for another method call.


silvabox:
Well done Natalie - your code is clean and your tests all pass - I would just check that your tests cover all of the airport's responsibilities.  You've earned an architect badge!

:checkered_flag:

PR #89, opened 2015-07-24
Airport Challenge

Yorkshireman:


lib/airport.rb
```diff
@@ -1,2 +1,22 @@
 class Airport
+
+	attr_reader :planes, :traffic_controller, :capacity
+	attr_accessor :weather
+
```
danldb: * Consider if these should be exposed as public methods.
* There seems to be a lot of variables for a single class - this is a code smell that maybe your class does not have clear responsibility.

lib/airport.rb
```diff
@@ -1,2 +1,22 @@
...
+	end
+
+	def receive_plane(plane) 
+		@planes << plane
+		return
+	end
+
+	def full?
+		true if @planes.count == @capacity
+	end
```
danldb: this is not a boolean as it returns nil if not full.

lib/airport.rb
```diff
@@ -1,2 +1,22 @@
...
+	def initialize 
+		@planes = []
+		@traffic_controller = TrafficController.new(self)
+		@weather = "Glorious sunshine"
+		@capacity = 5
+	end
+
+	def receive_plane(plane) 
+		@planes << plane
+		return
```
danldb: you have an attr_reader for planes but you aren't using it. should be `planes << plane` instead of `@planes << plane`

lib/airport.rb
```diff
@@ -1,2 +1,22 @@
 class Airport
+
+	attr_reader :planes, :traffic_controller, :capacity
+	attr_accessor :weather
+
+	def initialize 
+		@planes = []
+		@traffic_controller = TrafficController.new(self)
```
danldb: Very tightly coupled with TrafficController

lib/plane.rb
```diff
@@ -1,2 +1,67 @@
...
+
+		@flying = false
+
+		airport.receive_plane(self)
+
+		@destination = nil
+		@location = airport
+		self.pilot.permission_to_land = false
+		return self
+	end
```
danldb: Vastly overcomplicated, heavily dependent on pilots, airports, weather etc.

Should be:

```
def land
  fail if not flying?
  self.flying = false
end

private
attr_writer :flying
```

lib/plane.rb
```diff
@@ -1,2 +1,67 @@
...
+		end
+		
+		plane = self.location.planes.index(self)
+		self.location.planes.slice!(plane)
+
+		@destination = destination
+		@flying = true
+		@location = nil
+		self.pilot.permission_to_take_off = false
+	end
```
danldb: Again, very much overcomplicated.

lib/traffic_controller.rb
```diff
@@ -0,0 +1,32 @@
+class TrafficController
+	attr_reader :location
+
+	def initialize(airport)
+		@location = airport
+	end
+
+	def generate_weather
+		possibilities = ["Glorious sunshine", "Glorious sunshine", "Stormy, like hell on earth"]
+		self.location.weather = possibilities.shuffle.first
```
danldb: Could be replaced with possibilities.sample

lib/traffic_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+	def grant_permission_to_take_off?(pilot)
+		weather = generate_weather
+
+		if pilot.plane.destination == self.location || self.location.weather == "Stormy, like hell on earth"
+			return false
+		else
+			return true
+		end
+	end
```
danldb: Very tightly coupled class - again, overcomplicating the problem.

spec/pilot_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+		it 'can land a plane' do
+			airport = Airport.new
+			airport.stub(:weather).and_return("Glorious sunshine")
+      		plane = Plane.new(airport)
+      		plane.pilot.request_to_land(airport)
+      		plane.pilot.land_plane(airport)
+
+      		expect(plane).to be_landed
+    	end
+
```
danldb: Fix your indentation

spec/pilot_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    		# 	airport.stub(:weather).and_return("Glorious sunshine")
+    		# 	plane.pilot.request_to_land(airport)
+    		# 	airport.stub(:weather).and_return("Stormy, like hell on earth")
+    		# 	plane.stub_chain(:pilot, :request_to_land).with(an_instance_of(Airport)) do 
+    		# 		false
+    		# 	end
+    		# 	plane.pilot.request_to_land(airport)
+    		# 	expect(plane.pilot.permission_to_land).to be false
+    		# end
+    	end
```
danldb: remove unused code

spec/plane_spec.rb
```diff
@@ -1,29 +1,231 @@
...
+      airport = Airport.new
+      airport2 = Airport.new
+      airport.stub(:weather).and_return("Glorious sunshine")
+      plane = Plane.new(airport)
+      plane.pilot.request_to_land(airport) 
+      plane.land(airport)
+      plane.pilot.request_to_take_off(airport2)
+      airport.stub(:weather).and_return("Stormy, like hell on earth")
+      expect{plane.take_off(airport2)}.to raise_error "Cannot take off when the weather is stormy"
+    end
```
danldb: No need for 2 airports

spec/plane_spec.rb
```diff
@@ -1,29 +1,231 @@
...
+  it 'cannot be flying and landed at the same time' do 
+  	expect(subject).to_not be_landed
+  end
+
+  context 'after it has taken off' do 
+    it 'is flying' do
+      airport = Airport.new
+      airport.stub(:weather).and_return("Glorious sunshine")
+      airport2 = Airport.new
+      plane = Plane.new(airport)
```
danldb: use doubles instead of real planes

spec/plane_spec.rb
```diff
@@ -1,29 +1,231 @@
...
+      airport = Airport.new
+      airport.stub(:weather).and_return("Glorious sunshine")
+      plane = Plane.new(airport)
+      plane.pilot.request_to_land(airport) 
+      plane.land(airport)
+      plane.pilot.request_to_take_off(airport2)
+      plane.take_off(airport2)
+      expect(plane.location).to be_nil      
+    end
+  end
```
danldb: tests are way too long - it is not clear what you are testing. Remember you should only be testing a single method in isolation, and it should be clear from reading the test exactly what it is you are testing.

danldb:
:construction: 

Sorry - this is not a pass.

Only address the brief - I can see you have done a huge amount of work, but you wrote code that was not asked for. Keep it simple - the more lines of code you write, the more lines of code you then have to maintain. Watch your encapsulation - make use of private methods. Feel free to arrange a 121 if you would like to discuss further.

PR #90, opened 2015-07-24
testing initial pull

Teeohbee:


lib/airport.rb
```diff
@@ -1,2 +1,42 @@
+require_relative 'plane'
+
 class Airport
+  attr_reader :planes
+  attr_reader :capacity
+
```
danldb: Do you want these attr_readers to be public?

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+  def initialize
+    @planes = []
+    @capacity = 10
+  end
+
+  def take_off
+    fail 'Airport empty. No planes to taxi' if empty?
+    fail 'Severe weather. Unable to take off' if stormy?
+    planes.pop
+  end
```
danldb: Great method

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+
+  def landing(plane)
+    fail 'Airport full. Unable to land' if full?
+    fail 'Severe weather. Unable to land' if stormy?
+    # plane.status=('grounded')
+    planes << plane
+  end
```
danldb: Again, this method is fantastic, the only thing that is missing is `plane.land`.

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+
+  def stormy?
+    chance_of_storm = rand(10)
+    if chance_of_storm <= 8
+      false
+    else
+      true
+    end
+  end
+
```
danldb: This should be extracted into a module or class, and injected into the airport object.

lib/plane.rb
```diff
@@ -1,2 +1,16 @@
...
+  end
+
+  # def land
+  #   @status = 'grounded'
+  # end
+
+  # def take_off
+  #   @status = 'flying'
+  # end
+
```
danldb: Why did you comment this out? Looks like good code to me.

lib/plane.rb
```diff
@@ -1,2 +1,16 @@
 class Plane
+  attr_accessor :status
+
+  def initialize
+    @status = 'flying'
+  end
```
danldb: `@status` is unclear, and only returns one of 2 states. If a variable can only have 2 states, it is better to use a boolean, i.e. `@flying = true`.

danldb:
:checkered_flag: 

I'm going to award a pass here - you have demonstrated excellent OOD, even though you commented some of your code out! Not sure what happened there. Let me know if you have questions.

PR #92, opened 2015-07-24
updated the readme a bit

christopheralcock:
really just doing an early pull request to see what happens, sorry if you got excited.


christopheralcock:
oh does my original pull request track changes?

README.md
```diff
@@ -1,48 +1,562 @@
+[![Build Status](https://travis-ci.org/christopheralcock/airport_challenge.svg?branch=master)](https://travis-ci.org/christopheralcock/airport_challenge)
```
silvabox: I think this is the longest airport challenge README we've ever seen... :wink:

lib/plane.rb
```diff
@@ -1,2 +1,10 @@
 class Plane
```
silvabox: Erm, I think if you read the spec, the plane should have a status of flying or not...

spec/airport_spec.rb
```diff
@@ -1,48 +1,67 @@
...
-#
-# If the airport is full then no planes can land
-
 describe Airport do
 
-  describe 'take off' do
-    xit 'instructs a plane to take off'
+  describe 'launch' do
+    it 'launches a specific plane' do
+      subject.instance_variable_set(:@weather, :stormy)
```
silvabox: You should avoid reaching into internal implementations in your tests - i.e. stubbing instance variables.  Much better to delegate everything to an attribute reader, which can be stubbed.

spec/airport_spec.rb
```diff
@@ -1,48 +1,67 @@
...
+    end
+  end
 
-    xit 'releases a plane'
+  describe 'launch' do
+    it 'raises an error when you try to launch a plane that isn\'t there' do
+      subject.instance_variable_set(:@weather, :stormy)
+      subject.instance_variable_set(:@hanger, [:plane1])
+      expect { subject.launch(:plane2) }.to raise_error
+      'That plane isn\'t here'
```
silvabox: `That plane isn't here` is not part of the expectation - it's just a separate line of code

lib/airport.rb
```diff
@@ -1,2 +1,60 @@
...
+# I appreciate that the above might not
+# strictly return 3% stormy, but close
+# enough for jazz
+  attr_reader :capacity
+# attr_reader :weather
+
+  def initialize
+    @capacity = DEFAULT_CAPACITY
+    @hanger = []
+    @weather = self.weather
```
silvabox: Who sets the `@weather` variable?  Is is set here - or in the `weather` method?

lib/airport.rb
```diff
@@ -1,2 +1,60 @@
...
+  end
+
+  def land(plane)
+    if at_capacity? && bad_weather?
```
silvabox: Try inline `if`s for better readability

spec/airport_spec.rb
```diff
@@ -1,48 +1,67 @@
...
   end
 
-  describe 'landing' do
-    xit 'instructs a plane to land'
+  describe 'launch' do
+    it 'raises error when you try to launch a plane in a storm' do
+      subject.instance_variable_set(:@weather, :stormy)
+      subject.instance_variable_set(:@hanger, [:plane])
+      expect { subject.launch(:plane) }.to raise_error
+      'It\'s too dangerous to launch!'
```
silvabox: As above - this line is not part of the expectation

lib/airport.rb
```diff
@@ -1,2 +1,60 @@
...
+      fail "There's no room here and it's too dangerous to land!"
+    end
+    if bad_weather? && !at_capacity?
```
silvabox: These conditionals are getting difficult to read.  What if there was a third condition?  You only need to report one reason for the failure at this stage.  If you want to keep track on multiple failures, there are better ways to do it, which we will see in week 4.

silvabox:
Hi Chris, thanks for submitting your weekend challenge!  There are some issues with your structure which would stop me from allowing it into production, but they're good points to learn.

Please study the common issues here:
https://github.com/makersacademy/airport_challenge/wiki/Common-issues and see how these relate to your own code.

:construction:

PR #93, opened 2015-07-24
Haroon airport challenge

haroonhassan:


lib/airport.rb
```diff
@@ -1,2 +1,45 @@
+
 class Airport
+
+DEFAULT = 1
+
```
danldb: careful of naming - this should clearly state what it is the default of. I.e. `DEFAULT_CAPACITY`.

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+	def plane_out plane # evaluate a request to take off
+
+		if self.weather_state 
+			@planes.delete(plane)
+			return true # Toewer: Yes
+		else
+			return false # Tower: No
+		end
+	end
```
danldb: You have overcomplicated this. Returning true or false implies that another class is relying on that true/false result, and goes beyond the single responsibility of this method. Google `tell don't ask - OOD`

Again,  pay attention to method naming `plane_in plane`, when read as plain english, suggests maybe putting one plane inside another one.

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+
+	def plane_size
+		@planes.size
+	end
+
+	def weather_state #
+		@weather_state = WeatherReport.new.state
+	end
+
+
```
danldb: Your class is littered with `@`. Make a private attr_reader and use that instead.

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
...
+	end
+
+	def land airport # Pilot: can I land
+		if airport.plane_in(self)
+			@flying = false # Tower: Yes
+		else 
+			@flying = true # Tower: No
+		end
+		@flying 
+	end
```
danldb: Here you can see you class is very tightly coupled with the airport - it relies on the airport to not only land the plane, but also to return true or false. This makes it doubly difficult to reuse a plane object to land on something other than an airport (maybe later we will want it to also land on an aircraft carrier?).

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
...
+
+	def take_off airport # Pilot: Can I take off
+		@flying = airport.plane_out(self) # Tower: Maybe
+	end
```
danldb: Again, you are depending on the airport to set the status, which is creating more dependency. It should be something like this:

```
def take_off
  fail 'already flying' if flying
  self.flying = true
end

private

attr_accessor :flying
```

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+		end
+	end
+
+	def plane_size
+		@planes.size
+	end
+
+	def weather_state #
+		@weather_state = WeatherReport.new.state
+	end
```
danldb: Use depency injection as we discussed today (I am not considering this when awarding pass/fail - just a note for reference :) )

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+
+	def initialize
+		@capacity = DEFAULT
+		@planes = []
+		@weather_state = true
+	end
+
+	def capacity
+		@capacity
+	end
```
danldb: You could use attr_reader :capacity - it is a shortcut, and evaluates to exactly the same as the capacity method you have written here.

danldb:
:construction: 

Good attempt, but I am afraid this is a fail. The main problem is one of encapsulation - 'tell don't ask' and use private methods. Your classes are too tightly coupled. Book a 121 if you want to discuss further.

PR #95, opened 2015-07-24
initial pull request

emily-jane:


lib/airport.rb
```diff
@@ -1,2 +1,45 @@
+require_relative 'plane'
+
 class Airport
+
+DEFAULT_CAPACITY = 20
+
+attr_reader :capacity
+
+def initialize
```
danldb: Would be nice to accept an argument for the capacity in case you want a non-default airport.

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
+require_relative 'plane'
```
danldb: You need to fix your text editor - indentation should be set to 2 spaces. This is really hard to read in certain text editors as the tabs are so long.

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+attr_reader :capacity
+
+def initialize
+	@planes = []
+	@capacity = DEFAULT_CAPACITY
+	@weather
+end
+
+def weather(n = rand(5))
+	if n == 4
```
danldb: weather method doesn't need an argument

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+		@weather = "stormy"
+	else
+		@weather = "sunny"
+	end
+end
```
danldb: Look at refactoring weather into a module or class and then injecting it when required.

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+
+def take_off
+	fail 'No planes available' if empty?
+	fail 'Stormy, can not take off' if @weather == "stormy"
```
danldb: Excellent guard clauses

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+	plane = @planes.pop
+	plane
+end
```
danldb: Great method

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+
+def land plane
+	fail 'Airport is full' if full?
+	fail 'Stormy, can not land' if @weather == "stormy"
+	@planes << plane
+end
+end
+
+public
```
danldb: did you mean to use `private` instead of public?

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+		@weather = "stormy"
+	else
+		@weather = "sunny"
+	end
+end
+
+def take_off
+	fail 'No planes available' if empty?
+	fail 'Stormy, can not take off' if @weather == "stormy"
+	plane = @planes.pop
```
danldb: instead of referring directly to the instance variable (using the `@` symbol) it is better practice to use a private attr_reader. Ideally you shouldn't see too many `@` in your code.

lib/plane.rb
```diff
@@ -1,2 +1,17 @@
...
+  end
+
+  def landing
+    @status = "landed"
+  end
+
+  def taking_off
+    @status = "flying"
+  end
 end
```
danldb: Instead of `@status = 'flying'` I would use `@flying = true`. The reason for this is that status can only have 2 states, so it is more efficient to use a boolean. This is also clearer - status can mean lots of different things, so the variable name is unclear - flying is absolutely descriptive, so your colleagues will know exactly what the variable refers to.

danldb: Use an attr_reader instead of `@` - make it private if you do not want to expose it as a public method.

spec/airport_spec.rb
```diff
@@ -13,21 +13,43 @@
...
-    xit 'instructs a plane to take off'
+  let (:plane){double(:plane)}
+
+  describe '#take_off' do
+    it 'instructs a plane to take off' do
+      expect(subject).to respond_to :take_off
+    end
+
+    it 'releases a plane' do
+      terminal = Airport.new
```
danldb: defining terminal is unnecessary - you can just use subject

spec/airport_spec.rb
```diff
@@ -13,21 +13,43 @@
...
+      plane = :plane
```
danldb: you already have a plane - you defined it in your let statement above. This line is unnecessary.

spec/airport_spec.rb
```diff
@@ -40,9 +62,20 @@
...
+        expect { terminal.take_off }.to raise_error 'Stormy, can not take off'
+      end
 
-      xit 'does not allow a plane to land'
+      it 'does not allow a plane to land' do
+        terminal = Airport.new
+        plane = :plane
+        terminal.weather(4)
+        expect { terminal.land plane }.to raise_error 'Stormy, can not land'
+      end
```
danldb: You have not stubbed out your weather here - ideally this would be solved by extracting the functionality out into a module or class.

spec/plane_spec.rb
```diff
@@ -16,14 +16,29 @@
...
+    plane = Plane.new
+    expect(plane.status).to eq 'flying'
+  end
+
+  it 'can land' do
+    expect(subject).to respond_to :landing
+  end
+
+  it 'is landed after landing' do
+    plane = Plane.new
```
danldb: use subject instead

danldb:
:checkered_flag: 

Congrats - you pass! 

Your OOD is good, and you have made a good attempt at using doubles. Pay attention to the syntax - you made errors by using unnecessary arguments, and not using attr_reader methods.

 I recommend that you refactor as per the inline notes - and please please fix your indentation, this was really hard for me to mark as the code was so broken up. 

Well done :)

PR #101, opened 2015-07-24
Initialized plane and expected it to be flying when created

RichardCharman:


README.md
```diff
@@ -1,70 +1,20 @@
 Airport Challenge
 =================
```
sjmog: Getting Started please!

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
+require_relative 'plane'
 class Airport
+  DEFAULT_CAPACITY = 5
```

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+  
+  attr_accessor :capacity
```
sjmog: `attr_reader` for Encapsulation purposes

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
+require_relative 'plane'
 class Airport
+  DEFAULT_CAPACITY = 5
+  
+  attr_accessor :capacity 
+  
+  def initialize(capacity= DEFAULT_CAPACITY)
```
sjmog: Nice DEFAULT setting - you have a random space after `capacity` though

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+    @planes = []
+    @capacity = capacity
+    @weather = "Sunny"
```
sjmog: Interesting to see the Weather as part of an Airport.

Single Responsibility --> this airport "manages planes **and** randomises the weather". It has two responsibilities.

Split `Weather` out into a `Weather` module. You can then `include` that module in this class.

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+  end
+  
+  def weather
+    @weather
```
sjmog: Encapsulation!

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+  end
+
+  def take_off plane
+    raise "There are no planes at #{self.class.name}" if empty?
+    raise "Weather is stormy. You may not take off from #{self.class.name}" if weather == "Stormy"
+    @planes.delete (plane)
```
sjmog: Avoid exposing instance variables like this

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+    @weather = "Sunny"
+  end
+  
+  def weather
+    @weather
+  end
+
+  def take_off plane
+    raise "There are no planes at #{self.class.name}" if empty?
+    raise "Weather is stormy. You may not take off from #{self.class.name}" if weather == "Stormy"
```
sjmog: :+1: guard conditions

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+  def land plane
+    raise "#{self.class.name} is full" if full?
+    raise "Weather is stormy. You may not land at #{self.class.name}" if weather == "Stormy"
+    @planes << plane
+    plane.land
+  end
+  
+  def random_weather(skies=rand(7))
+    if skies == 3
+      @weather = 'Stormy'
```
sjmog: Where possible use symbols instead of strings. They have lower memory overhead as they are immutable.

```ruby
if skies == 3
  weather = :stormy
#etc.
```

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+      @weather = 'Sunny'
+    end
+  end
+  
+  private
+  
+  attr_accessor :planes
+  
+  def full?
+    planes.count >= capacity
```
sjmog: Lovely method as it returns t/f v clearly

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+    if skies == 3
+      @weather = 'Stormy'
+    else
+      @weather = 'Sunny'
+    end
+  end
+  
+  private
+  
+  attr_accessor :planes
```
sjmog: :+1: for putting this in the private interface

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+  
+  def full?
+    planes.count >= capacity
+  end
+  
+  def empty?
+    planes.count == 0
```
sjmog: Could use `planes.empty?`

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
 class Plane
+  
+  def initialize
+    @flying = true
+    @landed = false
```
sjmog: This is possibly up to interpretation, @danldb - I suggest just using `@flying`, then defining methods:

```ruby
def flying?
  @flying
end

def landed?
  !flying?
end
```

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
 describe Airport do
 
   describe 'take off' do
-    xit 'instructs a plane to take off'
-
-    xit 'releases a plane'
+# xit 'instructs a plane to take off'
```
sjmog: You can remove comments - and should! Comments are 'code smells' - they indicate when something ain't right. Code should read for itself (not so much the problem here)

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
+    it 'instructs a plane to take off' do
+      expect(subject).to respond_to :take_off
+    end
+# xit 'releases a plane'
+    it 'releases a plane' do
+      plane = Plane.new
+      subject.land plane
+      expect(plane.flying?).to eq false
```
sjmog: `expect(plane).to be_flying` --> RSpec predicate syntax

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
-    xit 'instructs a plane to take off'
-
-    xit 'releases a plane'
+# xit 'instructs a plane to take off'
+    it 'instructs a plane to take off' do
+      expect(subject).to respond_to :take_off
+    end
+# xit 'releases a plane'
+    it 'releases a plane' do
+      plane = Plane.new
```
sjmog: This should be a double! `Plane` does not belong in the `Airport` spec.

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
+      subject.land plane
+      expect(plane.flying?).to eq false
+      subject.take_off plane
+      expect(plane.flying?).to eq true
+      expect(@planes).to eq nil
```
sjmog: Where is `@planes` coming from? It isn't defined anywhere in this tes.

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
-    xit 'instructs a plane to land'
-
-    xit 'receives a plane'
+# xit 'instructs a plane to land'
+    it 'instructs a plane to land' do
+      expect(subject).to respond_to(:land).with(1).argument
+    end
+#xit 'receives a plane'
+    it 'receives a plane' do
+      plane = Plane.new
```
sjmog: Double!

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
+      subject.land plane
+      expect(plane.landed?).to eq true
```
sjmog: cf. Predicate syntax

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
+    end
   end
 
   describe 'traffic control' do
     context 'when airport is full' do
```
sjmog: Excellent use of context

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
-      xit 'does not allow a plane to land'
+# xit 'does not allow a plane to land'
+      it 'does not allow a plane to land' do
+        plane = Plane.new
+        subject.capacity.times {subject.land(plane)}
```
sjmog: I love this! `capacity.times` FTW

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
+      expect(plane.landed?).to eq true
+    end
   end
 
   describe 'traffic control' do
     context 'when airport is full' do
-      xit 'does not allow a plane to land'
+# xit 'does not allow a plane to land'
+      it 'does not allow a plane to land' do
+        plane = Plane.new
```
sjmog: Dooooble

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
+        subject.capacity.times {subject.land(plane)}
+        expect{ subject.land(plane) }.to raise_error "#{subject.class.name} is full" if :full?
```
sjmog: Why would you check if full again? You've already landed (capacity) planes.

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
 
     context 'when weather conditions are stormy' do
-      xit 'does not allow a plane to take off'
-
-      xit 'does not allow a plane to land'
+# xit 'does not allow a plane to take off'
+      it 'does not allow a plane to take off' do
+      plane = Plane.new
+      subject.land plane
+      subject.random_weather(3)
```
sjmog: This is probably not the way you want to stub randomness - you've built your method so it accepts a parameter especially so you can make it 'not random' in this test. Don't build for tests!

You want to _stub_ this method so it always returns `:stormy`. Check out method stubs.

spec/airport_spec.rb
```diff
@@ -14,35 +15,69 @@
...
+      expect{ subject.take_off plane }.to raise_error "Weather is stormy. You may not take off from #{subject.class.name}"
```
sjmog: If you do want to use `subject.class` you can use `described_class` in RSpec tests.

spec/plane_spec.rb
```diff
@@ -16,14 +16,40 @@
...
-  xit 'is landed after landing'
-
-  xit 'can take off'
-
-  xit 'is flying after take off'
-
+# xit 'is flying when created'
+  it 'is expected to be flying when created' do
+    plane = Plane.new
+    expect(plane).to respond_to :flying?
```
sjmog: This doesn't tell me that it's flying! It just tells me it knows if it's flying or not.

`expect(plane).to be_flying`

spec/plane_spec.rb
```diff
@@ -16,14 +16,40 @@
...
-
-  xit 'is landed after landing'
-
-  xit 'can take off'
-
-  xit 'is flying after take off'
-
+# xit 'is flying when created'
+  it 'is expected to be flying when created' do
+    plane = Plane.new
```
sjmog: Use `subject`

spec/plane_spec.rb
```diff
@@ -16,14 +16,40 @@
...
+
+# xit 'can land'
+  it 'can land' do
+    expect(subject).to respond_to :land
+  end
+
+# xit 'is landed after landing'
+  it 'is landed after landing' do
+    plane = Plane.new
+    expect(plane).to respond_to :landed?
```
sjmog: Same thing as above - this doesn't tell me that it's landed, only that it could be.

spec/plane_spec.rb
```diff
@@ -16,14 +16,40 @@
...
+# expect(plane.landed).to eq false
+  end
+  
+  it 'cannot be flying and landed' do
+    plane = Plane.new
+    plane.land
+    expect(plane.landed?).to eq true
```
sjmog: This test doesn't test for this.

You also need `expect(subject).not_to be_flying` (including a refactor of line 40)

spec/plane_spec.rb
```diff
@@ -16,14 +16,40 @@
...
+  it 'can take off' do
+    expect(subject).to respond_to :take_off
+  end
+    
+# xit 'is flying after take off'
+  it 'is flying after take off' do
+    plane = Plane.new
+    plane.take_off
+# expect(plane.landed?).to eq false
+    expect(plane.flying?).to eq true
```

sjmog:
This is good stuff but you need to read up on Doubles, Mocks and Stubs and use them.

Some shorthand RSpec syntax too.

I'll pass this now, but please do refactor using my comments and let me know when you're done.

:checkered_flag:

lib/plane.rb
danldb: Totally agree - a single instance variable that returns a boolean, and with two getters (one getter method to cover each state).

PR #104, opened 2015-07-26
pull request

winnieau:
requesting to pull changes on airport challenge


spec/airport_spec.rb
```diff
@@ -12,22 +12,38 @@
 # If the airport is full then no planes can land
 
 describe Airport do
+  it { is_expected.to respond_to :dock.with(1).argument }
+  it { is_expected.to respond_to :take_off }
```
sjmog: Good 1-liners and super clear

spec/airport_spec.rb
```diff
@@ -12,22 +12,38 @@
...
 
-  describe 'take off' do
-    xit 'instructs a plane to take off'
+  describe '#take_off' do
```
sjmog: Good use of hash syntax for describing instance methods

spec/airport_spec.rb
```diff
@@ -12,22 +12,38 @@
...
+    it 'instructs a plane to take off'
+    subject.take_off
+    expect(@plane).to be_landed
+  end
```
sjmog: Slightly weird indentation - check out Parm's amazing alignment gem

spec/airport_spec.rb
```diff
@@ -12,22 +12,38 @@
...
 describe Airport do
+  it { is_expected.to respond_to :dock.with(1).argument }
+  it { is_expected.to respond_to :take_off }
 
-  describe 'take off' do
-    xit 'instructs a plane to take off'
+  describe '#take_off' do
+    it 'instructs a plane to take off'
+    subject.take_off
+    expect(@plane).to be_landed
```
sjmog: What is @plane?

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
+require_relative 'plane'
 class Airport
+  DEFAULT_CAPACITY = 20
```

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
...
+    attr_accessor :capacity
```
sjmog: This can just be a `reader`

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+
+  def initialize
+    @planes = []
+    @capacity = DEFAULT_CAPACITY
```
sjmog: I think if we always set `capacity` to this constant, there isn't really any point in having a `capacity` attribute at all, because we could just use the constant `DEFAULT_CAPACITY`. If you DO want the user to be able to change the capacity, you should offer an argument to `initialize` like this:

```ruby
def initialize(capacity)
  @capacity = capacity
end
```

Then if you want to add a default as well:

```ruby
def initialize(capacity=DEFAULT_CAPACITY) # This will let you do airport = Airport.new without any args and it'll default to DEFAULT_CAPACITY
  @capacity = capacity
end
```

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+  end
+
+  def take_off
+    fail 'Plane is already flying' if landed_planes.empty?
```
sjmog: For extensibility: this error should say `Cannot takeoff: airport empty`

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+    @planes.delete landed_planes.pop
```
sjmog: Avoid using instance variables directly in methods. Use a reader (like the one you have) to do it instead.

sjmog: This is a really clever method though!

sjmog: Kinda unnecessary b/c the Airport `@planes` attribute shouldn't be storing flying planes.

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+    fail 'Inclement weather' if bad_weather
```
sjmog: This condition needs to go above as a guard condition b/c at the moment if it's bad weather a plane just disappears

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+  end
+
+  def dock plane
+    fail 'Airport is full' if full?
+    fail 'Inclement weather' if bad_weather
```
sjmog: So good! Guard conditions really neat.

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+    @planes << plane
```
sjmog: You _could_ use a reader here because you are using the shovel operator `<<` which modifies something in-place.

```ruby
planes << plane
```

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+
+private
+attr_reader :planes
+
+def flying_planes
+  @planes.select { | plane | plane.flying?}
+
+def landed_planes
+  @planes.reject { | plane | plane.flying?}
+end
```
sjmog: I don't think you need these.

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+
+def full?
+  @planes.count >= capacity
```
sjmog: :+1: (again you could use a `reader` instead of the `@instance variable`)

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+end
+
+def bad_weather?
+  @planes.reject { |plane | plane.stormy?}
+end
```
sjmog: I think this needs a bit of re-working because it doesn't make too much sense that `plane`s would be `stormy`

lib/plane.rb
```diff
@@ -1,2 +1,38 @@
...
 class Plane
+
+  def initialize
+    @flying = false
+    @weather = []
+  end
+
+  def can_land?
+    @flying && self.stormy? == false
+  end
```
sjmog: Maybe put the weather in the Airport.

lib/plane.rb
```diff
@@ -1,2 +1,38 @@
...
+
+  def landed?
+    !@flying
```
sjmog: :+1: but could use a reader

lib/plane.rb
```diff
@@ -1,2 +1,38 @@
...
+  end
+
+  def can_take_off?
+    !@flying && self.stormy? == false
```
sjmog: Could do `!flying && !self.stormy?`

lib/plane.rb
```diff
@@ -1,2 +1,38 @@
...
+  end
+
+  def reported_flying
```
sjmog: Method naming is very clear and so your code is quite easy to understand

lib/plane.rb
```diff
@@ -1,2 +1,38 @@
...
+    @flying = true
+  end
+
+  def flying?
+    @flying
+  end
```
sjmog: You could use `alias_method` which would do this `alias_method flying?, flying

lib/plane.rb
```diff
@@ -1,2 +1,38 @@
...
+
+  def stormy?
+    if @weather == 0
+      false
+    else
+      true
+    end
+  end
```
sjmog: This entire method could be condensed to
```ruby
def stormy?
  @weather != 0
end
```

sjmog:
I love your method naming, your awareness of the domain (e.g. Weather into a separate class). I feel like the lack of RSpec was a real hinderance to getting this great so once we've fixed that, can you try submitting again please?

:construction:

PR #106, opened 2015-07-26
Initial commit

ajitsy:


ajitsy:
I've not fully completed the challenge. I've yet to incorporate plane.rb with airport.rb - a key link is the fact that when a plane is created it should be flying, whereas my program assumes that when a plane is created it is stored at the airport. You'll notice that the plane.rb class is essentially redundant because I've not incorporated it within the airport class. Time allowing I'm confident I would have completed the challenge but unfortunately had to leave starting the challenge late.

ajitsy:
For some reason I can't get the "when airport is full" > "does not allow a plane to land" test working. I've submitted with that test pending.

spec/airport_spec.rb
```diff
@@ -13,36 +13,61 @@
 
 describe Airport do
 
-  describe 'take off' do
-    xit 'instructs a plane to take off'
-
-    xit 'releases a plane'
+  it 'has a default capacity' do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
```
sjmog: :+1:

spec/airport_spec.rb
```diff
@@ -13,36 +13,61 @@
...
   end
 
+  describe '#take_off?' do
```
sjmog: Great use use of hash syntax for methods

spec/airport_spec.rb
```diff
@@ -13,36 +13,61 @@
...
+    it 'instructs a plane to take off' do
+      expect(subject).to respond_to :take_off?
+    end
```
sjmog: Could use 1-liner for this

spec/airport_spec.rb
```diff
@@ -13,36 +13,61 @@
...
+
+
+    it 'releases a plane' do
+      expect(subject.instance_variable_get(:@airport)).to eql([])
```
sjmog: :(

spec/airport_spec.rb
```diff
@@ -13,36 +13,61 @@
...
+      expect { subject.take_off?}.to raise_error 'Can not take off'
```
sjmog: Refactor this into several tests using `context` keyword.

spec/airport_spec.rb
```diff
@@ -13,36 +13,61 @@
...
+
   describe 'landing' do
-    xit 'instructs a plane to land'
+    it 'instructs a plane to land' do
+      expect(subject).to respond_to :landing?
+    end
 
-    xit 'receives a plane'
+    it 'receives a plane' do
+      expect(subject.stub(:full?))
```
sjmog: This doesn't need an expect around it. If you want to stub something you can just go ahead and stub it like this:
```ruby
allow(subject).to receive(:full?).and_return(true)
```

spec/airport_spec.rb
```diff
@@ -13,36 +13,61 @@
...
+      expect { subject.landing?}.to raise_error 'Can not land plane'
```
sjmog: This test describes only the scenario where the airport is full, but claims to test `receives a plane` in totality.

```ruby
context 'airport full' do
  it 'raises an error when landing a plane' do
    allow(subject).to receive(:full?).and_return(true)
    expect { subject.land_plane(plane) }.to raise_error "Cannot land: Airport full"
  end
end
```

spec/airport_spec.rb
```diff
@@ -13,36 +13,61 @@
...
+    end
   end
 
   describe 'traffic control' do
-    context 'when airport is full' do
-      xit 'does not allow a plane to land'
-    end
+      context 'when airport is full' do
+        xit 'does not allow a plane to land' do
+          subject.capacity.times { subject.landing? Plane.new }
```
sjmog: OK, but I don't know why `landing?` is a predicate method when it's designed to do more than just answer `true` or `false`.

I also don't like seeing the `Plane` class in your `Airport` spec. You should mock the `plane` instance with a double.

spec/airport_spec.rb
```diff
@@ -13,36 +13,60 @@
...
+      #
+      # This will require stubbing to stop the random return of the weather.
+      # If the airport has a weather condition of stormy,
+      # the plane can not land, and must not be in the airport
 
-    context 'when weather conditions are stormy' do
-      xit 'does not allow a plane to take off'
+      context 'when weather conditions are stormy' do
+        it 'does not allow a plane to take off' do
+          allow(subject).to receive(:weather?) {"stormy"}
```
sjmog: Perfect way to stub it!

BUT

Why does the Airport also handle the weather? You should split the weather methods out to a `Weather` module.

Other thing - no reason for a predicate here as you're returning a string.

Finally, prefer symbols (like `:stormy`) to strings. They're smaller, immutable, take less space in memory, neater, etc.

spec/airport_spec.rb
```diff
@@ -13,36 +13,60 @@
...
+          expect { subject.landing?}.to raise_error 'Can not land plane'
```
sjmog: :+1:

spec/airport_spec.rb
```diff
@@ -13,36 +13,58 @@
...
+        end
 
-      xit 'does not allow a plane to land'
+        it 'does not allow a plane to land' do
+          allow(subject).to receive(:weather?) {"stormy"}
+          expect { subject.take_off?}.to raise_error 'Can not take off'
```
sjmog: :+1:

spec/plane_spec.rb
```diff
@@ -16,14 +16,24 @@
 
 describe Plane do
 
-  xit 'is flying when created'
+  it 'is flying when created' do
+    expect(subject).to be_flying
```

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
 class Plane
+
+  def initialize
+    @flying = true
+  end
+
+  def can_take_off?
+    not @flying
```
sjmog: Generally we should avoid long-term logicals in statements like this. A bang would be better.

Long term logicals are better in long-term conditionals:

```ruby
if not @flying
  true
else
  false
end
```
(as a contrived example)

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
...
+  end
+
+  def flying?
+    @flying = true
+  end
```
sjmog: This is a setter but written as a predicate

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
...
+
+  def can_land?
+    @flying
+  end
+
+  def taken_off?
+    @flying
+  end
+
```
sjmog: This whole class could be
```ruby
class Plane
  attr_reader :flying
  def initialize
    @flying = true
  end

  def landed?
    !flying
  end
end
```

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
+require_relative 'plane'
```
sjmog: :+1: great use of `require_relative`

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+
 class Airport
+
+  DEFAULT_CAPACITY = 10
```
sjmog: :+1: good use of constant

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+
+  attr_accessor :capacity
+
+    def initialize(capacity= DEFAULT_CAPACITY)
```
sjmog: Love this default value

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
+require_relative 'plane'
+
 class Airport
+
+  DEFAULT_CAPACITY = 10
+
+  attr_accessor :capacity
```
sjmog: I think this could be an `attr_reader` to avoid over-exposing stuff.

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+
+    def initialize(capacity= DEFAULT_CAPACITY)
+      @airport = []
```
sjmog: Probably call this `@planes` instead of `@airport` for two reasons:

1) It's bad practice to name an instance variable the same thing as the class that contains that instance variable (`airport.airport`...!)
2) This is going to store a bunch of `Plane` instances, so why not let me know that when I read through this?

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+      @capacity = capacity
+    end
+
+  def take_off?
+    fail 'Can not take off' if @weather = "stormy"
```
sjmog: Don't want to access instance variables directly: if you want to use `attr_reader` but don't want it to be a public reader, you can do this:

```ruby
private

attr_reader :weather
```

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+    fail 'Can not take off' if empty?
```
sjmog: Love the use of guard conditions: fail first!

sjmog: Lovely line

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+
+    if plane.can_take_off?
+      plane.taken_off? == true
```
sjmog: I think you meant to set the plane `@flying` attribute to `true` here.

```ruby
if plane.can_take_off?
  plane.flying = true
  planes.delete(plane)
end
```

sjmog: There is also no `plane` here

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+     fail 'Can not land plane' if full?
+
+     if plane.can_land?
+       airport << plane
+       plane.has_landed? == true
+     end
+   end
+
+  def weather?
+    @weather = ["stormy", "sunny", "sunny", "sunny", "sunny", "sunny"].sample
```
sjmog: Think about being DRY.

```ruby
rand(6) == 1 ? :stormy : :sunny
```

sjmog:
I think there's a lot to like here. It feels like you're grasping some concepts but not necessarily how they all fit together. You've got stubs and mocks which are right sometimes but not right other times, for example.

I think a big refactor is in order (but I'd suggest actually starting again - you'll be surprised how fast you can do it).

:construction:

PR #107, opened 2015-07-26
work so far

Aftermaths:


lib/airport.rb
```diff
@@ -1,2 +1,55 @@
...
+# So that I can avoid collisions
+# I want to be able to prevent airplanes landing when the airport if full
+
+# As an air traffic controller
+# So that I can avoid accidents
+# I want to be able to prevent airplanes landing or taking off when the weather is stormy
+
 class Airport
+  D_CAPACITY = 50
+
```
danldb: Don't abbreviate unnecessarily, `DEFAULT_CAPACITY` removes all doubt as to what you are referring to.

lib/airport.rb
```diff
@@ -1,2 +1,55 @@
...
+  attr_accessor :planes
+  attr_accessor :capacity
+  attr_accessor :weather
+
```
danldb: poor encapsulation - these accessors should be private

lib/airport.rb
```diff
@@ -1,2 +1,55 @@
...
+  def initialize(capacity = D_CAPACITY, weather = 'sunny')
+    @capacity = capacity
+    @planes = []
+    @weather = weather
+  end
+
+  def take_off
+    fail 'Weather is bad for take off' if weather == 'stormy'
+    planes.last.off
+    @planes.pop
```
danldb: You use the attr_reader on line 33 and then forget to use it on line 34. Ideally, if you have an attr_accessor, you should never see `@` in your code.

lib/airport.rb
```diff
@@ -1,2 +1,55 @@
...
+  end
+
+  def empty?
+    @planes.empty?
+  end
+
+  def full?
+    @planes.count >= capacity
+  end
+
```
danldb: empty? and full? should probably be private.

lib/airport.rb
```diff
@@ -1,2 +1,55 @@
...
+  def change_weather(num = rand(100))
+    num <= 30 ? @weather = 'stormy' : @weather = 'sunny'
+  end
 end
```
danldb: Extract this into a module or class and inject it.

lib/plane.rb
```diff
@@ -1,2 +1,34 @@
...
+# I want to be able to prevent airplanes landing when the airport if full
+
+# As an air traffic controller
+# So that I can avoid accidents
+# I want to be able to prevent airplanes landing or taking off when the weather is stormy
+
 class Plane
+  def initialize(status = 'flying')
+    @status = status
+  end
```
danldb: use a boolean instead of `@status`. I have no idea what you mean by status - it would be much clearer to use `@flying = true`.

lib/plane.rb
```diff
@@ -1,2 +1,34 @@
...
+
+  def land
+    @status = 'landed'
+  end
+
+  def off
+    @status = 'flying'
+  end
```
danldb: Be wary of method names like `off`. I don't know whether this switched the plane off, makes the plane take off, or maybe off is an abbreviation... would be much clearer to call this `take_off`.

spec/airport_spec.rb
```diff
@@ -40,9 +71,18 @@
...
 
-      xit 'does not allow a plane to land'
+      it 'does not allow a plane to land' do
+        plane = double :plane, land: 'landed'
+        subject.change_weather(29)
+        expect { subject.landing plane }.to raise_error('Weather is bad for landing')
+      end
     end
   end
 end
```
danldb: didn't _really_ stub out the weather, but your solution was ingenious. We were hoping to see you use `allow`.

danldb:
:checkered_flag: 

Very good! Nice OOD. I have made a few comments, and would recommend a refactor at the end of the course.

PR #109, opened 2015-07-26
Completed airport challenge

tealpaintedduck:


lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+    elsif weather == "stormy"
+      raise "Denied, bad weather."
+    else
+      plane.landing
+      planes << plane
+    end
+  end
+
+  def take_off
+    raise "No planes here at the mo." unless !empty?
```
silvabox: Avoid testing negatives - prefer `if empty?`

lib/plane.rb
```diff
@@ -1,2 +1,18 @@
 class Plane
-end
+  attr_reader :status
+
+  def initialize
+    @status = "flying"
+  end
+
+  def landing
```
silvabox: command methods should be verbal - i.e. `land` rather than `landing` and `take_off` rather than `taking_off`

spec/airport_spec.rb
```diff
@@ -13,21 +13,60 @@
...
+  before(:each) do
+    allow(subject).to receive(:weather) { "sunny" }
+  end
+
   describe 'take off' do
-    xit 'instructs a plane to take off'
 
-    xit 'releases a plane'
+    it 'instructs a plane to take off' do
+      expect(subject).to respond_to(:take_off)
```
silvabox: The test does not match the description.  You are not testing that the plane is told to take off - only that there is a take off method

spec/plane_spec.rb
```diff
@@ -15,15 +15,28 @@
...
-  xit 'can land'
-
-  xit 'is landed after landing'
-
-  xit 'can take off'
-
-  xit 'is flying after take off'
-
+  it "can't land unless flying" do
+    subject.instance_variable_set(:@status, "landed")
```
silvabox: It should be possible to test a plane without exposing it's internal implementation (e.g. instance variables).

silvabox:
Hi Sarah, thanks for submitting your airport challenge!  You've done well - but there are a few minor points to address.

Please study the common issues here: https://github.com/makersacademy/airport_challenge/wiki/Common-issues and see how these relate to your own code.

:checkered_flag:

PR #111, opened 2015-07-26
Final Submission

dregules:


spec/plane_spec.rb
```diff
@@ -15,15 +15,45 @@
...
-
-  xit 'can take off'
-
-  xit 'is flying after take off'
-
+  let(:airport) { double(:airport, { :full? => false }) }
+
+  it 'is flying when created' do
+    expect(subject.landed).to be false
+  end
```

spec/plane_spec.rb
```diff
@@ -15,15 +15,45 @@
...
+
+  describe ' #land' do
+    it 'lands at the airport' do
+      subject.land(airport)
```
sjmog: This suggests that "The Plane lands the Airport", which is a bit weird.
You'll run into dependency issues this way round. It makes more sense to have "Airport lands the Plane".

In general, you should [depend on abstract classes](http://www.informit.com/articles/article.aspx?p=1946176&seqNum=3).

The general rule of thumb is to _depend on things that change less often than you do_ . So here, `Plane` is much less likely to change that `Airport` in future, so `Airport` should depend on `Plane`.

spec/plane_spec.rb
```diff
@@ -15,15 +15,45 @@
...
+      expect(subject.airport).to be airport
+    end
+
+    it 'is landed after landing' do
+      subject.land(airport)
+      expect(subject.landed).to be true
+    end
+  end
```

spec/plane_spec.rb
```diff
@@ -15,15 +15,45 @@
...
+    end
+
+    it 'removes airport trace' do
+      subject.land(airport)
+      subject.take_off
+      expect(subject.airport).to be nil
+    end
+
+    it 'cannot take off if flying' do
+      subject.landed = false
```
sjmog: This is `false` when initialized anyway

spec/plane_spec.rb
```diff
@@ -15,15 +15,45 @@
...
+      expect { subject.take_off }.to raise_error "Plane cannot take off, it's already flying!"
```

spec/airport_spec.rb
```diff
@@ -12,37 +12,71 @@
 # If the airport is full then no planes can land
 
 describe Airport do
+  let (:flying_plane) { double(:flying_plane, { :landed => false }) }
+  let (:landed_plane) { double(:landed_plane, { :landed => true }) }
+
+  it 'has a plane' do
+    expect(subject.planes).not_to be_empty
+  end
```
sjmog: I would expect this class to be empty when initialized.

spec/airport_spec.rb
```diff
@@ -12,37 +12,71 @@
...
 
   describe 'take off' do
-    xit 'instructs a plane to take off'
+    it 'instructs a plane to take off' do
+      subject.weather = 'sunny'
+      subject.take_off_order
+      expect(subject).to respond_to(:take_off_order)
+    end
```
sjmog: :+1:

spec/airport_spec.rb
```diff
@@ -12,37 +12,71 @@
...
   end
 
   describe 'landing' do
-    xit 'instructs a plane to land'
+    it 'instructs a plane to land' do
+      subject.weather = 'sunny'
+      subject.landing_order(flying_plane)
+      expect(subject.planes).to include flying_plane
+    end
```
sjmog: Slightly confusing naming here

spec/airport_spec.rb
```diff
@@ -12,37 +12,71 @@
...
+      subject.weather = 'sunny'
+      expect(flying_plane).to receive(:landed=).with(true)
+      subject.landing_order(flying_plane)
+      subject.receive(flying_plane)
+    end
 
-    xit 'receives a plane'
+    it 'fails to receive a plane without instruction' do
+      expect { subject.receive flying_plane }.to raise_error "Plane not instructed to land"
+    end
```
sjmog: :+1: but I don't think this is the _simplest_ solution. Planes landing at airports could do it in one if you did:

```ruby
airport = Airport.new
plane = Plane.new # => flying == true (or landed == false)
airport.land(plane) # => tells the plane to change its flying or landed status
```

spec/airport_spec.rb
```diff
@@ -12,37 +12,71 @@
...
-    # The weather must be random and only have two states "sunny" or "stormy".
-    # Try and take off a plane, but if the weather is stormy,
-    # the plane can not take off and must remain in the airport.
-    #
-    # This will require stubbing to stop the random return of the weather.
-    # If the airport has a weather condition of stormy,
-    # the plane can not land, and must not be in the airport
+      it 'responds to full' do
+        expect(subject).to respond_to :full?
+      end
```
sjmog: You could use RSpec 1-liner syntax to do this.

spec/airport_spec.rb
```diff
@@ -12,37 +12,71 @@
...
+
+      it 'does not allow a plane to land' do
+        subject.weather = 'sunny'
+        expect { subject.capacity.times { subject.landing_order(flying_plane) } }.to raise_error "Airport is full"
```
sjmog: Beautiful

spec/airport_spec.rb
```diff
@@ -12,37 +12,71 @@
...
+      end
+    end
 
     context 'when weather conditions are stormy' do
-      xit 'does not allow a plane to take off'
+      it 'responds to weather' do
+        expect(subject).to respond_to :weather
+      end
```
sjmog: Probably unnecessary: your line 72 addresses this

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
...
 class Plane
+
+  attr_accessor :landed, :airport
+
+  def initialize state = false
+    @landed = state
+  end
+
+  def land airport
+    @landed = true
```
sjmog: You should avoid accessing state directly, and either _mutate_ state via an `attr_reader` or set state using an `attr_writer`. For instance:

```ruby
def land
  @landed = true
end
```

should become:

```ruby
attr_accessor :landed
def land
  landed = true
end
```

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
...
+    @airport = airport
+  end
+
+  def take_off
+    raise "Plane cannot take off, it's already flying!" if !landed
```
sjmog: Great use of guard conditions :+1:

lib/airport.rb
```diff
@@ -1,2 +1,47 @@
+require_relative 'plane'
+
 class Airport
+
+  attr_accessor :planes, :capacity, :weather
+
+  def initialize capacity = 5
+    @capacity = capacity
+    @weather = self.weather
+    @planes = Array.[](Plane.new true)
```
sjmog: You are directly depending on `Plane` here which is unnecessary.

sjmog: Plus, I don't really understand what this line does.

lib/airport.rb
```diff
@@ -1,2 +1,47 @@
...
+  end
+
+  def take_off_order
+    fail "Bad Weather - cannot take off for now" if @weather == 'stormy'
```
sjmog: You have an `attr_accessor` for weather, so don't need to directly access the instance variable

lib/airport.rb
```diff
@@ -1,2 +1,47 @@
...
+    @planes.last.take_off
+  end
+
+  def landing_order plane
+    if full?
```

lib/airport.rb
```diff
@@ -1,2 +1,47 @@
...
+      fail "Airport is full"
+    elsif @weather == 'stormy'
+      fail 'Bad Weather - cannot land for now'
+    else
+      @planes << plane
+    end
+  end
```
sjmog: This very long conditional (using `elsif` - horror) could be refactored to use guard conditions.
```ruby
fail "Airport is full" if full?
fail "Bad Weather - cannot land for now" if weather == 'stormy'
planes << plane
```

lib/airport.rb
```diff
@@ -1,2 +1,47 @@
...
+
+  def receive plane
+    if planes.include?(plane) != true
```
sjmog: This is a very complex line and could be much clearer. `!= true` is a strong code smell to me.

lib/airport.rb
```diff
@@ -1,2 +1,47 @@
...
+      fail "Plane not instructed to land"
+    else
+      @planes.last.landed = true
+    end
+  end
```
sjmog: See comment about guard conditions above.

lib/airport.rb
```diff
@@ -1,2 +1,47 @@
...
+
+  def release
+    @planes.pop
+  end
+
+  def full?
+    planes.length >= capacity
+  end
```
sjmog: Love this method. But it could be a private method

lib/airport.rb
```diff
@@ -1,2 +1,47 @@
...
+
+  def weather
+    @weather = ['stormy', 'sunny'].sample
+  end
```
sjmog: I don't think `weather` belongs in `Airport`. This says "The Airport can change the weather", which breaks the Single Responsibility Principle.

Instead, extract this method to a `Weather` module and either `include` it in `Airport` or call it directly:

```ruby
module Weather
  def set_weather
    rand(5) > 3 ? :stormy : :sunny
  end
end
```
which is then called with

```ruby
Weather.set_weather # => gives a random weather scenario as a symbol
```

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
+require_relative 'airport'
+
 class Plane
+
+  attr_accessor :landed, :airport
+
+  def initialize state = false
+    @landed = state
+  end
```
sjmog: `state` is not a very clear variable name. Most of your other variable names are great though :)

sjmog:
@dregules this is a good challenge! I think you made things harder for yourself because of your dependency direction (`Plane` depending on `Airport`). Check my comments and see what you think.

:checkered_flag:

PR #113, opened 2015-07-26
Airport Challenge finished

sivanpatel:


spec/airport_spec.rb
```diff
@@ -40,9 +55,20 @@
     # the plane can not land, and must not be in the airport
 
     context 'when weather conditions are stormy' do
-      xit 'does not allow a plane to take off'
+      it 'will not allow a plane to take off' do
```
silvabox: You need to ensure it is sunny before landing the plane otherwise your test will fail randomly.

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+
 class Airport
+  DEFAULT_CAPACITY = 50
+  attr_accessor :capacity
+  attr_reader :weather
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+    @weather = ['sunny', 'sunny', 'sunny', 'sunny', 'stormy'].sample
```
silvabox: Does the weather never change at this airport!?

silvabox:
Hi Sivan, thanks for submitting your airport challenge!  Overall, you code is nice and clean and it's a good submission.  The main issue is that the weather doesn't ever change!  But that should be a fairly simple to fix.

Please study the common issues here: https://github.com/makersacademy/airport_challenge/wiki/Common-issues and see how these relate to your own code.

:checkered_flag:

PR #122, opened 2015-07-26
Airport challenge

richgeog:


lib/airport.rb
```diff
@@ -1,2 +1,46 @@
...
 class Airport
+  attr_accessor :capacity
+
+  def initialize capacity
+    @capacity = capacity
+  end
+
+  DEFAULT_CAPACITY = 15
+
+  attr_reader :capacity
```
danldb: attr_reader is not needed when you already have an attr_accessor

lib/airport.rb
```diff
@@ -1,2 +1,46 @@
+require_relative 'plane'
+
 class Airport
+  attr_accessor :capacity
+
+  def initialize capacity
+    @capacity = capacity
+  end
```
danldb: not making use of your DEFAULT_CAPACITY

lib/airport.rb
```diff
@@ -1,2 +1,46 @@
...
+
+  def initialize
+    @plane = []
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def release_plane
+    @plane if not empty?
+    @plane.pop
+  end
```
danldb: I don't think this method is doing what you think it is doing.

Line 20 is completely vacuous - it does nothing. Only line 21 is doing anything at all.

lib/airport.rb
```diff
@@ -1,2 +1,46 @@
...
+  def initialize capacity
+    @capacity = capacity
+  end
+
+  DEFAULT_CAPACITY = 15
+
+  attr_reader :capacity
+
+  def initialize
+    @plane = []
```
danldb: should this be @planes?

lib/airport.rb
```diff
@@ -1,2 +1,46 @@
...
+
+  # def weather
+  #   number = rand(10)
+  #     if number.odd?
+  #       puts "it is sunny"
+  #     else
+  #       fail "Airport is closed"
+  #     end
+  # end
+
```
danldb: Never commit unused code or commented out code.

lib/plane.rb
```diff
@@ -1,2 +1,19 @@
 class Plane
-end
+
+  def initilaize
+    @flying = true
+  end
+
+  def flying?
+    @flying = true
+  end
```
danldb: This sets flying to true, so it will never return false. Should just return @flying.

lib/plane.rb
```diff
@@ -1,2 +1,19 @@
...
+
+  def landed?
+    !@flying
+  end
+
+  def take_off
+    true
+  end
```
danldb: This should set flying to true.

danldb:
:construction: 

Well done, Your OOD is very good - you have kept the functionality very simple. 

However, you make some errors with handling your instance variables that make the code unusable in its current state. Also you did not correctly implement the weather functionality. 

I would advise you to arrange a 121 with me so that we can go through this together.

Good attempt!

PR #123, opened 2015-07-26
Submitted airport challenge.

7091lapS:
Need refactoring but all tests pass


lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+    @planes = []
+    @capacity = capacity
+  end
+
+  def allow_to_land(plane)
+    self.randomize_weather
+    if self.randomize_weather == "stormy"
+      raise "Can't authorize landing, adverse weather conditions"
+    end
+    if @planes.size >= capacity
```
danldb: extract randomize_weather into a separate class or module.

danldb: use a private attr_reader instead of @planes

lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+  end
+
+  def allow_to_takeoff(plane)
+    self.randomize_weather
+    if self.randomize_weather == "stormy"
+      raise "Can't authorize take off, adverse weather conditions"
+    end
+    plane.takeoff
+    @planes -= [plane]
+    self
```
danldb: why do you need to pass this method an argument?

lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+  def randomize_weather
+    case rand(100)
+    when  1..75 then @weather = "sunny"
+    when 75..100 then @weather = "stormy"
+    end
+  end
+
+  attr_reader :planes
+  attr_reader :capacity
+
```
danldb: why do you need to expose capacity?

spec/airport_spec.rb
```diff
@@ -1,48 +1,70 @@
...
+          allow(subject).to receive(:randomize_weather).and_return "stormy"
+          expect {subject.allow_to_takeoff(plane)}.to raise_error "Can't authorize take off, adverse weather conditions"
+        end
 
-    context 'when weather conditions are stormy' do
-      xit 'does not allow a plane to take off'
 
-      xit 'does not allow a plane to land'
+        it 'does not allow a plane to land' do
+          allow(subject).to receive(:randomize_weather).and_return "stormy"
```
danldb: stubbing the subject is a code smell - you should be testing the actual behaviour of the subject. This is a code smell that should push you to create a separate object for the random event.

spec/plane_spec.rb
```diff
@@ -1,29 +1,34 @@
...
-#
-# Think about your implementation - does it allow a plane to be "flying" and landed?
-# Are you testing that?
+  it { is_expected.to respond_to :flying? }
 
-describe Plane do
 
-  xit 'is flying when created'
+  it 'is flying when created' do
+    Plane.new
```
danldb: Plane.new does nothing in this context

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
+
```
danldb: very nicely written class - clear responsibilities, short methods, well named.

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
...
 class Plane
+
+  attr_reader :status
```
danldb: This should be a private method as you are revealing the planes status through `flying` so you should not also reveal it's inner status.

lib/plane.rb
```diff
@@ -1,2 +1,22 @@
...
+
+  def initialize
+    @status = "flying"
+  end
+
+  def flying?
+    false if @status == "landed"
+    @status == "flying"
+  end
```
danldb: you have an `attr_reader :status` so you should use that - we should not see `@status` in your code - replace with `status`.

danldb:
:checkered_flag: 

Well done - your classes are not too tightly coupled, and you successfully stubbed the random event. Look at improving your encapsulation (use private methods!), and try to keep your methods to 5 lines or less.

PR #125, opened 2015-07-26
Retesh - Completed Challenge

reteshbajaj:


reteshbajaj:
Method stub sorted

reteshbajaj:
Method

lib/airport.rb
```diff
@@ -1,2 +1,59 @@
...
+
+  def initialize (weather = DEFAULT_WEATHER, capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+    @weather = weather
+  end
+
+  def weather(set)
+    @weather = set
+  end
```
danldb: use an attr_accessor instead of @weather =

lib/airport.rb
```diff
@@ -1,2 +1,59 @@
 class Airport
+
+attr_reader :planes,:capacity, :weather
+
```
danldb: These could probably be private

lib/airport.rb
```diff
@@ -1,2 +1,59 @@
...
+  end
+
+  def weather(set)
+    @weather = set
+  end
+
+  def possible_land?
+    weather_check == :sunny
+  end
+  ###not clear why and works but && doesnt
```

lib/airport.rb
```diff
@@ -1,2 +1,59 @@
...
+  ### check with Ben or Sam - the above gives nil if the weather is stormy
+
+  def request_land(plane)
+    fail if full?
+    plane.ground if possible_land?
+    planes<<plane
+  end
```
danldb: fine - semantically your method name should be in the imperative - so instead of `request_land` you should simply tell the airport to `def land plane`.

lib/airport.rb
```diff
@@ -1,2 +1,59 @@
...
+
+  def request_take_off(plane)
+    fail if empty?
+    plane.fly_away if possible_take_off?
+    planes.delete(plane)
+  end
+
```
danldb: everything below here should be private

lib/airport.rb
```diff
@@ -1,2 +1,59 @@
...
+
+  def empty?
+    planes.empty?
+  end
+
+  def weather_check
+    @weather = :stormy if Random::rand(1.0) < CHANCE_OF_STORM
+    @weather
+  end
```
danldb: extract this out to a class or module and inject it in

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
 class Plane
+
+
+attr_accessor :airborne
+
```
danldb: this should be private

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
...
+def initialize
+  @airborne = :up
+end
```
danldb: use a boolean instead of :up

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
...
+  self
+  #self
+end
+
+def ground
+  @airborne = :downed
+  self
+  #self
+end
+
```
danldb: you have an attribute accessor, but you never use it! Remove all the `@` from your code.

spec/airport_spec.rb
```diff
@@ -10,26 +10,49 @@
 # and how that is implemented.
 #
 # If the airport is full then no planes can land
-
 describe Airport do
 
-  describe 'take off' do
-    xit 'instructs a plane to take off'
+let(:plane){ double :plane, airborne?:true}
+let(:airport){Airport.new}
```
danldb: use subject instead of airport

spec/airport_spec.rb
```diff
@@ -10,26 +10,50 @@
...
+  it { is_expected.to respond_to(:request_land).with(1).argument }
+
 
-    xit 'releases a plane'
+  describe '#request_land' do
+    it 'instructs a plane to land and receives plane' do
+    allow(plane).to receive(:ground)
+    airport.request_land(plane)
+    expect(airport.planes.pop).to eq plane
   end
```
danldb: airport.planes.pop is accessing the planes internal instance variable, which should be private. Instead, test the behaviour that you expect observe from outside the class. In this case, something like:

```
expect(plane).to receive :ground
```

spec/airport_spec.rb
```diff
@@ -10,26 +10,50 @@
...
 
-    xit 'receives a plane'
+  describe '#request_take_off' do
+    it 'instructs a plane to take off, then releases a plane' do
+    allow(plane).to receive(:ground)
+    allow(plane).to receive(:fly_away)
+    airport.request_land(plane)
+    airport.request_take_off(plane)
+    expect(airport).to be_empty
   end
```
danldb: again, you are testing methods that should be private.

spec/airport_spec.rb
```diff
@@ -10,26 +10,50 @@
...
 
-  describe 'traffic control' do
-    context 'when airport is full' do
-      xit 'does not allow a plane to land'
-    end
 
+    context 'traffic control' do
+
+      it 'checks airport capacity when landing' do
+      allow(airport).to receive(:weather_check).and_return(:sunny)
```
danldb: good stub

spec/plane_spec.rb
```diff
@@ -16,14 +16,25 @@
 
 describe Plane do
 
-  xit 'is flying when created'
+  let(:plane){ Plane.new }
+  let(:landed_plane) { Plane.new.ground }
+  let(:flying_plane) {landed_plane.fly_away}
```
danldb: use `subject (:landed_plane) { Plane.new.ground } ` instead. This would push you to write a different context for each subject.

spec/plane_spec.rb
```diff
@@ -16,14 +16,25 @@
...
-  xit 'can take off'
 
-  xit 'is flying after take off'
+  it 'is flying when created' do
+    expect(plane).to be_flying
+  end
+
+  it 'can land; the landed plane status is downed' do
+    expect(landed_plane).to_not be_flying
+  end
```
danldb: instead of using landed_plane, just use a regular plane, and land it in the setup phase of your test:
```
it 'can land' do
  subject.ground
  expect{subject.take_off}.not_to raise_error
end
```

spec/plane_spec.rb
```diff
@@ -16,14 +16,25 @@
...
+
+  it 'can take off and then the plane status is up' do
+    expect(flying_plane).to be_flying
+  end
 
 end
+
```
danldb: You missed some test cases. 

Can you land a plane that is already landed? Or make a plane take off when it is already flying?

Remember - test first, test the bad path as well as the good path. That will help you to cover all the edge cases.

danldb:
:checkered_flag: 

Well done!

PR #130, opened 2015-07-28
Airport challenge advance 

GabyML:
From: Gabriela May Lagunes


Progress-Reached.md
```diff
@@ -0,0 +1,8 @@
+It was possible to express the first and second user story in code.
+The third and the fourth were not compleated because:
+Once a plane lands on the airport, the capacity comes back to the default capacity. I was not able to find the source of this error.
+I am still not capable of applying doubles in rspec. Therefore I had to use relative classes and it was not possible to cover the requirements for the random weather.
+
+To be done: Understand and apply doubles. 
+
+July 27th, 2015
```
danldb: Very useful to have this information, thanks! In future you could put that into the main README.md file.

lib/airport.rb
```diff
@@ -1,2 +1,54 @@
...
+
+	attr_accessor :capacity
+
+	#The following lines of code initialize the values needed to define the capacity of the airport
+	
+	DEFAULT_CAPACITY = 1
+
+	#Objects from the class plane needed. To be substituted with doubles.
+	def initialize (capacity=DEFAULT_CAPACITY)
+	plane = Plane.new
```
danldb: You do not need to create a plane here.

lib/airport.rb
```diff
@@ -1,2 +1,54 @@
...
+	@planes = [] 
+	@capacity = capacity
+	end
+
+	#The following method let a plane to land on an airport. 
+	#The status of the plane changes from flying to not flying.
+	#One plane is added to the airport.
+	#If the plane already landed the method can not be called again over the same plane.
+	def landing (plane)
```
danldb: be careful with method names - They are like a message or command for the object, so prefer the Imperative form rather than the Continuous (i.e. land is better than landing).

lib/airport.rb
```diff
@@ -1,2 +1,54 @@
...
+		fail 'This plane has already landed' if plane.flying? == false
+		fail 'The airport is full. No landings are possible.' if full?
+		plane = plane.landed?
+		@planes << plane
+		puts 'Your flight has landed'
```
danldb: Do not use `puts` or `print` in your submitted code anymore. They are only used in debugging, or specific times when you want a command-line interface.

lib/airport.rb
```diff
@@ -1,2 +1,54 @@
...
+
+	end
+
+	#The following method let a plane to take off from an airport.
+	#The status of the plane changes from not flying to flying.
+	#One plane is substracted from the airport.
+	#If the plane already took off, the method can not be called again over the same plane.
```
danldb: Documentation is excessive - the tests provide enough documentation for this code.

lib/airport.rb
```diff
@@ -1,2 +1,54 @@
...
+	def take_off (plane)
+		fail 'This plane has already departed' if plane.flying? == true
+		plane = plane.departed?
+		@planes[@planes.size - 1] = true
```
danldb: Not sure what you are trying to do on lines 37 and 38.

lib/airport.rb
```diff
@@ -1,2 +1,54 @@
...
+		@planes.delete(true)
+		puts 'Your flight has departed'
+	end
```
danldb: You almost have this correct, but in its current state it doesn't quite work.

Instead of lines 38 and 39 you could just put `planes.delete(plane)`

lib/airport.rb
```diff
@@ -1,2 +1,54 @@
...
+
+	#The following method let a plane to take off from an airport.
+	#The status of the plane changes from not flying to flying.
+	#One plane is substracted from the airport.
+	#If the plane already took off, the method can not be called again over the same plane.
+	def take_off (plane)
+		fail 'This plane has already departed' if plane.flying? == true
+		plane = plane.departed?
+		@planes[@planes.size - 1] = true
+		@planes.delete(true)
```
danldb: You have created an `attr_reader :planes`, so use it. You should refer to `planes` instead of `@planes`.

lib/airport.rb
```diff
@@ -1,2 +1,54 @@
...
+	private
+
+	attr_reader :planes
+
+	def full?
+		planes.count >= capacity
+	end
+
+
 end
```
danldb: Watch your indentation - it should be set to 2 spaces - this code is very difficult to read in certain text editors.

lib/plane.rb
```diff
@@ -1,2 +1,23 @@
...
+	def landed? #Method that state that the plane is not flying
+		subject = false
+		@flying = false
+	end
+
+	def departed?#Method that state that the plane is flying
+		subject = true
+		@flying = true
+	end
+
```
danldb: `landed?` should be `land` and `departed` should be `depart`. These methods should not have a `?` as they are commands, not questions.

`subject = true` is not doing anything and should be removed.

spec/airport_spec.rb
```diff
@@ -1,33 +1,24 @@
...
-
-# A plane currently in the airport can be requested to take off.
-#
-# No more planes can be added to the airport, if it's full.
-# It is up to you how many planes can land in the airport
-# and how that is implemented.
-#
-# If the airport is full then no planes can land
 
 describe Airport do
```
danldb: To get started with a double use `let(:plane){ double :plane }`, then you can use the double instead of a real plane.

spec/plane_spec.rb
```diff
@@ -1,29 +1,53 @@
...
+	it 'is not flying after landing' do
+		airport = Airport.new
+		airport.landing (subject) 
+		expect(subject.flying?).to be false
+		end
+	#Third test: when the plane is in the airport, it's state should be not flying, then it should respond to landed?
+
+	it 'can not land (again) when is not flying' do
+		airport = Airport.new
+		subject.landed?
```
danldb: You should not be testing the airport here - in plane_spec.rb you should only be testing planes.

danldb:
:construction: 

Hey Gabi - I see you struggling with using doubles, and with separating your classes. Let's book a 121 to see if we can fix this - get in touch on slack to book a time.

PR #139, opened 2015-09-05
#land #take_off methods added

timhyson:


lib/airport.rb
```diff
@@ -1,2 +1,34 @@
...
+
+  DEFAULT_CAPACITY = 10
+
+  def initialize
+    @planes = []
+  end
+
+  def capacity
+    DEFAULT_CAPACITY
+  end
```
danldb: I like this - it is unorthodox, but it encapsulates the instance from its constant, which is not a bad idea.

lib/airport.rb
```diff
@@ -1,2 +1,34 @@
...
+
+  def release(plane)
+    fail 'airport empty' if empty?
+    @planes.pop
+  end
```
danldb: no argument needed - as per houndci.

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
 class Plane
+  attr_accessor :flying
+
+  def initialize
+    @flying = 'flying'
+  end
```
danldb: this should return a boolean

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
...
+  # def flying
+  #   @flying
+  # end
+  def flying?
+    @flying
+    true if @flying == 'flying'
+  end
```
danldb: just return @flying

lib/plane.rb
```diff
@@ -1,2 +1,27 @@
...
+  def landed?
+    @flying
+    true  if @flying == 'landed'
+    false if @flying == 'flying'
+  end
```
danldb: just return `not @flying`

danldb:
:construction: 

Very good attempt - you have been so successful in removing dependency, that in the end your airplane had zero dependency on the airport. Unfortunately, without dependency there is no communication.

`airport.land(plane)` should send the message to the plane to change its status to landed.

PR #141, opened 2015-09-05
Jongmin's weekend challenge 1

jongmin141215:


README.md
```diff
@@ -50,20 +54,20 @@ For overriding random weather behaviour, please read the documentation to learn
...
+
 BONUS
 * Write an RSpec **feature** test that lands and takes off a number of planes...
 
 Note that is a practice 'Tech Test' of the kinds that employers use to screen developer applicants.  More detailed submission requirements/guidelines are in [CONTRIBUTING.md](CONTRIBUTING.md)
 
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first. 
+Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first.
 
 * **Submit a pull request early.**  There are various checks that happen automatically when you send a pull request.  **You should pay attention to these - the results will be added to your pull request**.  Green is good.
```
danldb: You should update the README.md

domain_model
```diff
@@ -0,0 +1,30 @@
...
+|pilot                  |           |
+|air traffic controller |           |
+|plane                  | land      |
+|plane                  | take_off  |
+|plane                  |flying?    |
+|plane                  |landing?   |
+|airport                |plane_land |
+|airport                |plane_leave |
+|airport                |full?      |
+|weather                |sunny?     |
```
danldb: This is great - it should be in the README.md

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
+require_relative 'plane'
+require_relative 'weather'
+
 class Airport
+  attr_reader :capacity, :planes
+  def initialize(capacity = 50)
```
danldb: use a constant for the default capacity

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
...
+    @capacity = capacity
+    @planes = []
+  end
+
+  def clear_for_takeoff(plane)
+    fail 'Too stormy' if weather_report == :stormy
```
danldb: weather_report should return true or false

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
...
+    fail 'Airport empty' if empty?
+    plane.take_off
+    planes.delete(plane)
+    plane
```
danldb: line 16 is not necessary as `planes.delete(plane)` will return the plane.

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
...
+  def clear_for_landing(plane)
+    fail 'Too stormy' if weather_report == :stormy
+    fail 'Airport full' if full?
+    plane.land
+    planes << plane
+  end
+
+  def weather_report
+    Weather.new.weather_report
+  end
```
danldb: use depndency injection

lib/plane.rb
```diff
@@ -1,2 +1,15 @@
 class Plane
+  attr_reader :plane_status
+  def initialize
+    @status = [:flying, :landed]
+    @plane_status = @status[0]
+  end
```
danldb: why not @flying = true instead of using an array?

lib/weather.rb
```diff
@@ -0,0 +1,15 @@
...
+    rand(101)
+  end
+  def condition
+    return 1 if storm_probability > 90
+    0
+  end
+  def weather_report
+    @weather[condition]
+  end
+end
```
danldb: should only have a single public method - everything else should be private. If it is only going to return one of 2 states, then you should prefer a boolean - that way all the methods that rely on true/false can be utilised without any conversion.

danldb:
:checkered_flag: 

Well done Jongmin - an excellent submission! There were a few issues, see inline comments.

PR #142, opened 2015-09-05
working planes and airport

phillipclarke29:


lib/airport.rb
```diff
@@ -1,2 +1,26 @@
+require_relative 'plane'
 class Airport
+
+  DEFAULT_CAPACITY = 20
+
+  attr_reader :capacity
+  def initialize (capacity=DEFAULT_CAPACITY)
```
roidriscoll: Have to agree with Hound on that one ;)

spec/airport_spec.rb
```diff
@@ -13,23 +14,40 @@
...
-  describe 'landing' do
-    xit 'instructs a plane to land'
+    it 'counts the number of planes in the aiport' do
+      expect(subject.count_planes).to eql 0
+    end
+
+    it 'should allow a plane to land' do
+      plane = Plane.new
+      subject.clear_to_land(plane)
+      puts plane.flying
```
roidriscoll: Is this a remnant of a debugging session?

spec/airport_spec.rb
```diff
@@ -13,23 +14,40 @@
...
+      expect(plane.flying).to eql("landed")
+    end
 
-    xit 'receives a plane'
-  end
+    it 'should allow a plane to take off' do
+      plane = Plane.new
+      puts plane #why does this not show the flying variable
```
roidriscoll: That's a great question - I have done a bit of research but nothing turned up! I will ask around the other coaches to see if anyone has an explanation.

The big question though, is why? :smile:

spec/airport_spec.rb
```diff
@@ -13,23 +14,40 @@
...
-  end
 
-  describe 'landing' do
-    xit 'instructs a plane to land'
+    it 'counts the number of planes in the aiport' do
+      expect(subject.count_planes).to eql 0
+    end
+
+    it 'should allow a plane to land' do
+      plane = Plane.new
```
roidriscoll: It would be really nice to see you using doubles here so that you can test your classes in isolation.

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+    @planes = []
+    @capacity = capacity
+    @today_weather
+    #@planes << Plane.new("landed")
```
roidriscoll: Remember to delete comments when you are done with them.

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+  end
+
+  def clear_to_land(plane)
+    if @planes.count >= @capacity
+    raise 'Airport at Capacity'
+  elsif self.weather_conditions == 'stormy'
+    raise 'Stormy!'
```
roidriscoll: Agree with hound!

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+    else
+    plane.land
+    @planes << plane
+    end
+  end
```
roidriscoll: So the other day I mentioned in one of our stand ups about `elsif` being a code smell. Do you think you could refactor this method to reduce the complexity of this conditional?

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+
+
+  def clear_to_take_off(plane)
+
+    plane = @planes.pop
+    plane.take_off
+  end
+
+  def planes
```
roidriscoll: +1

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+    @planes
+  end
+
+  def count_planes
+    @planes.count
+  end
+
+  def weather_conditions
+  @today_weather = Weather.new.stormy
```
roidriscoll: Here we are introducing a DEPENDENCY! How about we inject that in `initialize` so that `weather_conditions` looks like this:
```
def weather_conditions
  todays_weather.stormy? # my kingdom for an apostrophe!
end
```

lib/airport.rb
```diff
@@ -1,2 +1,45 @@
...
+    raise 'Stormy!'
+    else
+    plane.land
+    @planes << plane
+    end
+  end
+
+
+  def clear_to_take_off(plane)
+
```
roidriscoll: Empty line! :broken_heart:

lib/plane.rb
```diff
@@ -1,2 +1,26 @@
...
+
+  def take_off
+    fail 'already flying' if @flying == 'flying'
+    @flying = "flying"
+  end
+
+  # def destination
+  #   fail 'you cannot have a plane in the air without a destination' if @destination == ''
+  # @destination
+  # end
```
roidriscoll: Remember to delete code you aren't using, rather than comment them out. It looks neater, and also you are less at risk of accidentally uncommenting them with an errant keystroke!

It's a much better practice to create a branch for feature experiments, and merge the branches when you are sure you want to integrate them into your application.

lib/plane.rb
```diff
@@ -1,2 +1,26 @@
...
+    @flying = flying
+  end
+
+  def flying
+    @flying
+  end
+
+  def land
+    fail 'already landed' if @flying == 'landed'
+    @flying = "landed"
```
roidriscoll: So many `@`s!

lib/weather.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+  def stormy
+    conditions
+    if @number == 5
+    'stormy'
+    else
+    'sunny'
+    end
+  end
+
```
roidriscoll: This code is quite hard to read, despite being quite simple due to poor formatting. You need a blank line between 9-10, and indentation is so important! Use the tab key on lines 13 and 14 would make this much easier to read.

lib/weather.rb
```diff
@@ -0,0 +1,19 @@
+class Weather
+
+  def initialize
+  @number
```
roidriscoll: you don't need `initialize` here, as all it's doing is creating `@number` which happens again on line 8, with a value (here it's automatically assigned `nil`)

Also you should really wrap `@number` in an `attr_reader`

lib/weather.rb
```diff
@@ -0,0 +1,19 @@
+class Weather
```
roidriscoll: `Weather` really needs to be a module!

spec/airport_spec.rb
```diff
@@ -13,23 +14,40 @@
 
 describe Airport do
 
-  describe 'take off' do
-    xit 'instructs a plane to take off'
+  let(:weather){double:weather, {conditions: 'stormy'}}
+  let(:weather1){double:weather1, {conditions: 'sunny'}}
```
roidriscoll: Agree with hound, plus `weather1` isn't a particularly clear name for this double: `sunny_weather` would be much more useful. Also, this doesn't seem to be used anywhere?

spec/airport_spec.rb
```diff
@@ -13,23 +14,40 @@
...
+      plane = Plane.new
+      subject.clear_to_land(plane)
+      puts plane.flying
+      expect(plane.flying).to eql("landed")
+    end
 
-    xit 'receives a plane'
-  end
+    it 'should allow a plane to take off' do
+      plane = Plane.new
```
roidriscoll: `Plane` should be a double here, so that you can test your class in isolation!

spec/airport_spec.rb
```diff
@@ -39,10 +57,27 @@
...
+        if subject.weather_conditions == weather.conditions
+        expect  { subject.clear_to_take_off(plane) }.to raise_error 'Stormy!'
+        end
+      end
+
+
+      it 'does not allow a plane to land' do
+        plane = Plane.new
+        if subject.weather_conditions == 'stormy'
+        expect  { subject.clear_to_land(plane) }.to raise_error 'Stormy!'
```
roidriscoll: A little more verbosity on your error messages will go a long way!

roidriscoll:
Hi Phillip! I've left a few pointers here, but I am sure after your progress this week, these amendments shouldn't be a problem for you. Remember also that tidy code is a lot easier to read (and comment on), so I have time to leave comments on more interesting things!

:construction:

PR #143, opened 2015-09-05
Zhivko Siderov: Airport Challenge

zsid:


lib/aircontroller.rb
```diff
@@ -0,0 +1,38 @@
...
+    fail "Bad weather conditions" if weather_stormy?
+    plane_landed(*plane)
+  end
+
+  def permission_to_take_off(*plane)
+    fail "No planes at the airport" if plane_hangar_empty?
+    fail "Bad weather conditions" if weather_stormy?
+    plane_took_off(*plane)
+  end
+
```
danldb: Over complicated - the traffic controller module is adding a layer of complexity that is not asked for.

lib/aircontroller.rb
```diff
@@ -0,0 +1,38 @@
...
+  private
+
+  def plane_landed(*plane)
+    x = plane.each { |plane| plane.landing }
+    planes << x.flatten
+    planes.flatten!
+  end
```
danldb: Not sure why this module has this responsibility. The plane shoudl be responsible for its own state.

lib/airport.rb
```diff
@@ -1,2 +1,14 @@
...
+
 class Airport
+  include AirController
+  include PlaneHangar
+
+  def initialize(capacity = PlaneHangar::DEFAULT_CAPACITY)
+    self.capacity = capacity
+    self.planes
+  end
+
```
danldb: Trying to be too clever - the airport is a simple class that has the responsibility to store and release planes. This feels over-engineered.

lib/plane.rb
```diff
@@ -1,2 +1,17 @@
...
+  end
+
+  def landing
+    @flying = false
+  end
+
+  def take_off
+    @flying = true
+  end
+
```
danldb: great - I would have used a `flying?` method instead of the attr_reader in order to show that it will return a boolean.

lib/planehangar.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def plane_hangar_full?
+    planes.length >= capacity
+  end
+
+  def plane_hangar_empty?
+    planes.empty?
+  end
+
+end
```
danldb: I don't think this module is necessary as it is not DRYing up your code in any way - in fact it means that you have one more thing to search through if you ever want to make amendments to your airport class.

lib/plane.rb
```diff
@@ -1,2 +1,17 @@
 class Plane
+
+  attr_reader :flying
+
+  def initialize
+    @flying = true
+  end
+
+  def landing
```
danldb: `def land` would be better - use the imperative instead of the continuous. The method would command a plane to land.

danldb:
:construction: 

Over complicated - I would recommend you to refactor this during lab week or after the course. Follow the brief :)

PR #144, opened 2015-09-05
Harriet's Airport Challenge

harrietc52:


lib/airport.rb
```diff
@@ -1,2 +1,40 @@
+require './lib/plane.rb'
+require './lib/weather.rb'
+
 class Airport
+  include Weather
+
+  DEFAULT_CAPACITY = 5
```
ptolemybarnes: Nice :+1:

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
...
+
+  attr_reader :planes, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def plane_land(plane)
+    fail 'Plane has already landed' if @planes.include?(plane)
```
ptolemybarnes: An attr_reader for @planes has been set so:
`fail 'Plane has already landed' if planes.include?(plane)`
would be cleaner here.

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
...
+    fail 'Cannot currently land at airport' if full?
+    fail 'Cannot land if weather is stormy' if stormy? == true
```
ptolemybarnes: We needn't check that `stormy? == true` as #stormy? returns the boolean we want anyway.

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
...
+    planes << plane
+    'Plane is landed'
```
ptolemybarnes: Not bad. It's good that your @plane_land method doesn't return the `planes` array. :-)

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
...
+  DEFAULT_CAPACITY = 5
+
+  attr_reader :planes, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def plane_land(plane)
```
ptolemybarnes: Not sure about the naming of this method. In my opinion, simply `#land` would be cleaner. What if, for instance, it was decided that the airport should have helicopters land at it? Then we would have: `airport.plane_land(helicopter)`, which would be a bit weird.

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
...
+    fail 'Plane has already landed' if @planes.include?(plane)
+    fail 'Cannot currently land at airport' if full?
+    fail 'Cannot land if weather is stormy' if stormy? == true
+    planes << plane
+    'Plane is landed'
+  end
+
+  def plane_take_off(plane)
```
ptolemybarnes: Same as above on the naming here

lib/airport.rb
```diff
@@ -1,2 +1,40 @@
...
+    fail 'Cannot currently take off' if empty?
+    fail 'Cannot land if weather is stormy' if stormy? == true
```
ptolemybarnes: Same as above on running a boolean check on the #stormy? method.

lib/plane.rb
```diff
@@ -1,2 +1,26 @@
 class Plane
+
+ def initialize
+   @flying = true
+   "Plane is flying"
```
ptolemybarnes: The initialize method is a rare case where we have no control over the return value. Calling `Plane.new` will always return an instance of plane. So this string is redundant.

lib/plane.rb
```diff
@@ -1,2 +1,26 @@
...
+ end
+
+ def flying?
+   @flying
+ end
+
+ def can_land
```
ptolemybarnes: Not sure about the naming of `can_land`. It seems like a question: "Can this plane land?" In which case the answer yes or no would be represented by a boolean. But then the convention would be to do `land?`. Annyywayyy the method is actually landing a plane! So maybe just `plane.land` would be nice?

lib/plane.rb
```diff
@@ -1,2 +1,26 @@
...
+    fail 'the plane cannot land if it is not flying' if landed?
+    @flying = false
```
ptolemybarnes: It would be better to access these instance variables via a private attribute writer.

lib/plane.rb
```diff
@@ -1,2 +1,26 @@
...
+    true
```
ptolemybarnes: Nice. :+1: Well done on considering what the return value of the method is.

lib/plane.rb
```diff
@@ -1,2 +1,26 @@
...
+  end
+
+ def landed?
+   !@flying
```
ptolemybarnes: Again, access this instance variable via a reader method.

lib/plane.rb
```diff
@@ -1,2 +1,26 @@
...
+ end
+
+ def can_take_off
```
ptolemybarnes: Same as a bove on naming

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+module Weather
+
+  def stormy?
```
ptolemybarnes: Think about what the equality check on line 5 is returning. This method could be refactored to:
```
def stormy?
  Random.rand(5) == 0
end

File: spec/airport_spec.rb [hunk 15]
```diff
@@ -1,48 +1,78 @@
...
-describe Airport do
 
   describe 'take off' do
-    xit 'instructs a plane to take off'
+    it 'instructs a plane to take off' do
+      is_expected.to respond_to(:plane_take_off)
+    end
 
-    xit 'releases a plane'
+    it 'releases a plane' do
```
ptolemybarnes: There's no expectation in this test.

spec/airport_spec.rb
```diff
@@ -1,48 +1,78 @@
...
+
   describe 'landing' do
-    xit 'instructs a plane to land'
+    it 'instructs a plane to land' do
+      is_expected.to respond_to(:plane_land)
+    end
 
-    xit 'receives a plane'
+    it 'receives a plane' do
+      allow(subject).to receive(:stormy?).and_return(false)
```
ptolemybarnes: Again, not expectation here.

spec/airport_spec.rb
```diff
@@ -1,48 +1,78 @@
...
 
+
   describe 'traffic control' do
     context 'when airport is full' do
-      xit 'does not allow a plane to land'
+      it 'does not allow a plane to land' do
+        allow(subject).to receive(:stormy?).and_return(false)
+        subject.capacity.times { subject.plane_land(Plane.new)}
+        expect { subject.plane_land(Plane.new)}.to raise_error
+        'Cannot currently land at airport'
```
ptolemybarnes: Hmm I can see a pattern with these strings here... >_> I can sort of see what they're doing for you but it's quite unconventional.

spec/airport_spec.rb
```diff
@@ -1,48 +1,78 @@
...
+      it 'does not allow a plane to take off' do
+        allow(subject).to receive(:stormy?).and_return(false)
+        subject.plane_land(plane)
+        allow(subject).to receive(:stormy?).and_return(true)
+        expect{subject.plane_take_off(plane)}.to raise_error
+        'Cannot take off if weather is stormy'
+      end
     end
   end
 end
```
ptolemybarnes: Nice tests! Apart from the random strings.

spec/plane_spec.rb
```diff
@@ -1,30 +1,40 @@
...
+
+ describe '#take_off'
+
+ it 'can take off' do
+   expect{subject.can_take_off}.to raise_error
+   'cant take off when already flying'
+ end
+
+  it 'is flying after take off' do
+    subject.can_land
```
ptolemybarnes: In fact I would go as far as to say that you should never write a method in Ruby starting with 'can'. It's always going to end up being unconventional. There may be some cases of boolean methods that could start with 'can' but not otherwise.

spec/plane_spec.rb
```diff
@@ -1,30 +1,40 @@
...
+    subject.can_take_off
+    expect(subject.flying?).to eq true
```
ptolemybarnes: RSpec gives us some nice sugar for doing this sort of test:
```ruby
expect(subject).to be_true
```

spec/airport_spec.rb
```diff
@@ -1,48 +1,78 @@
...
+      is_expected.to respond_to(:plane_land)
+    end
 
-    xit 'receives a plane'
+    it 'receives a plane' do
+      allow(subject).to receive(:stormy?).and_return(false)
+      subject.plane_land(plane)
+    end
+
+     it 'can\'t land same plane twice' do
```
ptolemybarnes: Use double quotes rather than escaping apostrophes.

spec/airport_spec.rb
```diff
@@ -1,48 +1,78 @@
...
+     end
   end
 
+
   describe 'traffic control' do
     context 'when airport is full' do
-      xit 'does not allow a plane to land'
+      it 'does not allow a plane to land' do
+        allow(subject).to receive(:stormy?).and_return(false)
+        subject.capacity.times { subject.plane_land(Plane.new)}
```
ptolemybarnes: Why not use the double of plane here?

spec/plane_spec.rb
```diff
@@ -1,30 +1,40 @@
...
+  end
+
+ describe '#land' do
+ it 'airport can_land plane' do
+   expect(subject).to respond_to :can_land
+ end
+
+ it 'can land' do
+   subject.can_land
+   subject.landed?
```
ptolemybarnes: This line isn't necessary.

ptolemybarnes:
:airplane: Nice submission! :airplane: There's some great stuff here. See my comments and let me know if you have any questions. ;-)
:checkered_flag:

PR #145, opened 2015-09-05
Dan St Paul: Airport Challenge

danielstpaul:


README.md
```diff
@@ -42,29 +33,41 @@ So that I can avoid accidents
...
+| :-----------------: | :------------------: |
+| Pilot               |                      |
+| Air Traffic Control |                      |
+| Plane               | status(flying/landed)|
+| Airport             | capacity(full?)      |
+| Airport             | allow_take_off       |
+| Airport             | allow_land           |
+| Airport             | weather(stormy/sunny)|
+
+Airport <-- allow_take_off        --> Plane.take_off --> Plane(status: flying)
```
danldb: nice!

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
+require_relative 'plane'
+
 class Airport
+
+  Capacity = 20
+
+  attr_reader :capacity, :planes
```
danldb: don't need an instance variable to return the constant. Constant should be SCREAMING_SNAKE_CASE. I love when I get to type SCREAMING_SNAKE_CASE as it is definitely the best case name.

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+
+  def allow_take_off
+    fail "Sorry, bad weather" if rand_num == 1
+    fail "No planes at the airport" if empty?
+
+    plane = @planes.pop
+    @planes.delete(plane)
+    plane.take_off
+    plane
+  end
```
danldb: This would perform the same functionality:
```
def launch
  # insert fails here
  plane = planes.pop
  plane.take_off
end
```
Be careful of your naming - methods should usually be in the imperative as they are generally commands.

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+
+  def allow_land(plane)
+    fail "Sorry, bad weather" if rand_num == 1
+    fail "Airport is full pilot" if full?
+    plane.land
+    @planes << plane
+  end
```
danldb: good

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+
+  def forecast
+    return "stormy" if rand_num == 1
+    "sunny"
+  end
```
danldb: return a boolean instead of a string

lib/airport.rb
```diff
@@ -1,2 +1,49 @@
...
+
+  def rand_num
+    rand(2)
+  end
```
danldb: ideally should be isolated into a module or class, otherwise shoudl definitely be private

lib/plane.rb
```diff
@@ -1,2 +1,18 @@
 class Plane
+
+  attr_reader :status
```
danldb: use `flying?` instead of status

spec/airport_spec.rb
```diff
@@ -12,22 +12,71 @@
 # If the airport is full then no planes can land
 
 describe Airport do
+  let(:plane){double(:plane, land: nil, take_off: nil)}
 
-  describe 'take off' do
-    xit 'instructs a plane to take off'
+  before do
+    subject.stub(:rand_num).and_return(0)
+  end
```
danldb: stub is deprecated, use allow instead. 

it is a code smell to stub subject - better to isolate the random behaviour in its own class or module.

danldb:
:construction: 

Good, but not quite a pass. I love the README.md describing your CRC cards, and the process you went through. I would also mention how long it took you to complete the challenge, and how much previous experience you had.

PR #146, opened 2015-09-05
airport challenge completed

bat020:


lib/airport.rb
```diff
@@ -1,2 +1,60 @@
...
+
+  def has_plane?(plane)
+    plane.landed_at?(self)
+  end
+
+  def full?
+    plane_count >= CAPACITY
+  end
+
+  def permission_to_take_off
```

lib/airport.rb
```diff
@@ -1,2 +1,60 @@
...
+    weather_report == :stormy ? :stormy : :ok
+  end
+
+  def permission_to_land
+     if full?
```

lib/airport.rb
```diff
@@ -1,2 +1,60 @@
...
+    @plane_count += 1
+  end
+
+  def decrement_plane_count
+    @plane_count -= 1
+  end
+
+  def weather_report
+    Kernel.rand < STORM_CHANCE ? :stormy : :sunny
+  end
```
danldb: Returns 1 of 2 states, so ideally should return a boolean - that way you get access to all the methods that rely on a true/false status. `def stormy?` could return true or false.

lib/airport.rb
```diff
@@ -1,2 +1,60 @@
+require_relative 'plane'
+
 class Airport
+
+  CAPACITY = 5
+  STORM_CHANCE = 0.1
+
+  def initialize
+    @plane_count = 0
+  end
```
danldb: an array of planes would have made life easier for you

lib/plane.rb
```diff
@@ -1,2 +1,30 @@
 class Plane
+
+  def initialize
+    @status = :flying
+  end
```
danldb: `@flying = true` would have been better - again, if you have a choice of 2 states, prefer a boolean.

lib/plane.rb
```diff
@@ -1,2 +1,30 @@
...
+
+  def flying?
+    status == :flying
+  end
```
danldb:
```
def flying?
  @flying
end
```

lib/airport.rb
```diff
@@ -1,2 +1,60 @@
...
+
+  def increment_plane_count
+    @plane_count += 1
+  end
+
+  def decrement_plane_count
+    @plane_count -= 1
+  end
+
+  def weather_report
```
danldb: should be isolated and extracted into a module

lib/plane.rb
```diff
@@ -1,2 +1,30 @@
...
+
+  def land_at(airport)
+    fail 'Airport is full' if airport.permission_to_land == :full
+    fail 'Too stormy' if airport.permission_to_land == :stormy
+    self.status = airport
+  end
+
+  private
+
+  attr_accessor :status
```
danldb: Excellent that you have made this private

spec/airport_spec.rb
```diff
@@ -1,48 +1,109 @@
...
-    xit 'releases a plane'
   end
 
-  describe 'landing' do
-    xit 'instructs a plane to land'
+  context 'in sunny weather' do
+
+    before :each do
+      ensure_sunny_weather
+    end
```
danldb: good

spec/airport_spec.rb
```diff
@@ -1,48 +1,109 @@
...
+
+    it 'releases a plane that it instructs to take off' do
+      plane = double :plane, take_off_from: nil, landed_at?: false
+      subject.request_take_off(plane)
+      expect(subject).not_to have_plane(plane)
+    end
+
+    it 'gives permission to planes that want to land' do
+      plane = double :plane
```
danldb: you have made a plane, but not used it

spec/airport_spec.rb
```diff
@@ -1,48 +1,109 @@
...
+      expect(subject.permission_to_land).to eq :ok
+    end
```
danldb: permission to land does not appear to do anything

lib/airport.rb
```diff
@@ -1,2 +1,60 @@
...
+  end
+
+  def request_landing(plane)
+    plane.land_at(self)
+    increment_plane_count
+  end
+
+  def has_plane?(plane)
+    plane.landed_at?(self)
+  end
```
danldb: violating 'tell don't ask'.

lib/plane.rb
```diff
@@ -1,2 +1,30 @@
...
+    @status = :flying
+  end
+
+  def flying?
+    status == :flying
+  end
+
+  def take_off_from(airport)
+    fail 'Too stormy' if airport.permission_to_take_off == :stormy
+    self.status = :flying
```
danldb: seems like you are doubling the dependency. The responsibility of launching a plane should come from the airport, not the plane. The plane is responsible for taking off, i.e. changing its status to flying. The airport is responsible for launching it and removing it from its list of planes.

lib/plane.rb
```diff
@@ -1,2 +1,30 @@
...
+  end
+
+  def landed_at?(airport)
+    status == airport
+  end
+
+  def land_at(airport)
+    fail 'Airport is full' if airport.permission_to_land == :full
```
danldb: this fail belongs in the airport class

spec/airport_spec.rb
```diff
@@ -1,48 +1,109 @@
...
-# If the airport is full then no planes can land
+  def ensure_sunny_weather
+    sunny_rand = Airport::STORM_CHANCE / 2 + 0.5
+    allow(Kernel).to receive(:rand).and_return(sunny_rand)
+  end
 
-describe Airport do
+  def ensure_stormy_weather
+    stormy_rand = Airport::STORM_CHANCE / 2
+    allow(Kernel).to receive(:rand).and_return(stormy_rand)
```
danldb: Never stub Kernel. Never never. Really, never.

danldb:
:construction: 

You have over complicated this. Your biggest enemy is simplification, or lack thereof. There are some ingenious solutions (stubbing kernel...) that have become necessary because you made fundamental errors in the design of the program. I would suggest that you redo this challenge during lab week or after the end of the course.

PR #149, opened 2015-09-05
Harry Venables Airport Challenge

hvenables:


lib/airport.rb
```diff
@@ -1,2 +1,38 @@
+require './lib/plane.rb'
+require './lib/weather.rb'
+
 class Airport
+  include Weather
+
```
danldb: dependency injection for the weather would have been amazing - but it is fantastic that you have isolated the random behaviour in a module.

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+    @planes = []
+    @capacity = capacity
+  end
+
+  def plane_take_off(plane)
+    fail 'Cannot currently take off' if empty? || stormy?
+    planes.delete(plane)
+    plane.take_off
+    plane
+  end
```
danldb:
```
fail 'Cannot currently take off' if empty? || stormy?
plane.take_off
planes.delete(plane)
```
Should do the same thing in one less line of code - however the trade off is that the message to take off is being passed to the plane after it is being removed from the list. For me, I think that would be a fair trade off, but you may disagree!

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+
+  def plane_land(plane)
+    fail 'Plane cannot currently land at airport' if full? || stormy?
+    plane.land
+    planes << plane
+  end
+
+  private
+
```
danldb: good use of `private`

lib/plane.rb
```diff
@@ -1,2 +1,14 @@
...
+  def land
+    @flying = false
+  end
+  def flying?
+    @flying
+  end
+  def take_off
+    @flying = true
+  end
 end
```
danldb: Great, but no guard clauses - what happens if I tell a flying plane to take off?

lib/weather.rb
```diff
@@ -0,0 +1,8 @@
+module Weather
+  def stormy?
+    weather_randomiser <= 2 ? true : false
+  end
+  def weather_randomiser
+    rand(10)
+  end
+end
```
danldb: really great, well done.

spec/airport_spec.rb
```diff
@@ -1,48 +1,69 @@
...
-#
-# No more planes can be added to the airport, if it's full.
-# It is up to you how many planes can land in the airport
-# and how that is implemented.
-#
-# If the airport is full then no planes can land
-
 describe Airport do
 
+  let (:plane) {double('plane', flying?: true, land: false, take_off: true) }
```
danldb: Too many methods in this double - consider using `allow` in relevant contexts.

spec/airport_spec.rb
```diff
@@ -1,48 +1,69 @@
...
+
   describe 'take off' do
-    xit 'instructs a plane to take off'
+    it 'instructs a plane to take off' do
+      is_expected.to respond_to(:plane_take_off)
+    end
+
+    it 'releases a plane' do
+      allow(subject).to receive(:stormy?).and_return(false)
+      subject.plane_land(plane)
```
danldb: stubbing subject is a bit of a code smell as it means you are modifying the object you are trying to test.

danldb:
:checkered_flag: 

Well done! I notice that coveralls is red - I think that is our fault as we have the settings too strict.

A strong pass.

PR #150, opened 2015-09-05
Lucy Airport Challenge

lucetzer:


lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+
+DEFAULT_CAPACITY = 10
+
+  def initialize
+    @planes = []
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def release_plane
+    return false if empty? || stormy?
```
danldb: This currently fails silently - I would raise an error in this situation.

lib/airport.rb
```diff
@@ -1,2 +1,42 @@
...
+    @planes.pop
+  end
+
+  def dock(plane)
```
danldb: I don't know if planes really _dock_

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+    fail "Airport unavailable" if full? || stormy?
+    @planes << plane
+  end
+
+  def stormy?
+    rand(40) == 6 ? "stormy" : "sunny"
+  end
```
danldb: Ideally this would be isolated in a module or class and then injected into the class

lib/plane.rb
```diff
@@ -1,2 +1,23 @@
...
+  end
+
+  def landed?
+    !@flying
+  end
+
+  def take_off
+    @flying = true
+  end
+
```
danldb: Looks fantastic - although you have not used any guard clauses - what happens if you instruct a landed plane to land?

spec/airport_spec.rb
```diff
@@ -39,10 +75,23 @@
...
 
-    context 'when weather conditions are stormy' do
-      xit 'does not allow a plane to take off'
+    context "when weather conditions are stormy" do
+
+      it "tests the stormy method" do
+        # subject.stub(:forecast) {5}
+        srand(40)
+        expect(subject.stormy?).to eq("stormy")
+      end
```
danldb: In this case you could use `allow(subject).to receive(:stormy?).and_return(:false)`. I wouldn't try to test a random event.

spec/plane_spec.rb
```diff
@@ -17,14 +17,30 @@
...
+    expect(subject.flying).to eq(true)
+  end
+
+  it "can land" do
+    is_expected.to respond_to(:land)
+  end
+
+  it "has landed" do
+    subject.land
+    subject.landed?
```
danldb: Line 32 is unnecessary

spec/plane_spec.rb
```diff
@@ -17,14 +17,30 @@
...
+    expect(subject).to be_landed
+  end
+
+  it "can take off" do
+    is_expected.to respond_to(:take_off)
+  end
+
+  it "is flying after take off" do
+    subject.landed?
```
danldb: Line 41 is unnecessary

danldb:
:checkered_flag: 

Well done Lucy - that is a good pass!

I would look at using dependency injection for the weather functionality - that could be a good learning exercise for you. In terms of grasping the concepts of week one, I am very impressed :)

PR #153, opened 2015-09-06
trwh submission

trwh:


lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+
+  def initialize
+    @planes = []
+  end
+
+  def planes?
+    @planes
+  end
+
+  def receive(plane, weather)
```
roidriscoll: Here would be an indication for me, that `Weather` would be better as a module instead of a class. You would be able to then have access to the `stormy?` method without passing it as an argument which would clean up line 16 nicely :)

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+    fail 'The airport is full.' if full?
+    fail 'Plane cannot land due to bad weather conditions.' if weather.stormy?
+
+    plane.land
+    @planes << plane
```
roidriscoll: A small point, but the variable name for your collection of planes whilst logical on it's own, suffers by being terse in this application.collection_of_planes or hangar might clear this issue up. Naming things in coding is one of the more difficult and subtle skills to pick up but leads to extremely readable code in the long run!

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
+require_relative './weather.rb'
+
 class Airport
+  CAPACITY = 10
+
+  attr_reader :capacity
```
roidriscoll: This seems to be unused - surprised that hound didn't pick up on it...

lib/airport.rb
```diff
@@ -1,2 +1,37 @@
...
+
+  def initialize
+    @planes = []
+  end
+
+  def planes?
```
roidriscoll: Agree with hound on this one.

lib/plane.rb
```diff
@@ -1,2 +1,16 @@
 class Plane
+
+  def initialize
+    @flying = true
+  end
+
+  def flying?
```
roidriscoll: Again, better to use accessor methods than leave instance variables scattered around. Read this excerpt: https://github.com/makersacademy/course/blob/master/pills/wrapping_instance_variables_in_accessor_methods.md

lib/plane.rb
```diff
@@ -1,2 +1,19 @@
 class Plane
+
+  def initialize
+    @flying = true
```
roidriscoll: Assigning `@flying` here is redundant - you are doing it again on line 16 in a much more meaningful way, so you could refactor your initialize method as follows:
```
def initialize
  take_off
end
```
One more thing to add - since your `take_off` method changes state of the object, it is a convention to add a bang to the end of the method name: `take_off!` would be more idiomatic Ruby.

spec/airport_spec.rb
```diff
@@ -14,20 +15,61 @@
...
   describe 'take off' do
-    xit 'instructs a plane to take off'
 
-    xit 'releases a plane'
+    it 'instructs a plane to take off' do
+      expect(subject).to respond_to(:release)
+    end
+
+    it 'releases a plane' do
+      subject.receive(Plane.new, (double :good_weather, stormy?: false) )
```
roidriscoll: No doubt at this point you'll be aware of this, but here you are using a double to control the random factor in your test. However, there is an opportunity here to use a double in place of `Plane.new` so that you are testing your `Airport` class in isolation from other classes (namely Plane)!

spec/airport_spec.rb
```diff
@@ -14,20 +15,61 @@
...
-    xit 'instructs a plane to land'
 
-    xit 'receives a plane'
+    it 'instructs a plane to land' do
+      expect(subject).to respond_to(:receive).with(2).argument
+    end
+
+    # it 'receives a plane' do
+    #   expect(subject).to respond_to(:receive).with(1).argument
+    # end
```
roidriscoll: It's good practice to remove comments from code - try to reserve them for explaining unusual and unavoidable behaviour (usually stemming from dependency on an external service).

spec/airport_spec.rb
```diff
@@ -14,20 +15,61 @@
...
+    end
+
   end
 
   describe 'traffic control' do
+
+    context 'when the airport is empty' do
+
+      it 'raises an error if a plane is requested for release' do
+        expect{ subject.release(double :good_weather, stormy?: false) }.to raise_error 'No planes available.'
```
roidriscoll: You could reduce the length of this line by creating a `let` statement which gives you a `good_weather` double.

spec/plane_spec.rb
```diff
@@ -17,14 +17,34 @@
...
+  end
+
+  it 'can land' do
+    expect(subject).to respond_to(:land)
+  end
+
+  it 'is landed after landing' do
+    plane = Plane.new
+    plane.land
+    expect(plane).to_not be_flying
```
roidriscoll: Is there a reason why you've deviated from using `subject` here? There are arguments for and against each way, but it's important to be consistent throughout your tests.

roidriscoll:
Hi Tim, great work here - I am sure that Battleships week has already taught you more than this review will, but at the same time have a look at the points raised and feel free to refactor :)

:checkered_flag:

PR #154, opened 2015-09-06
Adrian Booth Airport Challenge 

Adrian1707:


lib/airport.rb
```diff
@@ -1,2 +1,44 @@
 class Airport
-end
+
+	DEFAULT_CAPACITY = 30
```
spike01: Remember to set your editor to 2 space tabs, not 4

lib/plane.rb
```diff
@@ -1,2 +1,24 @@
+require_relative 'airport'
+
 class Plane
-end
+
+	def initialize
+		@flying = true 
+		"Plane is flying"
```
spike01: Generally, try to avoid returning strings just to log stuff out in your final code - fine when you're initially building your code to help you see what's going on, but make sure you take them out, especially if they're in your `initialize` method! `initialize` should build your object and nothing else.

spec/airport_spec.rb
```diff
@@ -1,48 +1,56 @@
...
-#
-# No more planes can be added to the airport, if it's full.
-# It is up to you how many planes can land in the airport
-# and how that is implemented.
-#
-# If the airport is full then no planes can land
+  	# let(:plane) { Plane.new }
 
-describe Airport do
+  	let(:plane) {double :plane}
```
spike01: Be consistent with your spacing - below you have an extra space inside your braces and here you don't. Code is usually nicer to read and less exhausting if your formatting is the same throughout

spec/airport_spec.rb
```diff
@@ -1,48 +1,56 @@
...
+  	before(:each){ allow(plane).to receive(:land) }
+  	before(:each){ allow(plane).to receive(:take_off) }
+
+  	before do 
+  		allow(subject).to receive(:stormy?).and_return(false) #stubs out each method and keeps code DRY.
```
spike01: Remember to remove comments before submitting, unless your code is doing something that really really needs to be explained - in which case, it probably needs refactoring.

spec/airport_spec.rb
```diff
@@ -1,48 +1,53 @@
...
+  		allow(subject).to receive(:stormy?).and_return(false) #stubs out each method and keeps code DRY. 
+  	end 
+
+	describe "landing" do 
+
+		it { is_expected.to respond_to(:instruct_to_land).with(1).argument}
+		it { is_expected.to respond_to(:stormy?)}
+
+		it "planes array increase by 1 for each landing" do 
+			expect {subject.instruct_to_land(plane)}.to change{subject.plane_count.length}.by(1)
```
spike01: Again, formatting

spec/plane_spec.rb
```diff
@@ -1,30 +1,30 @@
...
 
-  xit 'is flying when created'
+	end
 
-  xit 'can land'
+	describe '#new' do
 
-  xit 'is landed after landing'
+		it 'Plane should be flying' do 
+			expect(Plane).to respond_to(:new)
```
spike01: This test isn't particularly meaningful - you could create the plane, and then check whether it's `flying?` in the test instead

spec/plane_spec.rb
```diff
@@ -1,30 +1,30 @@
...
-
 describe Plane do
+	
+		it { is_expected.to respond_to(:land)}
+		it { is_expected.to respond_to(:take_off)}
+		it { is_expected.to respond_to(:flying?)}
+
+		it "plane lands at airport and stops flying" do 
+			subject.land
+			expect(subject.flying?).to eql(false)
```
spike01: This is fine, but make sure you know the difference between the different equality matchers - if you have to check for true/false, you would normally use `expect(something).to be_truthy`

spec/airport_spec.rb
```diff
@@ -1,48 +1,56 @@
...
+		end 
 
-  describe 'take off' do
-    xit 'instructs a plane to take off'
+		it "Raise error when airport is full" do
+			subject.capacity.times{subject.instruct_to_land double(:plane, land: nil)}
```
spike01: Don't be afraid of splitting a line like this down into multiple lines if it helps legibility:
```ruby
subject.capacity.times do 
  subject.instruct_to_land double(:plane, land: nil)
end
```

spike01: Also, you can see it's a bit weird that you're returning a string from your `land` method, seeing as your double has to return `nil` - but good that you put those two things together

lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+	end 
+
+	def instruct_to_take_off(plane)
+		fail "Plane cannot take-off. Wait until wind slows down" if stormy? 
+		@planes.delete(plane)
+		plane.take_off 
+	end 
+
+	def plane_count
+		planes = @planes
```
spike01: Is there a reason that you have to reassign this? And does the method name really describe what this does? I would have expected this to be `@planes.count`

lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+	end 
+
+
+	def full?
+		@planes.count >= capacity 
+	end 
+
+	def stormy?
+		rand(1..100) >= 70
+	end
```
spike01: Is it really the responsibility of the airport to know what the weather is? Or could this maybe be split into something else...?

spike01:
Good challenge submission - a few minor slips and watch your formatting, but otherwise grab a sticker :checkered_flag:

PR #155, opened 2015-09-06
michael_roger

Benaud12:


lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+
+  DEFAULT_CAPACITY = 10
+
+  attr_reader :capacity, :planes, :allow_landing, :weather
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+    @allow_landing = true
+    @weather = Weather.new
```
roidriscoll: Inject this as an argument! Also, `Weather` should be a module :smile:

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+
+  private
+
+  def full?
+    return true if planes.count == capacity
+    false
+  end
+
+  def weather_update
+    @weather = Weather.new
```
roidriscoll: DEPENDENCY ALERT :rotating_light:

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+  end
+
+  def allow_take_off?
+    weather_update
+    return false if weather.stormy?
+    true
+  end
+
+  def receive_plane(plane)
+    planes << plane if allow_landing?
```
roidriscoll: A small point, but the variable name for your collection of planes whilst logical on it's own, suffers by being terse in this application.collection_of_planes or hangar might clear this issue up. Naming things in coding is one of the more difficult and subtle skills to pick up but leads to extremely readable code in the long run!

lib/plane.rb
```diff
@@ -1,2 +1,36 @@
...
+  end
+
+  def land(airport)
+    fail "Already landed" unless flying
+    fail "Airport refuses landing request" unless airport.allow_landing?
+    @flying = false
+    @location = airport
+    airport.receive_plane(self)
+  end
+
```
roidriscoll: This is quite complex for what should be a pretty simple class. Not necessarily a bad thing if you have introduced extra functionality, but I guess this is just a note to say be careful of over-engineering.

lib/weather.rb
```diff
@@ -0,0 +1,22 @@
+class Weather
+
+  attr_reader :sunny
+
+  def initialize
```
roidriscoll: It's a bit of a code smell to have a conditional statement in `initialize`, so to extract that out into a separate method would be preferable :smile:

spec/airport_spec.rb
```diff
@@ -1,48 +1,152 @@
...
+      allow(subject).to receive(:weather_update) {
+          @weather}
+    end
+
+    it 'allows planes to land' do
+      expect(subject).to be_allow_landing
+    end
+
+    it 'receives a plane' do
+      num_planes = subject.planes.count
```
roidriscoll: Don't abbreviate if you don't have to! Remember, as Rubyists we're striving to write code that a non-programmer could read.

spec/airport_spec.rb
```diff
@@ -1,48 +1,152 @@
...
+    before(:each) do
+      allow(subject.weather).to receive(:stormy?) {false}
+      allow(subject).to receive(:weather_update) {
+          @weather}
+    end
+
+    it 'releases specific plane' do
+      p1 = double :p1
+      p2 = double :p2
+      p3 = double :p3
```
roidriscoll: Even tests deserve good variable names!

lib/airport.rb
```diff
@@ -1,2 +1,52 @@
...
+    @allow_landing = false if full?
+  end
+
+  def release_plane(plane)
+    fail 'Plane not at this airport' unless planes.include?(plane)
+    return unless allow_take_off?
+    while planes.last != plane
+      planes.rotate!
```
roidriscoll: I enjoyed reading that.

lib/plane.rb
```diff
@@ -1,2 +1,36 @@
...
+  def initialize
+    @flying = true
+  end
+
+  def land(airport)
+    fail "Already landed" unless flying
+    fail "Airport refuses landing request" unless airport.allow_landing?
+    @flying = false
+    @location = airport
```
roidriscoll: In this case I think both `flying` and `location` could do with being `attr_accessor`s

roidriscoll:
Nice work Michael. I've left a few pointers for you - but a great start to the course.

:checkered_flag:

PR #156, opened 2015-09-06
Airport Challenge

dbatten4:


lib/airport.rb
```diff
@@ -1,2 +1,48 @@
 class Airport
+
```
student-progress-tracker-dev: Has approx 6 statements (TooManyStatements)

lib/airport.rb
```diff
@@ -1,2 +1,48 @@
...
+  def make_take_off(plane)
+    fail 'Plane has already taken off' if plane.flying?
+    fail 'The weather is too stormy' if stormy?
+    plane.take_off
+    @planes.delete(plane)
+    plane
+  end
+
+  private 
+
```
student-progress-tracker-dev: Doesn't depend on instance state (UtilityFunction)

tansaku:
Some great work here @dbatten4 - basics are very solid

Some tips:

*) remove all commented code before submission 
*) ensure your indentation is consistent or you will find it hard to debug and we will find it hard to find you a job - hiring partners really care about this
*) avoid having public attr_accessors and writers wherever possible - we must avoid allowing collaborator objects to directly access our internal state
*) pull weather into a module in order to reduce airport responsibilities

fix those up (and all the inline comments from hound etc.) and you can get the checkered flag 

:construction:

PR #157, opened 2015-09-06
D Bebb - Airport challenge

davidbebb:


spec/airport_spec.rb
```diff
@@ -1,48 +1,52 @@
...
-    # the plane can not land, and must not be in the airport
-
     context 'when weather conditions are stormy' do
-      xit 'does not allow a plane to take off'
-
-      xit 'does not allow a plane to land'
+      it 'does not allow a plane to take off' do
+        plane=double(:plane, land: :landed, take_off: :flying)
+        allow(subject).to receive(:report_weather).and_return(:stormy)
+        expect { subject.launch_plane(plane) }.to raise_error 'Weather is stormy, do not take off'
```
davidbebb: Line length due to error message string

spec/airport_spec.rb
```diff
@@ -1,48 +1,52 @@
...
+      end
+
+      it 'does not allow a plane to land' do
+        plane=double(:plane, land: :landed, take_off: :flying)
+        allow(subject).to receive(:report_weather).and_return(:stormy)
+        expect { subject.land_plane(plane) }.to raise_error 'Weather is stormy, do not land'
```
davidbebb: Line length due to error message string

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
 class Airport
+
+  attr_reader :capacity
+
+  def initialize
+    @capacity=5
```
roidriscoll: 5, is a magic number.
Yeah 5 (in this case), is a magic number.

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+    @planes_at_airport=[]
```
roidriscoll: Nice variable naming. Very readable. We like.

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+    @weather=:sunny
```
roidriscoll: Good use of a symbol. Also, PLEASE PUT SPACES AROUND '='. I BEG YOU.

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+    fail 'Weather is stormy, do not land' if report_weather==:stormy
+    fail 'Plane has already landed' if plane.status==:landed
+    plane.land
+    @planes_at_airport<<plane
+  end
+
+  def launch_plane(p)
+    fail 'Weather is stormy, do not take off' if report_weather==:stormy
+    fail 'Plane is not here' if @planes_at_airport == @planes_at_airport-[p]
+    p.take_off
```
roidriscoll: `p`? What is `p`?
If you were looking to make this method more readable, you could have:
`def launch(plane)`

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+    @planes_at_airport-=[p]
+  end
+
+  def report_weather
+    if rand(0..99)==0 then @weather=:stormy else @weather=:sunny end
```
roidriscoll: So a better way would be to put the weather in a module called `Weather`.
But if you were to continue along this route,  I would put the rand function in a method of its own, place `@weather` in an `attr_accessor` (because you want to avoid accessing state as much as possible) and then find a nicer way to put the remains on a single line. That said, better to go over to two lines if it avoids having a cumbersome conditional on one line.

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+  end
+
+  private
+
+  def full?
+    if (@planes_at_airport.length+1)>@capacity then true else false end
```
roidriscoll: Please don't put if/else statements on one line unless you are following this syntax:
` condition_to_be_checked ? do_this_if_true : do_this_otherwise`

lib/plane.rb
```diff
@@ -1,2 +1,20 @@
+
+
 class Plane
+
+  attr_reader :status
```
roidriscoll: `attr_accessor` would be more appropriate here. Your eye will soon become trained to spot a sea of `@`s a mile away.

spec/airport_spec.rb
```diff
@@ -1,48 +1,79 @@
...
-
   describe 'take off' do
-    xit 'instructs a plane to take off'
+    it 'instructs a plane to take off' do
+      expect(subject).to respond_to(:launch_plane)
+    end
 
-    xit 'releases a plane'
+    it 'releases a plane' do
+      plane=double(:plane, land: :landed, take_off: :flying, status: :flying)
```
roidriscoll: Put this in a `let` statement to clean up your code considerably!

spec/airport_spec.rb
```diff
@@ -1,48 +1,79 @@
...
+      allow(subject).to receive(:report_weather).and_return(:sunny)
+      subject.land_plane(plane)
```
roidriscoll: see how this line would read much more nicely if you removed `_plane` from the method name?

spec/airport_spec.rb
```diff
@@ -1,48 +1,79 @@
...
+      allow(plane).to receive(:status).and_return(:landed)
+      expect(subject.launch_plane(plane)).to eq([])
+    end
+
+    it 'does not release a plane that is not here' do
+      allow(subject).to receive(:report_weather).and_return(:sunny)
+      p1=double(:plane, land: :landed, take_off: :flying, status: :flying)
```
roidriscoll: `p1`? Tests need good variable naming too... :cry:

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+    @planes_at_airport=[]
+    @weather=:sunny
+  end
+
+  def land_plane(plane)
+    fail 'Do not land, airport is full' if full?
+    fail 'Weather is stormy, do not land' if report_weather==:stormy
+    fail 'Plane has already landed' if plane.status==:landed
+    plane.land
+    @planes_at_airport<<plane
```
roidriscoll: `@planes_at_airport` should be an `attr_accessor`

roidriscoll:
Hi David, good stuff - left some pointers for you, see what you make of it. I'm confident that you'll be able to refactor this to a very high standard!

:checkered_flag:

davidbebb:
Excellent - thanks Roi.

 Have a good weekend, read rest of this on Monday.

D


RE variable names - these were refactored to reduce line length in the face
of hound.

It might be worth considering more guidance on when to ignore hound (I'm
coming to the conclusion that line length is less important than
readability, so unless it's a line that's too big to fit on a screen,
ignore the 80 char guidline if it means clearer method/variable names - Is
this  a fair summary?).

PR #158, opened 2015-09-06
Challenge added

sarahkristinepedersen:


tansaku:
Great work getting this in @sarahkristinepedersen 

I think you can see some basic hound issues you need to fix above.  If you update and re-push to this branch you should be able to fix them.

A few other things not picked up by hound

*) indenting - you must indent your methods by two spaces inside classes - we can't get you a job if you don't :-/
*) you could try using a ternary statement or array shuffle in your weather module

but overall this is pretty damn solid.  If you want to get advanced you could try to minimize your reference to instance variables by using private attr_*s, but fix the hound issues and indenting and I'll give you a checkered flag ....

:construction:

PR #160, opened 2015-09-06
pull1

archiefielding:


tansaku:
some great work here @archiefielding 

To finish this up you just need to 

* avoid public attr_accessor and attr_writer - so that we don't expose internal state to collaborators
* don't indent private (and associated methods) that extra two spaces
* move weather into module for SRP
* avoid creating Planes in Airport --> use doubles

There's also a little room for DRYing out your boolean statements (if/unless) and your specs if you want to get advanced, but that's not required for the checkered flag.

Fix up the outstanding issues and you're good to go

:construction:

PR #161, opened 2015-09-06
Airport challenge submission 1

aaronkendall:


tansaku:
Hi Aaron,

Great work getting this in.  Some tips

1) when submitting delete all commented code
2) CONSTANTS are in capitals so Ruby knows they are constants
3) weather should be pulled out to a module so it's not the responsibility of the airport and airport can focus on planes
4) fix all the issues mentioned by hound above
5) remove this line from airport.rb `require '~/Projects/airport_challenge/lib/plane.rb'` you should use require_relative ...
6) you have a problem of a vacuous test  - check the plane spec with the two expect statements - one of them is not testing your application code.  can you tell which one?

but otherwise great job on stubbing the weather, nice identing (critical for getting a job)

Fix the above and you can have a checkered flag on this :-)

:construction:

PR #162, opened 2015-09-06
airport challenge

Proskurina:


lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+    fail "FULL" if full?
+    plane.land
+    planes << plane
+  end
+
+  def traffic_control
+    fail "The weather is stormy!" if weather_stormy?
+  end
+
+  def weather_stormy?
```
roidriscoll: Should `Airport` be responsible for the weather? I know we've talked about it this week, so just to reiterate, I think you would benefit greatly by extracting this out into a module.

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
 class Airport
+
+  attr_reader :planes
+  attr_reader :capacity
```
roidriscoll: Here you could save yourself a line:
`attr_reader :planes, :capacity`

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+
+  def initialize(capacity = 10)
```
roidriscoll: Here `10` is a magic number. Better to extract that value into a constant with some meaning, `DEFAULT_CAPACITY = 10` for example.

`def initialize(capacity = DEFAULT_CAPACITY)` then becomes much more readable!

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+
+  def plane_take_off
+    traffic_control
+    planes[0].take_off
+    planes.shift
+  end
+
+  def plane_landing(plane)
+    traffic_control
+    fail "FULL" if full?
```
roidriscoll: It's good to be a little more verbose with error messages: something like "Airport is full" would be much more useful from a user's perspective.

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+  attr_reader :capacity
+
+  def initialize(capacity = 10)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def plane_take_off
+    traffic_control
+    planes[0].take_off
```
roidriscoll: `planes.first.take_off` would read more nicely here :)

lib/airport.rb
```diff
@@ -1,2 +1,38 @@
...
+    planes.shift
+  end
+
+  def plane_landing(plane)
+    traffic_control
+    fail "FULL" if full?
+    plane.land
+    planes << plane
```
roidriscoll: A small point, but the variable name for your collection of planes whilst logical on it's own, suffers by being terse in this application. `collection_of_planes` or `hangar` might clear this issue up. Naming things in coding is one of the more difficult and subtle skills to pick up but leads to extremely readable code in the long run!

lib/plane.rb
```diff
@@ -1,2 +1,25 @@
 class Plane
+
+  attr_reader :fly
+
+  def initialize
+    @fly = true
```
roidriscoll: Assigning `@fly` here is redundant - you are doing it again on line 22 in a much more meaningful way, so you could refactor your `initialize` method as follows:
```
def initialize
  take_off
end
```

lib/plane.rb
```diff
@@ -1,2 +1,25 @@
...
+  def land
+    @fly = false
+  end
+
+  def landed?
+    ! flying?
+  end
+
+  def take_off
+    @fly = true
```
roidriscoll: It's important to avoid accessing state directly. Instead you should change `:fly` to be an `attr_accessor` and then `take_off` would read:
```
def take_off
  fly = true
end
```

lib/airport.rb
Proskurina: Thank you so much! All noted! 

I extracted Weather in my exp branch (pushed it to my repo on Monday morning). Will change it to a module.

roidriscoll:
Hi Yana, I have left some comments for you! Great work! :checkered_flag:

PR #163, opened 2015-09-06
Completed Airport Challenge

adellanno:


tansaku:
some great work here @adellanno 

some tips

* avoid attr_accessor being public so that our collaborator objects cannot directly manipulate our internal state
* always twospace indent code inside any chunk ending with 'end' - indenting is critical to you being able to debug, and its a big no no with the hiring partners
* move your weather code into a module for SRP
* avoid Plane.new in airport spec - prefer a double

fix up this and the hound issues and it's checkered flag time :-)

:construction:

PR #165, opened 2015-09-06
Alex Lemons airport_challenge

alexlemons1:


lib/airport.rb
```diff
@@ -1,2 +1,26 @@
...
+  def instruct_takeoff(plane)
+    plane.take_off
+    @planes.delete(plane)
+  end
+
+  def instruct_landing(plane)
+    @planes << plane
+    plane.flying = false
+    plane.current_airport = self
+    return nil
```
roidriscoll: So what hound means by this, is the `nil` is returned implicitly. If you aren't sure what this means I recommend searching for this term :)

lib/plane.rb
```diff
@@ -1,2 +1,32 @@
+require_relative 'airport'
+
 class Plane
+
+  attr_accessor :flying
+  attr_accessor :current_airport
+
+  def initialize
+    @flying = true
+    @current_airport = nil
```
roidriscoll: Instance variables are initialized when first assigned, so when you create `@current_airport` here, you're basically repeating half the work that line 21 does, when you assign `@current_airport = airport`

lib/plane.rb
```diff
@@ -1,2 +1,29 @@
...
+
 class Plane
+
+  def initialize
+    @flying = true
+    @current_airport = nil
+  end
+
+  def flying?
+    return @flying
```
roidriscoll: Again, implicit return on `@flying`.

lib/plane.rb
```diff
@@ -1,2 +1,32 @@
+require_relative 'airport'
+
 class Plane
+
+  attr_accessor :flying
+  attr_accessor :current_airport
```
roidriscoll: You could save yourself a line here by declaring the following:
`attr_accessor :flying, :current_airport`

lib/airport.rb
```diff
@@ -1,2 +1,26 @@
+require_relative 'plane'
+
 class Airport
+
+ attr_accessor :planes
```
roidriscoll: Indent this line

lib/plane.rb
```diff
@@ -1,2 +1,25 @@
...
+
+  def land(airport)
+    if @flying == false; fail "This plane has already landed"; end
+    @flying = false
+    @planes << self
+    return nil
+  end
+
+  def take_off
+    if @flying == true; fail "This plane is already airbourne"; end
```
roidriscoll: There is a better way to do this: http://www.thechrisoshow.com/2009/02/16/using-guard-clauses-in-your-ruby-code/

lib/plane.rb
```diff
@@ -1,2 +1,32 @@
...
+    @flying = false
+    airport.planes << self
+    @current_airport = airport
+    return nil
+  end
+
+  def take_off
+    if @flying == true; fail "This plane is already airbourne"; end
+    @flying = true
+    @current_airport.planes.delete(self)
```
roidriscoll: Be careful here. It may well be that your plane has too many responsibilities in this case.
Also here you are allowing a plane to reach into a data structure that would normally belong to an airport, and run methods on it.

spec/airport_spec.rb
```diff
@@ -13,14 +13,29 @@
...
-    xit 'releases a plane'
+  describe 'capacity' do
+    it 'should take one argument' do
+    expect(Airport).to respond_to(:new).with(1).argument
+    end
   end
 
+  describe 'take off' do
+    it 'instructs a plane to take off' do
+      plane = Plane.new
```
roidriscoll: Here it would be better to use a double for Plane, because otherwise we are not testing Airport in isolation.

spec/airport_spec.rb
```diff
@@ -13,14 +13,29 @@
...
+      airport = Airport.new
+      plane.land(airport)
+      subject.instruct_takeoff(plane)
+      expect(plane.flying?).to eql(true)
+      end
```
roidriscoll: So here we have a couple of issues - look at what you're testing specifically. You are looking at a method on Airport but your expectation is on the Plane. It would be better to see that you are checking on the internal state of the Airport to see a change as a result of your method, and also be checking that Airport has made a call to an external object.

spec/plane_spec.rb
```diff
@@ -17,14 +17,43 @@
...
-  xit 'can land'
-
-  xit 'is landed after landing'
-
-  xit 'can take off'
-
-  xit 'is flying after take off'
+  it 'is flying when created' do
+    plane = Plane.new
+    expect(plane.flying?).to eql(true)
```
roidriscoll: I can see here that you are using `subject` in some places, and `plane = Plane.new` in others. While there are arguments for each, the important thing here is to pick one, and stick with it for the duration of the codebase :)

roidriscoll:
Hi Alex, you have a few things to address here - please remember to look out for repeated examples of anything I have flagged here, as I try to mention things once and trust you to recognise it elsewhere, as opposed to me pointing out every violation :)

alexlemons1:
Hi Roi,

Thanks for the feedback. I have made some of the smaller changes. I feel
like whole thing needs remodelling, so I will make some of the larger
changes at a later date.

Thanks !

Alex

tansaku:
:construction:

PR #167, opened 2015-09-06
Matt Paul - Airport Challenge v1

matt-paul:
Hello,

I am a bit disappointed with this to be honest.  I got distracted creating an extremely over complicated weather class, which then greatly confused matters, and also had a general feeling of 'too many balls in the air', so one of my main takeaways is to try to work on keeping a logical mind when getting frustrated with a situation, checking one thing at a time and considering properly why something is (not!)functioning in the way I hope it to.

I certainly learned a fair amount though, and felt that some of my learning from the week was solidified, however I definitely need to come back to stubs and doubles with a clearer head.  

I have also learnt that sometimes it is probably best to start again rather than try to fix something that should be in the bin :-)

Matt


User Stories.txt
```diff
@@ -0,0 +1,15 @@
...
+So that I can set off for my specified destination
+I would like to be able to take off from the appropriate airport
+
+As an air traffic controller
+So that I can avoid collisions
+I want to be able to prevent airplanes landing when the airport if full
+
+As an air traffic controller
+So that I can avoid accidents
+I want to be able to prevent airplanes landing or taking off when the weather is stormy
```
danldb: use .md instead of .txt

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+  def initialize
+    @capacity = DEFAULT_CAPACITY
+    @planes = []
+  end
+
+  attr_reader :capacity, :planes  # => nil
+
+  def release_plane(plane)
+    fail "No take off allowed while stormy" if Weather.new.stormy?
+   planes.pop
```
danldb: watch your indentation

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+
+  def initialize
+    @capacity = DEFAULT_CAPACITY
+    @planes = []
+  end
+
+  attr_reader :capacity, :planes  # => nil
+
+  def release_plane(plane)
+    fail "No take off allowed while stormy" if Weather.new.stormy?
```
danldb: use dependency injection to uncouple airport from weather

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+    planes << plane
+  end
+
+  def full?
+    if planes.length >= @capacity
+      true
+    else
+      false
+    end
+  end
```
danldb: this whole method could be written as so:
```
def full?
  planes.length >= @capacity
end
```

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+
+  def stormy?
```

lib/airport.rb
```diff
@@ -1,2 +1,43 @@
...
+   "Cleared for takeoff"
+  end
+
+  def receive_plane(plane)
+    fail "Airport full" if full?
+    planes << plane
+  end
+
+  def full?
+    if planes.length >= @capacity
```

lib/plane.rb
```diff
@@ -1,2 +1,31 @@
...
+    @flying = true
+  end
+
+  def status?
+    if @flying
+      :airborne
+    else
+      :landed
+    end
+  end
```
danldb: unnecessary method

spec/airport_spec.rb
```diff
@@ -1,34 +1,42 @@
...
+    end
+
+
+    it 'receives a plane' do
+      expect(subject).to respond_to :receive_plane
+    end
+
+
+    it 'increases the number of planes within the airport when an aircraft lands' do
+      expect{subject.receive_plane(double :plane)}.to change{subject.planes.length}.by(1)
```
danldb: declare the plane beforehand instead of inline.

spec/airport_spec.rb
```diff
@@ -39,10 +47,21 @@
     # If the airport has a weather condition of stormy,
     # the plane can not land, and must not be in the airport
 
-    context 'when weather conditions are stormy' do
-      xit 'does not allow a plane to take off'
+  describe 'weather' do
 
-      xit 'does not allow a plane to land'
+    it 'can be stormy' do
+      expect(subject).to respond_to :stormy?
```
danldb: not really testing the weather

danldb:
:construction: 

This is not quite ready yet - I would recommend redoing this, either during lab week or alternatively during after the course.

PR #168, opened 2015-09-06
Basic plane functions and tests

MoeSadoon:


lib/plane.rb
```diff
@@ -1,2 +1,28 @@
...
+
+	def land
+		@flying = false
+	end
+
+	def take_off
+		@take_off = true
+	end
+
 end
```
danldb: Awesome - very simple, clean code. Well done!

lib/plane.rb
```diff
@@ -1,2 +1,28 @@
...
+		!@flying
+	end
+
+	def land
+		@flying = false
+	end
+
+	def take_off
+		@take_off = true
+	end
```
danldb: Set your indentation to 2 spaces instead of tabs

lib/plane.rb
```diff
@@ -1,2 +1,28 @@
+
+# Assume all newly creates planes are flying
```
danldb: This seems unnecesary as your code is self explanatory

spec/plane_spec.rb
```diff
@@ -1,30 +1,27 @@
...
-# Are you testing that?
 
 describe Plane do
 
-  xit 'is flying when created'
-
-  xit 'can land'
+it "is expected to fly when created" do
+	expect(subject.flying?).to be_truthy
+end
```
danldb: `expect(subject.flying?).to be true`

spec/airport_spec.rb
```diff
@@ -13,20 +12,22 @@
 
 describe Airport do
 
-  describe 'take off' do
-    xit 'instructs a plane to take off'
+  it 'instructs a plane to take off'
+  expect(subject).to respond_to :take_off("Plane")
```
danldb: 'expect(subject).to respond_to(:take_off).with(1).argument`

danldb:
:construction: 

Looks like you got stuck on a single test and then did not continue. Ideally in that situation you would isolate that single test and see if you can move on. I would recommend you to revisit Airport challenge at a later stage as there is still a lot that you have not covered.

PR #169, opened 2015-09-07
Airport Challenge Submission - Alex Person

AlexPerson:


tansaku:
some good work here @AlexPersons

Some issues:

* `require '~/projects/airport_challenge/lib/plane.rb'` is breaking your build - you need `require_relative`
* airport spec should use doubles instead of Plane.new
* weather and/or capacity should be added to make this have some meat

let's talk if you're unsure how to proceed

:construction:

PR #171, opened 2015-09-07
Week one - Airport Challenge

Harryandrew:


lib/airport.rb
```diff
@@ -1,2 +1,37 @@
+require_relative "plane"
+require_relative "weather"
+
 class Airport
+  include Weather
+  attr_accessor :capacity, :hangar, :planes
+
```
giusepped: Be careful when using attr_accessor and use it sparingly. For instance, you can already alter the default capacity of your airport when initializing the class so you don't need that here. 
Also, you don't want to give full access of your hangar to anyone.

README.md
```diff
@@ -1,6 +1,8 @@
 Airport Challenge
 =================
 
+Week one Makers Academy Airport Challenge.
+
 Instructions
 ---------
```
giusepped: Always try to update the README file maybe with "Get started" instructions on how to install your app on another computer (e.g. clone it and how to run it).

lib/plane.rb
```diff
@@ -1,2 +1,16 @@
 class Plane
+  attr_accessor :status
+
```
giusepped: Again, be careful with attr_accessor because it gives full access to your property to anyone. They could even change the class your status to a integer or a hash.

lib/plane.rb
```diff
@@ -1,2 +1,16 @@
...
+  end
+
+  def land
+    @status = :landed
+  end
+
+  def take_off
+    @status = :flying
+  end
+
```
giusepped: Maybe have an error that a plane that is already flying cannot take off and a plane that has already landed cannot be landed again (with the corresponding tests).

spec/airport_spec.rb
```diff
@@ -1,48 +1,81 @@
...
+    expect(subject.hangar).to eq []
+  end
+
+
+  describe "when a plane lands" do
 
-  describe 'take off' do
-    xit 'instructs a plane to take off'
+      it "it is landed" do
+        plane = Plane.new
```
giusepped: Because you are initializing a new Plane here, if you modify your Plane class, that will also affect your Airport tests. 

This is why it is useful to use a double for the plane. A double is a dummy object that you can use for testing. Initialize like this:
let(:plane) { double(:plane) }
Because the double does not have the methods of your plane class, you can allow it to receive those methods like this:
allow(plane).to receive(:land)

spec/airport_spec.rb
```diff
@@ -1,48 +1,81 @@
...
+        allow(subject).to receive(:stormy?).and_return(false)
+        allow(plane).to receive(:status).and_return(:landed)
```
giusepped: I see you have allowed plane to receive a method BUT plane in this case is not a double, it is an actual instance of Plane. Also, it is better to use the actual method "land" instead of changing :status. This right now works because you have an attr_accessor for status but it won't work when you take that out. 

So when you do use a double, use your "land" method instead:
´allow(plane).to receive(land)´

lib/plane.rb
```diff
@@ -1,2 +1,16 @@
 class Plane
+  attr_accessor :status
+
+  def initialize
+    @status = :flying
+  end
```
danldb: use a boolean `@flying = true`

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+module Weather
+  def stormy?
+    true if rand(21) == 1
+  end
+end
```
danldb: awesome - this is really great encapsulation.

lib/airport.rb
```diff
@@ -1,2 +1,37 @@
...
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @hangar = []
```
danldb: the convention for a list of planes would be to call it `planes`

lib/airport.rb
```diff
@@ -1,2 +1,37 @@
+require_relative "plane"
+require_relative "weather"
+
 class Airport
+  include Weather
```
danldb: could use dependency injection instead of including the module

danldb:
:checkered_flag: 

Great submission! Well done :)

PR #172, opened 2015-09-07
Airport Challenege Tests Pass - nathanielgreen

nathanielgreen:


lib/airport.rb
```diff
@@ -1,2 +1,58 @@
...
+  DEFAULT_CAPACITY = 20
+
+  attr_accessor :capacity, :planes
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def instruct_to_take_off
```
student-progress-tracker-dev: Has approx 6 statements (TooManyStatements)

tansaku:
some great work here @nathanielgreen 

Some tips:

*) make sure you fix all the issues mentioned above, in particular the line length - research "guard clauses" to see how you can slim down those long methods
*) avoid using attr_accessor unless it's private (we don't want to give that much control over our instance variables to our collaborator objects)
*) ideally you'd pull your weather component out into a module so we have single responsibilities for weather and airport
*) plane_spec could be slimmed down with a bit of be_truthy
*) you've got some other methods that could be very slimmed down e.g.

```
def flying?
    if @flying == true
      true
    else
      false
    end
  end
```

and I'm not sure why you are returning nil in 

```
  def land
    @flying = false
      return nil
  end
```
 
also please watch your indenting and make sure it is 2-space consistent or we will not be able to find you a job.

Fix the majority of the above and I can give you a checkered flag

:construction:

PR #173, opened 2015-09-07
Partial Solution for Airport Challenge

catherinestevenson:


tansaku:
some good work here @catherinestevenson 

some key changes to get this fixed up

* ensure two space indenting inside any chunk ending with 'end' - needs to be consistent to help you debug and for us to help find you a job - hiring partners really care about this
* plane shouldn't need to refer to airport internally
* DEFAULT_CAPACITY should be inside the airport class

let's look at this together if we can find a moment

:construction:

PR #174, opened 2015-09-07
Week 1 Challenge - Gavin

gavinwcheng:


lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+
+  def initialize
+    @capacity = CAPACITY
+    @planes = []
+    @instructed_to_take_off = false
+    @instructed_to_land = false
+  end
+
+  def sky_clear
+    if rand(2) == 1; @sky_clear = true else @sky_clear = false end
```
roidriscoll: As hound says, if you are trying to put an `if` statement on one line, instead of:
```
if rand(2) == 1; @sky_clear = true else @sky_clear = false
```
it would be better practice to instead say:
```
rand(2) == 1 ? @sky_clear = true : @sky_clear = false

File: lib/airport.rb [hunk 2]
```diff
@@ -1,2 +1,44 @@
+require_relative "plane"
+
 class Airport
+  CAPACITY = 2
```
roidriscoll: So you have an instance variable called `@capacity` and a constant called `CAPACITY` and you are putting one inside the other. Perhaps there is a more clear name for the constant? 

For instance, `@capacity = DEFAULT_CAPACITY` might read more nicely.

lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+  attr_reader :capacity
+
+  def initialize
+    @capacity = CAPACITY
+    @planes = []
+    @instructed_to_take_off = false
+    @instructed_to_land = false
```
roidriscoll: It isn't necessary to declare the `@instructed_to_take_off` or `@instructed_to_land` at point of initialisation, as you are declaring and assigning them in other methods.

lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+  end
+
+  def sky_clear
+    if rand(2) == 1; @sky_clear = true else @sky_clear = false end
+  end
+
+  def instruct_take_off
+    self.sky_clear
+    fail "Sky is not clear for take off" if @sky_clear == false
```
roidriscoll: this line could be wrapped in a method. You are making multiple calls to this same logic elsewhere so that would tidy it up nicely.

lib/plane.rb
```diff
@@ -1,2 +1,15 @@
 class Plane
+  attr_reader :flying
+
+  def initialize
+    @flying = true
+  end
```
roidriscoll: Note the repetition on line 13? You could refactor your `initialize` method by placing `take_off` on line 5.

lib/airport.rb
```diff
@@ -1,2 +1,44 @@
...
+    fail "No instruction for take off" if @instructed_to_take_off == false
+    @instructed_to_take_off = false
+    plane.take_off
+    @planes.delete plane
+  end
+
+  def instruct_landing
+    self.sky_clear
+    fail "Sky is not clear for landing" if @sky_clear == false
+    fail "Airport is full" if @planes.count >= capacity
```
roidriscoll: There's a large amount of repetition when it comes to `@sky_clear == false` - perhaps it would be good to wrap this check in a method?

roidriscoll:
Hi Gavin,

So there's still a way to go before this challenge can be closed but I have left some suggestions that should help you proceed. You mentioned in the standup that you weren't sure how to get the stubs working, but it seems that you've yet to get to that stage. One piece of advice that might help, would be for you to extract your weather functionality into a separate module. From there you can inject it into your Airport class as a dependency at point of initialisation - which will make stubbing much easier.

Until the challenge is complete, I will leave this here: :construction: - once you have completed the tasks, a coach can leave the finished emoji :smile:

PR #175, opened 2015-09-07
my first pull request 

TJQKAs:
my first pull request


tansaku:
Hi @TJQKAs 

Great work getting this in.  Don't worry too much about the 20 "violiations", at least I mean don't feel too bad - they are pretty easy to fix.  However, you do need to fix them.

Line length and indentation are very important in terms of getting an interview and then a job.  You MUST fix them. If you are unsure how, please ask.  In general it's just a question of following the pattern of two space indenting and moving your comments around, switch {} to do ... end where necessary and you'll be fine :-)

Don't use attr_accessor unless you can possibly avoid it - we've talked about minimizing the public interface of the class.

More seriously your tests are failing.  That definitely needs to be taken care of.  The key thing here is that you haven't stubbed the weather randomness to ensure that the tests are stable.   Do you have any idea how to do this?

:construction:

PR #183, opened 2015-10-16
Story amends

silvabox:
@tansaku @pitchinvasion @roidriscoll @sjmog  I've reworded the stories a little bit - I'm hoping they'll be clearer?


tansaku:
:+1:

sjmog:
I think these need to be even clearer, there are some slightly weird things here still. Do we need e.g. the one about planes taking off from airports that they are at?

PR #185, opened 2015-10-17
Airport_Challenge

DeathRay1977:


README.md
```diff
@@ -1,27 +1,44 @@
...
+* A Hangar (Collection of Planes). - Read only (but writable via public methods)
+
+Planes are very things. They have the following internal state.
 
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-* If you do not submit a pull request, we will not be able to see your progress
+* An ID - Such as "BA535" - Read Only.
```
danldb: This was not asked for in the brief

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :capacity, :name
+
+  def initialize(name, weather, capacity = DEFAULT_CAPACITY)
+    @name = name
+    @hangar = []
+    @capacity = capacity
+    @weather = weather
+  end
```
danldb: Good use of dependency injection - well done.

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
+require 'weather'
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :capacity, :name
+
+  def initialize(name, weather, capacity = DEFAULT_CAPACITY)
+    @name = name
```
danldb: Airport does not need a name

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    @hangar = []
+    @capacity = capacity
+    @weather = weather
+  end
+  def land plane
+    raise "Plane #{plane.id} Cannot Land. Already Landed!" if already_landed? plane
```
danldb: This error seems to be more associated with the internal status of the plane, rather than the status of the airport.

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    raise "Plane #{plane.id} Cannot Land. Airport Full!" if full?
+    raise "Plane #{plane.id} Cannot Land. Bad Weather!" if bad_weather?
+    plane.flying = false
```
danldb: directly amending the internal state of the plane. This is more a problem of the plane class (it shouldn't allow direct manipulation of its state) than the airport class, but you can see here an example of how dependencies spread problems from one class to another.

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    plane.location = name
```
danldb: This is unnecessary - not asked for in the brief.

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+  def take_off plane
+    raise "Plane #{plane.id} Cannot Take Off. Bad Weather!" if bad_weather?
+    raise "Plane #{plane.id} Cannot Take Off. Already Taken Off!" if already_flying? plane
+    raise "Plane #{plane.id} Cannot Take Off. Unknown Plane!" if !in_airport? plane.id
+    raise "Plane #{plane.id} Cannot Take Off. Wrong Airport #{plane.location}" if wrong_airport? plane
+    plane.location = nil
+    remove plane
+  end
+  def hangar
+    @hangar.dup.freeze
```
danldb: unnecessary freeze here

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    # Is a plane with this ID currently landed
+    # at this airport?
+    @hangar.any? {|plane| plane.id == id}
+  end
+  def remove plane
+    location = @hangar.find_index {|p| p.id == plane.id}
+    @hangar[location].flying = false
+    @hangar.delete_at location
+    not in_airport? plane.id
+  end
```
danldb: very complicated. Could be replaced with 
```ruby
hangar.delete(plane)
plane.take_off
```

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    in_airport?(plane.id) || !plane.flying
+  end
+  def already_flying? plane
+    plane.flying
+  end
+  def bad_weather?
+    @weather.stormy?
+  end
+  def wrong_airport? plane
+    plane.location != name
```
danldb: not asked for in the brief

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    not in_airport? plane.id
+  end
+  def full?
+    @hangar.count >= @capacity
+  end
+  def already_landed? plane
+    in_airport?(plane.id) || !plane.flying
+  end
+  def already_flying? plane
+    plane.flying
```
danldb: this is a pure dependency - it adds nothing to the class, just calls another object.

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+  def hangar
+    @hangar.dup.freeze
+  end
+
+  private
+  def in_airport? id
+    # Is a plane with this ID currently landed
+    # at this airport?
+    @hangar.any? {|plane| plane.id == id}
```
danldb: This works and is a good solution to the problem of finding a specific plane in an airport. However, this was never asked for in the brief.

lib/plane.rb
```diff
@@ -0,0 +1,9 @@
+class Plane
+  attr_reader :id
+  attr_accessor :flying, :location
```
danldb: You should rarely-almost-never use a public attr_accessor. It allows another class to replace :flying with a :banana if it so chooses.

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
+  def initialize
+    @possible_conditions = [:stormy, :sunny]
```
danldb: This should be a constant

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
+  def stormy?
+    weather_conditions == :stormy
+  end
```
danldb: good

spec/airport_spec.rb
```diff
@@ -0,0 +1,136 @@
+require 'airport'
+
+describe Airport do
+
+  let(:airport_name) { "LHR" }
+  let(:weather) { double(:weather) }
+  subject { Airport.new(airport_name, weather) }
+  let(:flying_plane) { double(:plane, :id => 'BA535', :flying => true) }
```
danldb: Airport should not have knowledge of the plane id, or the planes internal :flying status

spec/airport_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+  let(:grounded_plane) { double(:plane, :id => 'BA535', :flying => false) }
+  before(:each) do
+      allow(flying_plane).to receive(:flying=).with(anything).and_return(anything)
+      allow(flying_plane).to receive(:location=).with(anything).and_return(anything)
+      allow(weather).to receive(:stormy?).and_return(false)
+  end
+
+  context 'Showing the contents of the Hangar' do
+    it { is_expected.to respond_to(:hangar)}
```
danldb: redundant test

spec/airport_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+
+    it 'should return an array when asked for the contents of the hangar' do
+      expect(subject.hangar).to be_instance_of Array
```
danldb: This is quite strong and therefore brittle. There is a subtle difference between `be_instance_of` and `be_empty`. The former ensures that if you ever decide to use a hash (or a more complicated object) instead of an array, you have to change your test. The latter means the test will still work, as long as the object responds to `.empty?`.

danldb:
:construction: 

Thanks for the submission. You have really strong syntactic skills, and you seem to have really nailed TDD. Also great understanding of Dependency Injection. You failed in two big areas. The first is encapsulation - strive at all times to reduce the amount of knowledge that one class has of another class. The second big issue was not sticking to the brief. If you imagine that these challenges are like a tech test, the employer may be asking for something that will fit a specific interface - it is better not to make assumptions about their future needs. In this case it has led to you writing very overcomplicated code with many inter-class dependencies. That could mean the difference between passing a tech test and failing. 

Sorry - I know that brief comments in the code can come across harsh and that is not at all my intention. I can see that you are having fun with the code and seeing where it takes you, definitely keep doing that - just be aware that these challenges will look like tech tests in your portfolio, so do try to stick to the brief :)

Let me know if you want to discuss.

lib/airport.rb
DeathRay1977: Sorry. I misinterpreted the brief. When it says

```
As an air traffic controller
So that I can ensure safe take off procedures
I want planes only to take off from the airport they are at

```
I assumed that to mean that I had to be able to identify a specific plane and a specific airport. Otherwise
```
schipol = Airport.new
heathrow = Airport.new
jumbo = Plane.new
jumbo2 = Plane.new

schipol.land(jumbo)
heathrow.land(jumbo2)
heathrow.take_off(jumbo)
```
would work.

PR #188, opened 2015-10-17
Tom Barnes' Airport Challenge

forty9er:


lib/plane.rb
```diff
@@ -0,0 +1,52 @@
+require_relative './weather.rb'
+
+class Plane
+
+  include Weather
+
+  attr_reader :location
+  attr_accessor :flying
```
yevdyko: Improper use of attr_accessor

lib/airport.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class Airport
+
+  include Weather
+
+  attr_reader :name, :planes, :capacity
+
+  DEFAULT_CAPACITY=10
+
+  def initialize name, capacity=DEFAULT_CAPACITY
```
danldb: Look into Ruby's 'keyword arguments' syntax.

lib/airport.rb
```diff
@@ -0,0 +1,21 @@
+require_relative './weather.rb'
+
+class Airport
+
+  include Weather
```
danldb: Use Dependency Injection to add your module, instead of using a mixin.

lib/airport.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  attr_reader :name, :planes, :capacity
```
danldb: Do `planes` and `capacity` need to be public methods?

lib/airport.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def initialize name, capacity=DEFAULT_CAPACITY
+    @name=name
+    @planes=[]
+    @capacity=capacity
+  end
+
+  def full
+    @planes.length>=capacity ? true : false
+  end
```
danldb: `#full` should be a private method

lib/plane.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def initialize
+    @location=nil
+    @flying=true
+  end
+
+  def land airport
+    if @flying==true
+      raise 'Cannot land while airport is full' if airport.full
```
danldb: This error depends upon the internal state of the airport - that is knowledge the plane class should not have.

lib/plane.rb
```diff
@@ -0,0 +1,52 @@
...
+      raise 'Cannot land while weather is stormy.' if airport.weather==:stormy
```
danldb: Again, this guard clasue should exist in the airport, not the plane.

I would recommend doing some research on the idea of 'Encapsulation in OOD'. The basic idea is that objects should know as little as possible about other objects - thus reducing dependency. The overall result of this is modular, reusable and maintainable code.

lib/plane.rb
```diff
@@ -0,0 +1,52 @@
...
+      complete_landing(airport)
+    else 
+      'This plane is already on the ground.'
```
danldb: Don't return a string, raise an error.

lib/plane.rb
```diff
@@ -0,0 +1,52 @@
...
+    end 
+  end
+
+  def complete_landing airport
+    @location=airport.name
```
danldb: There was no need for a location variable, you are going beyond the brief.

lib/plane.rb
```diff
@@ -0,0 +1,52 @@
...
+    @flying=false
+    airport.planes << self
+    "The plane has landed at #{airport.name}"
```
danldb: don't return strings - that would be saved for a user interface, not for an internal object.

lib/plane.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+    else  
+    'The plane is already flying'
+    end  
+  end
+
+  def complete_take_off airport
+    @location=nil
+    @flying=true
+    airport.planes.delete(self)
```
danldb: Breaking 'law of demeter'. Also directly modifying internal variables of another object - that is like reaching right in and playing with its privates!

lib/weather.rb
```diff
@@ -0,0 +1,8 @@
+module Weather
+
+  def weather
+    weather=rand(4)
+    weather==1 ? :stormy : :sunny
+  end
+
+end
```
danldb: AWESOME - love the use of a module for this.

spec/plane_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+describe Plane do 
+
+  describe 'initialization' do
+    it { is_expected.to respond_to (:location) }
+
+    it { is_expected.to respond_to (:flying) }
+  end
+
+  describe '#take_off' do
+    it { is_expected.to respond_to (:take_off) }
```
danldb: respond_to tests are redundent

spec/weather_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  describe '#weather' do
+  
+    it 'is expected to be sunny or stormy' do
+      weather_options=[:sunny, :stormy]
+      expect((weather_options).include?(@dummy.weather)).to be true
+    end 
+  end
+end
```
danldb: This is not how I would test the module - but as you are using it as a mixin, then you do need to mix it in.

danldb:
:construction: 

Thanks for the submission Tom. You have really got well developed Ruby skills - your are writing clean, concise code already, at the end of week 1 - well done! There are a bunch of OOD principles that you are violating (you have probably never seen or heard of them before, so don't be too concerned!). Use this week to focus on OOD - that is, SOLID principles and Encapsulation. Things will start to fall into place more once you start to recognise these patterns.

PR #189, opened 2015-10-17
dan's airport challenge

d9nny:


lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+require_relative 'weather'
+require_relative 'plane'
+
+class Airport
+	DEFAULT_CAPACITY = 10
+
+	attr_reader :capacity
```
danldb: I don't think this should be a public attr_reader - you could use a private reader instead

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+		raise 'Too stormy to take off' if stormy_weather?
+		plane.has_taken_off; @planes.delete plane if airport_is_holding(plane) 
+	end
+
+	private
+
+	attr_reader :planes
+			
+		def full?
+    		@planes.count >= @capacity
```
danldb: you have an attr_reader for @capacity and @planes, so you should use them:

```ruby
def full?
  planes.count >= capacity
end
```

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+    	end
+
+    	def stormy_weather? 
+    		Weather.new.stormy? == true
+    	end
```
danldb: Use dependency injection for your weather:

```ruby
def initialize(weather: Weather.new)
  @weather = weather
end

def stormy_weather?
  weather.stormy?
end
```

The advantage of this is if you later write an amazing upgrade for the weather class, you don't need to change your airport code - you just add an argument at the point of initialize.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    	def airport_is_holding plane
+    		@planes.include? plane
+    	end
```
danldb: Not sure of the relevance of this method. Seems unnecessary.

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+	end
+
+	def has_landed
+		@flying = false 
+	end
+
+	def has_taken_off
+		@flying  = true
+	end
+end
```
danldb: This looks great, and is exactly what I was hoping to see in a Plane class. In terms of your naming, ideally use the imperative mood (like giving a command) for methods. You don't need to be polite `#flying?` would be better than `#is_flying?`, `#land` is better than `#has_landed`, `#take_off` is better than `#has_taken_off`.

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+		@flying = true
+	end
+
+	def is_flying?
+		@flying
+	end
+
+	def has_landed
+		@flying = false 
+	end
```
danldb: Do youneed some guard clauses for your edge cases? What if the plane has already landed?

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
...
+	end
+
+	def stormy?
+		@random_outlook == :stormy
+	end
+	
+	private
+	
+	OUTLOOKS = [:stormy, :sunny, :sunny, :sunny]
+end
```
danldb: Looks great

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'airport'
+
+describe Airport do 
+	let(:plane) { double :plane }
+	let(:weather) { double :weather}
+	subject(:airport) { described_class.new }
+	subject(:airport1) { described_class.new(100) }
```
danldb: Love seeing the explicit subject. You could be _more_ explicit i.e. `big_airport` instead of `airport1`.

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'airport'
+
+describe Airport do 
+	let(:plane) { double :plane }
+	let(:weather) { double :weather}
```
danldb: You never use this weather double. Ideally you would use dependency injection, so you could actually pass the double instead of Weather.new.

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+        context "When a capacity is not specified" do
+            it "it uses DEFAULT_CAPACITY" do
+      			expect(airport.capacity).to eq(Airport::DEFAULT_CAPACITY)
+            end
+        end
+    end
+
+    before(:each) do
+		    allow(subject).to receive(:stormy_weather?).and_return(false)
+			allow(plane).to receive(:is_flying?).and_return true
```
danldb: Fix your indentation

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+		it "prevents a plane landing when its not flying" do
+			allow(plane).to receive(:is_flying?).and_return false
+			expect{subject.land(plane)}.to raise_error "Plane is not flying"
+		end
+		it "prevents a plane landing when the airport is full" do
+			allow(subject).to receive(:full?).and_return(true)
+			expect{subject.land(plane)}.to raise_error "Airport is full"
+		end
+		it "prevents a plane landing when the weather is stormy" do
+			allow(subject).to receive(:stormy_weather?).and_return(true)
```
danldb: avoid stubbing subject - use Dependency Injection instead

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+			allow(plane).to receive(:has_taken_off)
+		end	
+
+		it "instructs a plane to take_off only if its in the airport" do
+			allow(plane).to receive(:has_landed)	
+			subject.land(plane)
+			allow(plane).to receive(:is_flying?).and_return false
+			allow(plane).to receive(:has_taken_off)
+			subject.take_off(plane)
+			expect(subject.send(:planes)).to eq []
```
danldb: You have a lot of `allow` in here. That is a sign that maybe you have coupled your classes too much.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+
+	attr_reader :capacity
+
+	def initialize(capacity = DEFAULT_CAPACITY)
+		@capacity = capacity
+		@planes = []
+	end
+
+	def land plane
+		raise 'Plane is not flying' if !plane.is_flying?
```
danldb: line 15 is actually a guard clause for the plane class, and should be in the Plane class.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+		raise 'Airport is full' if full?
+		raise 'Too stormy to land' if stormy_weather?
+		plane.has_landed; @planes << plane
```
danldb: Avoid semi-colon in ruby. You'll get plenty of semi-colon fun in JavaScript ;-)

spec/plane_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+describe Plane do
+	
+	context "On initialise" do
+		it 'the plane starts flying' do
+		    expect(subject.is_flying?).to eq true
+		end
+	end
+
+	it 'has_landed changes flying to false' do
+		subject.has_landed
```
danldb: indentation

spec/weather_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'weather'
+
+describe Weather do
+	
+	it "It returns whether it is stormy or not" do
+		allow(Weather::OUTLOOKS).to receive(:sample).and_return(:stormy)
+		expect(subject.stormy?).to eq true
+	end
```
danldb: Vacuous test. It is very hard to test Random, and often not worth it. That is one of the reasons we isolate this behaviour into a separate object.

danldb:
:construction: 

Good submission - excellent Plane class, you managed to keep it really simple. Overcomplicated the Airport class a little bit. Good work extracting the weather into a separate object, you just needed one more step (dependency injection) for this to be flawless.

One thing that is very important - please watch your indentation. It is set to 4 spaces at the moment and is hard to read in certain text editors. Set indentation to 2 spaces by default. Make sure you correctly tab inside methods - weird indentation looks messy and can literally mean the difference between getting an interview and not getting an interview.

PR #190, opened 2015-10-17
first 3 statements

jrose111284:


lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  attr_reader :capacity, :planes
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+   @planes = []
+   @capacity = capacity
+
+  end
+
+  def land_safely(planes, working=true)
```
danldb: should `planes` be `plane`? `working = true` is doing nothing right now.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+   raise 'airport is full' if full? || stormy?
+   @planes << planes
+  end
+
+  def full?
+   @planes.count >= @capacity
+  end
+
+  def take_off_safely(plane)
```
danldb: Your argument is doing nothing at the moment. The method works perfectly well without it, so maybe you could just remove the argument.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+   r.rand(10...20)
+   if r == 12
+     true
+   else
+   false
+   end
+  end
+
+  def flying?
+    true
```
danldb: hard coded - so your plane is ALWAYS flying, and can never land.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def non_flying?
+    true
```
danldb: yet at the same time, your plane is always landed and never flying. Your plane is a paradox.

lib/planes.rb
```diff
@@ -0,0 +1,11 @@
+class Planes
```
danldb: should be singular

lib/planes.rb
```diff
@@ -0,0 +1,11 @@
...
+  attr_writer :working
+  def initialize
+   @working = true
+  end
+  
+ def working?
+   working
+ end
```
danldb: Watch your indentation - this is really important if you want to be a professional writer of code. Exactly like you would not expect a professional writer to make a mess of their page structure, a professional coder makes their code easy to read. Indentation is really important.

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  describe '#land_safely' do
+   it 'raises a error when theres no room to land' do
+   AirPort::DEFAULT_CAPACITY.times do
+   subject.land_safely Planes.new
+  end
+
+  expect { subject.land_safely Planes.new }.to raise_error 'airport is full'
+    end
+  end
```
danldb: Are you missing an `it 'does something' do ` line?

danldb:
:construction:

Thanks for the submission Jay. There were a lot of issues with this. My advice would be to redo this challenge from the beginning - HOWEVER don't do it now. Wait until Lab Week to revisit this, but definitely _do_ revisit it.

PR #191, opened 2015-10-17
Andrew Dowell (Ronin)

k0zakinio:


lib/airport.rb
```diff
@@ -0,0 +1,45 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+DEFAULT_CAPACITY = 20
+
+  attr_reader :capacity, :airport_name
```
danldb: :airport_name is unnecessary and goes beyond the brief

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def initialize(airport_name='undefined', capacity=DEFAULT_CAPACITY,conditions=Weather.new)
```
danldb: https://robots.thoughtbot.com/ruby-2-keyword-arguments

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    @hangar = []
+    @capacity = capacity
+    @airport_name = airport_name
+    @conditions = conditions
+  end
+
+  def land(plane)
+    raise "Stormy weather!  Try again later." if condition_check == :stormy
```
danldb: could have a private `#stormy?` method that returns true or false, instead of condition_check == :stormy

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    raise "Airport is full!" if full?
+    plane.land
+    plane.update_location @airport_name
+    @hangar << plane
+  end
```
danldb: Great, apart from not needing location.

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def takeoff(plane)
+    raise "Stormy weather!  Try again later." if condition_check == :stormy
+    raise "Plane is not in this airport, either land here first or add a new plane to the airport!" unless @hangar.include?(plane)
+    plane.takeoff
+    @hangar.delete_if {|elem| elem == plane }
```
danldb: You could just use `@hangar.delete(plane)`

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def add_new_planes(plane)
+    plane.update_location(@airport_name)
+    @hangar << plane
+  end
```
danldb: This method is unnecessary.

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  private
+  def full?
+    @hangar.length >= @capacity
+  end
+
+  def condition_check
+    @conditions.weather_check
+  end
```
danldb: good, but return a boolean instead of a symbol. Booleans have many uses built-in, many methods are built to understand booleans. Arbitrary symbols don't have access to anywhere near that built-in functionality.

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
...
+    @flying = false
+    @location = nil
+  end
+
+  def flying?
+    @flying
+  end
+
+  def takeoff
+    fail "Already flying!" if @flying
```
danldb: You have a  reader method `#flyng?` so use it. This line should read:

`fail "Already flying!" if flying?`

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def land
+    fail "Already on the ground!" unless @flying
+    @flying = false
+  end
+
+  def update_location(location)
+    @location = location
+  end
```
danldb: Update location is not needed.

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
...
+  attr_reader :current_weather
+
+    def initialize
+      @current_weather = weather_check
+    end
+
+    def weather_check
+      Random.rand(101/PERCENT_CHANCE_OF_STORMS).zero? ? :stormy : :sunny
+    end
+end
```
danldb: Excellent

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
+  PERCENT_CHANCE_OF_STORMS = 10
+
+  attr_reader :current_weather
```
danldb: I don't think you need current_weather - why not just call weather_check?

spec/airport_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'airport'
+
+describe Airport do
+  let (:plane1) { double(:plane, land: false, takeoff: true, update_location: "undefined")}
+  let (:plane2) { double(:plane, land: false, takeoff: true, update_location: "undefined")}
+  let (:weather) { double(:weather, weather_check: :sunny)}
```
danldb: awesome - great use of weather double

spec/airport_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'airport'
+
+describe Airport do
+  let (:plane1) { double(:plane, land: false, takeoff: true, update_location: "undefined")}
+  let (:plane2) { double(:plane, land: false, takeoff: true, update_location: "undefined")}
```
danldb: Why not just use 1 plane?

spec/airport_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  let (:weather) { double(:weather, weather_check: :sunny)}
+  let (:test_capacity_30) { Airport.new('test_capacity', 30) }
+  let (:test_full) { Airport.new('test_full',0) }
```
danldb: these should be subjects in specific contexts. i.e. `subject(:full_airport){ Airport.new('test_full', 0) }`

spec/airport_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  before (:each) { allow(subject).to receive(:condition_check).and_return(:sunny)}
```
danldb: Noooooo you should not be stubbing subject! You were doing so well with the dependency injection, you shouldn't need to stub subject.

spec/airport_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  before (:each) { allow(test_capacity_30).to receive(:condition_check).and_return(:sunny)}
+  before (:each) { allow(test_full).to receive(:condition_check).and_return(:sunny)}
```
danldb: Not sure what these are doing?

spec/airport_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  describe '#land' do
+    it { is_expected.to respond_to(:land).with(1) }
```
danldb: redundant respond_to

spec/airport_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+  end
+
+  describe '#stormy_weather?' do
+    it 'prevents planes from landing' do
+      allow(subject).to receive(:condition_check).and_return(:stormy)
+      expect{subject.land(:test)}.to raise_error "Stormy weather!  Try again later."
+    end
+    it 'prevents planes from taking off' do
+      allow(subject).to receive(:condition_check).and_return(:stormy)
```
danldb: don't stub subject, just inject a weather double

danldb:
:checkered_flag:

Good job. You went beyond the brief with location. Good attempt at Dependency Injection, which you got almost perfect, but then failed to inject your weather double correctly. In general this was a strong submission.

PR #192, opened 2015-10-17
Airport challenge

blancopado:


lib/plane.rb
```diff
@@ -0,0 +1,15 @@
+class Plane
```
yevdyko: There are no unit tests for this methods,

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+		
+		# it 'refuses to land planes if weather is stormy' do
+		# 	expect {subject.land(plane)}.to raise_error("There is a storm")
+		# 	expect {subject.land(plane)}.not_to raise_error
+		# end
+	end
+	
+	describe '#take_off' do
+		#let(:plane) { double(:plane) }
+		it {is_expected.to respond_to(:take_off).with(1) }
```
yevdyko: Redundant respond_to test.

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+	
+	describe '#initialize' do
+	 	it 'has a default capacity of 50 planes' do
+	 		expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+	 	end
+	end
+
+	describe '#land' do
+		#let(:plane) { double(:plane) }
+		it {is_expected.to respond_to :land }
```
yevdyko: Redundant respond_to test.

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+		
+		it 'refuses to land planes if the airport is full' do
+			plane = double(:plane, landed: false)
+			described_class::DEFAULT_CAPACITY.times { subject.land(plane) }
+			expect {subject.land(plane)}.to raise_error("The airport is full")
+		end
+		
+		# it 'refuses to land planes if weather is stormy' do
```
yevdyko: Commented-out code.

lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+# class Weather
```
yevdyko: Commented-out code

spec/feature_spec.rb
```diff
@@ -0,0 +1,77 @@
+require './lib/plane'
+require './lib/airport'
+
+
+#puts "First feature test: As an air traffic controller
```
yevdyko: Commented-out code

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+	attr_reader :capacity, :planes_array
+	DEFAULT_CAPACITY = 50
+
+	
+	def initialize(capacity=DEFAULT_CAPACITY)
+		@capacity = capacity
+		@planes_array = [] # we can hold 50 planes in this array	
+	end
+
+	def land(plane)
```
yevdyko: The solution contains redundant code

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'plane'
+
+class Airport
+	attr_reader :capacity, :planes_array
+	DEFAULT_CAPACITY = 50
+
+	
+	def initialize(capacity=DEFAULT_CAPACITY)
+		@capacity = capacity
+		@planes_array = [] # we can hold 50 planes in this array
```
danldb: convention for a list of planes would be to call it `@planes`

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'plane'
+
+class Airport
+	attr_reader :capacity, :planes_array
```
danldb: I don't think you need a public attr_reader for capacity

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+		@capacity = capacity
+		@planes_array = [] # we can hold 50 planes in this array	
+	end
+
+	def land(plane)
+		fail "The airport is full" if @planes_array.length >= capacity
+		fail "There is a storm" if stormy
+		plane.landed
+		@planes_array << plane
+	end
```
danldb: Great `#land` method

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+
+	def take_off(plane)
+		fail "The plane did not land on this airport" if !@planes_array.include?(plane)
+		fail "A flying plane cannot take off" if plane.flying
```
danldb: this `fail` is actually related to the plane, and should live in the plane class.

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+		plane.fly
+		@planes_array.select { |i| i == plane }.pop
+	end
+
+	def stormy
+		random = Random.new
+		num = random.rand(1..100)
+		num >= 95
+	end
```
danldb: you could write this as 
```ruby
def stormy?
  rand(1..100) >= 95
end
```

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+#  p plane = Plane.new
+
+#  p airport = Airport.new(20)
+
+#  p airport.land(plane)
+
+#  p airport.take_off(plane)
+
+
+# p airport.stormy
```
danldb: Do not commit commented out code - always remove comments before you commit.

lib/plane.rb
```diff
@@ -0,0 +1,15 @@
...
+
+	def initialize
+		@flying=true
+	end
+	
+	def fly
+		@flying = true
+	end
+
+	def landed
```
danldb: For method names, use the imperative form (like giving a command) instead of past tense. In this case, `def land` would be better than `def landed`.

lib/plane.rb
```diff
@@ -0,0 +1,15 @@
+class Plane
+	attr_reader :flying
```
danldb: As flying is a boolean value, it is better to write your attr_reader by hand:
```ruby
def flying?
  @flying
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+		 	allow(subject).to receive(:stormy).and_return(true)
+		 	expect {subject.land(plane)}.to raise_error("There is a storm")
+		end
+	end
+		# it 'refuses to take off a plane that is flying' do
+		# 	plane = double(:plane, fly: true)
+		# 	#allow(plane).to receive(:fly).and_return(true)
+		# 	#allow(subject).to receive(land(plane))
+		# 	expect {subject.take_off(plane)}.to raise_error("A flying plane cannot take off")
+		# end
```
danldb: You are running into problems because that error should live in the plane class. So to raise this error, you have to create a plane double that returns `flying: true`. You did this, but then you did not actually land that plane in the airport, so the error would not be raised. I would not try and fix this - instead, raise the error in your plane class.

spec/plane_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'plane'
+
+describe Plane do
+	describe '#fly' do 
+		it {is_expected.to respond_to :fly }
+	end
+	describe '#landed' do
+		it {is_expected.to respond_to :landed }
+	end
```
danldb: I think you could go a bit further with this. I woudl like to see that you have tested the correct return values of your methods, instead of just testing respond_to.

danldb:
:construction: 

You kept it simple, which is really the main point of this challenge, so well done! This was a good attempt. There are a couple of things to amend, but i would wait until Lab week to revisit, not urgent to relook at this.

PR #201, opened 2015-10-18
Rob's airport challenge changes

youngmanr:


spec/plane_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  subject { Plane.new }
+  it 'A flying plane cannot take off' do
+    allow(airport).to receive(:request_take_off).with(subject)
+    expect{ subject.take_off(airport) }.to raise_error 'Already flying'
+    expect( subject.airport ).to be_nil
+  end
+
+  let(:airport) { double(:airport) }
+  let(:airport2) { double(:airport) }
+  subject { Plane.new }
```
blancopado: It's not necessary to set subject Plane on 'describe Plane'

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
+class Airport
+  DEFAULT_CAPACITY = 1
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @planes = Array.new
```
danldb: I would prefer `@planes = []` but that is really just a style thing.

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
...
+    @capacity = capacity
+  end
+
+  def request_landing( plane)
+    raise "Airport is full" if full?
+    @planes.push plane
```
danldb: Again, I would probably prefer `@planes << plane` but it is purely a style thing - this code functions fine.

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
+class Airport
+  DEFAULT_CAPACITY = 1
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @planes = Array.new
+    @capacity = capacity
```
danldb: good use of default capacity

lib/plane.rb
```diff
@@ -0,0 +1,30 @@
+require_relative "airport"
+
+class Plane
+  attr_reader :airport
+
+  def initialize()
+    @flying = true
+    @airport = nil
```
danldb: Plane doesn't need knowledge of airport

lib/plane.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def land( airport)
+    raise "Not flying" if !flying?
```
danldb: `plane.land(airport)` semantically reads as though the airport is flying around and the plane is being commanded to land it. Fails the sanity test.

lib/plane.rb
```diff
@@ -0,0 +1,30 @@
...
+    airport.request_landing(self)
+    @flying = false
+    @airport = airport
+  end
+
+  def take_off(airport)
+    raise "Already flying" if flying?
+    raise "Plane cannot take off from requested airport" if (@airport != airport)
```
danldb: This guard clause would be unneccesary of the plane had no knowledge of airport.

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def request_take_off( plane)
+    raise "Plane is not at airport" if !@planes.include?plane
+    @planes.delete(plane)
+  end
+
+  private
+
+  def full?
```
danldb: good use of private

spec/plane_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'plane'
+
+describe Plane do
+  it { is_expected.to respond_to :land}
+  it { is_expected.to respond_to :take_off}
```
danldb: These respond_to tests are useful when building the class, but can be removed once you have more tests on those same methods.

danldb:
:construction: 

Really good effort Rob. I can't award a pass just yet as you have not yet implemented the weather, but I am happy with your progress. Take care with your encapsulation -  you made good use of private methods to hide the internal workings of your objects, but in the end your plane had too much knowledge of the airport - ideally the plane would have no mention of airport anywhere.

PR #212, opened 2015-10-19
Week 1. Airport Challenge

yevdyko:


spec/plane_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      expect(subject.status).to eq :flying
+    end
+
+    it 'reports a default plane location' do
+      expect(subject.location).to eq :air
+    end
+  end
+
+  describe '#land' do
+    it 'is landing' do
```
forty9er: Redundant respond_to test

spec/plane_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      expect(subject.status).to eq :landed
+    end
+
+    it 'reports a plane location after landing' do
+      subject.land
+      expect(subject.location).to eq :airport
+    end
+  end
+
+  describe '#take_off' do
```
forty9er: Redundant respond_to test

spec/airport_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(subject.capacity).to eq described_class::DEFAULT_CAPACITY
+    end
+
+    it 'has a variable capacity when specified' do
+      airport = Airport.new(50)
+      expect(airport.capacity).to eq 50
+    end
+  end
+
+  describe '#land' do
```
forty9er: Redundant respond_to test

spec/airport_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect { subject.land(plane) }.to raise_error 'Plane can\'t land. Airport is full!'
+    end
+
+    it 'raises an error when a plane makes landing in stormy weather' do
+      allow(subject).to receive(:stormy?) { true }
+      expect { subject.land(plane) }.to raise_error 'Plane can\'t land. Weather is stormy!'
+    end
+  end
+
+  describe '#take_off' do
```
forty9er: Redundant respond_to test

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
+class Plane
+  attr_accessor :status, :location
```
forty9er: exposed internal implementation

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+class Airport
+  DEFAULT_CAPACITY = 25
+
+  attr_accessor :capacity
```
forty9er: exposed internal implementation

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @planes.pop
+  end
+
+  private
+
+  def full?
+    @planes.length >= capacity
+  end
+
+  def stormy?
```
forty9er: Airport class should not have responsibility for generating weather conditions.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    rand(0..100) > 95
+  end
```
danldb: agreed - you should use dependency injection to inject a weather object into the airport

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def wrong_airport?(plane)
+    @planes.include?(plane)
+  end
```
danldb: unnecessary method

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def initialize
+    @status = :flying
+    @location = :air
+  end
+
+  def land
+    @status = :landed
+    @location = :airport
+  end
```
danldb: so you have too many ways to change state. Your attr_accessor provides full access to your instance variables, and your land method does not make use of your accessor. `@location` is completely unnecessary. `@status` would be better if it was `@landed` and was true/false. `true` and `false` are boolean values that allow a whole host of built in functionality, if you only have 2 states you should resort to a boolean rather than two arbitrary symbols.

spec/airport_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+
+  describe '#land' do
+    it 'lands a plane' do
+      expect(subject).to respond_to(:land).with(1).argument
+    end
+
+    it 'raises an error when an airport is full' do
+      allow(subject).to receive(:stormy?) { false }
```
danldb: `allow(subject).to receive xxx` is a code smell - you should avoid stubbing subject. Instead, use a weather object, and replace it with a double in your tests.

danldb:
:construction: 

It's pretty good, Yev. Just needs a little bit of work - use dependency injection to inject your weather object into the airport. Remove all mention of location - I think that is unnecessary, although I can see where you got the idea from.

PR #234, opened 2015-11-24
updates rubric for consistency with others

### LINKED ISSUE #232: update rubric for consistency with others
we have modified the rubric style over the course of the last six weeks in terms of tone of voice and a few extra components such as progress checking, and so we should bring our earlier rubrics into line with that

tansaku:
closes #232

### DIFF
--- a/docs/review.md
+++ b/docs/review.md
@@ -1,9 +1,9 @@
 # Introduction
 Welcome to doing your first code review!  Firstly, don't worry - you are not expected to have all the answers. The following is a code-review scaffold for Airport Challenge that you can follow if you want to.  These are common issues to look out for in this challenge - but you may decide to take your own route.
 
-Either way we'd very much appreciate you submitting the form, even if it's just to say that you didn't use it :-)
+Either way we'd very much appreciate you submitting the [form](http://goo.gl/forms/kFRD3a9Dlg), even if it's just to say that you didn't use it :-)
 
-Please use [this form](http://goo.gl/forms/kFRD3a9Dlg) to tick off where your reviewee has successfully has successfully incorporated these guidelines!  This form helps us get an overall picture of how the whole cohort is doing - it's not an assessment of an individual student.
+If you can, please use [this form](http://goo.gl/forms/kFRD3a9Dlg) to tick off where your reviewee has successfully has successfully incorporated these guidelines!  This form helps us get an overall picture of how the whole cohort is doing - it's not an assessment of an individual student.
 
 # Step 0: Checkout and Run tests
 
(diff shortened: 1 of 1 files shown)


docs/review.md
```diff
@@ -1,17 +1,35 @@
 # Introduction
 Welcome to doing your first code review!  Firstly, don't worry - you are not expected to have all the answers. The following is a code-review scaffold for Airport Challenge that you can follow if you want to.  These are common issues to look out for in this challenge - but you may decide to take your own route.
 
-If you don't feel comfortable giving technical feedback at this stage, try going through this guide with your reviewee and review the code together.
+Either way we'd very much appreciate you submitting the form, even if it's just to say that you didn't use it :-)
```
sjmog: This is before the form is linked or mentioned? Maybe rearrange to after next para

docs/review.md
```diff
@@ -28,7 +46,7 @@ $ irb
  => #<Airport:0x007fafdb81ea88 @capacity=1, @planes=[], @weather=#<Weather:0x007fafdb81ea60>>
 2.2.3 :002 > plane = Plane.new
  => #<Plane:0x007fafdb0041b8>
-2.2.3 :003 > airport.land plane
+2.2.3 :003 > plane.land(airport)
```
sjmog: I think this is meant to be `airport.land(plane)`

docs/review.md
```diff
@@ -87,78 +107,81 @@ subject(:airport) { described_class.new }
 let(:plane) { double :plane }
 
 it 'has the plane after landing' do
-  allow(:plane).to receive(:land)
-  airport.land(plane)
+  airport.clear_for_landing(plane)
```
sjmog: The exemplar and video currently use `land` which is IMO a good pattern to encourage them to use (`land` all the way down)

docs/review.md
```diff
@@ -87,78 +107,81 @@ subject(:airport) { described_class.new }
...
   end
 end
 ```
 
-Does every implementation in the code have associated unit tests?  For example, if you take off a specific plane:
+Does every implementation in the code have associated unit tests?  For example, if you clear a specific plane for take off:
 
 ```ruby
-def take_off(plane)
+def clear_for_take_off(plane)
```
sjmog: Ditto `take_off`

sjmog:
Boop. Will need to re-make the exemplar video next week along with the exemplar.

sjmog:
@tansaku any thoughts on my suggestions here?

tansaku:
@sjmog yes - added another form link - reverted to old code examples to be consistent with current exemplar, since I think they are less immediately important than getting the rubric consistent with the other rubrics etc.

I do want to come back to the issue of the exemplar code, but on the next cycle.

PR #236, opened 2015-11-24
adds detailed rubric requirements to challenge readme

### LINKED ISSUE #230: add challenge requirements clarity
Feedback from students suggested they would like more clarity on the requirements or their code

tansaku:
closes #230

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,16 +15,20 @@ Steps
 
 1. Fill out your learning plan self review for the week: https://github.com/makersacademy/learning_plan_november2015 (edit week 1 - you can edit directly on Github)
 2. Fork this repo, and clone to your local machine
-3. run the command `gem install bundle`
+3. Run the command `gem install bundle`
 4. When the installation completes, run `bundle`
 3. Complete the following task:
 
 Task
 -----
 
-We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
+We have a request from a client to write the software to support an air traffic controller managing the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
 
 ```
+As an air traffic controller 
+So I can provide waiting passengers with information 
+I want to ask a plane if it has landed 
+
 As an air traffic controller 
 So I can get passengers to a destination 
 I want to instruct a plane to land at an airport and confirm that it has landed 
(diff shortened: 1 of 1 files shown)


docs/review.md
```diff
@@ -87,78 +107,81 @@ subject(:airport) { described_class.new }
...
 
-The above example _could_ be combined into one test, but this is not good practice:
 ```ruby
-it 'instructs the plane to land and then has the plane' do
-  expect(plane).to receive(:land)
-  subject.land plane
-  expect(subject.planes).to include plane
+it 'requests landing clearance and location is airport after landing' do
+  expect(airport).to receive(:clear_for_landing)
+  subject.land airport
```
danldb: use named subject?

docs/review.md
```diff
@@ -87,78 +107,81 @@ subject(:airport) { described_class.new }
...
 
 ```ruby
-describe 'landing planes' do
-  it 'instructs the plane to land' do
-    expect(plane).to receive(:land)
-    subject.land plane
+describe Plane do
+  it 'clears landing with the airport' do
+    expect(airport).to receive(:clear_for_landing)
+    plane.land airport
```
danldb: I'm not sure that plane needs to have knowledge of airport - seems like unnecessary dependency?

docs/review.md
```diff
@@ -176,69 +199,72 @@ it 'does not allow plane to take off' do
...
   allow(weather).to receive(:stormy?).and_return true
   message = 'Unable to land due to stormy weather'
-  expect { airport.land(plane) }.to raise_error message
+  expect { plane.clear_for_landing(airport) }.to raise_error message
+end
+
+describe 'a plane can land after storm has cleared' do
   allow(weather).to receive(:stormy?).and_return false
-  expect { airport.land(plane) }.not_to raise_error
+  expect { plane.clear_for_landing(airport) }.not_to raise_error
```
danldb: `plane.clear_for_landing` or `plane.land`? This is inconsistent.

docs/review.md
tansaku: I now believe that plane does need to have knowledge of airport - I think that otherwise we end up with a monolithic airport controlling all the planes - driving from user stories I think this relationship makes better sense for the domain

danldb: I would say the danger of a monolithic airport is less than the danger that you might want to land helicopters or spaceships at some point in the future, and allowing the use of a duck type that responds to `#land` could make that easier. I can't really see any need for the plane to know about the airport - unless the user stories particularly call for that. To be honest I think the only thing that really _needed_ to be changed here was the user story:
```
-As an air traffic controller
-So that I can ensure safe take off procedures
-I want planes only to take off from the airport they are at
```
 That I was very glad to see you removed :)

docs/review.md
sjmog: :+1:

docs/review.md
sjmog: I agree with Dan - plus this will mean that the majority of students will not receive appropriate feedback for what they will do. I think we have to work with what we normally get.

PR #241, opened 2015-11-28
Emma's challenge

emmabeynon:


tansaku:
You had your airport_challenge reviewed by **Mateja Popovic**.
### The good points are:
* Code avoids magic numbers

* Handle randomness in tests

### Additional comments:
Overall this was a very good attempt at the weekend challenge. It's clear that Emma has a good understanding of using doubles, writing tests to achieve proper coverage of her code, and has most of the business logic in order.

Emma could amend the following to get her code into the best shape possible:

- removing weather to its own class
- using before statements in her rspec tests
- moving more logic into private methods
- make the stormy function random, as opposed to having to be set
- instruct which plane to take off, similarly to when it lands


 

PR #242, opened 2015-11-28
Create test

Tmgree:


tansaku:
You had your airport_challenge reviewed by **Chuka Ebi**.
### The good points are:
* Declares a subject rather than a let of the class under test
* The code follows Ruby naming conventions
* Code avoids magic numbers

* README should be updated
* Handle randomness in tests

### Additional comments:
Test suite needs work, Needs to make a weather class/module. Needs to make some methods private, use mocks and stubs.

PR #243, opened 2015-11-28
readme changed

jelgar1:


tansaku:
You had your airport_challenge reviewed by **Rajeev Hejib**.
### The good points are:
* Code avoids magic numbers
* Never defines `attr_reader` then accesses an instance variable directly.

* README should be updated
* Avoid Vacuous tests
* Handle randomness in tests
* Obey the Single Responsibility Principle (SRP)

PR #247, opened 2015-11-28
landing and taking off functionality started

gareth4192:


tansaku:
You had your airport_challenge reviewed by **Deon Tan**.
### The good points are:
* Uses named subject rather than implicit subject (i.e. `subject.do_something`), Uses subject and let appropriately rather than local variables
* Code avoids magic numbers

* README should be updated
* Avoid Vacuous tests

### Additional comments:
Some areas can be refactored but otherwise excellent attempt at Airport - great stubbing of random weather.

Doubles used correctly too.

PR #250, opened 2015-11-28
airport challenge - start

allimac:


tansaku:
You had your airport_challenge reviewed by **Ivan Sathianathan**.
### The good points are:
* Code avoids magic numbers

* README should be updated
* Improper use of attr_*

### Additional comments:
Generally very clean and easy to understand code. Main point is to refactor so that weather functionality is removed from Airport and placed into separate weather class. Otherwise, conceptually very sound. 

PR #251, opened 2015-11-28
tom_airport_challenge

trbradley:


tansaku:
You had your airport_challenge reviewed by **George Maddocks**.
### The good points are:
* Code avoids magic numbers

* README should be updated
* Handle randomness in tests

### Additional comments:
Write a good readme. Needs to stub testing of randomness. Perhaps import weather into module. Also change 'landplane' method's name. Otherwise very good - clean code, well tested with few redundancies. 

PR #252, opened 2015-11-28
MA: Airport Challenge - My Solution

gimi-q:


tansaku:
You had your airport_challenge reviewed by **Harry Venables**.
### The good points are:
* Code avoids magic numbers

* Avoid Vacuous tests

PR #253, opened 2015-11-28
Airport challenge

saratateno:


lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'plane'
+require_relative 'weather_module'
```
DeathRay1977: Don't have to include the word 'module' in the filename. The convention would be to just call it 'weather.rb'

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'plane'
```
DeathRay1977: No need to require plane.rb since we never instantiate a object of this class. The calls to plane.landplane and plane.takeoff will work later on in the code without this. [RE: Duck Typing]

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail 'Cannot land. Stormy weather.' if is_stormy?
+    fail 'Cannot land. Plane has already landed.' if landed(plane)
+    fail 'Cannot land. Airport is full.' if full?
+    plane.landplane
```
DeathRay1977: No need to call this method in Plane 'landplane' since we know it is a method on Plane anyway. Would be more idiomatic to just call it land(). As in plane.land()

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    planes << plane
+  end
+
+  def takeoff(plane)
+    fail 'Cannot instruct takeoff. Stormy weather.' if is_stormy?
```
DeathRay1977: It would be more idiomatic to call is_stormy? just plain ole stormy?

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  private
+
+  def landed(plane)
+    !plane.flying
+  end
+
+  def full?
+    planes.size == capacity
```
DeathRay1977: Usually safer to say >= rather than just ==

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def missing(plane)
```
DeathRay1977: Does this get used anywhere? If not, it is dead code.

spec/airport_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      expect(Airport.new(30).capacity).to eq(30)
+    end
+
+    it 'if no argument is passed, set\'s capacity to default capacity' do
+      expect(subject.capacity).to eq(Airport::DEFAULT_CAPACITY)
+    end
+  end
+
+  context 'behaviour when the weather is fine' do
+    before {allow(Kernel).to receive(:rand).with(100) {99}}
```
DeathRay1977: Nice stubbing!

spec/airport_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+        plane = double(:plane, landplane: false, flying: true)
+        2.times do
+           planes_array << plane
+           subject.land(plane)
+        end
+        expect(subject.planes).to match_array(planes_array)
+      end
+    end
+  end
+
```
DeathRay1977: Maybe also some tests to check that landing a plane increases to number of planes by 1 and takeoff decreases it by 1?

spec/airport_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+describe Airport do
+
+  describe '#initialize' do
+    it 'sets capacity to the argument passed' do
+      expect(Airport.new(30).capacity).to eq(30)
+    end
+
+    it 'if no argument is passed, set\'s capacity to default capacity' do
+      expect(subject.capacity).to eq(Airport::DEFAULT_CAPACITY)
```
DeathRay1977: Best not to use ```subject``` in tests. Better to give it a name 

```ruby
subject(:airport) {described_class.new}
```
Then you can use airport instead of subject, like so:

```ruby
expect(airport.capacity).to eq(Airport::DEFAULT_CAPACITY)
```

DeathRay1977:
An excellent submission! Well done!

tansaku:
You had your airport_challenge reviewed by **Harriet Craven**.
### The good points are:
* Uses named subject rather than implicit subject (i.e. `subject.do_something`), Uses subject and let appropriately rather than local variables
* Never defines `attr_reader` then accesses an instance variable directly.

PR #254, opened 2015-11-29
First user test

DanielaGSB:


tansaku:
You had your airport_challenge reviewed by **Samuel Overloop**.
### The good points are:
* Code avoids magic numbers

* README should be updated
* Avoid Vacuous tests
* Handle randomness in tests
* Improper use of attr_*

### Additional comments:
Don't worry about an incomplete submission. Coding on your own for the first time is not an easy feat to pull of and makes you appreciate pair programming even more :-) The review is based on your codebase as I first saw it. Please feel free to check on me with your updated codebase and we can look together at ticking more boxes.

Two suggestions:
1) Stick religiously to TDD and test what actually has to be tested, e.g. every public method and every line in it that returns a value or changes a state.
2) Go through your code again and try to understand the use of doubles. For example, `stormy?` is a an instance method of `Weather`, so a mock of the stormy weather is done on Weather and not on Plane.

PR #255, opened 2015-11-29
updated airport_challenge

Htunny:


tansaku:
You had your airport_challenge reviewed by **Amy Yang**.
### The good points are:
* Booleans are returned implicitly where appropriate
* Never defines `attr_reader` then accesses an instance variable directly., Never defines `attr_accessor` that should be `attr_reader`

* README should be updated
* Avoid Vacuous tests
* Handle randomness in tests

### Additional comments:
- Improving test coverage (test the opposite of expectation whenever possible) 
- reading rspec error messages to continue test drive the code with every single change 
- Using context to help organize rspec tests, especially when testing for random conditions 
- Indentation of code for clarity 

PR #259, opened 2015-11-29
Airport challenge - unfinished

klssmith:


tansaku:
You had your airport_challenge reviewed by **Rebecca Collins**.
### The good points are:
* Declares a subject rather than a let of the class under test
* Code avoids magic numbers

* Avoid Vacuous tests
* Handle randomness in tests

PR #261, opened 2015-11-29
Airport Richard 

ric9176:


tansaku:
You had your airport_challenge reviewed by **Mahmud Hussain**.
### The good points are:
* Declares a subject rather than a let of the class under test, Uses named subject rather than implicit subject (i.e. `subject.do_something`), Uses subject and let appropriately rather than local variables
* Code avoids magic numbers

* README should be updated
* Handle randomness in tests

### Additional comments:
Checkout attr_reader in private scope. 
Test random numbers. 

PR #265, opened 2015-11-30
My Airport Challenge

Adzz:


tansaku:
You had your airport_challenge reviewed by **Alaan Rahim**.
### The good points are:
* Code avoids magic numbers

* README should be updated

### Additional comments:
- Update README.MD
- Refactor tests; implement context blocks.
- Extract Weather functionality out of Airport class and into it's own WeatherReporter/Weather class.
- Extract error messages into their own method.
- Change status of plane upon instantiation.
- Use named subject.

PR #266, opened 2015-11-30
For pull request

charlieperson:


tansaku:
You had your airport_challenge reviewed by **Mari-Ann**.
### The good points are:
* Declares a subject rather than a let of the class under test, Uses subject and let appropriately rather than local variables
* Code avoids magic numbers

* Handle randomness in tests

### Additional comments:
I appreciate that Charlie had a super packed weekend and all things considered it wasn't that bad. I have advised Charlie that he needs to separate out versions of his code instead of having 10000 lines of commented out code that isn't labeled properly! 

PR #267, opened 2015-11-30
airport challenge

TY231618:


tansaku:
You had your airport_challenge reviewed by **Fergus Lemon**.
### The good points are:

* Avoid Vacuous tests
* Handle randomness in tests

### Additional comments:
I also reviewed the weekend challenge for (Github username) 'willibaur', both he and Tony made a good crack at the challenge.  They both struggled with dependency injection, stubbing behaviour and interaction between classes but generally very good.

PR #289, opened 2016-01-23
Working through airport_challenge

RPiper93:


lib/weather.rb
```diff
@@ -0,0 +1,6 @@
+class Weather
+  def stormy?
+    true if rand(0...6) == 0
```
ggwc82: Does this line ever return true if the condition is satisfied?

features/feature_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    expect { airport.take_off(plane) }.to raise_error "Can't take off due to stormy weather"
+  end
+
+  it "can't land when stormy" do
+    airport = Airport.new
+    plane = Plane.new
+    allow_any_instance_of(Weather).to receive(:stormy?).and_return(true)
+    expect { airport.land(plane) }.to raise_error "Can't land due to stormy weather"
+  end
+
```
ggwc82: Feature test for multiple planes landing and taking off confirmed working in irb, need to implement it in the feature_spec.rb

README.md
```diff
@@ -1,79 +1,23 @@
...
+    2.2.3 :002 > plane = Plane.new
+      => #<Plane:0x007fb95388c3b8 @landed=false> 
+    2.2.3 :003 > airport.land(plane)
+      => [#<Plane:0x007fb95388c3b8 @landed=true>] 
+    2.2.3 :004 > airport.land(plane)
+      RuntimeError: This plane has already landed.
+    2.2.3 :005 > airport.take_off(plane)
+      => #<Plane:0x007fb95388c3b8 @landed=false> 
+    2.2.3 :006 > airport.take_off(plane)
+      RuntimeError: This plane is not at this airport.
```
ggwc82: Glad to see a good summary of the project, and also instructions included. A good guide to creating READMEs is this http://stackoverflow.com/questions/2304863/how-to-write-a-good-readme

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'airport'
+
+describe Airport do
+  subject(:airport) { Airport.new }
```
ggwc82: 1. Is this the correct form or usage of subject?
2. When referenced as airport, does it create a new instance every time?
Follow up for both of us!

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  it "can't have a plane take-off twice" do
+    airport.land(plane)
+    airport.take_off(plane)
+    expect { airport.take_off(plane) }.to raise_error "This plane is not at this airport."
+  end
+
+  it "can't have a plane landing in more than one place" do
+    airport.land(plane)
+    allow(plane).to receive(:landed).and_return(true)
+    airport_2 = Airport.new
```
ggwc82: Would it be better to put this as a new let rather than leave it in the code block, so it could possibly be reused again?

features/feature_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    plane = Plane.new
+    airport.land(plane)
+    allow_any_instance_of(Weather).to receive(:stormy?).and_return(true)
+    expect { airport.take_off(plane) }.to raise_error "Can't take off due to stormy weather"
+  end
+
+  it "can't land when stormy" do
+    airport = Airport.new
+    plane = Plane.new
+    allow_any_instance_of(Weather).to receive(:stormy?).and_return(true)
```
ggwc82: Lines 62 and 69 are identical, can these be placed as a let at the top? No, because it would fail all previous tests which rely on non-stormy weather. 

Suggestion - what if you seperated the weather method into the weather class entirely, then you can declare multiple stubs (one for stormy, one for sunny) and call them appropriately in your it blocks.

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'airport'
+
+describe Airport do
+  subject(:airport) { Airport.new }
+  let(:plane) { double(:plane, :landed => false, :landed= => true)}
+
+  it "can't have a plane land twice" do
```
ggwc82: Step 4 - Use symbols instead of strings

Use describe within a describe for method unit tests like this:
describe "#land" do

to group up similar unit tests

mechacoach:
You had your airport_challenge reviewed by **Godfrey Chiu**

### The good points are:

* Improper use of attr_*

### Additional comments:
In step 3, she used describe regularly, but never used context.

Used let doubles with code blocks instead of 'before' to reduce code repetition., avoidmagicnumberse.g.oncapacity: Code avoids magic numbers

PR #318, opened 2016-01-24
Refactoring branch

reissjohnson:


README.md
```diff
@@ -48,32 +31,18 @@ I want to prevent landing when the airport is full
...
+Function tests were performed in IRB with each unit test written in a spec file & ran through RSPEC.
 
-**BONUS**
+25-01-16 
 
-* Write an RSpec **feature** test that lands and takes off a number of planes
+RSpec is passing, the program covers all of the features asked however it still needs refactoring to make it a bit tidier.
 
-Note that is a practice 'tech test' of the kinds that employers use to screen developer applicants.  More detailed submission requirements/guidelines are in [CONTRIBUTING.md](CONTRIBUTING.md)
+Spec files are found in the ./spec folder with library files found in ./lib.
```
MatDrake: Include a wee bit about how to install the program; a short bit of IRB code explaining how it'd work - http://stackoverflow.com/questions/2304863/how-to-write-a-good-readme

spec/Airport_Spec.rb
```diff
@@ -0,0 +1,90 @@
+require 'Airport.rb'
+
+
+describe Airport do
+
```
MatDrake: Declaring a subject at the start using:
     let(:<variable_name>) { <class_you_want_it_to_be>.new }
will allow you to call the instance throughout, rather than creating a new one for each test (e.g. line 22,38 etc.)

spec/Airport_Spec.rb
```diff
@@ -0,0 +1,90 @@
...
+let(:plane){double :plane}
+let(:plane2){double :plane2}
+
+it 'has a default capacity' do
+  expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+
+it 'lands a plane' do
+  plane = double(:plane)
```
MatDrake: You can safely delete bit like "plane = double(:plane)" from each test as you've 'let(:plane) {double :plane}' at the start of your code. Using 'let' means you can now just use the variable 'plane' throughout.

spec/Airport_Spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  allow(airport).to receive(:current_weather).and_return(false)
+  airport.land(plane)
+  expect(airport.planes_landed).to eq [(plane)]
+  end
+
+it 'lets a specified plane takeoff' do
+  plane = double(:plane)
+  plane2 = double(:plane2)
+  airport = Airport.new
+  allow(airport).to receive(:current_weather).and_return(false)
```
MatDrake: Using 'context' blocks can help to create a scope for your tests. For this test and a few others, you stub for current_weather to return false; so you could have a "context 'when weather is not stormy' " block and declare it once in a before block.
This will help avoid repetition and can improve readability.

spec/weather_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'Weather.rb'
+
+describe Weather do
+
+it { is_expected.to respond_to :stormy?}
+
```
MatDrake: Include a unit test for your weather.stomy? to increase your test coverage

spec/Airport_Spec.rb
```diff
@@ -0,0 +1,90 @@
+require 'Airport.rb'
+
+
+describe Airport do
+
+it { is_expected.to respond_to(:land).with(1).argument }
```
MatDrake: Once your tests start to become more sophisticated when testing methods, you can delete these earlier 'respond_to' tests to clean up you code as they become redundant (they have to accept arguments inherently in your later tests).

lib/Airport.rb
```diff
@@ -0,0 +1,58 @@
...
+    @current_weather = Weather.new
+    @capacity = capacity
+  end
+
+  def current_weather
+    if @current_weather.stormy? == true
+      true
+    else
+      false
+    end
```
MatDrake: Avoid using long if statements and instead use the implicit return of booleans

lib/Airport.rb
```diff
@@ -0,0 +1,58 @@
...
+    raise "this plane has already landed" if @planes_on_the_ground.include?(plane)
+    raise "plane cannot land due to stormy conditions" if current_weather == true
+    raise "plane cannot land as airport is full" if capacity_hit == true
+     @capacity -= 1
+     @planes_on_the_ground << plane
+  end
+
+  def planes_landed
+    @planes_on_the_ground
+  end
```
MatDrake: Avoid public access to the internal objects like you plane array, as a user could modify this in a way your system wasn't built to accept.

lib/Airport.rb
```diff
@@ -0,0 +1,58 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def land plane
+    raise "this plane has already landed" if @planes_on_the_ground.include?(plane)
+    raise "plane cannot land due to stormy conditions" if current_weather == true
+    raise "plane cannot land as airport is full" if capacity_hit == true
```
MatDrake: This could be altered to follow the SRP, as this method is checking these conditions. Creating separate methods to check these conditions will mean that each method has one responsibility (one condition to check).

You could have a 'plane_landed?' method that ran the code '@planes_on_the_ground.include?(plane)' and then have your raises read along the lines of:

raise 'this plane has already landed' if plane_landed?

lib/Airport.rb
```diff
@@ -0,0 +1,58 @@
...
+  def current_weather
+    if @current_weather.stormy? == true
+      true
+    else
+      false
+    end
+  end
+
+  def capacity_hit
+    if @capacity == 0
```
MatDrake: Accessing the instance variable directly, rather than calling the reader method capacity

lib/Weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
+
+  def stormy?
+    if rand(100) <= 20
+      true
+    else
+      false
+    end
+  end
```
MatDrake: This method is a command and query. You could break this up into two separate ones by have one run the rand function and a second for the true/false.

lib/Airport.rb
```diff
@@ -0,0 +1,58 @@
...
+     @planes_on_the_ground << plane
+  end
+
+  def planes_landed
+    @planes_on_the_ground
+  end
+
+  def takeoff plane
+    raise "this plane has already taken off" unless @planes_on_the_ground.include?(plane)
+    raise "plane cannot take off due to stormy weather" if current_weather == true
```
MatDrake: As hound has sniffed out also, slightly over the 80 length for multiple line redundancy.

mechacoach:
You had your airport_challenge reviewed by **Matthew Drake**

### The good points are:

* Guard clauses are used where appropriate, removeallcommented-outcode: There is no commented-out code in the submission

* Handle randomness in tests
* Obey the Single Responsibility Principle (SRP)
* Improper use of attr_*

### Additional comments:
When refactoring, if you're unsure whether a line of code is needed, try deleting is and re-running your tests to see if it passes. Some of the rspec tests have things defined a second time, so could be removed., avoidmagicnumberse.g.oncapacity: Code avoids magic numbers

PR #327, opened 2016-03-05
Airport challenge

junyuanxue:


mechacoach:
You had your airport_challenge reviewed by **Alex Avlonitis**

### The good points are:

* Improper use of attr_*

### Additional comments:
Junyuan Xue did a really good job, covered all tests plus feature tests. Really neat and clever code, used dependecy injections and separate weather class. 
She could use encapsulation for the public array and SRP at some methods, but other than that it was really good.

README.md
```diff
@@ -4,88 +4,103 @@ Airport Challenge
...
-=  = ==(____MA____)
+=  = ==(____JX____)
           \_____\___________________,-~~~~~~~`-.._
-          /     o o o o o o o o o o o o o o o o  |\_
+          /     j u n y u a n - a i r w a y s   |\_
           `~-.__       __..----..__                  )
                 `---~~\___________/------------`````
                 =  ===(_________)
 
 ```
```
danldb: nice!

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class Airport
+
+  attr_reader :planes, :capacity
+  DEFAULT_CAPACITY = 30
+
+  def initialize(**opts)
+    @weather = opts[:weather] || Weather.new
+    @capacity = opts[:capacity] || DEFAULT_CAPACITY
+    @planes = []
```
danldb: I would lean towards [ruby key-word arguments](https://robots.thoughtbot.com/ruby-2-keyword-arguments) instead of (**opts).

lib/plane.rb
```diff
@@ -0,0 +1,33 @@
+class Plane
+
+  attr_reader :flying, :airport
+
+  def initialize(airport=Airport.new, flying=false)
```
danldb: why does a plane need an airport?

lib/airport.rb
junyuanxue: @danldb thanks for the suggestion and the link! now I know :)

lib/plane.rb
junyuanxue: @danldb that's because...when I was writing my feature spec, I realised that it makes more sense for planes to start in an airport rather than in the air. So I assigned each plane an initialization airport so they don't land in the same airport again. :P

PR #332, opened 2016-03-05
Challenge pull request

shaneoston72:


silvabox:
Nice work Shane, your code is clean and easy to read.  There's a bit of duplication of set up code in your tests, which you could improve with some `before` blocks.

PR #337, opened 2016-03-05
weekend challenge pull request

lorenzoturrino:


mechacoach:
You had your airport_challenge reviewed by **Lou Franklin**

### The good points are:

* Booleans are returned implicitly where appropriate, removeallcommented-outcode: There is no commented-out code in the submission

* Avoid Vacuous tests
* Improper use of attr_*

### Additional comments:
Good attempt. Just watch encapsulation and how methods are split up - guard statement location and whether better to return a boolean to trigger this. , avoidmagicnumberse.g.oncapacity: Code avoids magic numbers

silvabox:
Hi Lorenzo.  This is a good effort, but there are some unusual aspects to it.  Generally, we would push private methods to the bottom of the class rather than the top as it reads more naturally.  You need to be careful about breaking encapsulation - particularly reaching into the airport's planes array and directly testing against it.  It's also not clear why one should be able to change the weather.

One other thing to watch is your `check_current_status` method.  The logic here is peculiar because you start by knowing which status you want to check for (e.g. `:flying`), then pass this to a method which goes through a series of switches to find out which status you want to check for.

PR #342, opened 2016-03-05
Claudia Airport Challenge

claudiagreen:


mechacoach:
You had your airport_challenge reviewed by **EIrik**

### The good points are:

* README should be updated
* Improper use of attr_*


lib/airport.rb
```diff
@@ -0,0 +1,58 @@
...
+  def instruct_takeoff(plane)
+    fail "Airport empty." if empty?
+    fail "Storm." if storm?
+    fail "That plane isn't in that airport." if plane_in_airport(plane) == false
+    if plane.landed?
+      plane.takeoff
+      plane = @planes.pop
+      @planes.delete(plane)
+    else
+      "Plane has taken off."
```
silvabox: What does it mean to return a string from this method?

lib/airport.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def instruct_land(plane)
+    fail "Airport full." if full?
+    fail "Storm." if storm?
+    unless plane.landed?
+      plane.land
+      @planes << plane
+      plane
+    else
+      "Plane has landed."
```
silvabox: What does it mean to return a string from this method?

silvabox:
Well done Claudia - that's a pretty good effort.  You should avoid exposing your planes array with an `attr_reader`

PR #344, opened 2016-03-06
Progress satisfying the first user story

tobenna:


mechacoach:
You had your airport_challenge reviewed by **Chris Wynne**

### The good points are:

* Improper use of attr_*

### Additional comments:
Good stuff. Bit of research on use of attr would be good, and the use of context blocks to lay out your tests would be great. Overall really good though., avoidmagicnumberse.g.oncapacity: Code avoids magic numbers

silvabox:
Hi Tobenna, this is a pretty good effort - well done.  I've made a couple of comments above.

lib/airport.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def remove_from_airfield(plane)
+    @airfield.delete_if { |pl| pl == plane }
+  end
+
+  def at_capacity?
+    capacity == airfield.size
+  end
+
+  def can_land?
```
silvabox: by convention, a method with a `?` should return true or false; not true or raise an error.

lib/weather.rb
```diff
@@ -0,0 +1,25 @@
...
+  def stormy?
+    forecast == :stormy
+  end
+
+  private
+
+  WEATHER_SEED = { sunny: 19, stormy: 1 }.freeze
+
+  def weather_array
+    weather = []
```
silvabox: Is it necessary to recreate this array on every call to `weather_array`?

PR #348, opened 2016-03-06
airport challenge

hawksdoves:


mechacoach:
You had your airport_challenge reviewed by **Matthew Bridges**

### The good points are:

* Avoid Vacuous tests
* Handle randomness in tests
* Improper use of attr_*

### Additional comments:
ObjectSpace wizardry, cool code., avoidmagicnumberse.g.oncapacity: Code avoids magic numbers

docs/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+
+	def initialize(capacity=DEFAULT_CAPACITY)
+		@capacity = capacity
+		@storage = []
+		@weather = Weather.new
+	end
+
+	def in_airport?(plane)
+		storage.include?(plane)
+	end
```
danldb: ARGH my eyes! please fix your indentation. ;-)

docs/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+			plane.landed
+		else
+			announcement
+		end
+	end
+
+	def releases_plane(plane)
+		if airport_open?	
+			fail 'This plane is already in the air, so it cannot take off.' if plane.plane_status		
+			fail 'This plane is not at this airport, so it cannot take off from here.' if in_airport?(plane) == false
```
danldb: prefer:

`fail 'This plane is not at this airport, so it cannot take off from here.' if in_airport?(plane)`

(remembering that in_airport? returns a boolean already)

docs/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+			storage.delete(plane)
+			plane.flying
+		else
+			announcement
+		end
+	end
+
+	def plane_location(plane)
+		plane.plane_status ? 'This plane is flying.' : find_plane(plane)
```
danldb: Looks like feature envy - everything about this method regards the plane, so the responsibility is probably on the plane, not the airport.

docs/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+		found =	airports.select {|airport| airport.in_airport?(plane)}
+		found[0]
+	end
+
+	def full?
+		storage.length>=capacity 
+	end
+
+	def airport_open?
+	 	weather.random_weather == "not stormy"
```
danldb: prefer symbols to strings

docs/plane.rb
```diff
@@ -0,0 +1,34 @@
...
+
+	def take_off(from)
+		from.releases_plane(self)
+	end
+
+	def land(at)
+		at.accepts_plane(self)
+	end
+	
+	def plane_status
```
danldb: consider this method name. plane_status returns true or false. If I asked a plane it's status, would I expect it to say `true`

danldb:
Not bad. Broken YAGNI with the plane location thing. Slightly over-complicated, and more elements than were needed to solve the problem.

danldb:
also indentation is horrible. I would recommend you change the settings in your text editor - use spaces instead of tabs, 2 spaces per indent. Really important that you focus on this and get it good, bad indentation will prevent you from getting job interviews!

PR #352, opened 2016-03-06
land plane

sachinkaria:


lib/airport.rb
```diff
@@ -0,0 +1,30 @@
+require_relative 'plane'
+
+class Airport
+DEFAULT_CAPACITY = 1
+attr_reader :planes
+
+	def initialize (capacity = DEFAULT_CAPACITY)
+		@planes = []
+		@capacity = capacity
+	end
```
danldb: indentation :(

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+
+	def land(plane)
+		if full? then raise 'Sorry the airport is full'
```
danldb: `raise 'airport full' if full?`

lib/plane.rb
```diff
@@ -0,0 +1,27 @@
...
+
+	def initialize
+		@status = false
+	end
+
+	def landed
+		@status = true
+	end
+
+	def takenoff
```
danldb: methods should be commanding verbs, or nouns. In this case '#take_off' would be better.

spec/airport_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'airport'
+
+describe Airport do
+		plane = Plane.new
```
danldb: no real objects in your tests other than the subject

spec/airport_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+	describe 'landing plane' do
+
+
+      it 'confirms the plane has landed' do
+      	subject.land(plane)
+        expect(subject.planes.length).to eq 1
```
danldb: this indentation :(

PR #353, opened 2016-03-06
Airport Challange

jazzygasper:


mechacoach:
You had your airport_challenge reviewed by **Harry Gordon**

### The good points are:

* README should be updated
* Improper use of attr_*

### Additional comments:
The code is very good on the whole.
Could potentially remove errors to their own method.
Look at how a plane takes off and how the plane is removed form the storage array.
The test are now correctly being stubbed. It is necessary to refactor the airport spec to keep it dry.
An introduction of a feature test could be necessary, where no classes are being stubbed.  , avoidmagicnumberse.g.oncapacity: Code avoids magic numbers

PR #354, opened 2016-03-06
Airport Challenge

JojSh:


mechacoach:
You had your airport_challenge reviewed by **Scott Gledhill**

### The good points are:

* Booleans are returned implicitly where appropriate

* README should be updated
* Handle randomness in tests
* Improper use of attr_*

### Additional comments:
Complete all tests before refactoring to get more functionality,
Understand doubles and stubs better, 
eliminate respond_to test,
try and complete more code by escalating faster when stuck.

silvabox:
Hi Joj (sp?!)

Look like there's quite a bit of confusion in your code about what to test and you are breaking encapsulation all over the place.  Let me know if you want to go over it.

Ben

silvabox:
One huge clue is that you have references to `@capacity` all over your tests.  When writing tests you should think about _behaviour_, not _implementation_.   No-one outside of the airport should know about instance variables within it.  Is `capacity` the right name?

PR #357, opened 2016-03-07
New pull request after refactoring

yasgreen93:


silvabox:
Hi Yasmin, some nice structure in you code, well done.  I would advise you think about your naming of things e.g. `DEFAULT_VALUE` has no context and `allow_landing` doesn't really convey that the plane will actually land.  You should also look at your `Weather` class as it has three redundant attribute readers.

PR #359, opened 2016-04-15
add initial commit to generate pull request

tigretoncio:


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+class Airport
+
+  ERR_ALREADY_LANDED = "plane already landed"
```
tigretoncio: ok, looks like the solution is to freeze the String using
`ERR_ALREADY_LANDED = "plane already landed".freeze`

README.md
```diff
@@ -7,12 +7,64 @@ Airport Challenge
...
+  from (irb):6
+  from /home/sergio/.rvm/rubies/ruby-2.2.3/bin/irb:11:in `<main>'
+2.2.3 :007 > airport.takeoff(plane)
+ => #<Plane:0x00000001f96db0 @landed=false>
+
+```
+
+
+
+
```
danldb: Great readme, maybe a little long.

I would expect to see instructions for how to run the code and how to run the tests. You could also mention how long you have been learning to code, and how long you spent on this challenge.

lib/error_mgr.rb
```diff
@@ -0,0 +1,13 @@
...
+  LANDED = "plane alreay on ground".freeze
+  STORMY = "stormy weather".freeze
+  FULL = "capacity full".freeze
+
+  def check_errors(plane,land)
+    fail LANDED if land && in_airport?(plane)
+    fail NOT_FOUND unless land || in_airport?(plane)
+    fail STORMY if stormy?
+    fail FULL if full?
+  end
```
danldb: This is interesting - I would say at this stage that this is over-engineered, where a simple fail message would do. I would not `include` the module - it made it quite hard to figure out where '#check_errors` method was being called.

lib/plane.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def land
+    @landed = true
+  end
+
+  def takeoff
+    @landed = false
+  end
+
+end
```
danldb: Very nice

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
+require_relative "error_mgr"
+require_relative "weather"
+class Airport
+  include ErrorManager
+  include Weather
```
danldb: Remove coupling be injecting Weather via Dependency Injection.

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+  describe "planes standard behaviour" do
+    it "sends message to plane to take-off" do
+      allow(airport).to receive(:stormy?) { false }
+      airport.land(plane)
+      airport.takeoff(plane)
+      expect(plane).to have_received(:takeoff)
+    end
+
+    it "after take-off plane no longer at airport" do
+      allow(airport).to receive(:stormy?) { false }
```
danldb: Avoid stubbing your subject when possible. Use DI to inject a double instead.

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      expect{ airport.land(flying_plane)}.to raise_error Airport::FULL
+    end
+  end
+
+  describe "capacity overrride" do
+    it "changes the default capacity on initialize" do
+      random_capacity = Random.rand(5000)
+      subject = described_class.new random_capacity
+      expect(subject.capacity).to eq random_capacity
```
danldb: why is capacity public?

spec/support/weather_test_spec.rb
```diff
@@ -0,0 +1,11 @@
+shared_examples_for Weather do
+  it "STORM is value 9" do
+    expect(WeatherTest::STORM).to eq 9
+  end
+  it { is_expected.to respond_to(:stormy?) }
+
+  it "stubbing the value to STORM gets STORM (d'oh)" do
+    allow(subject).to receive(:stormy?) { true }
```
danldb: vacuous test - this test will never fail.

spec/support/weather_test_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+    expect(subject.stormy?).to eq true
+  end
+end
```
danldb: If your weather had been completely isolated from the rest of your code (i.e. not `include`d), then you wouldn't even need to test it. It is possible to test random behaviour, but difficult and expensive. Instead, isolate random behaviour in a tiny object that does nothing else, then leave it untested.

spec/plane_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+describe Plane do
+
+  it { is_expected.to respond_to(:landed?) }
+
+  context "when instructed to land" do
+
+    it 'is landed' do
+      subject.land
+      expect(subject.landed?).to be true
```
danldb: `landed?` should be private - test behaviour by launching a landed plane, or landing a flying plane.

lib/airport.rb
tigretoncio: Thanks for the feedback.  I am unable to understand at this stage why injecting would be better than including the classes, would be possible to develop that point?

spec/airport_spec.rb
tigretoncio: Thanks for the feedback.  If I understood it properly, from the Object Design strategy I took it was not possible to inject a double in this situation but I presume your feedback is to review the interaction strategy.  At this stage I see it much simpler to decouple the interaction just moving weather into a module and calling the method to generate weather than injecting weather into classes, but that's just showing my ignorance.  Will try to understand why is better.

spec/airport_spec.rb
tigretoncio: Thanks for the feedback.  Upon reflection the answer is because it allowed me an easy way to test it by calling directly the public interface on the subject.  
Agreed that capacity attribute should be private, and the test more sophisticated and based on behaviour, i.e., to land planes up to the overriden capacity of the airport +1 and expect the overcapacity exception.

spec/plane_spec.rb
tigretoncio: Agreed. Thank you

spec/support/weather_test_spec.rb
tigretoncio: yes, that's why I added the "d'oh" expression. Thank you.

spec/support/weather_test_spec.rb
tigretoncio: Thanks for the feedback.  I don't understand why is the case.  Creating a small class with my 2 methods and 2 lines of code seems to me exactly the same as using a module with exactly the same amount of code, so it escapes to my why using classes no test was needed, but by using the module I should have.  I will try to understand it.

lib/airport.rb
danldb: Good question.

Using modules as a mixing (i.e. using `include`) means that the Airport is always coupled to Weather. If you ever start need to have several types of weather, or you need to improve the weather module, you will need to make changes to Airport as well.

In this case there is also the question of responsibility - it is not a part of the Airports responsibility to be stormy.

spec/support/weather_test_spec.rb
danldb: Hmm that's a good question as well. I would examine the differences between a module and a class - what properties do they each have? They are actually very similar in many ways.

You can leave a class or a module untested if it *only* exhibits untestable behaviour. But in this case you is good practice to completely isolate it from your tested, reliable code.

PR #360, opened 2016-04-16
Solution for airport_challenge

omajul85:


lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+class Airport
+  attr_reader :capacity, :planes, :name
```

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  
+  DEFAULT_CAPACITY = 10
+  
+  def initialize(name, plane, capacity = DEFAULT_CAPACITY)
+    @name = name
+    @capacity = capacity
+    @planes = []
+    2.times { @planes << plane.new }
```
danldb: This seems unnecessary - an Airport is not responsible for building planes when initialised.

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  
+  def receive_plane(plane)
+    @planes << plane
+  end
+  
+  def release_plane(plane)
+    @planes.delete plane
+  end
+  
+  def ready_for_landing?
```

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+    check_weather == :sunny && !full?
+  end
+  
+  def ready_for_taking_off?
+    check_weather == :sunny
```

lib/plane.rb
```diff
@@ -0,0 +1,36 @@
+class Plane
+  attr_reader :status
+  
+  def initialize
+    @status = :landed
```
danldb: prefer `@landed = true` - it is more explicit (status could mean anything), and it benefits from boolean behaviour.

lib/plane.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+  
+  def land(airport)
+    begin
+      return "Plane #{self} is already landed" unless @status == :flying
+      raise "Cannot land" unless airport.ready_for_landing?
+    rescue RuntimeError => error
+      p error
```
danldb: no reason to rescue

danldb:
See comments inline.

Use DI for your weather. Focus on encapsulation - you are allowing classes to know too much about each other.

PR #361, opened 2016-04-16
airport_challenge 

missamynicholson:


lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def land(plane)
+    fail "Airport full" if full?
+    fail "Stormy weather" if stormy?
+     plane.land
+     planes << plane
+  end
+
+  def stormy?
```
danldb: could be private

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+DEFAULT_CAPACITY = 50
+
+  attr_reader :capacity
```

danldb:
:eyes:

PR #366, opened 2016-04-16
First basic test completed - airport can land a plane

jackrim1:


lib/weather.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'airport'
+
+class Weather
+
+def stormy?
+  random_weather == :stormy
+end
```
danldb: You really need to fix your indentation. Tech tests are automatically failed if the applicant can't properly indent their code. It's like submitting an application letter in one massive paragraph with no punctuation - you wouldn't do it in a letter, definitely don't do it in code.

danldb:
:eyes:

Build failing on travis because your tests are not passing. I would recommend you have another go at this.

jackrim1:
Ok, I think I should be able to make them pass now. I'll have a go this
weekend.

PR #367, opened 2016-04-16
Airport challenge

ollieh-m:


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
+class Airport
+
+	attr_reader :landed
+
+	DEFAULT_CAPACITY = 10
+
+	def initialize(capacity=DEFAULT_CAPACITY,weather=Weather.new)
```
danldb: Indentation!!! It hurts my eyes!

lib/plane.rb
```diff
@@ -0,0 +1,46 @@
...
+
+	attr_reader :location
+
+	def initialize
+		@landed = false
+		@location = :sky
+	end
+
+	def try_to_land(airport)
+		landed? ? "Plane is already landed" : airport.receive(self)
```
danldb: use a guard clause instead of a ternary:
```ruby
raise "plane is already landed" if landed?
airport.receive(self)
```

danldb:
Good use of DI.

Classes are way over-complicated. No need for try_to methods - remember tell don't ask.

PR #375, opened 2016-04-16
First user story - Plane lands and confirms landing

RobinHeathcote:


lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def land(plane)
+    fail 'Cannot land because of bad weather' if stormy?
+    fail 'Airport at capacity' if full?
+    fail 'Plane has already landed' if landed?(plane)
+    planes << plane
+  end
+
+  def landed?(plane)
```
danldb: unnecessary method

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def take_off(plane)
+    fail 'Plane is not in airport' if landed?(plane) == false
+    fail 'Cannot take off because of the bad weather' if stormy?
+    planes.delete(plane)
+
+  end
+
+  def departed?(plane)
```
danldb: unnecessary method

danldb:
Remember to test the behaviour - what is the job of the airport? It doesn't need methods to examine the contents of its internal array. 

Plane not finished. 

No DI for the weather.

PR #376, opened 2016-04-16
Initial commit

sitypop:


silvabox:
Nice work Sity, your code is clean and concise.  I would think about improving `Plane` though as it could be more domain-oriented.  When we talk about planes and airports, would I ask "what's that plane's landed status?" and would you reply "true"?  Often, talking about the domain helps to reveal the right names for things e.g.  `plane.landed?` and `plane.flying?`.

PR #377, opened 2016-04-16
Planes, airports and weather - need to explore edge cases

gerauf:


silvabox:
There's some nice ideas here and you've tried to extend the functionality which is good.  However, watch your encapsulation and think about your plane's responsibilities (e.g., is it responsible for all of those messages; particularly `complain`, which is worded in a way that implies it belongs to airport.  The language you use can often lead you in the wrong direction unwittingly).

I like that you've had a go using a block, but think more carefully about how you name it.  I had to read it a few times before I realized your intention was to return either an 'error' message, or the results of yielding the block.  `run_way_check` doesn't really convey this; it only implies the checking part.

PR #378, opened 2016-04-16
Airport Challenge

festinalent3:


spec/plane_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  #I want to prevent takeoff when weather is stormy
+
+  describe '#take_off' do
+    context 'plane can take off from airport' do
+      it 'it takes off' do
+        expect(subject.take_off).to eq true
+      end
+      it 'does not take off in stormy weather' do
+        allow(airport).to receive(:weather).and_return("stormy")
+        expect { subject.take_off }.to raise_exception "Can't take off in stormy weather"
```
festinalent3: That's just silly! ;) I prefer keeping my error message clear and specific

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
...
+    @airport = airport
+  end
+
+  def land(airport)
+    @airport = airport
+    @landed = true
+  end
+
+  def take_off
+    raise "Can't take off in stormy weather" if airport.weather == "stormy"
```
festinalent3: done!

lib/communicator.rb
```diff
@@ -0,0 +1,29 @@
...
+    fail "Can't land in stormy weather" if airport.weather == "stormy"
+    fail "Can't land in full airport" if airport.full?
+  end
+
+  def check_can_take_off(plane, airport)
+    fail "Can't take off when flying" if plane.flying?
+    fail "Can't take off from another airport" if plane.current_airport != airport
+    fail "Can't take off in stormy weather" if airport.weather == "stormy"
+  end
+
```
danldb: This module seems unnecessary - remember to use fewest possible components.

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def dock(plane)
+    check_can_dock(plane)
+    update_weather
+    planes << plane
+  end
+
+  def undock(plane)
+    check_can_undock(plane)
```
danldb: tell don't ask - here, you are asking.

lib/plane.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'communicator'
+
+class Plane
+  include Communicator
+
+  def initialize(airport = Airport.new)
+    self.landed_at(airport)
+    airport.dock(self)
+  end
```
danldb: Plane should not be responsible for building airports.

lib/plane.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def landed_at(airport)
+    @airport = airport
+    @flying = false
+    true
+  end
+
+  def is_flying
+    @airport = 'no airport'
+    @flying = true
```
danldb: no need for `@` when using attr_accessor

danldb:
A little over complicated. Too many components.
 Use DI instead of mixing in the weather - it reduces the coupling

lib/plane.rb
festinalent3: Yes, thanks, it was me being lazy when refactoring *bad emma*

lib/plane.rb
festinalent3: Completely agree! I wrote about it as a weird thing in the README file..

lib/airport.rb
festinalent3: Are you referring to for example method check_can_dock? Not completely sure what you mean, would really, really appreciate just a short explanation if you have the time! Thank you :)

lib/communicator.rb
festinalent3: True, my bad, I sort of finished the challenge in its simple form at first but decided I wanted to play around with modules and that led to the decision of creating this monster for absolutely no reason at all! :D

PR #379, opened 2016-04-16
airport challenge

pixelandpage:


silvabox:
Some good design ideas here, but you are breaking encapsulation in a number of places.  Using strings to represent status should be avoided, e.g. you set a plane's locations to `in_the_air` and then test for this value in airport: `if plane.location == "in_the_air"`.  Encapsulate this behaviour in plane and provide an interface e.g. `plane.flying?`.

Beware of superfluous boolean conditionals cluttering your code e.g.:
```
  def landed?
    if !@location.nil? && @location != "in_the_air"
      true
    else
      false
    end
  end
```
which is the same as:
```
  def landed?
    @location != "in_the_air"
  end
```

Finally, in the following:
```
  def safe_to_fly(weather = Weather.new.weather_now)
    if weather == 'stormy'
      false
    else
      true
    end
  end
```
It's great that you've extracted the weather to a separate class, but you're missing an opportunity to interact with it in airport.  This would be better:
```
  def safe_to_fly?(weather = Weather.new)
    !weather.stormy?
  end
```

PR #380, opened 2016-04-16
Airport Challenge - Adam Skuse

AdamSkuse:


lib/airport.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'plane.rb'
+
+class Airport
+
+DEFAULT_CAPACITY = 20
+
+attr_reader :capacity
+attr_accessor :landed
```

lib/plane.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'airport.rb'
+
+class Plane
+attr_accessor :landed
+attr_accessor :stormy
```
danldb: attire_accessor - basically these should never be public. They allow me to call `plane.landed = :banana` without any error.

danldb:
Pay attention to Encapsulation - I would suggest practice rewriting this without attr_accessors and without if/else conditionals.

PR #381, opened 2016-04-16
finished user story 1&2

naridas:


lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
+	def stormy?
+		@storm
+	end
+	def fine
+		@storm = false
+	end
+	def storm
+		@storm = true
+	end
```
danldb: no random weather

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+	def initialize(weather = Weather.new, capacity = DEFAULT_CAPACITY )
+	    @capacity = capacity
+	    @planes = []
+	    @weather = weather
+	end 
+
+	def land(plane)
+		fail 'There is a storm occuring at the airport' if bad_weather
+		fail 'Airport full' if full?
+		planes << plane
```
danldb: watch your indentation here

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'airport'
+
+describe Airport do 
+	let(:plane){ double :plane }
+	let(:weather){double :weather}
```
danldb: you could add a default method return here to save some duplication:
`let(:weather){ double :weather, stormy?: false }`

danldb:
:eyes:

Not quite finished.

I liked your use of doubles and dependency injection for the weather and airport.

PR #382, opened 2016-04-16
Second attempt at a pull request

nmrichards:
Did one yesterday but just realized it didn't go through for some reason.


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    @planes = []
+    @weather = weather
+    @capacity = capacity
+  end
+
+  def land_plane(plane)
+    fail "Too stormy for landing" if @weather.stormy?
+    fail "Airport full" if full?
+    fail "Already landed" if @planes.include?(plane)
+    plane.landed?
```
danldb: No reason to ask the plane if it has landed. Tell it to land, and trust that it will.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    @planes << plane
+    "The plane has landed"
+  end
+
+  def take_off(plane)
+    fail "Too stormy for take-off" if @weather.stormy?
+    fail "Plane cannot take off" unless @planes.include?(plane)
+    plane.flying
```
danldb: use verbs when giving commands, and nouns when asking for a return value. In this case, you would say plane.fly, telling the plane to fly

lib/plane.rb
```diff
@@ -0,0 +1,12 @@
+class Plane
+  attr_reader :landed
+  
+  def landed?
+    @landed = true
```
danldb: plane is always landed when you call `.landed?`. this should return the landed status of the plane, whether true or false.

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
+
+  def stormy?
+    random_weather
+  end
+
+  def random_weather
+    rand(20) <= 1 ? true : false
```
danldb: you could just write:
`rand(20) <= 1` as that already returns true or false

danldb:
👀 
I like your use of DI, it's well tested with doubles.
Work on your semantics - your code is quite hard to read at times because your method naming was non-conventional.

PR #383, opened 2016-04-16
initial commit

chriselevate:


lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def take_off
+    fail "Cannot take off" if stormy?
+    planes.pop
+  end
+
+  private
+
+  def stormy?(weather = Weather.new)
```
danldb: interesting - sometimes it can work very well to inject at this level. I found it a little difficult to immediately see where you had put the dependency, I usually look at the top in the initialise.

danldb:
Unfinished, but what you have is nice and simple. 
:eyes:

PR #387, opened 2016-04-17
The airport challenge

hazukitran:


lib/weather.rb
```diff
@@ -0,0 +1,19 @@
...
+    @number = rand(1..10)
+  end
+
+  def sunny?
+    number == 5 ? false : true
+  end
+
+  def stormy?
+    return false if sunny?
+    true
```
danldb: `not sunny` would work here

danldb:
Looks unfinished - no implementation of landed status of the plane (needs an instance variable).

What is there is nicely simplified and readable. Good use of DI.

PR #388, opened 2016-04-17
Initial commit - Airport Challenge

jackhardy1:


spec/plane_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "plane"
+require 'rspec/expectations'
+
+describe Plane do
+
+  it "instance variable 'landed' can be passed true/false as a value" do
+    plane = Plane.new
+    plane.landed = true
+    expect(plane.landed).to eq true
```
danldb: vacuous test

danldb:
Your Plane class is breaking encapsulation - public attr_accessor is to be avoided. I can easily call `plane.landed = :banana` without any error being raised.

Your plane class doesn't look like you used test-first. None of the logic has been tested - what happens if you try to land a plane that is already on the ground?

Use DI to inject your weather class.

PR #389, opened 2016-04-17
Airport Challenge

HannaEb:


silvabox:
The code in your airport and plane classes looks very clean and you've written plenty of tests which is good.  I wonder if some of your tests are a bit overcomplicated with all the stubbing.  Ideally, you should avoid stubbing the object under test to ensure you don't test don't pass accidentally.  e.g. Rather than stubbing the airport `stormy?` method, stub `Weather.stormy?` instead.

The one thing that does confuse me is the `plane` parameter in airport's `initialize` method.  It's only used in the `flying?` method: which is equally confusing - is an airport ever flying?

PR #390, opened 2016-04-17
Planes take off & land

macebake:


silvabox:
I like the idea behind `check_plane_status` as this can provide good encapsulation.  Unfortunately, your code breaks encapsulation elsewhere by exposing `planes`.  Your plane deletes itself from the airport's `planes` array and adds itself to the array when landing (this is the airport's responsibility).  Your `cleared_for_xxx` methods each take a `_plane` parameter, but it's never used, so should be removed.

Note also, in your tests, you are declaring some doubles, but then never using them!

PR #392, opened 2016-04-17
Attempt at solution for the airports challenge

rahulrama:
This is my attempt at the solution for the airports challenge presented to us during Week 1 of the Makers Academy Course. 

'Pull early and often' - this solution will have multiple commits that will update the pull request automatically along the way, so as to check conformity with the conditions imposed by the Hound.

Meow...

### DIFF
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,6 +1,6 @@
 class Plane
 
-attr_reader :landed
+attr_accessor :landed
 
   def initialize(landed = true)
     @landed = landed
@@ -12,6 +12,10 @@ attr_reader :landed
 
   def take_off
     @landed = false
+  end
+
+  def docked?
+    landed
   end
 
 end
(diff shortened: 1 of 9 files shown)


lib/plane.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize(landed = false)
+    @landed = landed
+  end
+
+  def land
+    @landed = true
+  end
+
+  def take_off
+    @landed = false
```
danldb: why use the `@` when you have an attr_accessor?

lib/plane.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def docked?
+    landed
+  end
+
+  private
+  attr_accessor :landed
```
danldb: This could be an attr_writter

danldb:
:eyes:
A few of the errors in airport actually belonged in Plane (they purely related to the state of the plane).

Use DI for the weather.

rahulrama:
Thanks for the feedback - i'll try and redo the exercise, and if not, atleast incorporate the changes into the current weeks'

PR #393, opened 2016-04-17
Working airport

josephchin19293:


silvabox:
I like the interaction between plane and airport, but your tests are a bit of a problem.  You don't need all of those `let` statements at the top: use `context` or `describe` blocks to organize your tests and only use `let` to create objects used in the tests.  To initialize behaviour, use `before` blocks.  You should definitely avoid `instance_variable_set` and `instance_variable_get` wherever possible.

PR #394, opened 2016-04-17
User stories satisfied 

RichardWatkins1:
Code/tests aren't elegant yet. But I thought I should submit this before having a go at refactoring.


danldb:
You seem to have done something very odd when initialising the git repo - you have pushed all your gems up to git, making a massive repo!

.bundle/config
```diff
@@ -0,0 +1,2 @@
+---
+BUNDLE_DISABLE_SHARED_GEMS: '1'
```
danldb: Maybe this could be your problem with the thousands of files. Looks like you installed your gems into the project directory instead of instilling them system wide.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+require_relative 'weather'
+
+class Airport
+
+	attr_reader :planes, :capacity, :weather
+
+	DEFAULT_CAPACITY = 50
+
+	def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
+		@capacity = capacity
```
danldb: beautiful DI!

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
+
+require 'airport'
+require 'plane'
+
+describe Airport do
+	let(:plane) {Plane.new}
+	let(:airport) {subject}
+	it 'says that it has landed' do 
+	allow(airport).to receive(:stormy?).and_return(false)
```
danldb: You should never stub the subject under test (that is, 'allow(subject).to receive'). Because you have stubbed the subject to always return false, you missed a bug in the code.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+	def land(plane)
+		fail 'the plane hasnt taken off yet' if plane.flying? == false
+		fail 'the weather is stormy' if stormy? == true
+		fail 'no spaces available' if full?
+		planes << plane
+	end
+
+
+	def release_planes
+		fail 'the weather is stormy' if stormy?
```
danldb: should be `weather.stormy?`.

Because you stubbed the `stormy?` method in your tests, you missed this bug.

spec/plane_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'plane'
+
+describe Plane do
+	let(:plane) {subject}
+  	it {is_expected.to respond_to :flying?}
+  	it {is_expected.to respond_to :report_landed}
+  	it 'reports the plane is flying' do
+   	plane.report_landed
```
danldb: Indentation!

PR #395, opened 2016-04-17
Airport Challenge 

Melodija:


silvabox:
There are some nice ideas in your code and it's mostly quite clean.  Be wary of vacuous tests though:
```ruby
      it "should return false if weather is stormy" do
        allow(subject).to receive(:safe?).and_return(false)
        expect(subject.safe?).to eq false
      end
```
doesn't test your object - just that RSpec stubbing works!
Try to choose more descriptive method names like `stormy?` instead of `safe?` so the code is closer to the domain.

In a couple of places you have a superfluous `if` statement in your tests e.g.:
'''ruby
expect{subject.land(plane)}.to raise_error "Plane already landed." if plane == plane
'''
the `if plane == plane` is irrelevant and should be removed.

PR #397, opened 2016-04-17
Airport challenge attempt towards completion

MaxProvin:


lib/airport.rb
```diff
@@ -0,0 +1,65 @@
...
+        end
+    end
+
+    def weather_set
+        if rand(4) == 0
+            weather = 'stormy'
+        else
+            weather = 'sunny'
+        end
+    end
```
danldb: duplication or weather_check?

lib/plane.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class Plane
+    
+    def landed?(airport)
+        if airport.hangar.include? self
+            true
+        else
+            false
+        end
+    end
```
danldb: `airport.hanger.include? self` returns true or false anyways, so no need for the if else statement

spec/plane_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'plane'
+require 'airport'
+
+describe Plane do
+    
+    subject(:plane) {described_class.new}
+    let(:airport) { double :airport }
+    
+    it { is_expected.to respond_to(:landed?) }
+
```
danldb: you haven't tested the logic - can a plane take off if it is already flying?

danldb:
:eyes:

PR #398, opened 2016-04-17
weather class

hnamitha1:


danldb:
:eyes:
Good use of DI. Nice, simple objects.

lib/plane.rb
```diff
@@ -0,0 +1,11 @@
+class Plane
+
+	def landed?
+		true
+	end	
+
+	def take_off?
+		true
```
danldb: The plane has no state - use an instance variable for landed or not.

spec/plane_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'plane'
+
+describe Plane do
+  
+  it {is_expected.to respond_to :landed?}
+  it {is_expected.to respond_to :take_off?}
+
```
danldb: You didn't test the logic - can a plane take off if it is already flying?

PR #399, opened 2016-04-17
Refactored and commented

ChrisKummelstedt:


ChrisKummelstedt:
Had added more comments during evening that I had not pushed.

danldb:
I would look at redoing this later in the course. Your objects are over complicated - many if/else statements. You have mixed presentation logic (e.g. "Too many planes on the tarmac, assume holding pattern") into your errors. Plane is unfinished. Weather object looks good - just need to use DI to inject it into the airport.

PR #400, opened 2016-04-17
First push and commit for early checks

knowerlittle:


silvabox:
Some nice clean code here, but watch out for vacuous tests.  `plane.landed` and `plane.takeoff` tests only test the return values from those methods - not that the state of `flight` has changed.  `landed` is an adjective - maybe `land` would be better; and `flying?` instead of `flight`.

Also, your weather test doesn't test the `Weather` class - it just tests the double.

Should your airport know about the internal implementation of `Weather`?  It currently asks `weather` for a number, which it must then understand and implement business logic upon (i.e. `weather.condition < 60`).  This should be encapsulated in `weather`.  i.e. `weather.stormy?`

knowerlittle:
Will do some refactoring. Thanks for the review!

PR #401, opened 2016-04-18
completed and checking for tests

Vollcode:
Aiming for good coverage!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,3 +8,4 @@ gem 'coveralls', require: false
 gem 'rubocop'
 gem 'reek'
 gem 'rake'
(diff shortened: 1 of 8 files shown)


silvabox:
Some nice clean code in places but there are some issues.
In your airport spec you have this: `expect(subject.arrival(:plane)).to eq([:plane])`.  This violates encapsulation and does not support duck typing - can you see why?

Also note that although your tests pass, you won't be able to run your code in IRB because you haven't implemented a `stormy?` method anywhere in your airport.  You should always run some manual acceptance tests at some point toward the end, just to be sure.

PR #403, opened 2016-04-18
Submission of weekend challenge - week 1

bpstein:


danldb:
Great, you have managed to really simplify the domain.

Crafted a very nice weather module, but they did not use Dependency Injection to pass it into the code - leading to problems where you had to stub the method call on the object itself.

PR #404, opened 2016-04-18
airport challenge 

harrywynnwill:


silvabox:
Your design and tests for `Plane` are a little confused.  You have a method called `storm` which changes the internal state of `weather` to either `true` or `false`.  When naming your objects, think about the language of the domain.  e.g. if I ask "what's the weather?" do you reply "true"?

PR #405, opened 2016-04-18
finished challenge

MariaRomero:
Still will like to find a way to refactor the code a bit and find a way to used mixins as I still don't really understand how they work.

### DIFF
--- a/lib/airplane.rb
+++ b/lib/airplane.rb
@@ -1,0 +1,8 @@
+class Airplane
+	def initialize
+  	end
+
+  	def land
+  		@flying = false
+  	end		
+end	
--- a/lib/airplane.rb
+++ b/lib/airplane.rb
@@ -1,8 +1,20 @@
 class Airplane
-	def initialize
+
+	attr_reader :flying
+
+	def initialize(flying = true)
+		@flying = flying
   	end
 
   	def land
   		@flying = false
   	end		
+
+  	def take_off
+  		@flying = true
+  	end
+
+  	def flying?
+  		@flying
+  	end	
 end	
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,10 +1,54 @@
 class Airport
-	def initialize
-		airport_array = []
+
+	DEFAULT_CAPACITY = 20
+
+	attr_reader :capacity
+	attr_reader :apron
+
+	def initialize(capacity = DEFAULT_CAPACITY)
+		@capacity = capacity
+		@apron = []
   	end
 
-  	def land(airplane)
-  		airport_array << airplane
+  	def land(airplane, weather)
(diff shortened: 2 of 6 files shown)


lib/airplane.rb
```diff
@@ -0,0 +1,20 @@
+class Airplane
+
+	attr_reader :flying
+
+	def initialize(flying = true)
+		@flying = flying
```
danldb: Indentation!

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+	attr_reader :capacity
+	attr_reader :apron
+
+	def initialize(capacity = DEFAULT_CAPACITY)
+		@capacity = capacity
+		@apron = []
+  	end
+
+  	def land(airplane, weather)
+  		if landing_permission(weather) == true && airplane.flying? == true
```
danldb: Extract this to a separate method

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
+class Airport
+
+	DEFAULT_CAPACITY = 20
+
+	attr_reader :capacity
+	attr_reader :apron
+
+	def initialize(capacity = DEFAULT_CAPACITY)
+		@capacity = capacity
+		@apron = []
```
danldb: why `@apron`? The conventional name for a list of planes would be `@planes`

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  	def land(airplane, weather)
+  		if landing_permission(weather) == true && airplane.flying? == true
+  			airplane.land
+  			@apron << airplane
+  		end
+  	end	
+
+  	def fly(weather)
+  		if take_off_permission(weather) == true
```
danldb: instead of using `if` statements in this way, use guard clauses:

```ruby
  raise 'bad weather' unless take_off_permission
  airplane.take_off
```

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+  			airplane = remove_plane
+  			airplane.take_off
+  			airplane
+  		end
+  	end	
+		
+	def full?
+		if @apron.count >= @capacity
```
danldb: This entire method could be written as so:
```ruby
def full?
  @apron.count >= @capacity
end
```

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+			true
+		else
+			false
+		end
+	end
+
+	def take_off_permission(weather)
```
danldb: could be private

danldb:
I like the way you simplified the problem down into small parts.  You could improve your Dependency Injection w/re to the weather. Please please fix your indentation!

PR #423, opened 2016-05-28
Challenging PR

vannio:
:wolf::x::sob:


lib/airport.rb
```diff
@@ -0,0 +1,22 @@
+class Airport
+  DEFAULT_CAPACITY = 20.freeze
```
vannio: Oh hound, you're going to be the death of me

PR #446, opened 2016-05-30
Cameron Epstein: Airport Challenge (Complete)

cameronepstein:


lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+#Airport Challenge: Weather 
+
+class Weather
+
+	#Random number generator with '1' representing a bad weather situation:
+
+	def forecast? 
+		x = rand(1...5)
+		x == 1
```
sjmog: Could this be DRYed up to one line?

PR #454, opened 2016-07-09
Arunas's Airport implementation

arukomp:
Initial commit with first user story tested and implemented.


arukomp:
This is my final code submission. Awaiting judgement 🙈

PR #475, opened 2016-07-11
Functionality complete, failing tests due to random weather

dan-cl:
I haven't been able to successfully double my random weather variable, and set it to stormy or fine as needed. Therefore certain tests are passing/failing depending on the random weather generated on running the test.


dan-cl:
Fixed Rspec tests

PR #525, opened 2016-09-30
Initial pull request

darrenf54321:


lib/airport.rb
```diff
@@ -0,0 +1,74 @@
...
+  attr_accessor :capacity
+  DEFAULT_CAPACITY = 5
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airport_fleet = []
+    @capacity = capacity
+  end
+
+  def random_weather
+    weather = { "stormy" => 20, "sunny" => 80 }
```
silvabox: could this be a constant?  This hash object is recreated on every call to `random_weather`

lib/airport.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  attr_accessor :capacity
+  DEFAULT_CAPACITY = 5
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airport_fleet = []
+    @capacity = capacity
+  end
+
+  def random_weather
```
silvabox: the `random_` prefix reveals internal implementation details.  When I call this method, I just want to know what the weather is, I don't care that it's random.

lib/airport.rb
```diff
@@ -0,0 +1,74 @@
...
+    weather = { "stormy" => 20, "sunny" => 80 }
+    choose_weighted(weather)
+  end
+
+  def land_plane(plane)
+    if random_weather == "sunny"
```
silvabox: Could this be refactored into a `sunny?` method?

lib/airport.rb
```diff
@@ -0,0 +1,74 @@
...
+    fail "error - that plane has already landed" if @@landed_planes.include? plane
+    @airport_fleet << plane && @@landed_planes << plane
```
silvabox: `&&` is a conditional expression, not a concatenation. Never use it in this way.

lib/airport.rb
```diff
@@ -0,0 +1,74 @@
+
+class Airport
+
+  @@flying_planes = []
```
silvabox: Avoid class variables.  Does an airport need to know about all flying planes everywhere in the world and similarly all landed planes?

lib/airport.rb
```diff
@@ -0,0 +1,74 @@
...
+  def random_weather
+    weather = { "stormy" => 20, "sunny" => 80 }
+    choose_weighted(weather)
+  end
+
+  def land_plane(plane)
+    if random_weather == "sunny"
+    fail "error - that plane has already landed" if @@landed_planes.include? plane
+    @airport_fleet << plane && @@landed_planes << plane
+    @@flying_planes.pop
```
silvabox: This is logically incorrect.  What does the `pop` method do?

lib/airport.rb
```diff
@@ -0,0 +1,74 @@
...
+      "Plane has landed"
+    else
+      "Plane cannot land as it is stormy"
+    end
+  end
+
+  def land_plane_full(plane)
```
silvabox: This doesn't make sense.  If I interact with an airport, why would I call `land_plane_full`?  I would just call `land_plane`.  If the airport is full, then `land_plane` should fail.

lib/airport.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def land_plane(plane)
+    if random_weather == "sunny"
+    fail "error - that plane has already landed" if @@landed_planes.include? plane
+    @airport_fleet << plane && @@landed_planes << plane
+    @@flying_planes.pop
+      "Plane has landed"
+    else
+      "Plane cannot land as it is stormy"
```
silvabox: Avoid string return values to convey success.  If the plane can't land, raise an error as you do if the plane is already landed.

lib/airport.rb
```diff
@@ -0,0 +1,74 @@
...
+    if airport_full?
+      "plane cannot land as the airport is full"
+    else
+      @@landed_planes << plane
+    end
+
+  end
+
+  def airport_full?
+    @@landed_planes.count >= @capacity
```
silvabox: If I have more than one airport, then `@@landed_planes.count` is the count of all planes at all airports.  So `airport_full?` could be true even if this airport is empty.

lib/airport.rb
```diff
@@ -0,0 +1,74 @@
...
+      fail "error - that plane is already in the air" if @@flying_planes.include? plane
+      @@flying_planes << plane
+      @@landed_planes.pop && @airport_fleet.pop
+      "the plane has taken off"
+    else
+      "plane cannot take off, it's stormy"
+    end
+  end
+
+  def choose_weighted(weighted)
```
silvabox: This is massively overcomplicated and logically incorrect.  Partition your logic according to the value of `target` (which is `rand(100)`)
<= 20 : "stormy" (20% of occurrences)
\> 20 and <= 80: "sunny" (60% of occurrences)
\> 80 : ???? (20% of occurrences)

PR #526, opened 2016-09-30
Weekend 1

DagmaraSz:


lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_accessor :capacity
+  DEFAULT_CAPACITY = 10
+  POSSIBLE_WEATHER = ["sun", "storm", "clouds", "fog", "rain", "snow"]
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def add(plane)
```
maryrosecook: Could this method name use the domain language to make it easier to understand? eg could it be called something like land?

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  def adverse_weather?
+    weather == "storm"
+  end
+
+  def full?
+    @planes.count == @capacity
+  end
+
+  private
+  def weather
```
maryrosecook: Could weather be abstracted out to its own class?

lib/plane.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def landing_message_formatted
+    @status[:message] + " at airport " + @status[:airport].to_s
+  end
+
+  def set_status(status, airport)
+    @status = {message: status, airport: airport}
+  end
+
+  def problems?(action, airport)
```
maryrosecook: Love that the code for this is in a separate method.  However, it feels like it has two responsibilities: checking a plane can land and checking a plane can take off.  Could the common parts be in this method and the uncommon parts be in the land and take_off methods?

lib/plane.rb
```diff
@@ -0,0 +1,62 @@
...
+      landing_message_formatted
+    else
+      @status[:message]
+    end
+  end
+
+  private
+
+  def landed?
+    @status[:message] == "Landed"
```
maryrosecook: I see what this code is trying for.  It's has an interesting approach.  Could we have a chat about its complexities?

PR #527, opened 2016-09-30
Initial pull request.

hanrattyjen:


lib/airport.rb
```diff
@@ -0,0 +1,44 @@
+require_relative 'plane'
+
+class Airport
+  attr_reader :planes, :capacity
```
silvabox: avoid exposing `planes` as an attribute

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+  def depart_plane(plane)
+    fail "Plane cannot take off, it is too stormy." if stormy?
+    fail "This plane has already taken off." if plane.landed? == false
+    fail "This plane cannot depart from this airport." if planes.include?(plane) == false
+    plane.confirm_takeoff
+    planes.delete(plane)
+    planes
+  end
+
+  def list
```
silvabox: this is redundant as `planes` is already available as a public method.

spec/feature_spec.rb
```diff
@@ -0,0 +1,28 @@
+require './lib/airport'
```
silvabox: this shouldn't be a `_spec.rb` file as it's not an RSpec test

spec/plane_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'plane'
+
+describe Plane do
+
+  it {is_expected.to respond_to :landed?}
+
+  describe 'initialization' do
+    it 'defaults to a flying plane' do
+      expect(described_class.new.landed?).to eq false
```
silvabox: `expect(subject).not_to be_landed` achieves the same thing.

PR #528, opened 2016-09-30
Airport challenge solution from Frances Maxwell

francesmx:


lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+  attr_accessor :capacity
+  attr_reader :airport_id, :planes_on_ground
+
+  def initialize (capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes_on_ground = []
+    @airport_id = self.object_id
+  end
+
+  def land (plane)
```
silvabox: Do not put a space between the method name and the first `(` in method defs

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+    check_for_landing_issues (plane)
```
silvabox: do not put a space between the method name and the `(` when calling a method

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+    plane.flying = false
+    plane.set_location (airport_id)
```
silvabox: What utility is there in the plane knowing the `object_id` of the airport?  You are losing information here, it would be more useful to pass the full airport object to plane.  It's the plane's responsibility to decide what it does with that.

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+
+private
+
+  def full?
+    planes_on_ground.count == capacity
+  end
+
+  def check_for_landing_issues (plane)
+    fail "Cannot land. Airport is full." if full?
+    fail "Cannot land. Plane isn't flying." if !plane.flying?
```
silvabox: Avoid negated `if` conditionals - prefer `unless` instead:
```ruby
fail "Cannot land. Plane isn't flying." unless plane.flying?
```

lib/plane.rb
```diff
@@ -0,0 +1,23 @@
...
+    @flying = true
+    @location = "Plane is flying"
+  end
+
+  def flying?
+    flying
+  end
+
+  def set_location(airport_id)
+    @location = "Plane is at airport #{airport_id}"
```
silvabox: Again, you are losing precision here by burying an id inside a string.  If you later needed to utilize this id, you'd have to engineer it back out of the string.

lib/plane.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def get_location
```
silvabox: avoid `get_` and `set_` style methods in Ruby.

spec/airport_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+describe Airport do
+
+  let(:landed_plane) { double :plane, flying?: false }
+  let(:flying_plane) { double :plane, flying?: true }
+
+  # Credit to Chet Sanghani for using let to instantiate plane object so that it's reusable
+  let(:plane) { Plane.new }
+
+  # Credit to Chet Sanghani for helping me out with the method stub in line 16
+  context "Checks plane's eligibility for landing" do
```
silvabox: This is not a context

spec/airport_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+    it "checks that the plane is flying" do
+      allow(landed_plane).to receive(:flying).with(no_args)
```
silvabox: Is this stub needed?  The airport calls `flying?` which is already mocked in the `let` statement.

spec/airport_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+      expect { subject.land(landed_plane) }.to raise_error "Cannot land. Plane isn't flying."
+    end
+
+    it "checks that the plane is not already in the airport" do
+      expect(subject.planes_on_ground.include?(plane)).to eq false
```
silvabox: Compare this test to its description.  This expectation only tests that a new airport does not contain a new plane.

spec/airport_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+    it "won't allow the plane to land if the weather is stormy" do
+      allow(subject).to receive(:weather) { "stormy" }
+      expect { subject.land(plane) }.to raise_error "Cannot land due to storm."
+    end
+
+    it "won't allow the plane to land if the airport is full" do
+      allow(subject).to receive(:weather) { "sunny" }
+      subject.capacity.times do
+        plane.flying = true
+        subject.land(plane)
```
silvabox: beware landing the same plane n times

spec/airport_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+      end
+      expect { subject.land(plane) }.to raise_error "Cannot land. Airport is full."
+    end
+
+  end
+
+  context "Instructs plane to land" do
```
silvabox: This is not a context.  Use `context` to group tests that share some state or condition, e.g. `context "When stormy"`

spec/airport_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+    it "instructs the plane to land (provided it's sunny)" do
+      allow(subject).to receive(:weather) { "sunny" }
+      subject.land(plane)
+      expect(plane).not_to be_flying
```
silvabox: Does this test that the airport instructed the plane to land?

spec/airport_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+      allow(subject).to receive(:weather) { "sunny" }
+      subject.land(plane)
+      expect(subject.planes_on_ground.include?(plane)).to eq true
+    end
+
+  end
+
+  context "Checks plane's eligibilty for take off" do
+
+    it "checks that the plane is not already flying" do
```
silvabox: Try to word your test descriptions according to behaviour, not implementation.  e.g.:
```ruby
describe 'take off` do
  it 'fails if the plane is not flying` do
     ...
  end
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+      expect { subject.takeoff(flying_plane) }.to raise_error "Cannot take off. Plane is flying."
+    end
+
+    it "checks that the plane is in the airport" do
+      allow(subject).to receive(:weather) { "sunny" }
+      subject.land(plane)
+      expect(subject.planes_on_ground.include?(plane)).to eq true
+    end
+
+    it "checks multiple airports to make sure the plane is definitely in THIS airport" do
```
silvabox: Does an airport check multiple airports?

spec/airport_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+      Heathrow = Airport.new
```
silvabox: Be careful of your syntax here.  You are defining a new constant `Heathrow` here, when you should only be defining a local variable (`heathrow`).  Ruby is case sensitive.

spec/airport_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+  describe "Airport's weather conditions are variable" do
+
+    it "checks that the weather is either sunny or stormy" do
+      weather_conditions = ["sunny", "stormy"]
+      expect(weather_conditions.include?(subject.weather)).to eq true
+    end
+
+    it "allows you to check the weather (forcing it to be sunny)" do
+      allow(subject).to receive(:weather) { "sunny" }
+      expect(subject.weather).to eq "sunny"
```
silvabox: This is not testing the airport's behaviour.  It is simply testing that the mock on the line before has worked.

spec/plane_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  describe "Plane has a default status" do
+
+    it "is flying by default" do
+      expect(subject).to be_flying
+    end
+
+  end
+
+  describe "Plane's status changes on taking off or landing'" do
```
silvabox: This is  a feature test

spec/plane_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    it "sets flying to true on taking off" do
```
silvabox: Does the plane set `flying?` or does the airport?  Whose responsibility are you testing?

PR #529, opened 2016-10-01
airport challenge CI

crsanghani:


spec/airport_spec.rb
```diff
@@ -0,0 +1,66 @@
+require 'spec_helper'
+
+describe Airport do
+
+    let(:flying_plane) {double :plane, flying?: true}
+    let(:landed_plane) {double :plane, flying?: false}
+    let(:plane) {Plane.new}
+    subject(:airport) { described_class.new }
+
+    context "Launch tests" do
```
silvabox: use `context` to group tests by some state or condition, e.g. 'when stormy'

When grouping by behaviour, prefer `describe`.  But `describe "Launch tests"` would be wrong as you are not describing the tests, you are describing the 'Launch' behaviour of `Airport`.  Think of each `it` block as a description of the described object.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
+require './lib/plane'
+
+class Airport
+
+    attr_accessor :capacity, :planes, :planes_landed
```
silvabox: avoid `attr_accessor` for anything other than simple properties that do not influence the object's behaviour.  Do not expose `planes` as an attribute at all.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    plane.flying = false
+    @planes_landed << plane
+    puts "#{plane} has landed"
+    plane
+  end
+
+  def launch(plane)
+    fail "Can't take off. Plane is flying." if plane.flying?
+    plane.flying = true
+    @planes_landed.pop
```
silvabox: This is logically incorrect.  Your tests are insufficient as they do not surface this problem.  Can you see the problem?

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    puts "#{plane} has taken off"
+  end
+
+  def full?
+    @planes_landed.count >= DEFAULT_CAPACITY
+  end
+
+  #set 6 for roughly once a week 30 for a month etc
+  def weather
+    rand(6) == 0 ? "stormy" : "sunny"
```
silvabox: use symbols instead of strings here: i.e. `:stormy` and `:sunny`

spec/airport_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        
+        it "let a grounded plane launch:" do
+            subject.land(plane)
+            subject.launch(plane)
+            expect(plane).to be_flying
```
silvabox: This is more like a feature test.  You are not isolating the airport's behaviour, instead you are testing the final outcome with a real plane.  What is the airport's responsibility here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        it "has a variable capacity" do
+            subject.capacity = 20
+            expect(subject.capacity).to eq 20
+        end
+
+        #credit FrancesMX for getting this block to work
+        it "wont allow a plane to land if airport is full" do
+            subject.capacity.times do
+                plane.flying = true
+                subject.land(plane)
```
silvabox: beware landing the same place n times.

PR #530, opened 2016-10-01
Added airport_spec.rb file

amidabrian51:
Initial pull request.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,1 +1,5 @@
-Class Airport
+class Airport
+  def confirm_to_land
+
+  end
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,6 +1,12 @@
 require_relative 'plane'
+
 class Airport
+
   def confirm_to_land?
     Plane.new
   end
+  def not_in_airport(plane)
+    plane
+  end
+
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,1 @@
+Class Airport
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,5 +1,6 @@
+require_relative 'plane'
 class Airport
-  def confirm_to_land
-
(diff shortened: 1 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,12 @@
+require_relative 'plane'
+
+class Airport
+
+  def confirm_to_land?
```
silvabox: methods suffixed with `?` are intended to return `true` or `false`

lib/airport.rb
```diff
@@ -0,0 +1,12 @@
...
+    Plane.new
+  end
+  def not_in_airport(plane)
+    plane
```
silvabox: I'm not sure what this method is intended to do - currently it only returns the argument that's passed to it.

spec/airport_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'airport'
+require 'plane'
+
+describe Plane do
```
silvabox: Tests for `Plane` should be in `plane_spec.rb`

spec/airport_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  it { is_expected.to respond_to :landing?}
+  it { is_expected.to respond_to :taken_off?}
+end
+describe Airport do
+  it { is_expected.to respond_to :confirm_to_land?}
+  it 'gets the plane to land' do
+    plane = subject.confirm_to_land?
+    expect(plane).to respond_to :landing?
```
silvabox: How is testing that a plane has a `landing?` method related to the airport's behaviour?

spec/airport_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  end
+
+  it 'confirms that the plan is not in the airport' do
+    plane = Plane.new
+    subject.not_in_airport(plane)
```
silvabox: This line is redundant as the expression is repeated in the following line.

PR #531, opened 2016-10-01
Airport Challenge

glynester:
Glenn's attempt at Airport Challenge.

Interim development at this time.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -24,10 +24,12 @@ class Airport
     end
   end
   
-  def take_off
+  def take_off(plane = planes[0])
+    fail "ERROR: There is no plane to take-off." if planes.empty?   # || plane.class.to_s != "Plane")
     @plane_status = "down"
     if airport_open?     #Checks weather on "take_off" call
-      planes.pop
+      planes.delete(plane)
+      #planes.pop
       @plane_status = "up"
     else
       puts "Weather is too stormy to allow planes to take-off." 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -3,10 +3,12 @@ require_relative 'weather'
 
 class Airport
   
-  attr_accessor :planes, :plane, :plane_status, :runway_open     
+  attr_accessor :planes, :plane, :plane_status, :runway_open, :airport_capacity     
   
-  def initialize
+  def initialize (capacity = 20)
     @planes = []
+    @airport_capacity = capacity    #Set to 20 if no capacity is provided.
+    @runway_open = true
   end
   
   def is_plane_at_airport?(plane)
@@ -16,8 +18,13 @@ class Airport
   def airport_open?
     weather_forecaster == "good" ? @runway_open = true : @runway_open = false
   end
+  
+  def is_airport_full?
+    planes.count == airport_capacity ? true : false
+  end
     
   def land_plane(plane)
+    fail "You cannot land because the airport is full." if is_airport_full?
     fail "ERROR: That plane has already landed." if is_plane_at_airport?(plane)
     fail "ERROR: There is no plane to land." if (plane.nil? || plane.class.to_s != "Plane")
     @plane_status = "up"
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,14 +1,15 @@
-#require_relative 'plane'
+require_relative 'plane'
 
-#class Airport
+class Airport
     
     
     
     
     
-#end
+end
 
 
+=begin
 #require_relative 'bike'
 require_relative 'plane'
 
@@ -45,3 +46,6 @@ class DockingStation
   end
 
 end
+
+
+=end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -14,6 +14,7 @@ class Airport
   end
     
   def land_plane(plane)
+    fail "ERROR: There is no plane to land." if (plane.nil? || plane.class.to_s != "Plane")
     @plane_status = "up"
     if airport_open?      #Checks weather on "land_plane" call
       planes << plane
@@ -35,14 +36,21 @@ class Airport
     
 end
 
-
-
+=begin
+#Cannot land a plane if there is no plane object.
 airport = Airport.new
-#puts "WEATHER FORECAST"
-#puts airport.airport_open?
-
 plane = Plane.new
-puts "PLANE INSTRUCTED TO LAND"
+airport.land_plane(plane)
+  #plane = "I'm a little aeroplane up in the sky"
+  #airport.land_plane(plane)
+  #plane = nil
+  #airport.land_plane(plane)
+=end
+
+=begin
+airport = Airport.new
+plane = Plane.new
+puts "PLANE INSTRUCTED TO LAND - will not if weather is stormy"
 airport.land_plane(plane)
 puts "Is the runway open?: #{airport.runway_open}"
 puts "Array of planes at airport: #{airport.planes.inspect}"
@@ -50,13 +58,13 @@ puts "Plane status is: #{airport.plane_status}"
 puts
 plane = Plane.new
 airport.planes << plane
-puts "PLANE INSTRUCTED TO TAKE-OFF"
+puts "PLANE INSTRUCTED TO TAKE-OFF - will not if weather is stormy"
 airport.take_off
 puts "Is the runway open?: #{airport.runway_open}"
 puts "Array of planes at airport: #{airport.planes.inspect}"
 puts "Plane status is: #{airport.plane_status}"
 
-
+=end
 
 
 =begin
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+  
+  def is_airport_full?
+    planes.count == airport_capacity ? true : false
+  end
+    
+  def land_plane(plane)
+    fail "You cannot land because the airport is full." if is_airport_full?
+    fail "ERROR: That plane has already landed." if is_plane_at_airport?(plane)
+    fail "ERROR: There is no plane to land." if (plane.nil? || plane.class.to_s != "Plane")
```
silvabox: avoid testing types in Ruby.   This line will make testing with doubles very difficult.

lib/airport.rb
```diff
@@ -0,0 +1,87 @@
...
+    @plane_status = "up"
+    plane.plane_status = "up"       #Sets plane status
+    if airport_open?                #Checks weather on "land_plane" call
+      planes << plane
+      @plane_status = "down"
```
silvabox: What is the local `@plane_status` for?

lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
+    
+    attr_accessor :plane_status     #plane_status is set by the airport.
```
silvabox: `plane_status` should not be set by the airport.  You don't want external code to have full control over this value - it could be set to anything.  Define an interface on plane instead, e.g.:
```ruby
def up
  @status = :up
end
```
Note the `plane_` prefix is redundant as it's implicit.

lib/airport.rb
```diff
@@ -0,0 +1,87 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  
+  attr_accessor :planes, :plane, :plane_status, :runway_open, :airport_capacity
```
silvabox: `planes` should not be an attribute at all, while the others `attr_reader` at best.

lib/airport.rb
```diff
@@ -0,0 +1,87 @@
...
+    if airport_open?                #Checks weather on "land_plane" call
+      planes << plane
+      @plane_status = "down"
+      plane.plane_status = "down"
+    else
+      puts "Weather is too stormy to allow planes to land." 
+    end
+  end
+  
+  def take_off(plane = planes[0])
```
silvabox: avoid specifying defaults in this way as it can create unpredictable results and requires far more test cases.  Better just to make the argument mandatory.

lib/airport.rb
```diff
@@ -0,0 +1,87 @@
...
+    fail "ERROR: There is no plane to take-off." if planes.empty? || !is_plane_at_airport?(plane)   # || plane.class.to_s != "Plane")
```
silvabox: This clause is difficult to read as it's easy to miss the `!` before `is_plane_at_airport?(plane)`
The initial test is also redundant as the plane can't be at the airport if the airport is empty; so this would be better:
```ruby
fail "ERROR: There is no plane to take-off." unless is_plane_at_airport?(plane)
```

lib/airport.rb
```diff
@@ -0,0 +1,87 @@
...
+    @airport_capacity = capacity    #Set to 20 if no capacity is provided.
+    @runway_open = true
+  end
+  
+  def is_plane_at_airport?(plane)
+    planes.include? plane
+  end
+  
+  def airport_open?
+    weather_forecaster == "good" ? @runway_open = true : @runway_open = false
```
silvabox: Do you need to memoize `@runway_open`?

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+def weather_forecaster
```
silvabox: This essentially defines a global function `weather_forecaster`.  Your methods should always belong to a class or module.

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+    describe "- Airport sends instructions to land." do
+        it { is_expected.to respond_to(:land_plane).with(1).argument}  
+        plane = Plane.new
+        it "Airport returns a status of 'down' when plane lands." do
+            allow(subject).to receive(:airport_open?){true}           #This removes random weather factor.
+            expect(subject.land_plane(plane)).to eq "down"
+        end    
+    end
+    
+    describe "- Plane instructed to take-off." do
```
silvabox: What's happened to your indentation?

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
+require 'airport'
+
+describe Airport do
+    
+    describe "- Airport sends instructions to land." do
+        it { is_expected.to respond_to(:land_plane).with(1).argument}  
+        plane = Plane.new
+        it "Airport returns a status of 'down' when plane lands." do
```
silvabox: `it` is the airport, so this should read: `it "returns a status of 'down' when plane lands."`

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
+require 'airport'
+
+describe Airport do
+    
+    describe "- Airport sends instructions to land." do
```
silvabox: This `describe` block is superfluous as it's not aiding readability or providing a scope

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        it { is_expected.to respond_to(:land_plane).with(1).argument}  
+        plane = Plane.new
```
silvabox: `plane = Plane.new` does not belong here.  It should be inside an `it` block, or in a `let` statement.

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        it "Airport returns a status of 'up' when plane takes-off." do
+            subject.planes << Plane.new
+            allow(subject).to receive(:airport_open?){true}           #This removes random weather factor.
+            expect(subject.take_off).to eq "up"
+        end
+    end
+    
+    describe "- Plane instructed to land when there is no plane." do
+        it "Plane does not land if there is no plane object. An error is thrown." do
+            plane = nil
```
silvabox: this line is redundant

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+            plane = "123456789"
+            expect{subject.land_plane(plane)}.to raise_error("ERROR: There is no plane to land.")
+        end
+    end
+    
+    describe "- Plane instructed to take-off when there is no plane." do
```
silvabox: All of these extra `describe`s make your tests difficult to read.  This one doesn't even make sense:
"- Plane instructed to take-off when there is no plane."
Try reversing the sentence: "When there is no plane, the plane is instructed to take off"

lib/airport.rb
```diff
@@ -0,0 +1,87 @@
...
+    fail "ERROR: That plane has already landed." if is_plane_at_airport?(plane)
+    fail "ERROR: There is no plane to land." if (plane.nil? || plane.class.to_s != "Plane")
+    @plane_status = "up"
+    plane.plane_status = "up"       #Sets plane status
+    if airport_open?                #Checks weather on "land_plane" call
+      planes << plane
+      @plane_status = "down"
+      plane.plane_status = "down"
+    else
+      puts "Weather is too stormy to allow planes to land."
```
silvabox: `puts` should only be used for logging and debugging.

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        it "Plane does not take off if there are no planes in the airport. An error is thrown." do
+            subject.planes = []     
+            expect{subject.take_off}.to raise_error "ERROR: There is no plane to take-off."
+        end
+    end
+    
+end
+
+describe Airport do
```
silvabox: Do we need to describe the `Airport` twice?

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+            @airport.take_off(@plane)   
+            expect{@airport.take_off(@plane)}.to raise_error "ERROR: There is no plane to take-off."
+        end
+    end
+    
+   
+    
+end
+
+describe Airport do
```
silvabox: ...three times?

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+    describe "random weather" do
+        
+        it "produces 'stormy' or 'fine' weather" do
+            airport = Airport.new
+            expect(airport.airport_open?).to satisfy { |v| [true,false].include? v }
```
silvabox: You are not testing random weather here - you are testing `airport_open?`  The two may well be related but nothing in this test proves that.

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        it "Produces an error message if a plane tries to land when the airport is full." do
+            airport = Airport.new
+            airport.airport_capacity.times {airport.planes << Plane.new}
+            expect {airport.land_plane(Plane.new)}.to raise_error "You cannot land because the airport is full."
+        end
+    end
+ 
+    describe "Edge cases:" do
+       before(:each) do         #Sets variables for the 2 tests below.
+            @plane = Plane.new
```
silvabox: Don't define instance variables anywhere in your RSpec files.  Use `let` and `subject`

PR #533, opened 2016-10-01
First stab

wirsindpapst:
instruct_to_land, land, instruct_to_takeoff, takeoff and check_status methods created

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,35 +1,41 @@
-
 class Airport
 
-# attr_accessor :clearance_for_landing,
+  DEFAULT_CAPACITY = 5
 
-def initialize
-#   @clearance_for_landing = true
-   @landed_planes = []
-   @clearance = true
-#   puts planes
- end
+  attr_reader :capacity # can maybe get rid of this - need to look into refactoring
 
-def clearance(weather)
-  @clearance = weather.stormy?
+def initialize(capacity = self.class::DEFAULT_CAPACITY)
+  @landed_planes = []
+  @capacity = capacity
+  @clearance = true
 end
 
-
 def instruct_to_land(plane)
-  fail "Clearance for landing not granted: stormy weather" if @clearance == false
+  fail "Clearance for landing not granted: stormy weather" if !@clearance
+  fail "Clearance for landing not granted: stormy weather" if at_capacity?
   plane.land
   @landed_planes << plane
   @landed_planes.last
 end
 
 def instruct_to_takeoff(plane)
-  fail "Clearance for landing not granted: stormy weather" if @clearance == false
+  fail "Clearance for takeoff not granted: stormy weather" if !@clearance
   plane.take_off
   @landed_planes.delete(plane)
 end
 
+def clearance(weather)
+  @clearance = !weather.stormy?
+end
+
 def confirm_status(plane)
   plane.status
+end
+
+private
+
+def at_capacity?
+  @landed_planes.count == capacity
 end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,32 @@
+
+class Airport
+
+# attr_accessor :clearance_for_landing,
+
+def initialize
+#   @clearance_for_landing = true
+   @landed_planes = []
+#   puts planes
+ end
+
+def clearance_for_landing
+  @clearance_for_landing
+
+end
+
+def instruct_to_land(plane)
+  plane.land
+  @landed_planes << plane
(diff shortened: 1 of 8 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
+class Airport
+
+  DEFAULT_CAPACITY = 5
+
+  attr_reader :capacity # can maybe get rid of this - need to look into refactoring
+
+def initialize(capacity = DEFAULT_CAPACITY)
```
silvabox: Indentation!

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+  fail "Clearance for takeoff not granted: stormy weather" if !@clearance
+  plane.take_off
+  @landed_planes.delete(plane)
+end
+
+def clearance(weather)
+  @clearance = !weather.stormy?
+end
+
+def confirm_status(plane)
```
silvabox: Is this method redundant?  I have to pass `plane` to it anyway, so why would I not just call `plane.status` myself?

lib/plane.rb
```diff
@@ -0,0 +1,25 @@
+require_relative 'airport'
+
+class Plane < Airport
```
silvabox: erm...?  How does a plane inherit from an airport?

lib/plane.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def initialize
+    @status = :in_flight
+  end
+
+  def land
+    raise "Cannot land: plane is already grounded" if @status == :landed
```
silvabox: You could refactor the conditional into another method to improve SRP and readability:
```ruby
raise "Cannot land: plane is already grounded" if landed?
```

lib/weather.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'airport'
+
+class Weather < Airport
```
silvabox: I think you're misunderstanding inheritance.

lib/weather.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  private
+
+  def randomiser
+    rand(@weather.length-1)
+  end
+
+  def weather_forecast
+    return @weather[randomiser]
```
silvabox: you could use `@weather.sample` instead.  `sample` is an array method that returns a random entry in the array.

spec/airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  describe '#initialize' do
+  it 'sets default capacity' do # need to stub this it uses default capacity
+    allow(plane).to receive(:land)
+    4.times{ subject.instruct_to_land(plane) }
+    expect(subject.instruct_to_land(plane)).to eq plane
+  end
+  it "allows a specific capacity to be set" do
+    allow(plane).to receive(:land)
+    airport = Airport.new(10)
+    10.times { airport.instruct_to_land(plane) }
```
silvabox: beware landing the same plane 10 times

spec/airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      subject.clearance(weather)
+      expect{ subject.instruct_to_land(plane) }.to raise_error "Clearance for landing not granted: stormy weather"
+    end
+    it 'will not allow a plane to land if the airport is at capacity' do
+      airport = Airport.new(1)
+      airport.instruct_to_land(plane)
+      expect{ airport.instruct_to_land(plane) }.to raise_error "Clearance for landing not granted: airport full"
+    end
+    it 'stores the plane in the airport' do
+      subject.instruct_to_land(plane)
```
silvabox: This line is redundant as the same expression is repeated on the next line

spec/airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect( subject.instruct_to_land(plane) ).to eq plane
```
silvabox: This only tests that the `instruct_to_land` method returns the `plane` not that the plane is now stored in the airport.

spec/airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  describe '#instruct_to_take off' do
+    before do
+      allow(plane).to receive(:land)
+      allow(plane).to receive(:take_off)
+      subject.instruct_to_land(plane)
+      plane.land
+    end
+    it 'removes the plane for the airport' do
+      expect( subject.instruct_to_takeoff(plane) ).to eq plane
```
silvabox: This and the following test are identical and neither test what the description suggests they should

PR #534, opened 2016-10-01
Initial pull request

crispinandrews:


spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  let(:plane) { double :plane }
+
+  it {is_expected.to respond_to(:plane)}
+
+  it {is_expected.to respond_to(:land).with(1).argument}
+
+  it 'allows you to check the weather is stormy' do
+    allow(subject).to receive(:weather) { 'stormy' }
+    expect(subject.weather).to eq 'stormy'
```
silvabox: This is a vacuous test since all you are testing here is that RSpec has correctly mocked the `weather` method in the line above.

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  it 'allows you to check the weather is fine' do
+    allow(subject).to receive(:weather) { 'fine' }
+    expect(subject.weather).to eq 'fine'
```
silvabox: as above - if you mock the method under test, you are only testing the mock.

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
+require_relative 'plane'
+
+class Airport
+
+attr_reader :plane
```
silvabox: This is never used - it should be removed

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+attr_accessor :flying, :status, :grounded_planes, :weather, :capacity
```
silvabox: apart from `capacity` these attributes are all unused and should be removed

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    fail 'Plane can not land as the weather is to stormy' if weather == 'stormy'
+    plane.flying = false
+    puts "Plane has landed"
+    @landed_planes << plane
+  end
+
+  def takeoff(plane)
+    fail 'The weather is stormy, no planes can take off' if weather == 'stormy'
+    fail 'The plane is not at this airport!' unless @landed_planes.include?(plane)
+    @landed_planes.pop
```
silvabox: `@landed_planes.pop` is logically incorrect.  Your tests are insufficient as they don't reveal this bug.  Can you see why?

lib/plane.rb
```diff
@@ -0,0 +1,14 @@
+require_relative 'airport'
+
+class Plane
+
+attr_accessor :flying, :status, :airport
```
silvabox: `status` and `airport` are unused and should be removed

lib/plane.rb
```diff
@@ -0,0 +1,14 @@
...
+  def initialize
+    @flying = nil
```
silvabox: You don't need to initialize instance variables to `nil` - they will be `nil` by default.  Avoid setting 'boolean' instance variables to `nil`; prefer `false`.

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+attr_reader :weather
```
silvabox: you shouldn't need attrs anywhere in your spec files.

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  it 'allows you to check the weather is fine' do
+    allow(subject).to receive(:weather) { 'fine' }
+    expect(subject.weather).to eq 'fine'
+  end
+
+  describe '#takeoff' do
+
+    it 'allows plane to takeoff from airport' do
+      allow(plane).to receive(:flying=).and_return false
```
silvabox: return values for attribute setters are generally irrelevant so the following is sufficient:
```ruby
allow(plane).to receive(:flying=)
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      allow(subject).to receive(:weather) { 'fine' }
+      subject.land(plane)
+      subject.takeoff(plane)
+      expect(plane).to be plane
```
silvabox: This test can never fail `expect(plane).to be plane`.  It's just saying "expect something to be itself"
One of the principles of TDD is to write a _failing_ test first.  If you don't see a test properly failing first, you can't be certain that it's passing for the right reasons.

spec/plane_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'plane'
+require 'airport'
+
+describe Plane do
+
+  let(:airport) { Airport.new }
+
+  it {is_expected.to respond_to :flying?}
+
+  it 'updates flying when landing' do
```
silvabox: This is testing the behaviour of `Airport`, not `Plane`.  The expectation is rightly on the status of plane, but it's the airport that makes that happen.

PR #535, opened 2016-10-01
First Commit

asifhafeez:


lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+require_relative 'weather'
+
+class Airport
+	
+	DEFAULT_CAPACITY = 50
+
+	attr_reader :landed_planes, :capacity
+
+	def initialize(capacity = DEFAULT_CAPACITY)
+		@landed_planes = []
```
maryrosecook: Watch out for idiomatic indentation style.  Have a look at a Ruby style guide for more info.

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+		@capacity = capacity
+	end
+
+	def land(plane, weather)
+		raise 'Unsuitable weather for landing!' if weather.stormy?
+		raise "No space for planes in airport" if full?
```
maryrosecook: Nice clear guard clause!

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+
+	def take_off(plane, weather)
+		raise 'Too stormy to take off!' if weather.stormy?
+		raise 'This plane is not currently landed here!' unless @landed_planes.include?(plane)
+		plane.take_off
+		@landed_planes.delete(plane)
+	end
+
+	def full?
+		true if @landed_planes.count >= @capacity
```
maryrosecook: Have a look at the logic of this method.  What does it return if the airport isn't full? Also, there's no need for the `true if` part.

lib/plane.rb
```diff
@@ -0,0 +1,19 @@
+class Plane
+
+	attr_accessor :flying
```
maryrosecook: Does the code need this?

spec/airport_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+	subject(:airport) { described_class.new }
+	let(:plane) { double (:plane) }
+	let(:weather) { double(:weather) }
+	let(:stormy_weather) { double(:weather) }
+
+	before(:each) do
+	 	@default_capacity = Airport::DEFAULT_CAPACITY
+	 	@different_capacity = rand(1...@default_capacity)
+	 	@error1 = 'Unsuitable weather for landing!'
```
maryrosecook: The variable names for these errors don't really explain what they are.  Sometimes, clarity is more important than avoiding repetition.

PR #536, opened 2016-10-01
Branch1

LawrenceHunt:
Weekend Homework... So far

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,4 @@
+require_relative 'plane'
+
+class Airport
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,23 +1,19 @@
 require_relative 'plane'
+require_relative 'weather'
 
 class Airport
+  attr_reader :planes, :capacity, :stormy_weather
 
-attr_accessor :planes, :capacity
+  DEFAULT_CAPACITY = 35
 
-DEFAULT_CAPACITY = 20
-
-  def initialize(capacity = DEFAULT_CAPACITY)
-    @capacity = capacity
+  def initialize(capacity=DEFAULT_CAPACITY)
     @planes = []
+    @capacity = capacity
+    @stormy = (Weather.new).stormy
   end
 
-
-def empty?
-  @planes.empty?
-end
-
-def full?
-  @planes.count >= @capacity
-end
+  def full
+    planes.count > capacity
+  end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,4 +1,23 @@
 require_relative 'plane'
 
 class Airport
+
+attr_accessor :planes, :capacity
+
+DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+
+def empty?
+  @planes.empty?
+end
+
+def full?
+  @planes.count >= @capacity
+end
+
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,7 +2,7 @@ require_relative 'plane'
 require_relative 'weather'
 
 class Airport
-  attr_reader :planes, :capacity, :stormy_weather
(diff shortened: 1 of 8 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,19 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  attr_reader :planes, :capacity, :stormy
```
silvabox: don't expose `planes` outside of this class.

lib/airport.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  DEFAULT_CAPACITY = 35
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+    @stormy = (Weather.new).stormy
```
silvabox: Does an airport only test the weather once and then never again?

lib/plane.rb
```diff
@@ -0,0 +1,31 @@
...
+  alias_method :landed?, :landed
+
+  def initialize
+    @landed = false
+  end
+
+
+    def take_off airport
+      fail "Sir, are you crazy? We're already airborne!" if not landed
+      fail "Sir, are you crazy? This plane isn't at that airport!" unless airport.planes.include?(self)
```
silvabox: all these calls to `airport` are a clear indication that this logic belongs in `Airport`

spec/airport_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  it "has a default capacity" do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+
+  it "can set the capacity" do
+    airport = Airport.new(1000)
+    expect(airport.capacity).to eq 1000
+  end
+
+  it "can tell 2 planes have landed" do
```
silvabox: Did you collaborate on this?  I've seen this same test in another submission.

spec/airport_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    plane1 = double(:plane)
+    allow(plane1).to receive(:land).with(subject).and_return(subject.planes << plane1)
+    plane2 = double(:plane)
+    allow(plane2).to receive(:land).with(subject).and_return(subject.planes << plane2)
+    plane1.land(subject)
```
silvabox: What happens if you delete:
```ruby
plane1.land(subject)
plane2.land(subject)
```
and rerun this test.  Will it pass?

spec/plane_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  let(:airport) {double(:airport, full: false, stormy: false, planes: [])}
+  let(:full_airport) {double(:airport, full: true, stormy: false, planes: [])}
+  let(:airport_with_storm) {double(:airport, full: false, stormy: true, planes: [])}
+
+  describe "#land" do
+
+    it "should be able to land" do
+      subject.land(airport)
+      expect(subject.landed?).to eq true
```
silvabox: `expect(subject).to be_landed`

spec/plane_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+describe Plane do
+  subject(:plane) {described_class.new}
+
+  let(:airport) {double(:airport, full: false, stormy: false, planes: [])}
+  let(:full_airport) {double(:airport, full: true, stormy: false, planes: [])}
+  let(:airport_with_storm) {double(:airport, full: false, stormy: true, planes: [])}
+
+  describe "#land" do
+
+    it "should be able to land" do
```
silvabox: avoid beginning tests with `should`.  Prefer `it "can land"` or similar.

lib/weather.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class Weather
+
+attr_accessor :stormy
+
+  def chance
+    rand(10)
+  end
+
+  def weather
```
silvabox: This leads to code like `weather.weather`, which feels wrong.  Really, I just want to call `weather.stormy?`.  The additional method call isn't needed, just calculate the weather on the fly.

PR #539, opened 2016-10-01
Initial commit: Plane and Airport classes set up

DavidAMAnderson:


lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+  DEFAULT_CAPACITY = 200
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    fail "Plane not in flight" if plane.on_ground
+    fail "Can't land when stormy" if !good_weather?
```
silvabox: Use `unless` to test negatives, it reads better:
```ruby
fail "Can't land when stormy" unless good_weather?
```
Here and elsewhere in the code.

lib/plane.rb
```diff
@@ -0,0 +1,9 @@
+class Plane
+
+  attr_accessor :on_ground
```
silvabox: for boolean attributes, prefer the `?` suffix.  You can't define these using `attr_reader`, you just have to define the method manually  (or create an alias for this attribute)

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def full?
+    @planes.count >= @capacity
+  end
+
+  def plane_in_airport?(plane)
+    @planes.include?(plane)
+  end
+
+  def find_plane(plane)
```
silvabox: This method is a bit confused.  It either raises an error, or returns the same object you pass into it.

PR #540, opened 2016-10-01
Ryanair

tadasmajeris:


lib/weather.rb
```diff
@@ -0,0 +1,27 @@
+class Weather
+  attr_reader :status, :chance
+
+  def initialize
+    @chance = generate
+  end
+
+  # storm chance: (1%..100%)
+  def get_status
```
silvabox: Why have `get_status` and `status`?  In Ruby, we generally avoid `get_` and `set_` style attributes as it's not the Ruby style.

PR #541, opened 2016-10-01
Pull request for the Weekend Challenge 1 - Airport Challenge

diok22:
Pull request created after the First User Story was successfully developed to the repo

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,2 @@
+class Airport
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -18,7 +18,7 @@ class Airport
   end
 
   def land(plane)
-    raise "No gates available" if @gates.count >= @gates_size
+    raise "No gates available - cannot land plane" if @gates.count >= @gates_size
     @gates << plane
     "The plane has landed"
   end
@@ -28,7 +28,5 @@ class Airport
     @gates.pop
     "The plane has departed"
   end
-
-
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -5,27 +5,41 @@ class Airport
   AIRPORT_CAPACITY = 231 # number of gates at London Heathrow - LHR
 
   attr_reader :gates, :gates_size
-  attr_accessor :stormy
+  attr_accessor :stormy_weather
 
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  AIRPORT_CAPACITY = 231 # number of gates at London Heathrow - LHR
+
+  attr_reader :gates, :gates_size
+
+  def initialize(gates_size = AIRPORT_CAPACITY)
+    @gates = []
+    @gates_size = gates_size
+  end
+
+  def stormy_weather
```
maryrosecook: Maybe add a question mark to the end of the method name to show it returns a boolean?

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    rand(1..10) > 7
+  end
+
+  def land(plane)
+    raise "The weather is stormy, plane cannot land" if stormy_weather == true
+    raise "No gates available - cannot land plane" if @gates.count >= @gates_size
+    raise "Plane is already landed" if @gates.include?(plane)
+    @gates << plane
+    "The plane has landed"
```
maryrosecook: No need to return strings to report what happened.  If this is needed for testing, consider returning the plane to indicate that everything went ok, or returning nothing and having the test check that no exceptions are thrown.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def take_off(plane)
+    raise "The weather is stormy, plane cannot takeoff" if stormy_weather == true
+    raise "There are no planes currently at the airport - No departures" if @gates.count == 0
```
maryrosecook: Can you add a method called something like full? to make this guard clause read more easily?

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  def land(plane)
+    raise "The weather is stormy, plane cannot land" if stormy_weather == true
+    raise "No gates available - cannot land plane" if @gates.count >= @gates_size
+    raise "Plane is already landed" if @gates.include?(plane)
+    @gates << plane
+    "The plane has landed"
+  end
+
+  def take_off(plane)
+    raise "The weather is stormy, plane cannot takeoff" if stormy_weather == true
```
maryrosecook: No need to check for `true`.

PR #547, opened 2016-10-01
Airport Challenge

tomnunn:


lib/airport.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'weather_forecast' # Is this necessary?
```
silvabox: As the file does not specifically reference `WeatherForecast`, no, it's not necessary.

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+
+class Airport
+  attr_reader :planes
+
+  def initialize(capacity, weather_forecast)
```
silvabox: Nice dependency injection!

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
```
silvabox: No behaviour for plane?

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  let(:plane) { double :plane }
+  let(:weather_forecast) { double :weather_forecast }
+
+  describe '#land' do
+    context 'when not stormy' do
+      before do
+        allow(weather_forecast).to receive(:stormy?).and_return false
+      end
+
+      it 'should instruct a plane to land' do
```
silvabox: Is that what this test tests for?

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        expect(airport).to respond_to(:land).with(1).argument
+      end
+
+      it 'should confirm a plane has landed' do
```
silvabox: Does it confirm a plane has landed?  This might be the right thing to test (I'm not saying it's right or wrong), but the test doesn't test it.

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        airport.land(plane)
+        expect(airport.planes).to include plane
+      end
+
+      context 'when full' do
+        it ' should prevent a plane from landing' do
+          20.times do
+            airport.land(plane)
```
silvabox: Beware landing the same plane 20 times...

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    context 'when not stormy' do
+      before do
+        allow(weather_forecast).to receive(:stormy?).and_return false
+      end
+
+      it 'should instruct a plane to take off' do
+        expect(airport).to respond_to(:take_off).with(1).argument
+      end
+
+      it 'should confirm that the plane in no longer at the Airport' do
```
silvabox: as above with `#land` make sure your tests actually match their descriptions.

spec/features_spec.rb
```diff
@@ -0,0 +1,53 @@
+describe 'These are all the user stories' do
```
silvabox: Great effort to include feature tests.  But bear in mind, the only feature you're actually testing here is that no error gets raised.  The user requirements are more detailed than that.

PR #548, opened 2016-10-01
Initial pull request.

IainDK:
...


lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'plane'
+
+class Airport
+
```
silvabox: Indentation!

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    
+attr_accessor :capacity
+DEFAULT_CAPACITY = 15
+
+def initialize(capacity = DEFAULT_CAPACITY)
+   @planes = [] 
+   @capacity = capacity
+end
+
+def flight_checker(plane)
```
silvabox: is this the right name for this method?  `flight` does not appear anywhere else in the domain.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @planes.delete(plane)
+end
+
+private
+
+def full?
+    @planes.count >= @capacity
+end
+
+def storm_weather?
```
silvabox: or just `stormy?`

lib/plane.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def initialize
+    @landed = false
+  end
+  
+  def landed?
+    @landed
+  end
+  
+  def land_status
```
silvabox: it's not clear from this method name that it changes the plane's status

spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'airport'
+
+describe Airport do
+    
+    describe '#initialize' do
+    end
+    
+    it 'has default capacity' do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
```
silvabox: Indentation!

spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+        
+    describe '#land()' do
+
+    it 'should know when a plane has landed' do
+    trial_plane = Plane.new
+    subject.land(trial_plane)
+    expect(subject.land(trial_plane).pop).to eq trial_plane
```
silvabox: Is `land` a command or query?  Here you are using it as both.  Either way, you should avoid returning the internal array here for this exact reason: by calling `pop` you've taken the plane out of the array.  If this code was somehow copied into production, that would be a big problem.

PR #549, opened 2016-10-02
Initial Pull Request

andygjenkins:


lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'plane'
+
+class Airport
+
+  DEFAULT_CAPACITY = 10
+
+  attr_reader :planes, :capacity
```
silvabox: don't expose `planes` to the outside world

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
+
+class Plane
+
+attr_accessor :flying
```
silvabox: maybe prefer a `flying?` method instead - it's more Ruby

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  subject(:airport) { described_class.new}
+  let(:plane) {double :plane}
+
+  before do
+    allow(plane).to receive(:landed)
+    allow(plane).to receive(:lift_off)
+  end
+
+  describe 'land' do
+    it 'tells a plane to land' do
```
silvabox: does it tell a plane to land?  That's not what your test actually tests for.

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      allow(airport).to receive(:stormy?).and_return false
+      airport.land(plane)
+      expect(airport.planes).to include plane
+    end
+  end
+
+  describe 'take-off' do
```
silvabox: a `describe` block surrounding one test is perhaps overkill

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    it 'instructs a plane to take-off' do
+      allow(airport).to receive(:stormy?).and_return false
```
silvabox: maybe this and the previous test could be in a `context` block `'when the weather is fine'` or similar.  Then this mock could be in a `before` block to avoid repetition

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      airport.land(plane)
+      airport.take_off(plane)
+      expect(airport.planes).to eq []
+    end
+  end
+
+  describe 'stormy?' do
```
silvabox: This is more like a context.  You are not describing the behaviour of the `stormy?` method here.  ie `context 'when stormy'`

spec/plane_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'plane'
+
+describe Plane do
+
+  it {is_expected.to respond_to :lift_off}
+  it {is_expected.to respond_to :landed}
+
+
```
silvabox: where are you testing the behaviour of `lift_off` and `landed`?

PR #550, opened 2016-10-02
Initial pull request

B-raw:


lib/airport.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  attr_reader :landed_planes, :capacity
```
silvabox: don't expose `landed_planes` outside of this class.

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+    @capacity = capacity
+    @stormy_weather = stormy_weather
+  end
+
+  def stormy_weather
+    (Weather.new).stormy_weather
+  end
+
+  def full
+    full = landed_planes.count > capacity
```
silvabox: `full = ` is a redundant local variable assignment

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+    @landed_planes = []
+    @capacity = capacity
+    @stormy_weather = stormy_weather
+  end
+
+  def stormy_weather
+    (Weather.new).stormy_weather
+  end
+
+  def full
```
silvabox: prefer the ruby idiom for predicate type methods: i.e. `full?`

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
+class Plane
+  attr_reader :landed
+  alias_method :landed?, :landed
+
+  def initialize
+    landed = false
+  end
+
+  def land airport
```
silvabox: don't omit parentheses in method definitions

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
...
+    airport.stormy_weather
```
silvabox: there are four calls to `airport` in this one method.  That's a very strong indication that this logic is in the wrong place.

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
...
+    airport.landed_planes << self
+  end
+
+  def take_off airport
+    airport.stormy_weather
+    fail "Plane cannot take-off when already in the air" if not landed
+    fail "Plane not landed at that airport" unless airport.landed_planes.include?(self)
+    fail "Plane can't take-off when weather is stormy" if airport.stormy_weather
+    @landed = false
+    airport.landed_planes.delete(self)
```
silvabox: As above - most of this belongs in `Airport`

lib/weather.rb
```diff
@@ -0,0 +1,12 @@
+class Weather
+
+  def stormy_weather
+    @stormy_weather = stormy_weather_randomiser
```
silvabox: `@stormy_weather` is redundant here as all calls to `stormy_weather` generate a new value

spec/airport_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    airport = Airport.new(1000)
+    expect(airport.capacity).to eq 1000
+  end
+
+  it "should know if 2 planes have landed" do
+    plane1 = double(:plane)
+    allow(plane1).to receive(:land).with(subject).and_return(subject.landed_planes << plane1)
+    plane2 = double(:plane)
+    allow(plane2).to receive(:land).with(subject).and_return(subject.landed_planes << plane2)
+    plane1.land(subject)
```
silvabox: try deleting these two lines:
```ruby
plane1.land(subject)
plane2.land(subject)
```
And rerunning this test.  Does it still pass?

spec/airport_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    plane2.land(subject)
+    expect(subject.landed_planes).to eq [plane1,plane2]
+  end
+
+  it "knows when it is full" do
+    41.times do
+      subject.landed_planes << Plane.new
+    end
+    expect(subject.full).to be true
```
silvabox: By naming the method `full?`, you'd be able to write this as:
```ruby
expect(subject).to be_full
```
which is quite satisfying

spec/plane_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  let(:airport) {double(:airport, :full=>false, :stormy_weather=>false, :landed_planes=>[])}
+  let(:full_airport) {double(:airport, :full=>true, :stormy_weather=>false, :landed_planes=>[])}
+  let(:stormy_airport) {double(:airport, :full=>false, :stormy_weather=>true, :landed_planes=>[])}
+
+  describe "#land" do
+
+    it "should know it is landed at an airport" do
+      subject.land(airport)
+      expect(subject.landed?).to eq true
```
silvabox: `expect(subject).to be_landed`

spec/plane_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+    it "should take-off then land and know it is landed" do
+      subject.land(airport)
+      subject.take_off(airport)
+      subject.land(airport)
+      expect(subject.landed).to eq true
+    end
+
+    it "should be in the airport array when landed" do
+      subject.land(airport)
+      expect(airport.landed_planes).to eq [subject]
```
silvabox: this is behaviour of the airport, not the plane.

lib/weather.rb
```diff
@@ -0,0 +1,12 @@
+class Weather
+
+  def stormy_weather
```
silvabox: `def stormy?` makes for better reading.

spec/feature_test_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'capybara/rspec'
+
+feature 'Functioning Airport' do
+
+  scenario 'I can land & take-off a plane' do
+    when_i_create_a_plane_and
```
silvabox: I haven't really seen this approach before, and I'm not very keen.  It's almost like Cucumber, but without all the benefits of Cucumber.

I would stick to the same RSpec syntax as we've used in unit tests.  Feature tests are different from unit tests in that they don't use doubles and you can comfortably have multiple expectations in one test.

`feature` is just an alias for `describe` and `scenario` is an alias for `it`, so all the normal RSpec stuff is exactly the same.

PR #555, opened 2016-10-02
Airport Challenge

lcbeh:


spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      subject.capacity.times {subject.land(plane)}
+      expect{subject.land(plane)}.to raise_error error
+    end
+
+    it "allows default capacity to be override" do
+      airport = Airport.new(40)
+      40.times {airport.land(plane)}
+      expect {airport.land(plane)}.to raise_error error
+    end
+
```
maryrosecook: Watch out for whitespace style.  Have a look at a Ruby style guide for more info.

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    it "doesn't accept new plane when full" do
+      subject.capacity.times {subject.land(plane)}
+      expect{subject.land(plane)}.to raise_error error
+    end
+
+    it "allows default capacity to be override" do
+      airport = Airport.new(40)
+      40.times {airport.land(plane)}
+      expect {airport.land(plane)}.to raise_error error
```
maryrosecook: It's nice that `error` avoids repetition.  However, could the variable have a more descriptive name?

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
+
+  WEATHER = [:good, :stormy, :good, :good, :stormy, :good, :good]
+
+  def check_weather
```
maryrosecook: This could probable be private to increase encapsulation.

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
+
+  WEATHER = [:good, :stormy, :good, :good, :stormy, :good, :good]
```
maryrosecook: It's nice to see the possible values in this array.  However, it feels hard to know how often it's going to be good weather and how often it's going to be stormy just by looking at the array.  Could you use rand() instead to make the ratio of good to stormy clearer?

PR #556, opened 2016-10-02
Mike's airport_challenge submission

calveym:
Just doing another check for hound and travis. Latest addition was coveralls test coverage integration and travis integration.


lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+  attr_accessor :weather
+
+  def initialize(weather = 10, capacity = 20)
+    @weather = weather
+    @plane_hangar = []
+    @capacity = capacity
+  end
+
+  def randomise_weather
+    @weather = Random.new.rand(10)
```
maryrosecook: It feels a little hard to understand how weather works because it's not encapsulated.  eg It's fine for the value of this number to drive whether the weather is good or bad, but it would be easier to read if it was accessed via a readable method like stormy?.

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  def takeoff(plane)
+    if weather?
+      raise 'Already flying!' if plane.flying? == true
+      index = @plane_hangar.index(plane)
+      temp_plane = @plane_hangar[index]
+      @plane_hangar.delete(plane)
+      temp_plane.flying = true
```
maryrosecook: This feels like it's breaking encapsulation a little.  Can Plane have a method that sets whether a plane is flying, rather than a variable being set from the outside.  (Ask, don't tell.)

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+      return temp_plane
+    else
+      raise('Cannot takeoff in bad weather')
```
maryrosecook: Can this be done with a guard clause?

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  def full?
+    if (@plane_hangar.size >= @capacity) then true else false end
```
maryrosecook: I think the `then ...` part isn't necessary.

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    else
+      raise('Cannot land in bad weather')
+    end
+  end
+
+  def takeoff(plane)
+    if weather?
+      raise 'Already flying!' if plane.flying? == true
+      index = @plane_hangar.index(plane)
+      temp_plane = @plane_hangar[index]
```
maryrosecook: Is temp_plane necessary?

lib/plane.rb
```diff
@@ -0,0 +1,11 @@
+class Plane
+  attr_accessor :flying
+
+  def initialize(flying = true)
+    @flying = flying
+  end
+
+  def flying?
+    return @flying
```
maryrosecook: No need to use the return keyword here.

PR #557, opened 2016-10-02
Airport Challenge

Katy600:


lib/plane.rb
```diff
@@ -0,0 +1,11 @@
+
+class Plane
+
+	def initialize (plane = true)
```
maryrosecook: I'm not sure I'm clear on what `plane` represents...

PR #558, opened 2016-10-02
initial correct pull request

jinimcoroneo:


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+class Airport
+
+def initialize(capacity)
```
maryrosecook: Watch out for indentation.  Have a look at a Ruby style guide.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+   @planes_at_airport = []
+end
+
+def land(plane)
+  raise "can't land plane, airport is full" if full?
+  raise "can't land plane due to stormy weather" if stormy?
+  @planes_at_airport << plane
+end
+
+def confirm_land(plane)
```
maryrosecook: Often best to avoid defining a method until you're ready to write the code to make it pass.  That way, the test for it will definitely fail.

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,73 @@
+require "airport.rb"
+require "plane.rb"
+
+describe "user_stories" do
```
maryrosecook: Great to have tests for the user stories if they aren't covered by the unit tests.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+def confirm_take_off(plane)
+end
+
+private
+
+def full?
+  @planes_at_airport.length >= @capacity
+end
+
+def stormy?
```
maryrosecook: Probably worth splitting weather code out into its own class to follow the SRP.

PR #559, opened 2016-10-02
Weekend challenge

MalinnaLeach:


lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @weather = "sunny"
+    @planes = []
+  end
+
+  def accept_plane(plane)
+    change_weather
```
maryrosecook: It feels like this method has two responsibilities: change the weather and land a plane.  Can the weather changing be moved out?

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @weather = "sunny"
+    @planes = []
+  end
+
+  def accept_plane(plane)
```
maryrosecook: "accept" feels like it's not in the language of the spec.  Maybe "land"?

lib/croissant.rb
```diff
@@ -0,0 +1,2 @@
+class Croissant
```
maryrosecook: Hahahahahaha.

lib/plane.rb
```diff
@@ -0,0 +1,25 @@
...
+  def initialize
+    @status = "flight in progress"
+    @airport = nil
+  end
+
+  def land(airport)
+    raise "A #{airport.class} is not an airport" unless airport.class == Airport
+    raise "This plane is not in flight" if @status == "landed"
+    airport.accept_plane(self)
+    @status = "landed"
```
maryrosecook: Clarity could be increased by changing `status` to `landed`.

lib/plane.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  attr_reader :status, :airport
+
+  def initialize
+    @status = "flight in progress"
+    @airport = nil
+  end
+
+  def land(airport)
+    raise "A #{airport.class} is not an airport" unless airport.class == Airport
```
maryrosecook: Just a reminder of the duck-typing stuff we talked about in the review.  Thanks again for showing your code!

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
+class Airport
```
maryrosecook: And probably worth separating out weather into a separate class for SRP.

PR #561, opened 2016-10-02
Airport Challenge

LaszloBogacsi:


lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+    @stormy_weather = false #it's sunny
+  end
+
+  def actual_weather(weather)
+    @stormy_weather = weather.storm #weather might change to stormy
+  end
+
+  def take_off
+    fail('No take off due to storm') if @stormy_weather
+      @planes.pop
```
maryrosecook: Watch out for this indentation.  Have a look at a Ruby style guide.

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def land(plane)
+    fail('Airport is full') if full?
+    fail('No landing due to storm') if @stormy_weather
+    @flight_log << plane.plane_id
```
maryrosecook: The flight log is nice.  However, it feels like it's outside the scope of the spec.

lib/plane.rb
```diff
@@ -0,0 +1,15 @@
+# contains plane class
+
+class Plane
+
+  @plane_inventory = 0
+  class << self
```
maryrosecook: This metaprogramming feels unnecessary for this challenge.  Do planes need a unique id?

lib/weather.rb
```diff
@@ -0,0 +1,16 @@
+# randomly sets the weather condition between sunny and stormy
+
+class Weather
+  attr_accessor :stormy
+
+  def storm
+    rand1 = Random.new
```
maryrosecook: Could this be broken down into several methods to make it clearer?

PR #564, opened 2016-10-02
Code for airport challenge

eunkenlow:


lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+  def initialize(capacity)
+    @plane = []
+    @weather = "sunny"
+    @capacity = 5
+  end
+
+  def receive_plane(plane)
+    raise("Airport is full, unable to land") if full?
+    raise("Bad weather, unable to land") if @weather == "stormy"
+      plane.instruct_land
```
maryrosecook: Watch out for code formatting.  Have a look at a Ruby style guide.

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+    if roll == 2
+      @weather = "stormy"
+    else
+      @weather = "sunny"
+    end
+  end
+
+  private
+
+  def roll
```
maryrosecook: Can this method have a description that describes what it does more clearly?

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
+require_relative 'plane'
+
+class Airport
```
maryrosecook: It feels like this class has two responsibilities - landing planes and looking at the weather.  It's worth moving the weather to its own class.

PR #565, opened 2016-10-02
First pull request

stuartcooper:


lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @weather = Weather.new
+    @capacity = capacity
+  end
+
+  def land(plane)
+    check_for_landing(plane)
+    plane.landed
```
maryrosecook: Maybe make the function name a verb rather than an adjective.  It'll read a little more clearly.

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    def stormy_weather
+      @weather.stormy
+    end
+
+    def plane_in_airport(plane)
+      @planes.include? plane
+    end
+
+    def check_for_landing(plane)
```
maryrosecook: Love this abstraction to a separate method!

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+class Weather
+
+attr_reader :stormy
+
+  def stormy
```
maryrosecook: It's helpful for readability to add a question mark to the name of methods that return a boolean eg stormy?

PR #566, opened 2016-10-02
Airport challenge

antonyCastineiras:


lib/airport.rb
```diff
@@ -0,0 +1,18 @@
...
+
+	def initialize(capacity=DEFAULT_CAPACITY,weather = "clear")
+		@landed_planes = []
+		@capacity = capacity
+		@weather = weather
+		@chance_of_storm = rand(10)
+	end
+
+	def stormy
+		if @chance_of_storm == 10 then @weather = "stormy" end
```
silvabox: this is not idiomatic ruby syntax - avoid.  Instead:
```ruby
@weather = "stormy" if @chance_of_storm == 10
```

lib/plane.rb
```diff
@@ -0,0 +1,27 @@
...
+	attr_accessor :current_airport
+
+	def initialize
+		@current_airport
+	end
+
+	def land(airport)
+		if airport.weather == "stormy"
+			"Stormy weather. Unable to land."
+		elsif airport.landed_planes.count >= airport.capacity
```
silvabox: most of this belongs inside `Airport` not plane.

lib/plane.rb
```diff
@@ -0,0 +1,27 @@
...
+			@current_airport = airport
+			airport.landed_planes << self
+		end	
+	end
+
+	def take_off
+		if @current_airport.weather == "stormy"
+			"Stormy weather. Unable to take off."
+		else
+			@current_airport.landed_planes.delete(self)
```
silvabox: This is poorly encapsulated.  Plane should not be able to remove itself from the airport's internal `landed_planes` array.

spec/airport_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+			expect(subject.chance_of_storm).to be_kind_of(Fixnum)
+		end
+
+	end
+
+	describe '#stormy' do 
+
+		it 'randomly changes the weather to stormy' do
+			subject.chance_of_storm = 10	
+			subject.stormy
```
silvabox: This method doesn't make sense.  Why should I need to call `stormy` before checking if it's stormy?  This indicates a problem with your design.

lib/plane.rb
silvabox: Avoid using string return values to indicate success/failure as they can be easily changed.  Better to raise an exception in this case as we're only interested in a message if something goes wrong.

spec/plane_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+		it 'prevents landing when airport capacity is full' do
+			20.times { subject.land(airport) }
+			expect(subject.land(airport)).to eq "Unable to land. Airport full."
+		end
+
+	end
+
+	describe '#take_off' do
+
+		it 'updates the current_airport variable to nil' do
```
silvabox: You are not testing a variable - that's an implementation detail; don't mention it in the test description.

PR #567, opened 2016-10-02
Airport challenge

fbell123:
Not complete, hit a stumbling block for most of the day

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,78 +13,23 @@ Airport Challenge
 
 ```
 
-Instructions
 ---------
 
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+Set each of the user stories to a test and then built the code one by one to match
+work them through.
 
-Steps
--------
-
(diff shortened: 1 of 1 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+    @planes << plane
+    "Plane has landed"
+  end
+
+  def take_off?(plane)
+    fail "Cannot fly due to weather" if weather.stormy?
+    if planes.include?(plane)
+      plane.flying
+      @planes.shift
+      "Plane has taken off"
```
maryrosecook: Try not to use strings to report on what happened in a function.  Perhaps return the plane to signal success? Or return nothing and assume everything went fine if there's no error?

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+    else
+      "Plane is already in the air"
```
maryrosecook: Perhaps use a guard clause for this?

lib/plane.rb
```diff
@@ -0,0 +1,17 @@
+class Plane
+
```
maryrosecook: Have a look at a Ruby style guide for how to use whitespace in classes.

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'airport'
```
maryrosecook: Try and develop test first.  Ideally, there's no code that doesn't have tests.  It's helpful for writing well-designed code as well as for writing correct code.

PR #568, opened 2016-10-02
Complete the Airplane Challenge - RED-GREEN-REFRACTOR

ManuCiao:


lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+
+attr_reader :capacity, :weather, :planes
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes =[]
+    @capacity = capacity
+    @weather = Weather.new
+  end
+
+#landing
```
maryrosecook: No need for these comments.  If the class is big enough to need them, it's too big :)

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def storm?
+    weather.stormy?
+  end
+
+  def full?
+    planes.count >= capacity
+  end
+
+  def include?(plane)
```
maryrosecook: Perhaps make this name more specific to the domain eg at_airport?

lib/plane.rb
```diff
@@ -0,0 +1,17 @@
+class Plane
+
+#plane status
+
+  def landed
+    @working = false
```
maryrosecook: Could working be called something more specific like `landed`?

lib/plane.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def taken_off
```
maryrosecook: Perhaps use a verb for this, rather than an adjective eg take_off

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
+
+#random weather forecast 0 = stormy 10 = Sunny
+  def stormy?
+    roll == 0
+  end
+
+  private
+
+  def roll
```
maryrosecook: Could this name be more descriptive of what the function does?

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
+
+#random weather forecast 0 = stormy 10 = Sunny
```
maryrosecook: Often a comment is a sign that code can be made clearer :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      describe "#takeoff" do
+
+        it "returns no planes are allowed to take off due to storm weather" do
+          subject.land(plane)
+          subject.stub(:storm?).and_return(0)
+          expect{ subject.takeoff(plane) }.to raise_error(RuntimeError, "Flight cancelled due to storm weather!")
+        end
+
```
maryrosecook: Watch out for your whitespace.  Have a look at a Ruby style guide.

PR #569, opened 2016-10-02
Pull Request WE Challenge 1

cgulli:
Completed first two user stories for WE Challenge 1 :/

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,23 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  airport = Airport.new
+  it 'lands a plane' do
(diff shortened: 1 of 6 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
```
maryrosecook: Nice clean tests.

spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  airport = Airport.new
```
maryrosecook: Watch out for setting variables that are used by several tests.  Can lead to them interfering with each other.

PR #570, opened 2016-10-02
First Pull Request

nspencer8:


spec/weather_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'weather'
+
+
+describe Weather do
+
+  it 'returns a new storm' do
+    allow(Weather).to receive(:stormy?).and_return(true)
+   end
```
maryrosecook: Try and use consistent formatting that follows a Ruby style guide!

lib/airport.rb
```diff
@@ -0,0 +1,13 @@
+
+class Airport
+
+attr_reader :plane
+
+  def land(plane)
+    @plane = plane
```
maryrosecook: Great that you started with a simpler (single plane) case first.

PR #571, opened 2016-10-02
Weekend Challenge

Swinston88:


lib/airport.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'plane'
+
+class Airport
+
+  # add DEFAULT_CAPACITY constant
+
+  attr_reader :planes, :weather
```
silvabox: Both of these are unused

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize
+    @planes = []
+  end
+
+  def clear_landing(plane)
+    if (self.weather) == 'sunny'
```
silvabox: refactor this expression to a separate method, so you can have `if sunny?`

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+      @planes << plane
+    else
+      raise "#{plane} unable to land due to poor weather conditions"
```
silvabox: try to place guard clauses at the beginning of the method for improved readability, i.e.:
```ruby
raise "#{plane} unable to land due to poor weather conditions" unless sunny?

@planes << plane
```

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'plane'
```
silvabox: As you don't explicitly reference `Plane` in this file, you don't actually need to require `plane.rb`

lib/plane.rb
```diff
@@ -0,0 +1,20 @@
+require_relative 'airport'
```
silvabox: As you don't explicitly reference `Airport` in this file, you don't actually need to require `airport.rb`

lib/plane.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class Plane
+
+  attr_reader :status
+
+  def initialize
+    @status = 'in flight'
```
silvabox: avoid using strings for status - use symbols instead.

spec/airport_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+describe Airport do
+
+    let(:plane) { Plane.new } # place in required areas to allow status differences
+    let(:airport) {Airport.new}
+
+  it { should respond_to :clear_landing, :clear_takeoff, :weather }
+
+  it 'clears for landing' do # add method stub for weather
+    plane.land(subject)
+    expect(subject.planes[0]).to eq plane
```
silvabox: don't break encapsulation by exposing the planes array outside airport.

spec/airport_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    plane.takeoff(subject)
+    expect(subject.planes).to be_empty
+  end
+
+  it 'sets the weather' do # fix - only accepts 'sunny' at present
+    expect(subject.weather).to eq ('sunny'||'stormy')
+  end
+
+  it 'denies permission to land' do # fix double/method stub
+    gatwick = double(:airport, weather: 'stormy')
```
silvabox: You're supposed to be testing the behaviour of `Airport`; you can't do that if you double the object under test!

PR #572, opened 2016-10-02
pull_request

abeddow91:


lib/airport.rb
```diff
@@ -0,0 +1,55 @@
+class Airport
+  require './lib/weather'
+  DEFAULT_CAPACITY = 10
+attr_reader :status, :take_off, :land, :departure_arr, :arrivals, :capacity, :conditions
+attr_writer :capacity
+
+def initialize (capacity=DEFAULT_CAPACITY, weather =Weather.new)
```
maryrosecook: Watch out for consistent code formatting.  Have a look at a Ruby style guide.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+  @departure_arr =[]
+  @arrivals=[]
```
maryrosecook: It feels like having a separate departure_arr and arrivals arrays is outside the scope of the spec.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
+class Airport
+  require './lib/weather'
+  DEFAULT_CAPACITY = 10
+attr_reader :status, :take_off, :land, :departure_arr, :arrivals, :capacity, :conditions
```
maryrosecook: Try and tidy up attr_readers that you're no longer using.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+def initialize (capacity=DEFAULT_CAPACITY, weather =Weather.new)
+  @departure_arr =[]
+  @arrivals=[]
+  # @conditions = rand(10)
+  @weather = weather
+  @capacity = capacity
+end
+
+def land(plane)
+  if @weather.stormy?
```
maryrosecook: Try rewriting these as guard clauses - it'll make things easier to read.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+    print  "Sorry, the airport is full"
+  else
+    @landed_plane = plane.landed
+    @arrivals << @landed_plane
+    confirm_landed
+  end
+end
+
+def confirm_landed
+  print  @arrivals.length >= 1 ? "Plane has landed" : "Landing failed!"
```
maryrosecook: This logging is probably outside the scope of the spec.  Also it feels like it will always say "Plane has landed" because of where it's called from.

lib/plane.rb
```diff
@@ -0,0 +1,17 @@
+class Plane
+attr_reader :status, :airborne
+
+def initialize(status='no status')
+  @status = status
+end
+
+def airborne
```
maryrosecook: airborne feels more like an adjective than a verb.  Perhaps take_off ?

PR #573, opened 2016-10-02
Airport_challenge pull request

powderham:


lib/airport.rb
```diff
@@ -0,0 +1,65 @@
...
+    add_plane(plane)
+    update_log("Plane #{plane} landed")
+    @flight_log.last
+  end
+
+  def take_off(plane)
+    raise 'Cannot take off due to bad weather' if stormy?
+    raise 'Plane is not in the airport' unless plane_exists?(plane)
+    remove_plane(plane)
+    update_log("Plane #{plane} took off")
```
maryrosecook: The flight log is nicely implemented! It feels like overkill.  It's also tricky that the tests rely on it.  Perhaps pass back the plane to indicate a successful landing, or just assume everything went ok if there are no exceptions.

lib/airport.rb
```diff
@@ -0,0 +1,65 @@
+require_relative 'plane'
+
+class Airport
+  attr_accessor :planes, :flight_log, :weather, :capacity
+
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @stormy = set_weather
```
maryrosecook: It feels like `set_weather` is not quite the right name, because it doesn't actually set the weather.

lib/airport.rb
```diff
@@ -0,0 +1,65 @@
...
+      @planes.length == @capacity
+    end
+
+    def update_log(message)
+      @flight_log << message
+    end
+
+    def remove_plane(plane)
+      runway = plane
+      @planes.delete_if {|plane| plane === runway}
```
maryrosecook: Have a look for the built in method for deleting specific objects from an array.

lib/airport.rb
```diff
@@ -0,0 +1,65 @@
...
+    def at_capacity?
+      @planes.length == @capacity
+    end
+
+    def update_log(message)
+      @flight_log << message
+    end
+
+    def remove_plane(plane)
+      runway = plane
```
maryrosecook: I see that you need a different name here for the plane.  runway doesn't feel like it describes what it is.

spec/airport_spec.rb
```diff
@@ -0,0 +1,84 @@
+require 'airport'
```
maryrosecook: Nice, clear, well organised tests!

PR #574, opened 2016-10-02
airport_challenge

Elizabeth555:


docs/feature_test.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def takes_off
+    5.times{airport.take_off(plane)}
+  end
+
+  def airport_is_empty
+    expect{airport.take_off(plane)}.to raise_error("Airport is empty")
+  end
+end
```
maryrosecook: Like the feature test!

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  MAX = 5
```
maryrosecook: Perhaps give this constant a more specific name?

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    @weather = Weather.new
+  end
+
+  def land(plane)
+    raise "Can't land plane :( " if weather.stormy?
+    raise "Airport is full" if full?
+    raise "Plane already landed" if present?(plane)
+    plane.landing
+    @planes << plane
+    return @planes
```
maryrosecook: No need for return keyword.

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    puts weather.stormy?
+  end
+
+private
+
+  def full?
+    @planes.length >= capacity
+  end
+
+  def present?(plane)
```
maryrosecook: Nice abstraction to a method!

PR #577, opened 2016-10-02
Airport Challenge

RoystonHenson:


lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    weather_good_today?
+    raise "That plane has already landed!" unless !@planes.include?(plane)
+    raise "You cannot land in a storm" unless weather_today_good?
```
maryrosecook: I'm not sure this method exists?

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+    raise "This airport is full!" unless !full?
+    confirm_landing
+    @planes << plane
+  end
+
+  def take_off(plane)
+    weather_good_today?
+    raise "That plane isn't in this airport!" unless @planes.include?(plane)
+    raise "You cannot take off in a storm!" unless weather_today_good?
+    confirm_take_off
```
maryrosecook: This logging, though helpful for debugging, should probably be removed in the final version.

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def confirm_take_off
+    puts "The plane has taken-off!"
+  end
+
+  def full?
+    return true if @planes.length >= @capacity
+  end
+
+  def weather_good_today?
```
maryrosecook: Perhaps move weather to its own class?

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+    @weather = (rand(11))
+    if @weather < 6
```
maryrosecook: Can this if be condensed to a single line?

lib/airport.rb
```diff
@@ -43,10 +42,6 @@ def full?
 
   def weather_good_today?
     @weather = (rand(11))
-    if @weather < 6
-      true
-    else
-      false
-    end
+    true unless @weather > 8
```
maryrosecook: Perhaps even simply `@weather <= 8`

RoystonHenson:
Thanks Mary :)


Last weekend I had this method setup to call random weather every time I tried to land/take-off a plane(thinking that weather can change throughout the day). But I forgot to commit it to github before experimenting, broke it and ran out of time to fix it and get the tests passing before having to submit the pull request.

I added it back as a method last night and was wondering if that method would be fine or is it still better to make weather a class?


best

Royston



________________________________

maryrosecook:
Having weather as a class reduces the number of responsibilities of the Airport, which is a good thing.  But having the weather in the Airport isn't the end of the world!

RoystonHenson:
Ok. Thanks again Mary!


best

Royston


________________________________

PR #578, opened 2016-10-02
Pull request for first weekend challenge.

fmlharrison:


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+    @weather = weather
+  end
+
+  def land(plane)
+    fail "This plane has already landed" unless plane.airborne?
+    fail "Plane unable to land, airport full" if full?
+    fail "Weather is too bad to land" if weather.stormy?
+    plane.landed
+    @planes << plane
+    "The plane landed safely"
```
maryrosecook: This feels like it's just there for the testing.  The test could either assume that everything's ok if no errors are raised.  Or it could return something that is more programatically meaningful (the landed plane?)

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def take_off(plane)
+    fail "This plane is aleady in the air" if plane.airborne?
+    fail "Can't take off due to stormy conditions" if weather.stormy?
+    if @planes.include?(plane)
+      plane.flying
+      @planes.delete(plane)
+      "The plane has taken off"
+    else
+      fail "This plane is not in the airport"
```
maryrosecook: Possibly move this to a guard clause.

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
+class Plane
+
+  def initialize(flying: true)
```
maryrosecook: Is there a reason you used a : here rather than =?

spec/airport_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+  end
+
+  describe '#land' do
+
+    context 'land the planes on a clear day' do
+
+      it 'tells a plane to land' do
+        expect(plane).to receive(:landed)
+        sunny_airport.land(plane)
```
maryrosecook: Very readable variable name - great!

spec/airport_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+       end
+
+       it 'rasie error when trying to land in stormy weather' do
+         expect{stormy_airport.land(plane)}.to raise_error "Weather is too bad to land"
+       end
+     end
+  end
+
+  describe '#take_off' do
+
```
maryrosecook: Watch out for code formatting.  Have a look at a Ruby style guide.

PR #580, opened 2016-10-02
Tom's Weekend1 Airport Challenge.

twp88:


lib/atc.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def take_off(plane)
+    @plane = plane
+    if @weather < 2
+      @land_arr = @land_arr.pop
+      @plane.in_airport = false
+      @plane.taken_off = true
+    else
+      print "Can't Take off! Too stormy!"
+      @plane.in_airport = true
```
maryrosecook: In OO, it's often better from objects to ask, rather than tell.  Can the airport tell the plane it has landed?

PR #581, opened 2016-10-02
Created first feature tests and methods for Land and Takeoff

stejgregory:
Here is the first pull request for Airport Challenge

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,10 +1,20 @@
 class Airport
 
   DEFAULT_CAPACITY = 25
-  attr_reader :capacity
+  attr_reader :capacity, :grounded_planes
+
+  def initialize (capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @grounded_planes = []
+  end
 
   def land(plane)
-    plane
+      fail "Airport is full. Unable to land." if full?
+      @grounded_planes << plane
+  end
+
+  def full?
+    @grounded_planes.count == @capacity
   end
 
   def takeoff(plane)
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,14 @@
+class Airport
+
+  DEFAULT_CAPACITY = 25
+  attr_reader :capacity
+
+  def land(plane)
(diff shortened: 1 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+      "Bing bong: Landed safe and sound!"
+  end
+
+  def full?
+    @grounded_planes.count == @capacity
+  end
+
+  def take_off(plane)
+    if empty?
+      "There are no grounded planes at the airport."
```
silvabox: Avoid using strings to indicate success/failure (i..e program control).  use `fail` at this stage to ensure failures can't be overlooked.

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    else
+    @grounded_planes.pop
```
silvabox: This is logically incorrect.  What does `@grounded_panes.pop` do?

spec/airport_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+describe Airport do
+
+  let(:plane) { double :plane }
+
+  it "accepts planes landing" do
+    expect(subject.land(plane)).to eq "Bing bong: Landed safe and sound!"
+  end
+
+  it "won't allow landing plane if airport full" do
+    subject.capacity.times{ subject.land(plane) }
```
silvabox: beware landing the same plane n times

spec/airport_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    expect{ subject.land(plane) }.to raise_error "Airport is full. Unable to land."
+  end
+
+  it "accepts planes taking off" do
+    subject.land(plane)
+    expect(subject.take_off(plane)).to eq "This is your captain speaking, we are off into the sky!"
+  end
+
+  it "doesn't allow planes to take off when airport is empty" do
+    expect(subject.take_off(plane)).to eq "There are no grounded planes at the airport."
```
silvabox: What if there was another plane at the airport other than `plane`?

PR #583, opened 2016-10-03
Airport Challenge submission

amaalali:
Submitted by Amaal Ali

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -10,6 +10,10 @@ class Airport
     @planes_at_airport << plane_name
   end
 
+  def takeoff(plane_name)
+    @planes_at_airport.delete(plane_name)
+  end
+
   def landed?(plane_name)
     @planes_at_airport.include?(plane_name)
   end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,21 @@
+require 'plane_file'
+
+class Airport
+
+  def initialize
+    @planes = Array.new
+  end
+
+  def land(plane_name)
+    @planes << plane_name
+    #plane_name.proceed_to_land
+  end
+
+  def which_planes
+    @planes
+  end
+
+  def landed?(plane_name)
+    @planes.include?(plane_name)
+  end
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -3,19 +3,14 @@ require 'plane_file'
 class Airport
 
   def initialize
-    @planes = Array.new
+    @planes_at_airport = Array.new
   end
 
-  def land(plane_name)
-    @planes << plane_name
-    #plane_name.proceed_to_land
-  end
-
-  def which_planes
-    @planes
+  def accept_plane(plane_name)
+    @planes_at_airport << plane_name
   end
 
   def landed?(plane_name)
-    @planes.include?(plane_name)
+    @planes_at_airport.include?(plane_name)
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,9 +1,10 @@
 require 'plane_file'
 
 class Airport
-
-  def initialize
+  DEFAULT_CAPACITY = 100
+  def initialize(capcity = DEFAULT_CAPACITY)
     @planes_at_airport = Array.new
+    @capacity = DEFAULT_CAPACITY
   end
 
   def accept_plane(plane_name)
@@ -16,5 +17,9 @@ class Airport
 
   def landed?(plane_name)
     @planes_at_airport.include?(plane_name)
+  end
+
+  def full?
+    @planes_at_airport.length >= @capacity
   end
 end
--- a/lib/plane_file.rb
+++ b/lib/plane_file.rb
@@ -2,17 +2,20 @@ require 'airport'
 
 class Plane
 
+  attr_reader :plane_is_landed
+
   def initialize(plane_is_landed = false)
     @plane_is_landed = plane_is_landed
   end
 
-  def plane_is_landed?
-    @plane_is_landed
-  end
-
-
   def proceed_to_land(airport_name)
     airport_name.accept_plane(self)
+    @plane_is_landed = airport_name.landed?(self)
+  end
+
+  def takeoff_from(airport_name)
+    fail 'Plane is not at this airport' unless airport_name.landed?(self)
+    airport_name.takeoff(self)
     @plane_is_landed = airport_name.landed?(self)
   end
 end
--- a/lib/plane_file.rb
+++ b/lib/plane_file.rb
@@ -4,6 +4,7 @@ class Plane
 
   attr_reader :plane_is_landed
 
+
   def initialize(plane_is_landed = false)
     @plane_is_landed = plane_is_landed
   end
--- a/lib/plane_file.rb
+++ b/lib/plane_file.rb
@@ -1,14 +1,14 @@
 class Plane
 
-  def initialize(plane_location = false)
-    @plane_location = plane_location
+  def initialize(plane_is_landed = false)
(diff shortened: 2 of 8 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+require 'plane_file'
+
+class Airport
+  DEFAULT_CAPACITY = 100
+  def initialize(capcity = DEFAULT_CAPACITY)
+    @planes_at_airport = Array.new
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def accept_plane(plane_name)
```
silvabox: `plane_name` implies I should pass the name of a plane, e.g. "Air Force One".  This argument is a plane object, so should be called `plane`

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+require 'plane_file'
+
+class Airport
+  DEFAULT_CAPACITY = 100
+  def initialize(capcity = DEFAULT_CAPACITY)
+    @planes_at_airport = Array.new
```
silvabox: `_at_airport` is probably redundant.  Just `@planes` would be easier to read (and type!)

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+require 'plane_file'
+
+class Airport
+  DEFAULT_CAPACITY = 100
+  def initialize(capcity = DEFAULT_CAPACITY)
```
silvabox: your `capcity`[sic] argument is unused.

lib/plane_file.rb
```diff
@@ -0,0 +1,28 @@
+require 'airport'
+require 'weather'
+
+class Plane
+
+  attr_reader :plane_is_landed
```
silvabox: `plane_is_` is implicit - just `:landed` would be better.

lib/plane_file.rb
```diff
@@ -0,0 +1,28 @@
...
+
+
+  def initialize(plane_is_landed = false)
+    @plane_is_landed = plane_is_landed
+  end
+
+  def proceed_to_land(airport_name, stormy = false)#weather = Weather.new)
```
silvabox: `airport_name` should be `airport`: you are passing the object; it doesn't have a name.

lib/plane_file.rb
```diff
@@ -0,0 +1,28 @@
...
+    fail 'Landing is not possible. All landing has been suspended due to stormy weather' if stormy
```
silvabox: Most of this logic belongs in the airport I feel.  `proceed_to_land` implies all of the checks have already been carried out.

lib/plane_file.rb
```diff
@@ -0,0 +1,28 @@
...
+    fail 'Landing is not possible. Airport full' if airport_name.full?
+    airport_name.accept_plane(self)
+    @plane_is_landed = airport_name.landed?(self)
+    puts "Plane has landed sucessfully" if @plane_is_landed
+  end
+
+  def takeoff_from(airport_name, stormy = false)#weather = Weather.new)
+    fail 'Plane is not at this airport' unless airport_name.landed?(self)
```
silvabox: again, most of this belongs in `Airport`, not `Plane`.

spec/airport_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'airport'
+require 'plane_file'
+
+describe Airport do
+
+  describe '#landed?()' do
+    it 'checks if a plane has landed' do
+      test_plane = Plane.new
+      test_airport = Airport.new
```
silvabox: why not use `subject`?

spec/airport_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      test_airport.accept_plane(test_plane)
+      expect( test_airport.landed?(test_plane) ).to eq true
+    end
+  end
+
+  describe '#accept_plane(plane_name)' do
+    it 'allows a plane to land and passes instruction to airport to add it to list of planes' do
+      test_plane = Plane.new
```
silvabox: use `let(:test_plane) { Plane.new }` to avoid repetition

spec/airport_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  describe '#landed?()' do
+    it 'checks if a plane has landed' do
+      test_plane = Plane.new
+      test_airport = Airport.new
+      test_airport.accept_plane(test_plane)
+      expect( test_airport.landed?(test_plane) ).to eq true
+    end
+  end
+
+  describe '#accept_plane(plane_name)' do
```
silvabox: Isn't this test identical to the previous one?

spec/airport_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    it 'allows a plane to land and passes instruction to airport to add it to list of planes' do
```
silvabox: Test the interface, not the implementation.  Nothing in your test refers to passing instructions to the airport to add it to a list of planes.  Nor should it - that's an internal detail private to `Airport`.  So describe your test accordingly.

spec/plane_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+describe Plane do
+
+  describe '#proceed_to_land()' do
+    it {is_expected.to respond_to (:proceed_to_land)}
+
+    it 'allows plane to sucesfully land' do
+      test_airport = Airport.new
+      test_plane = Plane.new
+      test_plane.proceed_to_land(test_airport)
+      expect( test_airport.landed?(test_plane) ).to eq true
```
silvabox: You are supposed to be testing `test_plane` but your expectation is on `test_airport`.

spec/plane_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    it 'except when airport is full' do
+      test_airport = Airport.new
+      Airport::DEFAULT_CAPACITY.times do
+        subject.proceed_to_land(test_airport)
```
silvabox: note you are landing the same plane multiple times at the same airport.

spec/plane_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      test_airport = Airport.new
+      test_plane.proceed_to_land(test_airport)
+      test_plane.takeoff_from(test_airport, false)
+      expect( test_plane.plane_is_landed ).to eq false
+    end
+  end
+
+  describe ' behaviour in stormy weather:' do
+    it 'takeoff' do
+      weather = double("Weather", :stormy? => true)
```
silvabox: this double is redundant - you never pass it anywhere

spec/plane_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      test_plane = Plane.new
+      test_airport = Airport.new
+      test_plane.proceed_to_land(test_airport, false)
+      expect { test_plane.takeoff_from(test_airport, weather.stormy?) }.to raise_error 'Take-off is not possible. All planes have been grounded due to stormy weather'
+    end
+
+    it 'takeoff' do
+      weather = double("Weather", :stormy? => true)
```
silvabox: again - redundant double

spec/weather_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+module Boolean; end
+class TrueClass; include Boolean; end
+class FalseClass; include Boolean; end
+
+
+describe Weather do
+  describe '#stormy?' do
+    it {is_expected.to respond_to (:stormy?)}
+
+    it 'output should be true/false' do
```
silvabox: interesting approach to testing the weather.  To improve it, remove all duplicates from the `weather_output` array and check that it's roughly equal to `[true, false]`.

Just think about how your test reads:
`it output should be true/false` would read better as:
`it should be true/false`

PR #586, opened 2016-10-03
Airport Challenge

Tomy8s:


lib/plane.rb
```diff
@@ -0,0 +1,25 @@
...
+      @state = 'grounded'
+      @airport = airport
+    end
+  end
+
+  def take_off
+    @state = 'flying'
+  end
+
+  def confirm
```
silvabox: This method name doesn't really relate to the value it returns

spec/airport_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'airport'
+
+describe Airport do
+  it 'has weather' do
+    expect(subject).to respond_to :weather
+  end
+  it 'has differing weather conditions' do
+    weather = []
+    100.times{ weather << subject.change_weather }
+    expect(weather).to include('clear') and expect(weather).to include('clear')
```
silvabox: duplicate expectation

spec/plane_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+describe Plane do
+    let(:airport) { double(:airport, weather: 'clear') }
+    let(:sunny_airport) { double(:airport, weather: 'clear') }
+    let(:stormy_airport) { double(:airport, weather: 'stormy') }
+  it 'lands somewhere' do
+    subject.land(airport)
+    expect(subject.state).to eq 'grounded'
+  end
+  it 'lands at a sunny airport' do
```
silvabox: Isn't this the same as the previous test?

spec/plane_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    subject.take_off
+    subject.land(sunny_airport)
+    expect(subject.state).to eq 'grounded'
+  end
+  it 'doesn\'t land at a stormy airport' do
+    subject.take_off
+    subject.land(stormy_airport)
+    expect(subject.state).to eq 'flying'
```
silvabox: Why is the plane still flying?  How can I query the system to find out what went wrong?

spec/plane_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  it 'confirms state after take off' do
+    subject.take_off
+    expect(subject.confirm).to eq 'flying'
+  end
+  it 'confirms state after landing' do
+    subject.land(airport)
+    expect(subject.confirm).to eq 'grounded'
+  end
+  it 'knows where it\'s grounded' do
+    location = ('a'..'z').to_a.shuffle[0,rand(10) + 1].join
```
silvabox: why so complicated?

PR #589, opened 2016-10-03
Weekend Challenge

AkramRasikh:
Able to store planes and allow them to fly out. capacity is not compromised but weather is not a factor :(

### DIFF
--- a/lib/airplane.rb
+++ b/lib/airplane.rb
@@ -1,2 +1,3 @@
 class Air_Plane
+  attr_reader :planes
 end
--- a/lib/airplane.rb
+++ b/lib/airplane.rb
@@ -1,0 +1,2 @@
+class Air_Plane
+end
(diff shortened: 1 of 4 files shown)


lib/airplane.rb
```diff
@@ -0,0 +1,3 @@
+class Air_Plane
```
maryrosecook: Use CamelCase for class names :)

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+  DEFAULT_CAPACITY = 5
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @airplanes = []
+  end
+
+  def set_capacity
+    input_capacity = gets.chomp
+    if input_capacity.empty?
+      @capacity
```
maryrosecook: Does this case need to be handled?

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    @airplanes.include? airplane
+  end
+
+  def instruct_take_off(airplane)
+  #  fail "Too stormy to take off" if stormy
+    @airplanes.delete(airplane)
+  end
+
+  private
+  def stormy?
```
maryrosecook: Try not to write code that isn't tested.  A good approach is to start with the test, then get it to pass.

PR #591, opened 2016-10-03
challenge pull request

James-SteelX:


lib/plane.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'airport'
+
+class Plane
+
+  attr_accessor :passengers
```
silvabox: `passengers` was not in the requirements

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def initialize
+    @passengers = rand(100..250)
+    @status = :Flying
+  end
+
+  def passengers?
+    @passengers
```
silvabox: methods suffixed with `?` should return `true` or `false`.  This returns a number and so will always be truthy.

spec/airport_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+
+   subject(:airport) { described_class.new }
+   let(:plane) {double :plane}
+
+    it { is_expected.to respond_to(:land_plane).with(1).argument }
+    it { is_expected.to respond_to(:take_off).with(1).argument   }
+    it { is_expected.to respond_to(:current_weather)             }
+
+    it 'gives the ability to change the weather' do
+      expect(airport.change_weather).not_to eq subject
```
silvabox: Why would `change_weather` ever be equal to the airport?

If you meant:
```ruby
expect(airport.change_weather).not_to eq subject.the_weather
```
Then this is still incorrect as `change_weather` could randomly select the same weather again.

spec/airport_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    end
+
+    it 'lets you know the current weather' do
+      expect(airport.current_weather).to eq subject.the_weather
+    end
+
+ describe '#initalization' do
+   subject(:airport) { described_class.new }
+    it 'has a default capacity' do
+      allow(airport).to receive(:stormy?).and_return(false)
```
silvabox: Is this mock relevant?

spec/plane_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+   end
+
+ describe 'status' do
+      it 'should give the current flight status' do
+        expect(subject.status).to be :Flying
+      end
+
+      it 'should change to Landed when plane lands' do
+        airport = airport = Airport.new(5, :Sunny)
```
silvabox: Are you testing the airport or the plane?

spec/plane_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        airport.land_plane(subject)
+        expect(subject.status).to be :Landed
+      end
+
+      it 'shouldn\'t have any passengers once landed' do
+        airport = airport = Airport.new(5, :Sunny)
```
silvabox: why `airport = airport = ...`?

PR #592, opened 2016-10-03
Pull request for week1 challenge

souljuse:
The code is working but the test is not. The problem seems to be the method stormy? which is defined in a different class, the weather one. I tried using doubles but I could not come to a conclusion


lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+	def land(plane)
+    fail 'Airport is full. The plane cannot land' if full?
+    fail 'The plane cannot land due to bad weather conditions' if @weather.stormy?
+    @planes << plane
+    #puts "The plane has landed"
+	end
+	attr_reader :plane
+  private :empty?, :full?
+end
```
silvabox: You must fix your whitespace and indenting.  This file is very difficult to read.  Look up some ruby style guides for more info.

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+  def full?
+    @planes.count >= @capacity
+  end
+	def land(plane)
+    fail 'Airport is full. The plane cannot land' if full?
+    fail 'The plane cannot land due to bad weather conditions' if @weather.stormy?
+    @planes << plane
+    #puts "The plane has landed"
+	end
+	attr_reader :plane
```
silvabox: This attribute is redundant

lib/plane.rb
```diff
@@ -0,0 +1,4 @@
+require_relative 'airport'
+class Plane
+
```
silvabox: Should the plane have behaviour?

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'airport'
+DEFAULT_CAPACITY = 20
```
silvabox: Why is this constant here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+describe Airport do
+  it { is_expected.to respond_to :take_off }
+  it { is_expected.to respond_to(:land).with(1).argument }
+  #mocking behaviour of the double
+  #we use the plane.new very often so we store it with a let statement
+  let(:plane) { double :plane }
+  describe '#land' do
+    it "lands a plane" do
+      plane = double(:plane)
```
silvabox: This is redundant as the `let` statement above already defines `plane`

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      expect(subject.land(plane)).to eq [plane]
+    end
+    it "should not let plane land if airport is full" do
+      subject.capacity.times { subject.land plane }
```
silvabox: bear in mind you are landing the same plane n times

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      expect { subject.land plane }.to raise_error ("Airport is full. The plane cannot land")
+    end
+    it "should not let plane land if the weather is stormy" do
+      weather = double(:weather, stormy?: true)
```
silvabox: you're correctly defining a double, but it has no subsequent involvement in the test.  Use dependency injection to pass it to the airport when the airport is created.

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  describe '#take_off' do
+    it "the airport should let planes take off" do
+      subject.land plane
+      expect(subject.take_off).to be plane
+    end
+      # it "should not release infinite planes" do
+      #   expect{ subject.take_off }.to raise_error ("There are no planes in the airport")
+      # end
+    it "should not let plane fly if the weather is stormy" do
+      weather = double(:weather, stormy?: true)
```
silvabox: See previous comment about weather double

PR #645, opened 2017-01-07
class plane and method land and take_off added

BasileKoko:


lib/airport.rb
```diff
@@ -0,0 +1,39 @@
+
+require_relative 'plane'
+
+
+class Airport
+  attr_reader :capacity
```
danldb: doen't need to be public - should be private

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    def full?
+      planes.count >= capacity
+    end
+
+private
+
+attr_reader :planes
+    def stormy?
+      rand(0..5) == 2
+    end
```
danldb: this should be isolated from any other code, maybe in a class or module. That way, you can ensure that untestable elements are isolated from the rest of your secure, safe code.

lib/plane.rb
```diff
@@ -0,0 +1,4 @@
+
+class Plane
+
+end
```
danldb: sadly not finished :(

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
+
+require 'airport'
+require 'plane'
+
+describe Airport do
+  it {is_expected.to respond_to(:land).with(1).argument}
+  it {is_expected.to respond_to(:take_off).with(1).argument}
+  it {is_expected.to respond_to(:full?)}
```
danldb: you can delete these respond_to once other tests cover the methods - the functionality is duplicated.

spec/plane_spec.rb
```diff
@@ -0,0 +1,8 @@
+
+require 'plane'
+
+describe Plane do
+  it " New insance class is Plane" do
+    expect(subject.class).to eq Plane
```
danldb: this is vacuous - not testing anything.

PR #646, opened 2017-01-07
Final submission

samjones1001:


lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+	
+	attr_reader :grounded, :capacity
+	DEFAULT_CAPACITY = 10
+
+	def initialize(capacity = DEFAULT_CAPACITY)
+		@grounded = []
+		@capacity = capacity
+	end
+
+	def receive_plane(plane)
```
danldb: extract the concretes - I would call this `receive(plane)` thus enabling future reuse of the code for helicopters, sea planes, drones, spaceships or whatever..

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+		raise "That plane is already grounded!" if !flying?(plane)
+		plane.on_ground
+		@grounded << plane
+	end
+
+	def release_plane(plane)
+		raise "Planes cannot take off during storms!" if stormy?
+		raise "That plane is not in the airport!" if !plane_present?(plane)
+		plane.in_air
+		@grounded.delete(plane)
```
danldb: use the attr_reader - you could put `grounded.delete(plane)`

lib/plane.rb
```diff
@@ -0,0 +1,16 @@
...
+		@airborne = true
+	end
+
+	def in_air
+		@airborne = true
+	end
+
+	def on_ground
+		@airborne = false
+	end
```
danldb: indentation went crazy here!

does the attr_reader need to be public?

spec/airport_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+describe Airport do
+
+	let(:airport) {Airport.new}
+	let(:plane)   {double :plane,airborne: true,on_ground: true,in_air: true}
+
+	context 'in good weather' do
+		
+		before(:each) do
+			allow(airport).to receive(:stormy?).and_return false
```
danldb: stubbing the subject with `allow .. to receive` is generally a bad idea, wth very few exceptions.

PR #647, opened 2017-01-07
plane mostly finished

JamesTurnerGit:


lib/air_traffic_control.rb
```diff
@@ -0,0 +1,20 @@
+#handles messages between airport and planes
```
danldb: should start with `# understands`

lib/air_traffic_control.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def take_off_plane plane, airport
+    return false if airport.stormy?
+    return false unless plane.landed?
+    plane.take_off airport
+    airport.plane_taken_off plane
+    true
+  end
+
+end
```
danldb: this class doesn't really have much behavior, and does not hold state

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+#manages an array and its size and has its own weather station
+require_relative "weather_station"
+
+class Airport
+  DEFAULT_CAPACITY = 5
+  attr_reader :weather_station, :planes
+  attr_accessor :capacity
+  
+  def initialize(init_capacity = DEFAULT_CAPACITY, station = WeatherStation.new)
```
danldb: very nice

lib/plane.rb
```diff
@@ -0,0 +1,35 @@
+#manages it's own state
```
danldb: `# understands how to land at an airport` would be clearer. Every object manages its own state.

PR #649, opened 2017-01-07
Airport-challenge

Justinio14:


lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+require_relative 'weather'
+
+class Airport
+
+  DEFAULT_CAPACITY = 40
+
+  attr_accessor :sunny, :arrivals, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @sunny = Weather.new.sunshine
```
danldb: airports weather can never change?

lib/plane.rb
```diff
@@ -0,0 +1,13 @@
+require_relative 'weather'
+
+class Plane
+
+  def land
+    true
```
danldb: always true? No functionality here.

PR #654, opened 2017-01-07
first user story update

thegooders:


lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+include Weather
+
+attr_reader :landing_bay, :capacity
```

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @landing_bay = []
+    @capacity = capacity
+  end
+
+  def land_plane(plane)
+    raise "Stormy Weather Alert: Landing not permitted" if Weather.stormy?
+    raise "Airport Full Alert: Landing not permitted" if at_capacity?
+    plane = plane
+    plane.change_status("Landed")
```
danldb: directly changing another objects internal state breaks encapsulation

lib/plane.rb
```diff
@@ -0,0 +1,12 @@
+class Plane
+
+attr_reader :landed_status
+
+def initialize(landed_status="Airbourne")
+  @landed_status = landed_status
```
danldb: prefer boolean to string:

`@landed = true`

lib/plane.rb
```diff
@@ -0,0 +1,12 @@
...
+end
+
+def change_status(new_status)
+  @landed_status = new_status
+end
```
danldb: keep tighter control over your internal state:
```ruby
def land
  fail "plane is already landed" if @landed
  @landed = true
end
```
Something like that ^^ prevents other objects messing things up.

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+module Weather
+
+STORM_PROBABILITY = 0.5#this is a floating point number between 1 and 0. The higher the number the higher the probability of storms. That is, if STORM_PROBABILITY = 1 then prob_threshold = 0 i.e. 100% chance that the weather is stormy ==> Weather.stormy? == true.
```
danldb: unnecessary comments

PR #655, opened 2017-01-07
first commit

schlattk:


lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    "plane has taken off"
+    end
+  end
+
+  def full?
+    planes.count == capacity - 1
+  end
+
+  def stormy?
+    Weather.new.forecast == "stormy"
```
neoeno: I like how you've given `Airport` the responsibility for deciding whether the weather is suitable for flying — great use of the SRP (Single Responsibility Principle).

lib/plane.rb
```diff
@@ -0,0 +1,10 @@
+class Plane
+
+  attr_accessor :status
+
+  def initialize
+
+    @status = "airborne"
+
+  end
+end
```
neoeno: I'd give plane the responsibility of deciding its own status here. Something like this:

```ruby
class Plane
  def initialize
    @status = :landed
  end

  def take_off
    @status = :airborne
  end

  def land
    @status = :landed
  end

  def airborne?
    @status == :airborne
  end

  private

  attr_accessor :status
end
```

It's nice for objects to take responsibility for their own internal state — having other objects like `Airport` reaching in and changing their instance variables directly makes it harder for a developer to understand what's going on (since in the example above `@status` can only change through one method — but in the original class it can be changed from anywhere in the application)

lib/tester.rb
```diff
@@ -0,0 +1,23 @@
...
+puts "there are now #{frankfurt.planes.count} planes at Frankfurt Airport"
+#puts "as an edge case we are asking #{mig} to land again"
+#puts "#{frankfurt.land(mig)}"
+#puts "we are asking #{spitfire} to take off although it is airborne"
+#puts "#{frankfurt.take_off(spitfire)}"
+puts "we are asking #{boeing} to take off again"
+puts "#{frankfurt.take_off(boeing)}"
+puts "there are now #{frankfurt.planes.count} planes at Frankfurt Airport"
+#puts "we are asking ten planes to land althoug remaining capacity is only 9"
+#puts "#{10.times do frankfurt.land(Plane.new)end}"
```
neoeno: I really like this — great way of automating your IRB stuff and giving a good "safety check" on the whole application. You'll learn a similar technique for this called integration testing later on!

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
+
+def forecast
+
+meteo = ["stormy","sunny","sunny","sunny","sunny"]
+
+meteo[rand(5)]
```
neoeno: Give this a read: https://ruby-doc.org/core-2.2.0/Array.html#method-i-sample

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
+
+def forecast
+
+meteo = ["stormy","sunny","sunny","sunny","sunny"]
```
neoeno: Might consider extracting this to a constant, something like this:

```ruby
class Weather
   METEO = ["stormy", "sunny", "windy"] # etc
  # ...
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+airport = Airport.new
+plane = Plane.new
+allow(airport).to receive(:stormy?) {false}
+airport.land(plane)
+expect(airport.planes).to include(plane)
+expect(airport.take_off(plane)).to include("plane has taken off")
+
+end
+it "should raise an error when landing and weather is stormy" do
```
neoeno: Really like these `it` descriptions. Give this a read to refine them even further: http://betterspecs.org/#should

spec/plane_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+  it "should have a status of 'airborne' when created" do
+      plane = Plane.new
+      expect(plane.status) == "airborne"
+      end
+  it "it should be possible to change status to 'landed'" do
+      plane = Plane.new
+      plane.status = "landed"
+      expect(plane.status) == "landed"
+    end
+end
```
neoeno: Do you find this file easier to read than `airport_spec.rb`? I find it much easier to read.

The indentation in a lot of the code here is quite inconsistent. This makes it hard to follow and understand — when we get to larger projects this is going to present more of a problem. You'll save yourself a lot of "pulling your hair out" moments if you get into the habit! Swing by my desk if you want me to show you some ways to automate that process.

PR #656, opened 2017-01-07
adds first user story landing plane

Putterhead:


lib/plane.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def landed?
+    true
+  end
+  
+  def airport
+  end
+
+
+end
```
danldb: shame this is not completed, you were on the right track

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+private
+
+  def full?
+    @planes.size >= @capacity
+  end
+
+  def stormy?
+    rand(1..6) > 4
+  end
+end
```
danldb: looks great - you kept it really simple. You should extract the weather out of the airport as it is a random, untestable element.

PR #657, opened 2017-01-08
Rick Clark pull request

rkclark:


lib/airport.rb
```diff
@@ -0,0 +1,39 @@
+class Airport
+
+  attr_reader :airport_name, :full, :planes, :capacity
+  alias_method :full?, :full
+
+  DEFAULT_CAPACITY = 3
+
+  def initialize(name, capacity=DEFAULT_CAPACITY)
+    @airport_name = name
+    @full = false
```
maryrosecook: Does `full` need to be stored and updated? Could it be derived when asked for?

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    set_full
+  end
+
+  def undock(plane)
+    @planes.delete(plane)
+    set_full
+  end
+
+  def has_plane_docked?(plane)
+    planes.include?(plane) ? true : false
```
maryrosecook: Does this line need `true : false`?

lib/plane.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def set_airborne(status)
+    @airborne = status
+  end
+
+  def sunny?
+    Weather.sunny?
+  end
+
+  def airport_full?(airport)
```
maryrosecook: I'm super into very short, explanatory methods.  However, this method, `dock_at_airport`, `undock_from_airport` and `docked_at_airport?` don't seem much clearer.  They aren't much shorter or less complex than the methods they wrap.

lib/weather.rb
```diff
@@ -0,0 +1,23 @@
+module Weather
+
+  @@sunny = true
```
maryrosecook: Is there a need to store the weather as state when it's recalculated every time it's asked for?

spec/plane_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+  describe ".land" do
+    context "plane is airborne" do
+      context "weather is sunny" do
+        before(:each) do
+          set_sunny(true)
+        end
+        it "can land" do
+          expect{plane.land(airport)}.not_to raise_error
+        end
```
maryrosecook: This code would be even easier to read if the it blocks were separated by a line of whitespace.

PR #658, opened 2017-01-08
Satisfied four out of six user stories

kathicks:
I've now satisfied four out of six of the user stories and decided that it was a good point to test the code with a pull request.

The code still does not have any weather state or exceptions based on the state of the weather.


lib/plane.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  private
+
+  def weather
+    Weather.new.check_weather
+  end
+
+  def land_exceptions(airport)
+    fail "Plane is already landed at #{airport}." if airport.planes.include? (self)
```
maryrosecook: These guard clauses are great! Super clear.  Also love that you've bundled them up into methods with super clear names like `land_exceptions`.  The only change I would suggest is to make the checks as simple as possible.  For example, could you extract `weather == "stormy"` to a method with a clear name?

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
+
+  def check_weather
+    chance < 85 ? "sunny" : "stormy"
```
maryrosecook: Could 85 be set as a constant so it can be changed more easily and is also self-documenting?

lib/plane.rb
```diff
@@ -0,0 +1,45 @@
+class Plane
+  attr_reader :landing_status
+
+  def initialize
+    @landing_status = nil
```
maryrosecook: When planes are created, it seems like they'll be in an indeterminate landing status.  Try to avoid allowing data to ever be in an error state.  Could landing status be initialised to a meaningful value?

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
+class Airport
```
maryrosecook: Great separation of responsibilities amongst the classes.

lib/plane.rb
KatHicks: @maryrosecook - thanks, really useful. hadn't thought about data being in an error state.

lib/plane.rb
KatHicks: @maryrosecook - yes, definitely. hadn't heard about guard clauses until the code review. been using them much more this week.

lib/weather.rb
KatHicks: @maryrosecook - again - yes definitely. been getting to grips with constants in the oystercard challenge.

KatHicks:
@maryrosecook - thanks for looking over my code - really appreciate it and comments are super helpful 😄

maryrosecook:
Sure thing! Glad it was helpful!

PR #662, opened 2017-01-08
Trial Pull

laurenrosie:


lib/action.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'plane.rb'
+require_relative 'airport.rb'
+require_relative 'weather.rb'
+
+class Action
```
maryrosecook: The Action class is an interesting idea.  Is there a reason this logic isn't in the Airport?

lib/action.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'plane.rb'
```
maryrosecook: No need for .rb when requiring.

lib/action.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def takeoff(plane, airport)
+    raise "Weather is too stormy" if airport.stormy?
+    raise "Plane not at airport" if !airport.has_plane?(plane)
+    plane.takeoff(airport)
+    airport.takeoff(plane)
+  end
+
+  def land_multiple(planes, airport)
```
maryrosecook: Could land_multiple use airport.land to reduce the duplication of code? eg the weather is checked in both methods.

lib/action.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'plane.rb'
+require_relative 'airport.rb'
+require_relative 'weather.rb'
+
+class Action
+
+  def land(plane, airport)
+    raise "Weather is too stormy" if airport.stormy?
```
maryrosecook: Lovely, clean code!

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  DEFAULT_CAPACITY = 2
+
+  def initialize(planes=[], capacity=DEFAULT_CAPACITY)
+    @planes = planes
+    @weather = get_weather
+    @capacity = capacity
+  end
+
+  attr_accessor :planes
```
maryrosecook: `attr_`s idiomatically go above all the methods.

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
+class Plane
+  def initialize(airport=nil)
```
maryrosecook: Does `airport` need to be passed in here?

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
...
+  def confirm_landing(airport)
+    self.current_airport == airport
+  end
+
+
+  def takeoff(airport)
+    self.current_airport= nil
+  end
+
+  def confirm_takeoff(airport)
```
maryrosecook: It's not always possible, but try to avoid having methods that are purely there to verify test conditions.

spec/action_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    it 'doesnt throw error when sunny' do
+      plane = Plane.new
+      airport = Airport.new
+      airport.weather= :sunny
+      puts "weather is #{airport.weather}"
+      expect{subject.land(plane, airport)}.not_to raise_error
+    end
+
+    # it 'throws error when airport full' do
```
maryrosecook: Try to avoid committing commented out code.

PR #663, opened 2017-01-08
First Pull Request

BenJohnCarson:


lib/plane.rb
```diff
@@ -0,0 +1,50 @@
...
+    
+    def land(airport)
+        landing_errors(airport)
+        @landed = true
+        park(airport)
+    end
+    
+    def take_off
+        take_off_errors
+        leave
```
danldb: I would tend to leave the guard clauses easily visible in the method

PR #664, opened 2017-01-08
airport_challenge pull request

bvjones:
completed bad weather function not applied to code.

### DIFF
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,3 +1,15 @@
 class Weather
 
+  def stormy?
+    set_weather(randomise)
+    @stormy
+  end
+
+  def set_weather(random_number)
+    @stormy = random_number >= 40 ? true : false
+  end
+
+  def randomise
+    rand(50)
+  end
 end
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,0 +1,3 @@
+class Weather
+
+end
--- a/spec/plane_spec.rb
+++ b/spec/plane_spec.rb
@@ -1,8 +1,14 @@
 require 'plane'
 
 describe Plane do
-  
-    it 'checks if plane is working' do
-      expect(subject).to respond_to(:working?)
+  plane = Plane.new
+
+  describe '#landing' do
+      it 'landing a plane in the airport' do
+      should respond_to(:landing)
+    end
+      it 'checking if plane is grounded' do
+      expect(subject.landing(false)).to eq true
+    end
   end
 end
--- a/spec/plane_spec.rb
+++ b/spec/plane_spec.rb
@@ -1,0 +1,8 @@
+require 'plane'
+
+describe Plane do
+  
+    it 'checks if plane is working' do
+      expect(subject).to respond_to(:working?)
+  end
+end
--- a/spec/plane_spec.rb
+++ b/spec/plane_spec.rb
@@ -4,11 +4,26 @@ describe Plane do
   plane = Plane.new
 
   describe '#landing' do
-      it 'landing a plane in the airport' do
+    it 'landing a plane in the airport' do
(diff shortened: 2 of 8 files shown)


lib/weather.rb
```diff
@@ -0,0 +1,15 @@
+class Weather
+
+  def stormy?
+    set_weather(randomise)
+    @stormy
+  end
+
+  def set_weather(random_number)
+    @stormy = random_number >= 40 ? true : false
```
neoeno: Also note that since `rand` with no arguments returns a random value between 0.0 and 1.0 — and since here we're going for a 1/5 chance of being stormy, you might equally do this with something like `rand < 0.2` — or to take it even further

```ruby
STORMY_THRESHOLD = 1.0 / 5.0 # Be careful with this, it only works if you keep it `5.0` rather than `5`!

def stormy?
  set_weather(rand)
  @stormy
end

def set_weather(random_number)
  @stormy = rand < STORMY_THRESHOLD
end
```

It all adds to the readability.

spec/plane_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+    it 'checking if plane has left' do
+      expect(subject.take_off(true)).to eq "Plane has sucessfully taken off!"
+    end
+    it 'error when plane is already grounded' do
+      expect{ subject.take_off(false) }.to raise_error("Cannot take off! Plane is already in the air")
+    end
+    # it 'error when taking off in bad weather' do
+    #   expect(subject.stormy?(true)).to eq "Cannot take off! Bad Weather"
+    # end
```
neoeno: Give this a read: https://github.com/makersacademy/airport_challenge/blob/master/docs/review.md#remove-all-commented-out-code

spec/plane_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'plane'
+
+describe Plane do
+  plane = Plane.new
```
neoeno: We don't use this as far as I can see. If we were to do:

```ruby
subject(:plane) { Plane.new }
```

Then we could say things like

```ruby
expect(plane.landing(false)).to eq "Plane has landed!"
```

Which is a little more readable than `subject.landing` etc

spec/plane_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  describe '#landing' do
+    it 'landing a plane in the airport' do
+      should respond_to(:landing)
```
neoeno: Give this a read: http://betterspecs.org/#expect — generally `should` is discouraged in RSpec tests these days. Just one more style thing to remember I guess!

spec/weather_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    it 'test is stormy is true' do
+      expect(subject.set_weather(40)).to eq true
+    end
+    it 'test is stormy is false' do
+      expect(subject.set_weather(39)).to eq false
+    end
+    it 'test if set_weather doesnt raise an error' do
+      expect { subject.stormy? }.not_to raise_error
+    end
+  end
```
neoeno: I like this — we've got as much of `Weather` testable as possible, isolating the randomness to a single method. There are other ways to do that, but this one is perfectly serviceable and clear.

neoeno: I especially like that even with `#stormy?` where you can't reliably test the return value you've tested that it doesn't raise an error! Really great to squeeze that last ounce of testing out of it.

To take this even further, in RSpec 3.4 you can do this:

```ruby
expect(subject.stormy?).to be(true).or be(false)
```

Overcomplicated? Perhaps! I'll let you decide whether you like it or not — haven't made up my mind yet. And in any case we're only using RSpec 3.2 in this project so we can't use it here!

PR #670, opened 2017-01-08
Week 1 Weekend Challenge Submission

thesedatedprince:
My solutions for the week 1 challenge.


lib/airport.rb
```diff
@@ -0,0 +1,53 @@
+require './lib/airplane.rb'
+require './lib/weather.rb'
+
+
+class Airport
+
+attr_reader :landed_planes
+attr_reader :capacity
+attr_accessor :initialize
```
danldb: `attr_accessor :initialize`- this shouldn't be in your code

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+    stormy?
+    full?
+    @landed_planes << plane
+    #print "The plane has landed."
+    print_landing
+  end
+
+  def takeoff
+    stormy?
+    full?
```
danldb: methods ending in `?` should be boolean i.e. return true or false. hiding your guard clauses makes this hard to read. I prefer:
```ruby
def takeoff
  fail "cannot takeoff if stormy" if stormy?
...
```

PR #674, opened 2017-01-08
Submission

frankiefy:
I have not 100% completed this challenge


lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+class Airport
+attr_accessor :planes
+
+  def initialize(capacity=1)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def recieve_plane(plane)
+    if full?
```
maryrosecook: Could this be a guard clause, rather than a conditional?

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+      fail 'airport full'
+    else
+      @planes.push(plane)
+      return @planes[-1]
```
maryrosecook: It's maybe clearer to just return the `plane` here, rather than getting it back out of the array.

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+  end
+
+  def takeoff_plane(plane)
+    @planes.delete(plane)
+    return @planes
```
maryrosecook: Be a little careful about giving internal state out from methods.  Code that calls `takeoff_plane` may start messing around with `@planes`, which could cause problems.

lib/plane.rb
```diff
@@ -0,0 +1,22 @@
+class Plane
+
+  attr_accessor :state
+  attr_accessor :weather
+
+  def initialize(state='flying')
+    @state = []
```
maryrosecook: Is there a need to keep a log of the states?

lib/plane.rb
```diff
@@ -0,0 +1,22 @@
...
+  def takeoff
+    @state.pop
+    @state.push('flying')
+  end
+
+  def land
+    @state.pop
+    @state.push('landed')
+  end
+
```
maryrosecook: Watch out for extra whitespace.

lib/weather.rb
```diff
@@ -0,0 +1,22 @@
+class Weather
+attr_accessor :weather
+
+  def initialize #upon initalizing it is given random weather 1/6 chance of storm
```
maryrosecook: Could this function be shorter? Perhaps consider using a Ruby range for the first conditional clause.  Perhaps consider using a ternary operator to succinctly communicate the idea: either this happens, or this happens.

lib/weather.rb
```diff
@@ -0,0 +1,22 @@
...
+    num = rand(0...5) #change this for definitive weather
+    if num == 0 || num == 1 || num == 2 || num == 3 || num == 4
+      @weather = 'clear'
+    elsif num == 5
+      @weather = 'stormy'
+    end
+  end
+
+
+  def clear?
```
maryrosecook: Could you make this shorter and clearer by returning the result of an equality comparison, rather than using a conditional?

PR #681, opened 2017-01-09
1st Pull request

dylanrhodius:


spec/plane_spec.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/plane'
+
+describe Plane do
+
+  it {is_expected.to respond_to :status?}
+  it {is_expected.to respond_to :landed}
+  it {is_expected.to respond_to :departed}
+
+  it 'returns the plane Object\'s status' do
```
neoeno: I like your dedication to consistency in your `it` descriptions here! It's a matter many programmers have fought and died over — personally I always use double quotes for strings with human language in, but it's a matter of style! 

Sometimes if I'm forced to use single quotes I'll use `’` (shift+alt+] on a mac) but people get very upset when I do that...

toDo.md
```diff
@@ -0,0 +1,4 @@
+# To Do List
+
+* Create .pryrc file to facilitate feature test
```
neoeno: Already done!

lib/plane.rb
```diff
@@ -0,0 +1,21 @@
+require_relative 'airport'
+
+class Plane
+
+def initialize (status = "Airborne.")
```
neoeno: Give this a read: https://github.com/makersacademy/airport_challenge/blob/master/docs/review.md#prefer-symbols-over-strings

lib/plane.rb
```diff
@@ -0,0 +1,21 @@
...
+  @status = status
+end
+
+  def status?
+    @status
+  end
+
+  def landed
```
neoeno: I'd consider calling this `land` — the idea is to put methods that "do something" in imperative case. Imagine it like "Oi Plane! Land!". It's a bit of a subtle distinction but it'll become clearer with time:

Some more info on the area of naming: https://github.com/makersacademy/airport_challenge/blob/master/docs/review.md#naming-convention-matching-the-domain-model

spec/airpot_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+    end
+
+    context 'when weather is stormy' do
+      before do
+        airport = Airport.new
+      end
+      it 'prevents a Plane object from taking off' do
+        allow(subject).to receive(:empty?) { false }
+        allow(subject).to receive(:weather) { "stormy" }
```
neoeno: It's very tempting to stub out methods on the class we're testing — but it's generally not a good idea as you want to verify the class functions as it is — by changing it you really testing something different. Does that make sense?

In this case, consider passing in a `Weather` instance in the initializer, like this:

```ruby
class Airport
  def initialize(planes = [], weather = Weather.new, capacity = DEFAULT_CAPACITY)
    @planes = planes
    @capacity = capacity
    @weather = weather.current_weather
  end
  # ...
end
```

And then call `current_weather` on that method.

Then in your tests you can do something like:

```ruby
context 'when weather is stormy' do
  let(:weather) { double :weather, current_weather: "stormy" }
  let(:planes) { [Plane.new] }
  subject(:airport) { Airport.new(planes, weather, Airport::DEFAULT_CAPACITY) }
  
  it 'prevents a Plane object from taking off' do
    expect { subject.take_off }.to raise_error "Plane cannot take off due to stormy weather."
  end
end
```

There are a few new techniques in there — let me know if anything doesn't make sense.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+private
+
+  def full?
+    @planes.count >= DEFAULT_CAPACITY
+  end
+
+  def empty?
+    @planes.empty?
```
neoeno: Give this a read: https://github.com/makersacademy/airport_challenge/blob/master/docs/review.md#avoid-defining-attr_reader-then-accessing-the-instance-variable-directly

lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+class Weather
+
+  def current_weather
```
neoeno: I'd consider calling this `weather` rather than `current_weather`. If you think about how it'll be called, it'll mostly be `weather.current` — so we already know it's about weather, and no need for the extra word in the method name.

spec/airpot_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+
+
+  describe '#land(plane):' do
+    context 'when weather is stormy' do
+      before do
+        @plane = Plane.new
+      end
+      it 'prevents a Plane object from landing' do
```
neoeno: Really like your `it` descriptions, context blocks, etc as you've done them in this exercise. Super impressive & will be similarly impressive to those you work with in future!

spec/airpot_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+    end
+  end
+
+
+
+  describe '#land(plane):' do
+    context 'when weather is stormy' do
+      before do
+        @plane = Plane.new
```
neoeno: Have a read on the RSpec `let` method for writing this in a more idiomatic way.

README.md
```diff
@@ -1,90 +1,27 @@
...
+[![Build Status](https://travis-ci.org/makersacademy/airport_challenge.svg?branch=master)](https://travis-ci.org/makersacademy/airport_challenge) [![Coverage Status](https://coveralls.io/repos/github/dylanrhodius/airport_challenge/badge.svg?branch=master)](https://coveralls.io/github/dylanrhodius/airport_challenge?branch=master)
+
 ```
         ______
         _\____\___
 =  = ==(____MA____)
           \_____\___________________,-~~~~~~~`-.._
           /     o o o o o o o o o o o o o o o o  |\_
+          `~-.__       __..----..__  DYLAN AIRLINES  )
```
neoeno: Mary and I found this very amusing

PR #684, opened 2017-01-09
Shereen's Airport Challenge

shezdev:
Halfway through, would have liked to get further but ran out of time :)

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -11,7 +11,7 @@ class Airport
     plane
   end
 
-  def prevent_take_off
+  def prevent_take_off(plane)
 
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,8 +2,13 @@
 require_relative "plane"
 
 class Airport
-def plane_landed?
-  Plane.new
-end
+
+  def plane_landed?
+    Plane.new
+  end
+
+  def plane_take_off
+  end
+
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -11,5 +11,8 @@ class Airport
     plane
   end
 
+  def prevent_take_off
+
+  end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -15,4 +15,7 @@ class Airport
     plane
   end
 
+  # def weather(type)
+  #   prevent_take_off if type == "stormy" end
+
 end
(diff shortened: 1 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,22 @@
+# /Users/shereen/Google Drive/Projects/airport_challenge/lib/airport.rb
+require_relative "plane"
+
+class Airport
+
+  def plane_landed?
```
maryrosecook: It feels like this method doesn't really check if a plane has landed.  Maybe think about how you want the Airport instance data to change when a plane lands.  Would you like to store the plane somewhere inside the airport instance. Then have `plane_landed?` look at that data to provide its answer.

lib/plane.rb
```diff
@@ -0,0 +1,14 @@
+# /Users/shereen/Google Drive/Projects/airport_challenge/lib/plane.rb
+
+class Plane
+
+  def confirm_landed?
```
maryrosecook: The idiom in Ruby is for methods with names that end in `?` to return booleans.

spec/airport_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+end
+
+describe Airport do
+  #2nd story / 1st test
+  it {is_expected.to respond_to :plane_take_off}
+  #2nd story / 3rd test
+  it { is_expected.to respond_to(:plane_take_off).with(1).argument }
+
+  #2nd story / 5th test
+  it 'returns an airborne plane' do
```
maryrosecook: Try to describe what part of the code is being tested.  eg `plane_take_off` is the method being run, but that isn't indicated in the text given to the describe or it blocks (`Airport` and `returns an airborne plane`).

lib/airport.rb
```diff
@@ -0,0 +1,22 @@
...
+    plane
+  end
+
+  def prevent_take_off(plane)
+    plane
+  end
+
+  def weather
+  #   prevent_take_off if type == "stormy" end
+  end
```
maryrosecook: I'd love to talk through your process for developing this code.  Could you grab me for a chat whenever you have a second? Thanks!

lib/airport.rb
```diff
@@ -0,0 +1,22 @@
+# /Users/shereen/Google Drive/Projects/airport_challenge/lib/airport.rb
+require_relative "plane"
+
```
maryrosecook: Not quite sure I understood the question you asked about the plane variable.  Perhaps we could talk about it in person?

PR #689, opened 2017-01-09
Weather test added

pedrocastanheira77:


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+  private
+  def airport_full
+    error_message = "Airport is full"
+    raise error_message if @planes.length >= @capacity
+  end
+
+  def airport_empty?
+    error_message = "No planes landed"
+    raise error_message if @planes.empty?
+  end
```
neoeno: There's some inconsistency here — one has a `?` on the end and one doesn't.

I'd consider googling up something like "predicate methods in ruby" to figure out which one you want.

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
+
+end
```
neoeno: I can imagine this caused you a little concern! I think the exercise expected you to actually instruct planes to take off, land, etc — so you might have had methods like `take_off` and `land` and then you could have raised an error when the plane was already in flight but tried to take off again — etc.

But with these conceptual exercises can be a little tricky to figure out what we mean — it's usually more straightforward when you have a real plane involved! Perhaps that's the answer — roleplay the weekend as if this is all really happening. Diagramming out how it works in real life might help.

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
+  def stormy?
+    rand(10) == 10 ? true : false
```
neoeno: Two things:

1. https://github.com/makersacademy/airport_challenge/blob/master/docs/review.md#use-implicit-return-of-booleans (Feel free to give this whole document a read — there's plenty of useful stuff there)
2. Try running this in irb:

```ruby
require_relative 'lib/weather'
weather = Weather.new
until weather.stormy?
  puts "Weather's still fine!"
end
```

What do you expect to happen? What actually happens?

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+require 'weather'
+require 'plane'
+
+
+describe Airport do
+
+  it {is_expected.to respond_to(:arrivals)}
+  it {is_expected.to respond_to(:departures)}
+  it {is_expected.to respond_to(:planes)}
+  it {is_expected.to respond_to(:capacity)}
```
neoeno: Give this a read https://github.com/makersacademy/airport_challenge/blob/master/docs/review.md#eliminate-redundant-respond_to-expectations

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  it 'arrival with proper weather' do
```
neoeno: Whenever you find yourself writing "with" or "if" in an `it` message — consider using context blocks. There's a little bit more on that [here](https://github.com/makersacademy/airport_challenge/blob/master/docs/review.md#use-context-and-describe-blocks-to-create-test-scopes)

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    airport = Airport.new
+    allow(airport.weather).to receive(:stormy?).and_return(false)
+    expect(airport.arrivals.empty?).to eq(false)
+  end
```
neoeno: Ah — I realise some more about the reason `arrivals` is so named. In general it's better not to mix commands with queries. There's some more on that [here](https://github.com/makersacademy/airport_challenge/blob/master/docs/review.md#separately-name-command-and-query-methods).

In this case we could have a method on `Airport` called `land_plane` which lands the plane and another separate method called something like `landed_planes` (or even possibly just `planes`).

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  it 'depature with stormy weater' do
+    airport = Airport.new
+    airport.planes << Plane.new
+    allow(airport.weather).to receive(:stormy?) { true }
+    error_message = "Bad weather conditions to depart"
+    expect {airport.departures}.to raise_error(error_message)
+  end
+
+  it 'airport empty' do
```
neoeno: The message of the `it` block here is a little obscure, we might consider something more readable & conventional, like:

```ruby
describe "#departures" do
  context "with no planes" do
    it "throws an error indicating that there are no planes to take off" do
      # ...
    end
  end
end
```

The idea is that you should be able to read the whole thing like a sentence. In this case:

> Airport #departures with no planes throws an error indicating that there are no planes to take off

The phrasing is a little clunky I'll admit — but you can see it's a bit clearer than

> Airport airport empty

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    error_message = "No planes landed"
+    expect {subject.departures}.to raise_error(error_message)
+  end
+
+  it 'does not land if airport full' do
+    airport = Airport.new
```
neoeno: Give this a read: https://github.com/makersacademy/airport_challenge/blob/master/docs/review.md#use-named-subject-with-described_class

user story
```diff
@@ -0,0 +1,59 @@
...
+
+------------
+
+As the system designer
+So that the software can be used for many different airports
+I would like a default airport capacity that can be overridden as appropriate
+
+Object | Messages
+System Designer |
+Airport | airport_capacity
```
neoeno: Really nice work here — like how you've thought a little about the classes & methods required to perform the task.

PR #691, opened 2017-01-11
Airport_challenge completed

gbhachu:


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require_relative 'plane'
+
+class Airport
+
+  attr_reader :capacity
```

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @planes= []
+    @capacity = capacity
+
+end
+
+    def land(plane)
+      raise 'Cannot land plane: Airport full' if full?
+      @planes << plane
+      raise 'Cannot land: weather is stormy' if stormy?
```
danldb: this guard clause is should be on line 16 - it isn't guarding anything when it is at the end of the methid - the plane has already been shoveled into the array.

lib/plane.rb
```diff
@@ -0,0 +1,8 @@
+class Plane
+
+  def plane
+    true
+  end
+
```
danldb: pointless method - no functionality

spec/airport_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'airport'
+
+describe Airport do
+
+  describe '#initialize' do
+  it 'returns DEFAULT_CAPACITY for airport if none is provided' do
+    expect{(Airport.new::capacity).to eq airport::DEFAULT_CAPACITY}
+  end
+  it 'sets capacity when specified on initializaion' do
+    expect{(Airport.new(10)::capacity).to eq 10}
```
danldb: I wouldn't test initialize unless it was doing something particularly outlandish. This si just more code to maintain.

spec/airport_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+end
+
+  describe '#land' do
+    it 'instructs plane to land' do
+      allow(subject).to receive(:stormy?).and_return false
```
danldb: don't stub subject

PR #694, opened 2017-02-19
Ashwini's Airport Challenge

AAMani5:
1. user stories to feature tests
2. feature tests to unit tests
3. made the tests pass by implementation
4. considered few edge cases
5. can see dependencies among classes, working on reducing it.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,4 +1,11 @@
 class Airport
+  attr_reader :capacity
+
+  def initialize
+    @capacity = 10
+  end
+
   def is_full?
+    # true or false
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -7,6 +7,11 @@ class Airport
   end
 
   def is_full?
-    # true or false
+    false
+    # true or false depending on number of planes & capacity
+  end
+
+  def permission_to_land(plane)
+    true
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,2 @@
+class Airport
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -9,13 +9,13 @@ class Airport
   end
 
   def is_full?
-    false
+    planes.count >= capacity
     # true or false depending on number of planes & capacity
   end
 
   def permission_to_take_off(plane)
     fail "Weather stormy" if weather.is_stormy?
-    fail "Invalid request" if plane.status == "In-flight"
+    fail "Invalid request" if plane.status == "In-Flight"
     plane.take_off(self)
     true
   end
@@ -24,10 +24,15 @@ class Airport
     fail "Airport at capacity" if is_full?
     fail "Weather stormy" if weather.is_stormy?
     fail "Invalid request" if plane.status == "Landed"
+    plane.land(self)
     true
   end
 
   def confirm_landing(plane)
     planes << plane if (plane.status == "Landed" && plane.airport == self)
+  end
+
+  def confirm_take_off(plane)
+    planes.delete(plane) if plane.status == "In-Flight" && plane.airport == self
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,8 +1,9 @@
 class Airport
   attr_reader :capacity
+  DEFAULT_CAPACITY = 10
 
-  def initialize
-    @capacity = 10
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
   end
 
   def is_full?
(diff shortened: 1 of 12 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    fail NotImplementedError, "Sorry, You are either flying or not at #{self.name}" if (plane.status == "In-Flight" || plane.location != self.name)
+    plane.take_off
+    confirm_take_off(plane)
+    true
+  end
+
+  def initiate_landing_procedure(plane)
+    fail NotImplementedError, "Airport at capacity" if is_full?
+    fail NotImplementedError, "Weather stormy" if weather.is_stormy?
+    fail NotImplementedError, "Sorry, You are already on ground" if plane.status == "Landed"
```
maryrosecook: This line could be a little clearer if `plane` had a `landed?` method.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    @name = name
+    @capacity = capacity
+    @weather = weather
+    @planes = []
+  end
+
+  def initiate_take_off_procedure(plane)
+    fail NotImplementedError, "Airport empty" if planes.count == 0
+    fail NotImplementedError, "Weather stormy" if weather.is_stormy?
+    fail NotImplementedError, "Sorry, You are either flying or not at #{self.name}" if (plane.status == "In-Flight" || plane.location != self.name)
```
maryrosecook: Is NotImplementedError the right exception to raise here? Perhaps have a look at the available Ruby exceptions to see if what it's normally used for, and whether another exception would be more descriptive.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+class Airport
+  attr_reader :capacity, :planes, :name
+  DEFAULT_CAPACITY = 10
+
+  def initialize(name, capacity = DEFAULT_CAPACITY, weather = Weather.new)
```
maryrosecook: Injecting Weather here really cleans up the tests.  I like it.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    @planes = []
+  end
+
+  def initiate_take_off_procedure(plane)
+    fail NotImplementedError, "Airport empty" if planes.count == 0
+    fail NotImplementedError, "Weather stormy" if weather.is_stormy?
+    fail NotImplementedError, "Sorry, You are either flying or not at #{self.name}" if (plane.status == "In-Flight" || plane.location != self.name)
+    plane.take_off
+    confirm_take_off(plane)
+    true
```
maryrosecook: Does this method need to return true? Maybe read up on the Ruby idioms around how to indicate that a method has run successfully.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def confirm_take_off(plane)
+    planes.delete(plane) if plane.status == "In-Flight"
+  end
+
+  def is_full?
+    planes.count >= capacity
+  end
+
+  attr_reader :weather
```
maryrosecook: It's great that this line increases encapsulation by making `weather` private.

lib/plane.rb
```diff
@@ -0,0 +1,21 @@
+class Plane
+  attr_reader :status, :location
+
+  def initialize
+    @location = "Factory" # this variable changes during the land or take_off instruction
+    @status = "Brand New" # plane can be in "Brand New", "Landed", "In-Flight" status
```
maryrosecook: It feels like the existence of Brand New means that it's possible that a plane can be neither on the ground or in the air.  One tip for program state is to never allow it to be in an erroneous state.

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
```
maryrosecook: Really nice, focused, concise class!

spec/airport_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+describe Airport do
+  before(:each) do
+    @weather_sunny = double(:weather, :is_stormy? => false)
+    @weather_stormy = double(:weather, :is_stormy? => true)
+  end
+  let(:london_airport) {Airport.new("LHR", 2, @weather_sunny)}
+  let(:london_plane) {Plane.new}
+  let(:new_york_airport ) {Airport.new("EWR", 5, @weather_sunny)}
+  let(:new_york_plane)  {Plane.new}
+  let(:chicago_airport) {Airport.new("ORD", 2, @weather_stormy)}
```
maryrosecook: As I think we discussed in the code review, this code could be in some context blocks to make the test file a little more readable.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+class Airport
+  attr_reader :capacity, :planes, :name
```
maryrosecook: Do `capacity` and `planes` need to be public? Could the tests instead check for behaviour rather than state?

PR #695, opened 2017-02-19
initial pull request

rorymcgit:
RSpec feature test added. Tests as individual files.


lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def take_off(plane)
+    @plane = plane
+    fail "#{@plane} is not in the airport." unless at_airport?
+    fail "Take off not permitted due to stormy weather." if stormy?
+    @plane_arr.delete_if { |p| p == @plane }
+  end
+
+  def confirm_takeoff(plane)
```
maryrosecook: Yes, I agree with what you said in the review that checking a plane is in the airport before taking it off is a good idea.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+  attr_reader :plane_arr
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @plane_arr = []
+  end
+
+  def instruct_to_land(plane)
+    @plane = plane
```
maryrosecook: Is this line necessary?

maryrosecook: It's also a good idea for guard clauses to be right at the top of a method.  This avoids letting the method partly do some work and then get interrupted if a failure case triggers.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+  include Weather
+
+  DEFAULT_CAPACITY = 10
+
+  attr_reader :plane_arr
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @plane_arr = []
```
maryrosecook: No need to specify the type of the variable.  The `s` in `planes` makes it clear it's a collection of stuff.  Because of Ruby's focus on duck-typing, the specific type of collection shouldn't matter.

spec/airportCapacity_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  describe "#capacity", :capacity_tests do
+    it "prevents planes from landing when airport is full" do
+      Airport::DEFAULT_CAPACITY.times{airport.instruct_to_land(Plane.new)}
+      message = "Landing not permitted as max capacity has been reached."
+      expect{airport.instruct_to_land(plane)}.to raise_error message
+    end
+
+    it "sets the capacity of the airport to a given value" do
+      bigger_airport = Airport.new(50)
+      expect(bigger_airport.capacity).to eq(50)
```
maryrosecook: Instead of checking the state, this test could check the behaviour.  eg It could check that it's not possible to land 51 planes.

lib/weather_class.rb
```diff
@@ -0,0 +1,18 @@
+class Weather
```
maryrosecook: Watch out for leaving in old code.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def confirm_landed(plane)
+    @plane = plane
+    fail "#{@plane} is not in the airport." unless at_airport?
+    "#{@plane} has landed."
+  end
+
+  private
+  def at_airport?
```
maryrosecook: Really good, clearly named helper methods.

spec/airportCapacity_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    it "allows default capacity to be overridden" do
+      bigger_airport = Airport.new(50)
+      49.times{bigger_airport.instruct_to_land(Plane.new)}
+      expect(bigger_airport.instruct_to_land(Plane.new)).to eq bigger_airport
+      # SHOULD I ACTUALLY TEST FOR LENGTH OF PLANE ARRAY?
```
maryrosecook: What you've done feels better because it tests behaviour, rather than state.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  attr_reader :plane_arr
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @plane_arr = []
+  end
+
+  def instruct_to_land(plane)
```
maryrosecook: Agree that `instruct_to_land` taking plane is more useful.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+    fail "Landing not permitted due to stormy weather." if stormy?
+    @plane_arr << @plane
+    self
+  end
+
+  def take_off(plane)
+    @plane = plane
+    fail "#{@plane} is not in the airport." unless at_airport?
+    fail "Take off not permitted due to stormy weather." if stormy?
+    @plane_arr.delete_if { |p| p == @plane }
```
maryrosecook: This could use just `delete(@plane`.

PR #701, opened 2017-02-19
Airport Challenge - WK1

tvfb85:


lib/airport.rb
```diff
@@ -0,0 +1,64 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  MAX_CAPACITY = 20
+  attr_accessor :capacity
+  attr_reader :planes
```
maryrosecook: As you said in your code review, having :capacity and :planes be accessible does break some of the encapsulation of the airport.  One solution is to think about testing behaviour, rather than state.  eg having tests that check that no more than 20 planes can be landed, rather than checking that the capacity variable is set to 20.

lib/weather.rb
```diff
@@ -0,0 +1,6 @@
+class Weather
+  def randomize(weight = 20)
+    rand <= weight/100
```
maryrosecook: randomise taking a weight is a cool way for it to support testing.  Another option is srand.

lib/plane.rb
```diff
@@ -0,0 +1,14 @@
+class Plane
+  attr_reader :in_flight, :in_airport
+
+  def flying
+    @in_flight = true
+    @in_airport = false
```
maryrosecook: Are both these vars required? Could they be combined into one?

lib/airport.rb
```diff
@@ -0,0 +1,64 @@
...
+    take_off_conditions(plane)
+    plane.flying
+    @planes.delete(plane)
+    confirm_take_off
+  end
+
+
+  private
+
+  def landing_conditions(plane)
```
maryrosecook: Could this method have a more descriptive name? The current name doesn't really indicate what happens if the landing conditions aren't met.

lib/airport.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def take_off(plane)
+    take_off_conditions(plane)
+    plane.flying
+    @planes.delete(plane)
+    confirm_take_off
+  end
+
+
```
maryrosecook: Too much whitespace here.

lib/airport.rb
```diff
@@ -0,0 +1,64 @@
...
+  def take_off_conditions(plane)
+    raise "Plane is already flying" if plane.in_flight
+    raise "Airport is empty" if airport_empty?
+    raise "Too stormy for take off" if stormy?
+    raise "Plane isn't located at this airport" if !@planes.include?(plane)
+  end
+
+  def confirm_landing
+    "Landing successful."
```
maryrosecook: Using strings as status indicators can be a bit brittle.  In Ruby, a lack of errors and nil *not* being returned is usually taken as a sign of success.

lib/airport.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def confirm_take_off
+    "Take off successful."
+  end
+
+  def stormy?
+    Weather.new.randomize
+  end
+
+  def airport_full?
```
maryrosecook: Great, clear, handy helper method.

lib/plane.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
+  def landed
```
maryrosecook: Method names should almost always be verbs (unless they return data).  eg `land` rather than `landed`.

spec/airport_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+
+    context 'with parameters' do
+      let(:airport) { Airport.new(45)}
+      it "should allow the airport capacity to be overridden when necessary" do
+        expect(airport.capacity).to eq(45)
+      end
+      it "should raise an error when the new capacity is exceeded" do
+        airport.capacity.times { airport.land(Plane.new) }
+        expect { airport.land(Plane.new) }.to raise_error("Airport is full")
+      end
```
maryrosecook: It's idiomatic to have a line of whitespace between tests.

PR #705, opened 2017-02-19
Airport Weekend Challenge Complete

JohnChangUK:


lib/airport.rb
```diff
@@ -0,0 +1,49 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  attr_reader :planes, :capacity, :weather
+
+  CAPACITY = 15
+
+  def initialize(capacity = CAPACITY)
+  @capacity = capacity
```
maryrosecook: Could the indentation be more consistent with Ruby idioms and the surrounding code?

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def confirm_boarding(plane)
+    raise 'This plane has boarded' if @planes.include?(plane)
+    "#{plane} has departed to its destination"
+  end
+
+  def confirm_landing(plane)
+    raise 'This plane is not at the airport' if !@planes.include?(plane)
+    "#{plane} has landed at its destination"
```
maryrosecook: Are these status strings necessary? Would it be possible for the tests to check the behaviour? (eg a test could land a plane and then confirm that it has landed by checking that it can take off again)

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+  def landing(plane)
+    raise 'No more room for planes' if airport_full?
+    raise 'Stormy weather, cannot land' if @stormy
+      plane.landed
+      @planes << plane
+      "The #{plane} has landed!"
+  end
+
+  def boarding
+    raise 'The airport has no more planes' if @planes.empty?
```
maryrosecook: Could the condition here use a helper method eg `raise 'The airport has no more planes' if airport_empty?`

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def landing(plane)
+    raise 'No more room for planes' if airport_full?
+    raise 'Stormy weather, cannot land' if @stormy
+      plane.landed
+      @planes << plane
+      "The #{plane} has landed!"
+  end
+
+  def boarding
```
maryrosecook: It feels like `boarding` doesn't quite convey the idea that the plan is actually taking off.

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+
+class Airport
+  attr_reader :planes, :capacity, :weather
+
+  CAPACITY = 15
+
+  def initialize(capacity = CAPACITY)
+  @capacity = capacity
+  @planes = []
+  @stormy = Weather.new.storm? == :stormy
```
maryrosecook: Could the Weather instance be injected so the Airport knows less about it?

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+      "The #{plane} has landed!"
+  end
+
+  def boarding
+    raise 'The airport has no more planes' if @planes.empty?
+    raise 'Sorry, weather too stormy' if @stormy
+      boarding_plane = @planes.pop
+      boarding_plane.flight
+      @planes.push(boarding_plane)
+      @planes.pop
```
maryrosecook: Why does the plane get popped and pushed and popped again?

spec/plane_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'plane'
+require 'airport'
+
+describe Plane do
+  subject(:plane) { described_class.new }
+  subject(:plane1) { described_class.new }
+  subject(:plane2) { described_class.new }
+
+it 'tests for a Plane class' do
+  expect(plane.class).to eq(Plane)
```
maryrosecook: Could this test check that the plane was created without testing the plane object's class? Ruby code usually avoids functioning based on an object's class.  This makes it more malleable and supports duck typing.  (Happy to talk about this in more detail.)

lib/plane.rb
```diff
@@ -0,0 +1,23 @@
...
+    @flying = false
+  end
+
+  def flight
+    @flying = true
+  end
+
+end
+
+  class Concorde < Plane
```
maryrosecook: It feels like Concorde wasn't in the user stories...

PR #740, opened 2017-04-14
Focus down the weekend challenge & remove likely config troubles

neoeno:
I wonder what happens if I PR this!


spec/spec_helper.rb
```diff
@@ -1,2 +1,17 @@
...
+  SimpleCov::Formatter::Console,
+  # Want a nice code coverage website? Uncomment this next line!
+  # SimpleCov::Formatter::HTMLFormatter
+)
+SimpleCov.start
+
+RSpec.configure do |config|
+  config.after(:suite) do
+    puts
+    puts "\e[33mHave you considered running rubocop? It will help you improve your code!\e[0m"
```
neoeno: Shut up you're useless.

neoeno: Actually I feel bad having said that now sorry hound

PR #741, opened 2017-04-15
Weekend challenge submission

adamjohnsnow:


spec/weather_spec.rb
```diff
@@ -0,0 +1,10 @@
+require './lib/weather'
+
+describe Weather do
+  let(:airport) { double :airport, name: 'the Airport' }
+  it '#weather can be #sunny or #stormy' do
+    weather = Weather.new(airport)
+    expect(weather.condition).to eq("Sunny").or(eq("Stormy"))
```
neoeno: This would pass if it always returned `"Sunny"`, right?

spec/plane_spec.rb
```diff
@@ -0,0 +1,33 @@
+require './lib/plane'
+
+describe Plane do
+
+  context 'when airport is sunny' do
+    let(:airport) { double :airport, name: "Test", full?: false, runway: [], land_plane: '', takeoff_plane: '' }
```
neoeno: Two of these attributes you've given to `double` aren't required — can you tell me which and why?

spec/plane_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    let(:weather) { double :weather, condition: "Sunny", check_weather: "Sunny"}
```
neoeno: And one of these attributes — but perhaps that's an easier question!

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    it "plane cannot #land_at a #full #airport" do
+      subject.runway = ["a", "a", "a", "a", "a", "a", "a", "a", "a", "a"]
+      expect { subject.landing_check(plane) }.to raise_error "Cannot land, airport full"
+    end
+
+  end
+
+  context 'when #airport is #stormy' do
```
neoeno: These `#` have a specific meaning in RSpec — to refer to instance methods — it's clever the way you've extended them here but it probably decreases readability sadly!

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    it "has a #runway" do
+      expect(subject.runway.class).to eq(Array)
+    end
+
+    it "can be #full" do
+      expect(subject).to respond_to(:full?)
+    end
+
+    it "recognised #runway with 10 as #full" do
+      subject.runway = ["a", "a", "a", "a", "a", "a", "a", "a", "a", "a"]
```
neoeno: Maybe give `["a"] * 10` a go :)

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+  attr_accessor  :current_weather, :name, :runway, :capacity, :landed_at
```
neoeno: Would it be possible to make some or all of these private? Any reason why we'd do that?

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def takeoff_check(plane)
+    @current_weather = generate_weather
+    fail "Cannot takeoff, weather at #{name} stormy" if @current_weather.condition == "Stormy"
+  end
+
+  def landing_check(plane)
+    @current_weather = generate_weather
+    fail "Cannot land, weather at #{name} stormy" if @current_weather.condition == "Stormy"
+    fail "Already landed at #{name}" unless plane.landed_at.nil?
```
neoeno: Here, `Airport` looks at one of `Plane`'s instance variables to see if it's nil. This feels like maybe something `Plane` should be doing.

spec/weather_spec.rb
adamjohnsnow: Yes, this was my attempt at checking that one of the two conditions were always returned, yet could still be random.

spec/plane_spec.rb
adamjohnsnow: Ah yes. I think I was just trying to replicate Airport as much as possible in desperation.

spec/airport_spec.rb
adamjohnsnow: Well I'll be damned!!

spec/airport_spec.rb
adamjohnsnow: Ah - I thought they referred to variables too.

lib/airport.rb
adamjohnsnow: Gotcha - my understanding of SRP and delegation is much better this weekend... I hope.
Thanks for your time on this, I hope you are able to have a look at this weekend's challenge and that I have shown some improvement.
Thanks

PR #742, opened 2017-04-15
Weekend 1 - Airport Challenge

JayWebDevCom:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,4 +9,8 @@ gem 'rubocop'
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-console', require: false, group: :test
 
+group :test do
+  gem 'capybara', '~> 2.4.3'
+end
+
 rvm rvmrc warning ignore
--- a/Gemfile
+++ b/Gemfile
@@ -8,3 +8,5 @@ gem 'rspec'
 gem 'rubocop'
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-console', require: false, group: :test
+
+rvm rvmrc warning ignore
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -7,19 +7,19 @@ class Plane
   end
 
   def land(weather, airport)
-    raise "Cannot land in #{weather.status} weather" if stormy(weather)
-    raise "I have already landed" if status == 'landed'
+    raise "Cannot land in stormy weather" if stormy(weather)
+    raise "I have already landed" if status == :landed
     raise "Airport is at capacity" if airport.full?
     airport.receive(self)
-    @status = 'landed'
+    @status = :landed
   end
 
   def take_off(weather, airport)
-    raise "Cannot take off in #{weather.status} weather" if stormy(weather)
+    raise "Cannot take off in stormy weather" if stormy(weather)
     raise "I am already in the air" if flying?
     raise "I am not in that airport" unless airport.planes.include?(self)
     airport.release(self)
-    @status = 'in air'
+    @status = :in_air
   end
 
   def confirm_status
@@ -27,10 +27,10 @@ class Plane
   end
 
   def flying?
-    @status == 'in air' ? true : false
+    confirm_status == :in_air
   end
 
   def stormy(weather)
-    weather.status == 'stormy' ? true : false
+    weather.status == :stormy
   end
 end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -2,7 +2,7 @@ class Plane
 
   attr_reader :status
 
-  def initialize(status = 'on ground')
+  def initialize(status = :on_ground)
     @status = status
   end
 
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,21 @@
+class Plane
+
+  def initialize(status = 'on ground')
+    @satus = status
+  end
+
+  def land(airport)
+    airport.planes << self
+    @status = 'landed'
+  end
+
+  def take_off(weather)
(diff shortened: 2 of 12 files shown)


Gemfile
```diff
@@ -8,3 +8,5 @@ gem 'rspec'
 gem 'rubocop'
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-console', require: false, group: :test
+
+rvm rvmrc warning ignore
```
neoeno: I think I know where you got this from — but I suspect this is the wrong place for the change. It stops `bundle install` working which is a good clue :)

lib/plane.rb
```diff
@@ -0,0 +1,39 @@
+class Plane
+
+  attr_reader :status
+
+  def initialize(status = :on_ground)
```
neoeno: It's interesting that you've got 3 states here (`:on_ground`, `:landed`, `:in_air`). Is this an example of a semantic difference which doesn't add value in the program? Or is there a good reason to have all three?

The danger is that it degrades readability, in that it's harder to understand how the program is supposed to work by reading it. A reader might ask: What impact does `:on_ground` have? I can't find any but maybe I'm missing something somewhere? Is `:on_ground` a subtype of `:landed`? If not, what's the difference?

lib/plane.rb
```diff
@@ -0,0 +1,39 @@
...
+    @status = status
+  end
+
+  def land(weather, airport)
```
neoeno: Good job on the dependency injection here! To take it to the next level, I'd consider adding `weather` in the constructor and then storing it in an instance variable. Partly convention, partly it reduces the number of arguments your methods have, which can make it a bit easier to follow.

spec/rspecfeaturetest_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+#
+#   def a_plane_can_take_off_from_that_airport
+#   	expect(klm.takeoff(weather, jfk)).to eq :in_air
+#   end
+#
+#   def a_plane_can_report_its_correct_takeoff_status
+#   	expect(klm.confirm_status_.to eq :in_air
+#   end
+#
+# end
```
neoeno: Commented out code is a big no-no I'm afraid! I'd suggest looking into `git stash`, or keeping in mind that if you delete code git you previously committed, it will keep it around in the commit history.

Nice attempt though — what went wrong here?

spec/weather_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+describe Weather do
+
+  subject(:weather) { described_class.new }
+
+  it 'responds to status' do
+    expect(weather).to respond_to :status
+  end
+
+  it 'varies randomly' do
+    allow(weather).to receive(:status).and_return(:sunny, :stormy, :sunny)
```
neoeno: Nice!

spec/plane_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  subject(:plane) { described_class.new }
+  let(:weather) { double :weather }
+  let(:airport) { double :airport }
+
+  describe "#land" do
+
+    before do
+      allow(airport).to receive_messages(:planes => [], :full? => false, :receive => [])
+      allow(weather).to receive_messages(:status => :sunny)
```
neoeno: Did you discover the `double` syntax where you can remove this line and replace line 7 with:

```
let(:airport) { double :airport, status: :sunny }
```

You could then move the let blocks to the specific `describe` or `context` blocks where you need them (just like you've got the `before` blocks here to only run in their right place)

JayWebDevCom:
Dear Kay, 

Thank you very for reviewing my Airport Challenge.
Your advice has been tremendously helpful.

Commit messages
Thanks for the link! 
I will be sure to make my commits descriptive, communicating more context about the changes that each one makes and why which relates to future maintainability. 
My take away from the link - 
Use consistent style and content across the team.
Sometimes more detailed commits can be made in a proper text editor.
Follow the seven rules and - 'if applied this commit will “..."'

Gemfile
I will be more careful with my gem file and what I write in it so as not to disrupt bundle.

Plane.rb and the 3 statuses - :on_ground, :landed, :in_air
Using `:on_ground` allows me to see that a plane is brand new and not in an airport.
This strategy was intended to help me ensure that a plane had to land at an airport before it could do anything else.
I wanted to be able to use `plane.confirm_status` to differentiate between a plane that had just been manufactured/initialized and a plane that was in an airport hangar as a result of having the land method called on it.
It certainly is not adding as much value as I intended you are right.

Feature_Spec.rb
My feature_spec needed debugging. I had an error - undefined method `feature` nameerror.
I have fixed this now by adding capybara to the gemfile and some RSpec.configure information to spec_helper.rb.
All tests run well now - except that my feature test raises a “weather is stormy exception” 25% of the time.
I also had some scope resolution errors in the featuretest file.
I will be sure not to commit any commented out files again. Noted!

Doubles
Thanks for the improved double-syntax. I can now avoid latter allow method calls.
I will definitely implement this, I hadn’t seen it before until your skills workshop/class code review.

Dependency Injection
I didn’t want an airport to be injected with weather at airport initialization. I thought that this might mean that an airport would therefore only ever have one weather condition.
I thought that I could separately generate several weather objects and these fixed-status weather objects could be passed to airports or landing method calls randomly.
I can see now that perhaps generating a weather object at a `land` or `take_off` method call might make more sense and mean that I can reduce the number of parameters that these methods need.
Next time I will add weather in the constructor and then store it in an instance variable and read its status from there.

Thanks again. Have a great weekend

Jaiye

PR #744, opened 2017-04-16
Weekend 1 - Airport Challenge

TimRobinson1:
Submitting my solution to the first weekend challenge.

It is able to land and take off planes from airports, the success of which depends on the variable weather conditions.  It also covers various edge cases such as invalid planes and attempting to set airport capacity to a negative value.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -6,10 +6,11 @@ class Airport
 
   DEFAULT_CAPACITY = 50
 
-  def initialize(weather = Weather.new)
+  def initialize(capacity = DEFAULT_CAPACITY)
+    fail 'invalid capacity' if capacity <= 0
     @planes = []
-    @weather = weather
-    @capacity = DEFAULT_CAPACITY
+    @weather = Weather.new
+    @capacity = capacity
   end
 
   def land(plane)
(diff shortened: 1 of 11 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,77 @@
...
+    capacity <= 0
+  end
+
+  def excess_planes?(capacity)
+    @planes.length > capacity
+  end
+
+  def invalid_plane?(plane)
+    !plane.instance_of?(Plane)
+  end
```
neoeno: I notice you changed this from testing it wasn't some other types. A good change! And an interesting thought in terms of an edge case to guard against. However, usually in Ruby we wouldn't write these sorts of checks, instead we let our program throw errors if someone passes in the wrong object (because it won't have the methods we need it to have). Compare this to the idea of duck typing we discussed earlier.

spec/weather_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  it { is_expected.to respond_to :stormy? }
+
+  it 'is more likely to be sunny than stormy' do
+    forecast = Weather::CONDITIONS
+    expect(forecast.count(:clear) > forecast.count(:storm)).to eq true
+  end
+
+  describe '#stormy?' do
+    it 'returns a boolean value' do
+      expect(weather.stormy?).to eq(true).or(eq(false))
```
neoeno: Can you think of a way to test that this doesn't always return true? As it is, I can change the code to this:

```ruby
class Weather

  CONDITIONS = [:storm, :clear, :clear, :clear, :clear, :clear]

  def stormy?
    true
  end
end
```

and the test still passes. A good test it one that truly tells us whether the method does what it is supposed to do. With randomness this is a bit harder, but it's possible to get close.

PR #745, opened 2017-04-16
Airport Challenge

sophieklm:


lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+
+class Airport
+
+  DEFAULT_CAPACITY = 50
+  DEFAULT_PLANES = 5
+
+  include Weather
+
+  attr_reader :planes
+  attr_accessor :capacity
```
ddosque: try to make your attr_readers and writers private.

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def take_off
+    fail "It is too stormy to fly" if stormy?
+    fail "There are no planes" if @planes.empty?
+    @planes[-1].fly
+    @planes.pop
+  end
+
+  def new_day
+    @stormy = generate_weather
```
ddosque: In this case, you `include` Weather, so you could just call generate_weather, which actually sets `@stormy`. new_day seems like duplication.

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def initialize(default = DEFAULT_CAPACITY, start_planes = DEFAULT_PLANES)
+    @planes = []
+    start_planes.times {@planes << Plane.new}
+    @capacity = default
+    @stormy = generate_weather
+  end
+
+  def land(plane)
+    fail "This plane has already landed" if plane.landed?
```
ddosque: This guard clause relies solely to the internal state of the plane, so it is the responsiblity of the plane class.

spec/weather_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'weather'
+
```
ddosque: As you used a mixin module for your weather, you qould ideally use shared examples to test the functionality of your module.

PR #746, opened 2017-04-16
Airport challenge

sulaimancode:


lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+class Weather
+
+  def condition(condition = { 1 => :sunny, 2 => :sunny, 3 => :sunny, 4 => :stormy })
```
neoeno: We might extract this to a constant, e.g.

```ruby
  CONDITIONS = { 1 => :sunny, 2 => :sunny, 3 => :sunny, 4 => :stormy }

  def condition
    @condition = CONDITIONS[rand(1..4)]
  end
```

You can actually reduce that even further, but I'll leave that for you to think about.

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+  end
+
+  describe 'weather condition' do
+
+    it 'should prevent landing when weather is stormy' do
+      allow(weather).to receive(:condition) { :stormy }
+      expect { subject.land(plane) }.to raise_error 'can not land, weather condition stormy'
+    end
```
neoeno: It might be interesting to think about `context` blocks here. Check out this for an example: http://betterspecs.org/#contexts

spec/plane_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'plane'
+
+describe Plane do
+  it '' do
+
+  end
```
neoeno: Probably worth getting rid of this or replacing it with something tokenistic like:

```ruby
it { is_expected.to be }
```

(http://www.relishapp.com/rspec/rspec-expectations/v/3-5/docs/built-in-matchers/be-matchers#be-matcher)

spec/weather_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'weather'
+
+describe Weather do
+
+  describe '#condition' do
+    it { is_expected.to respond_to :condition }
+
+    it 'should set weather condition depending on rand number' do
```
neoeno: Have a read of this: http://betterspecs.org/#should

Hint: you could word this as:

```ruby
it 'sets weather condition depending on rand number'
```

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  attr_reader :planes, :weather
+  attr_accessor :capacity
```
neoeno: Do all of these `attr_`s need to be public?

PR #752, opened 2017-04-16
Airport challenge

baree99:


lib/weather.rb
```diff
@@ -0,0 +1,18 @@
+# Understand how to create stormy weather randomly
```
neoeno: Now I remember! It was you who wrote the understands statements that I really liked!

lib/weather.rb
```diff
@@ -0,0 +1,18 @@
...
+
+module Storm
```
neoeno: I found this storm module idea really interesting — how do you think it worked out here?

lib/weather.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def chance_of_storm
+    rand(9)
+  end
+
+  def stormy?
+    chance_of_storm == 9
+  end
```
neoeno: I'd consider making `chance_of_storm` private, e.g.

```ruby
module Storm
  def stormy?
    chance_of_storm == 9
  end

  private

  def chance_of_storm
    rand(9)
  end
end
```

The idea is to really restrict how the rest of the program can interact with your class — a method that you know for certain is only being called in this class (because it's impossible to call it anywhere else) is much easier to change than a method that might be called elsewhere. Let's say you have to rename it — you only have to look in this file for the lines to change!

Bonus: you don't have to test private methods!

spec/airport_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        allow(plane).to receive(:location) { subject.name }
+        allow(subject).to receive(:stormy?) { true }
+        message = "Can't take off due to stormy weather"
+        expect { subject.take_off(plane) }.to raise_error message
+      end
+    end
+
+    describe "#change_capacity" do
+      it "changes the capacity of the airport" do
+        subject.capacity(5)
```
neoeno: It's a real shame when you changed it that it broke the tests! It was all green before! Make sure to always run RSpec before you commit.

spec/weather_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'weather'
+
+describe Weather do
+  describe "#chance_of_storm" do
+    it "returns a random number between 0 & 9" do
+      expect(subject.chance_of_storm <= 9).to eq subject.chance_of_storm >= 0
```
neoeno: This test is a little hard to figure out — is there a more straightforward way to put it?

lib/airport.rb
```diff
@@ -0,0 +1,57 @@
...
+class Airport
+
+  include Storm
+
+  DEFAULT_CAPACITY = 10
+
+  attr_reader :name, :planes_in_airport
+  attr_accessor :capacity
+
+  def initialize(name = "A#" + "#{self}"[-5..-2])
```
neoeno: This plane naming code is pretty neat! Looks like a real plane number!

spec/airport_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      expect { subject.land(plane) }.to output(text).to_stdout
+    end
+
+    it "lands a plane and assigns it to the airport" do
+      subject.land(plane)
+      expect(subject.planes_in_airport).to include plane.name
+    end
+  end
+
+    describe "#take_off" do
```
neoeno: Got some inconsistent indentation here! Certainly not the worst I've seen but easy to fix. Maybe give rubocop a run to show you all the inconsistent lines.

lib/airport.rb
```diff
@@ -0,0 +1,57 @@
...
+    plane.taking_off
+    print "The #{plane.name} took off succesfully"
+  end
+
+  def landing_criteria(plane)
+    fail "#{plane.name} already landed in this airport" if @planes_in_airport.include? plane.name
+    fail "#{plane.name} is not in the air" if plane.location != :up_in_the_air
+    fail "Can't land as the airport is full" if full?
+    fail "Can't land due to stormy weather" if stormy?
+  end
```
neoeno: Great method! Collecting all the pre-landing checks into one place makes it easy to see them all at once.

I'd consider the naming: one rule of thumb is that a method should be an 'action' — in this case I'd consider `check_landing_criteria`. That serves the really useful purpose of separating it from attributes like `capacity` — you can get an idea of whether it's a method or an attribute just via the name.

PR #758, opened 2017-04-17
Final pull request, pretty much there.

motri:


lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+  def take_off(plane)
+    fail "No taking off due to weather" if storms?
+    fail "The plane has already taken off" if plane.on_air?
+    fail "Plane not in this aiport" unless ground_fleet.include?(plane)
+    plane.flying
+    ground_fleet.delete(plane)
+  end
+
+  def in_airport?(plane)
+    ground_fleet.include?(plane) ? true : false
```
neoeno: Try removing the `? true : false` — do the tests still pass? If so why?

lib/plane.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'airport.rb'
+class Plane
+
+  def initialize
+    @status = 0
```
neoeno: I'd consider using symbols here instead of integers, compare:

```
def initialize
  @status = :on_ground
end

def landed?
  @status == :on_ground
end
```

It's easier to see at a glance what's going on, rather than `1` and `0` which the reader has to try to figure out what they mean by looking at the class and seeing how they're used.

spec/weather_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'weather'
+
+describe Weather do
+  it { is_expected.to respond_to :stormy? }
+
+  it 'should check if stormy' do
+    expect(subject.stormy?).to eq(true).or(eq(false))
```
neoeno: The next two tests cover this one I think, and do so better — great to start with this one! But I'd delete it after you've written the others.

spec/airport_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+      end
+    end
+
+  end
+
+  it { is_expected.to respond_to :take_off }
+
+  describe '#take_off' do
+
+    context 'when sunny' do
```
neoeno: Great use of context blocks here!

PR #762, opened 2017-04-17
Solution to week one's weekend challenge.

adc17:


lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+require './lib/plane'
+
+class Airport
+
+  attr_reader :planes, :climate
+  attr_accessor :capacity
+  DEFAULT_CAPACITY = 5
+
+  def initialize(climate = Climate.new)
+    raise 'Error: Parameter is not a climate.' unless climate.is_a?(Climate)
```
ddosque: avoid checking type in loosely typed languages - you prevent the evolution of duck types.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+require './lib/climate'
+require './lib/plane'
+
+class Airport
+
+  attr_reader :planes, :climate
+  attr_accessor :capacity
```
ddosque: no attr_accessor / reader / writer is required for this implementation. Avoid using these unless they are private.

lib/plane.rb
```diff
@@ -0,0 +1,36 @@
...
+    confirm_landing
+  end
+
+  def take_off
+    take_off_process
+    confirm_take_off
+  end
+
+  def confirm_landing
+    landed ? "Plane has landed." : "Plane has not landed."
```
ddosque: Avoid mixing presentation logic in your 'business' code.

spec/airport_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+      it 'raises error' do
+        expect { Airport.new(@not_a_climate) }.to raise_error 'Error: Parameter is not a climate.'
+      end
+    end
+  end
+
+  describe '#build_plane' do
+    it 'creates new plane in airport\'s dock' do
+      expect(airport.planes).to be_empty
+      airport.build_plane
```
ddosque: I'm pretty sure airports don't build planes. Check the brief.

spec/airport_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+      expect(airport.planes.count).to eq 1
+      expect(airport.planes[0]).to be_an_instance_of(Plane)
+    end
+
+    it 'takes an optional argument that specifies a custom plane' do
+      special_plane = double(:plane, :is_a? => true)
```
ddosque: custom planes are not mentioned in the brief, don't stray from it - you would be wasting your customer's time.

spec/airport_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+        expect { airport.build_plane(@non_plane) }.to raise_error "Error: The airport can only build planes."
+      end
+    end
+        
+  end
+
+  describe '#conditions' do
+    before(:example) do
+      @sunny_climate = double(:climate, :is_a? => true, :conditions => :sunny)
+      @airport = Airport.new(@sunny_climate)
```
ddosque: nice use of injection. If you weren't checking for type, you could have had a simpler double.

adc17:
Thanks @ddosque . For the record, I wasn't trying to be fancy—these are just bad solutions to the brief that I wish I could have done more simply. e.g. airports building planes—I couldn't think of another way to generate a plane inside an airport when weather was stormy, and I needed one inside to check whether it was prevented from leaving.

PR #764, opened 2017-04-17
First draft of airport_challenge

chrisjmit:


spec/airport_spec.rb
```diff
@@ -0,0 +1,68 @@
+require './lib/airport.rb'
+
+describe Airport do
+  subject(:airport) { described_class.new(25) }
+  let(:plane) { Plane.new }
+  let(:weather) { Weather.new }
+
+  context "A plane can land" do
+    it '#finds the plane that has been landed' do
+      allow(subject).to receive(:storm_outside?).and_return false
```
ddosque: stubbing the subject is generally a bad idea - you are no longer testing a true example of an Airport. In this case, you would be better to stub the Weather.

lib/plane.rb
```diff
@@ -0,0 +1,15 @@
...
+    @flying = true
+  end
+
+  def land_plane
+    @flying = false
+  end
+
+  def plane_take_off
+    @flying = true
+  end
```
ddosque: method naming - try not to repeat the class name in the method name, it prevents duck typing (what if you later have a helicopter that can also land and take off?)

PR #765, opened 2017-04-17
Airport_challenge 

danielemanuel:


lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :planes, :weather, :capacity
+
+   def initialize(capacity = DEFAULT_CAPACITY)
+     @capacity = capacity
+     @planes = []
+     @weather = weather
```
ddosque: you never actually use the weather object, as you are not injecting it at the top:
```
def initialize(weather = Weather.new, capacity = DEFAULT_CAPACITY)
  @capacity = capacity
  @planes = []
  @weather = weather
end
```

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  #  private
+
+    def full?
+      @planes.length >= @capacity
+    end
+
+    def stormy_weather?
+    rand(1..10) == 5 ? true : false
+    end
```
ddosque: Ah, I see you are using this instead of using your weather class. Would be better to use the separate object, as that wraps up and isolates that untestable portion of code.

PR #766, opened 2017-04-17
JessicaBarclay

JessicaBarclay:


spec/weather_spec.rb
```diff
@@ -0,0 +1,37 @@
+require './lib/weather'
+
+describe Weather do
+  it { is_expected.to be_an_instance_of(Weather) }
+  it { is_expected.to respond_to(:random_condition) }
+  it { is_expected.to respond_to(:stormy?) }
+  describe '#random_condition' do
+    it 'should return :sunny from CONDITIONS' do
+      allow(subject).to receive(:random_condition) { :sunny }
+      expect(subject.random_condition).to eq :sunny
```
ddosque: you are just testing your stub in these tests. I would not try to test the random nature of rand - it is a ruby kernel method, you can trust that it works.

PR #767, opened 2017-04-17
Ana's airport challenge

anaalta:


lib/airport.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+DEFAULT_CAPACITY = 20
+
+attr_accessor :capacity, :planes
```
ddosque: avoid public attr_accessor

lib/plane.rb
```diff
@@ -0,0 +1,13 @@
+require_relative 'airport'
+require_relative 'weather'
+
+class Plane
+  #def land
```
ddosque: shame this wasn't finished, looks like you are on the right tracks!

lib/weather.rb
```diff
@@ -0,0 +1,15 @@
...
+  def initialize
+  @weather = Random.rand(1..100)
+  end
+
+  def stormy?
+    if @weather<20
+      true
+    else
+      false
+    end
```
ddosque: could be rewritten:
```ruby
def stormy?
  @weather < 20
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  it { is_expected.to respond_to :stormy_weather?}
+  it { is_expected.to respond_to(:confirm_landing).with(1).argument}
+  it { is_expected.to respond_to :confirm_takeoff}
+
+  describe '#confirm_landing' do
+
+     context 'if weather is fine' do
+        it "should confirm landing by adding a new plane in the aiport" do
+        plane = Plane.new
```
ddosque: use doubles instead of real objects

PR #768, opened 2017-04-17
Airport Challenge pull request

DSeanGray:


lib/airport.rb
```diff
@@ -0,0 +1,24 @@
...
+    @planes << plane
+  end
+
+  def confirm_take_off
+    fail 'Cannot takeoff in stormy weather' if weather_status?
+    @planes.pop
+  end
+
+  def weather_status?
+    weather = Weather.new
```
ddosque: very nice! This is a very simple and effective way to isolate the random behaviour and separate the responsibilities. You could also have used dependency injection to allow the weather class to be changed, but your solution is excellent.

lib/plane.rb
```diff
@@ -0,0 +1,11 @@
+class Plane
+
+  def land
+    true
```
ddosque: why hard coded?

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
+
+ attr_accessor :stormy
```
ddosque: unnecessary accessor

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+require 'weather'
+
+ describe Airport do
+
+  it { is_expected.to respond_to :confirm_landed}
+  it { is_expected.to respond_to :confirm_take_off}
+  it { is_expected.to respond_to :weather_status? }
+
+  it "confirms landing" do
+    plane = Plane.new
```
ddosque: use doubles instead of a real plane

spec/plane_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'plane'
+
+describe Plane do
+  it { is_expected.to respond_to :land }
+  it { is_expected.to respond_to :take_off }
```
ddosque: not really testing anything here

PR #769, opened 2017-04-17
airport_challenge

milesillsley:


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+require './lib/plane'
+require './lib/weather'
+
+class Airport
+  DEFAULT_CAPACITY = 100
+  attr_accessor :capacity, :planes
```
ddosque: avoid public attr_accessors

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    plane.land
+    "Plane has landed."
+  end
+
+  def takeoff(plane)
+    fail "Plane is already flying." if plane.flying
+    fail "Too stormy for takeoff." if @weather
+    plane.takeoff
+    @planes.delete(plane)
+    "Plane has taken off."
```
ddosque: don't return presentation strings in business logic

spec/airport_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  it { is_expected.to respond_to(:land).with(1).argument }
+
+  it "Instructs a plane to land" do
+    expect(subject.land(plane)).to eq("Plane has landed.")
+  end
+
+  it "Won't land a plane that has already landed" do
+    allow(plane).to receive(:flying).and_return(false)
+    expect { subject.land(plane) }.to raise_error("Plane has aleady landed.")
```
ddosque: this error depends on internal state of the plane, not the airport. Guard clause (and test) is responsibility of plane not airport.

spec/airport_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'airport'
+
+describe Airport do
+  let(:weather) { double :weather, stormy?: false }
+  let(:plane) { instance_double Plane, :flying => true }
+  before { allow(plane).to receive(:land) }
+  before { allow(plane).to receive(:takeoff) }
+  subject { Airport.new weather }
```
ddosque: great use of dependency injection to mock the weather

PR #806, opened 2017-05-15
98.28% coverage even if the line indicated as not tested are inside private methods...

y0m0:


y0m0:
Adds feature test using capybara gem and runs 'A Plane lands into an airport' scenario

PR #808, opened 2017-06-11
Complete Airport Challenge

tbscanlon:


lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  DEFAULT_CAPACITY = 20
+
+  def initialize(args = { capacity: DEFAULT_CAPACITY })
+    @gates    = []
+    @weather  = Weather.new
+    @capacity = args.fetch(:capacity)
+  end
+
+  def instruct(args)
```
maryrosecook: (From group review.) The ability to pass an action feels like preparing for a specific future (not ideal), rather than avoiding making the future harder to adapt to (good).

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    raise WeatherError, "Inclement weather" unless check_weather
+    action, plane = args.fetch(:action), args.fetch(:plane)
+
+    action == "land" ? land(plane) : take_off(plane)
+  end
+
+  def report(action, plane)
```
maryrosecook: Is it possible to replace this case statement with a polymorphism?

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    puts case action
+         when "land" then "#{plane} has landed successfully."
+         when "take off" then "#{plane} has taken off successfully."
+         end
+  end
+
+  private
+  def check_weather
```
maryrosecook: It feels like the `check_weather` name is less intention revealing than `weather.good?`.  Is it needed?

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    weather.good?
+  end
+
+  def full?
```
maryrosecook: Really nice helper method!

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def report(action, plane)
+    puts case action
+         when "land" then "#{plane} has landed successfully."
+         when "take off" then "#{plane} has taken off successfully."
+         end
+  end
+
+  private
```
maryrosecook: Great job on encapsulating these implementation details.

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def full?
+    gates.count >= capacity
+  end
+
+  def land(plane)
+    raise PlaneError, "Plane not flying" if plane.grounded?
+    raise AirportError, "Airport is full" if full?
+    gates << plane
+    plane.grounded = true
```
maryrosecook: It feels like this is breaking the encapsulation of plane.  You've made `grounded?` already (a nice, encapsulating method) - you just need to use it!

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def take_off(plane)
+    raise PlaneError, "Plane already flying" unless plane.grounded?
+    leave_gate(plane)
+    plane.grounded = false
+    report("take off", plane)
+  end
+
+  def leave_gate(plane)
+    search = Proc.new { |ele| ele == plane }
```
maryrosecook: No need to pull out this Proc.  It's clearer to have the code in context.  Also, there's a Ruby builtin for selecting an element from an array.

lib/error.rb
```diff
@@ -0,0 +1,8 @@
+class WeatherError < StandardError
```
maryrosecook: No need to make special error classes for each class.  They don't add much information because knowing you have a weather error isn't super informative.  Error classes tend to exist to communicate what process has gone wrong, rather than what area of the code has the problem.  E.g. {}.fetch("nonexistent key") raises a KeyError, not a HashError.

spec/airport_spec.rb
```diff
@@ -0,0 +1,109 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { double("plane") }
```
maryrosecook: Nice doubling!

PR #812, opened 2017-06-11
Airport challenge attempt

terminalobject:
.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,13 @@
+class Airport 
+
+	attr_reader :planes
+
+	def initialize
+		@planes = []
+	end 
+
+	def land(plane)
+		@planes << plane
+  end 
+
+end 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,11 +1,11 @@
-require 'weather'
+require './lib/weather'
 
 class Airport 
 
 	DEFAULT_CAPACITY = 50
 
-	def initialize(capacity = DEFAULT_CAPACITY)
-		@planes = []
+	def initialize(capacity = DEFAULT_CAPACITY, planes = [])
+		@planes = planes
 		@weather = Weather.new
 		@capacity = capacity
 	end 
@@ -25,12 +25,13 @@ class Airport
   def release(plane)
   	raise "Plane already flying!" if plane.status == "Flying"
   	raise "Stormy conditions!" if stormy?
+  	raise "Plane not in the airport!" unless @planes.include?(plane)
   	@planes.delete(plane)
   	plane.take_off
   end 
 
 	def full?
-		planes.count >= 50
+		planes.count >= DEFAULT_CAPACITY
 	end 
 attr_accessor :planes, :capacity, :weather
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,13 +1,37 @@
+require 'weather'
+
 class Airport 
 
-	attr_reader :planes
+	DEFAULT_CAPACITY = 50
 
-	def initialize
+	def initialize(capacity = DEFAULT_CAPACITY)
 		@planes = []
+		@weather = Weather.new
+		@capacity = capacity
 	end 
 
-	def land(plane)
+	def stormy?
+		@weather == "Stormy"
+	end
+
+	def receive(plane)
+		raise "Plane already landed!" if plane.status == "Landed"
+		raise "The airport is full!" if full?
+		raise "Stormy conditions!" if stormy?
 		@planes << plane
+		plane.land
   end 
+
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+require './lib/weather'
+
```
maryrosecook: Make sure to Indent your code using Ruby idioms.  Here's a Ruby style guide with instructions on what the Ruby idioms are.

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  	raise "Stormy conditions!" if stormy?
+  	raise "Plane not in the airport!" unless @planes.include?(plane)
+  	@planes.delete(plane)
+  	plane.take_off
+  end 
+
+	def full?
+		planes.count >= DEFAULT_CAPACITY
+	end 
+attr_accessor :planes, :capacity, :weather
```
maryrosecook: Make sure to lay out your code according to Ruby idioms.  Accessors usually go at the top of the class, unless they are private.  The style guide (linked in the previous comment) will have the idioms for things like this.

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+	def receive(plane)
+		raise "Plane already landed!" if plane.status == "Landed"
+		raise "The airport is full!" if full?
+		raise "Stormy conditions!" if stormy?
+		@planes << plane
+		plane.land
+  end 
+
+  def release(plane)
+  	raise "Plane already flying!" if plane.status == "Flying"
```
maryrosecook: Could the condition here be refactored so it doesn't break the encapsulation of `plane`?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  	raise "Stormy conditions!" if stormy?
+  	raise "Plane not in the airport!" unless @planes.include?(plane)
```
maryrosecook: You could make this condition clearer by creating and using an appropriately named method to check if the plane is on the ground.

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  	@planes.delete(plane)
+  	plane.take_off
+  end 
+
+	def full?
```
maryrosecook: Really nice helper method!

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+		@status = "Landed"
+	end
+	
+	def land
+		self.status = "Landed"
+		self
+	end 
+
+	def take_off
+		self.status = "Flying"
```
maryrosecook: Definitely worth running Rubocop on your code and fixing the bits of code it suggests.  This will help make your code style very idiomatic.

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+	
+	attr_accessor :status
+	
+	def initialize	
+		@status = "Landed"
+	end
+	
+	def land
+		self.status = "Landed"
+		self
```
maryrosecook: It's unusual to return `self` in Ruby.  Though it does allow you to chain method calls (eg `plane.land.status`), it's hard to do consistently because some methods need to return another value.  This makes the interface of the object harder to understand.

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather 
+	
+	def initialize
+	  @weather = ["Sunny", "Sunny", "Sunny", "Stormy"].sample #I'm sorry.
```
maryrosecook: Does duplicating values in the array to control the probability of storms have a downside for readability?

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
...
+	end 
+
+	#private
+	
+	attr_reader :weather
```
maryrosecook: Could there be a method on `Weather` that let's someone find out if it's stormy without knowing about the fact that stormy is represented as a string?

spec/airport_spec.rb
```diff
@@ -0,0 +1,111 @@
+require 'airport'
+
+describe Airport do 
+
+	let(:plane) { double :plane }
```
maryrosecook: Really nice doubling!

spec/weather_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+describe Weather do
+
+	it { is_expected.to respond_to(:weather) }
+
+	it "is expected to be initialized with an admissible value" do
+		condition = Weather.new
+		expect(condition.weather).to eq("Sunny").or eq("Stormy")
+	end 
+
+	# describe "#sunny?" do
```
maryrosecook: Try not to leave commented in code when you commit.

PR #820, opened 2017-07-08
Airport challenge

ryandav:


dearshrewdwit:
These commit messages 😱

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'airport'
+
+describe Airport do
+  subject(:airport) { described_class.new }
+  let(:plane)       { double(:plane, :start_flying => true, :end_flying => false) }
+
+  context 'weather is good' do
+    before do
+      allow(airport.instance_variable_get(:@weather)).to receive(:stormy?).and_return false
```
dearshrewdwit: Does this feel right? (Hint: beware leading questions)

lib/airport.rb
```diff
@@ -0,0 +1,57 @@
...
+require_relative 'weather'
+
+class Airport
+
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+    @weather = Weather.new
```
dearshrewdwit: Is there a way we can inject this object in the initializer so the airport is blind to it/we can test/stub it more easily?

lib/airport.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def plane_count
+    planes.count
+  end
+
+  private
+  attr_reader :planes, :weather
+  DEFAULT_CAPACITY = 20
+  ERROR = { :stormy => "All flights cancelled due to stormy weather",
```
dearshrewdwit: I love this idea. But this is getting a little messy like this. 

I will push you to see how you can raise your own custom exceptions
http://blog.honeybadger.io/ruby-custom-exceptions/

spec/weather_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'weather'
+
+describe Weather do
+
+  describe "#stormy?" do
+    it 'should give a value of true or false' do
```
dearshrewdwit: This feels like a back up solution because of being unable to ensure 1 result consistently. 
How would you ensure 1 return value to be tested, and therefore 2 codepaths?

Essentially we need two tests because of two possible outcomes.

spec/plane_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'plane'
+
+describe Plane do
+
+  subject(:plane) { described_class.new }
+
+  it 'flying? method eq true when start_flying called' do
+    plane.start_flying
+    expect(plane.flying?).to eq true
```
dearshrewdwit: RSpec allows you to use predicate matchers like in this care

```ruby
expect(plane).to be_flying
# or
expect(plane).not_to be_flying
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    it 'should raise error and prevent plane landing if stormy weather' do
+      expect { airport.land(plane) }.to raise_error described_class::ERROR[:stormy]
+    end
+
+    it 'should raise error and prevent take_off(plane) if stormy weather' do
+      expect { airport.take_off(plane) }.to raise_error described_class::ERROR[:stormy]
+    end
+  end
+
+  it 'should have a default capacity that can be overridden' do
```
dearshrewdwit: how would you update this to ensure 1 expectation per test?

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    expect(airport.capacity).to eq Airport::DEFAULT_CAPACITY
+    airport.new_capacity(25)
+    expect(airport.capacity).to eq 25
+    airport = described_class.new(30)
```
dearshrewdwit: if this is the subject let's declare it as such and put it in its own context

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(airport.plane_count).to eq 0
+    end
+
+    it 'land(plane) should raise error when plane is not in airport' do
+      expect { airport.take_off(plane) }.to raise_error described_class::ERROR[:not_in_airport]
+    end
+
+    it 'should prevent land(plane) if airport full' do
+      described_class::DEFAULT_CAPACITY.times do
+        plane_unique = double(:plane, :start_flying => true, :end_flying => false)
```
dearshrewdwit: Is this needed if we have a double declared in a `let` at the top of the describe block?

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      airport.land(plane)
+      expect(airport.plane_count).to eq 1
+    end
+
+    it 'land(plane) should raise error when plane is already in airport' do
+      airport.land(plane)
+      expect { airport.land(plane) }.to raise_error described_class::ERROR[:already_in_airport]
+    end
+
+    it 'should instruct a plane to take off and remove from planes in airport' do
```
dearshrewdwit: How would you update this to have only 1 expectation per test?

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+describe Airport do
+  subject(:airport) { described_class.new }
+  let(:plane)       { double(:plane, :start_flying => true, :end_flying => false) }
+
+  context 'weather is good' do
+    before do
+      allow(airport.instance_variable_get(:@weather)).to receive(:stormy?).and_return false
+    end
+
+    it 'should instruct a plane to land and add to planes in airport' do
```
dearshrewdwit: Does this test that we're invoking a method `#start_flying` on the object called plane? How would you?

PR #832, opened 2017-07-09
airport_challenge

chuk-chuk:


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+  attr_reader :planes, :capacity, :weather
```
dearshrewdwit: what's the purpose of these readers?

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    @capacity = capacity
+    @weather = weather
+  end
+
+  def take_off(plane)
+    fail "Too stormy to take off today." if @weather.stormy?
+    @planes.pop
+  end
+
+  def land(plane)
```
dearshrewdwit: How does this method check if the plane has already been landed?

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+  attr_reader :planes, :capacity, :weather
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
+    @planes = []
+    @capacity = capacity
+    @weather = weather
+  end
+
+  def take_off(plane)
```
dearshrewdwit: How does this method check if the plane has already taken off?

spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'airport'
+require 'plane'
+
+
+describe Airport do
```
dearshrewdwit: This spec for airport, requires the two other files. How would you isolate these tests?

spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  subject(:airport) { described_class.new}
+  let(:plane) { Plane.new }
+  let(:weather) { Weather.new }
+
+  # it { is_expected.to respond_to(:take_off) }
```
dearshrewdwit: commmmmmentts

spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  it "confirms that a plane has landed" do
+    allow(weather).to receive(:stormy?) { false }
+    airport.land(plane)
+    expect(airport.planes).to eq [plane]
+  end
+
+  it "raises an error if airport is full" do
+    allow(weather).to receive(:stormy?) { false }
+    20.times { airport.land(plane) }
```
dearshrewdwit: Is there a way to remove this number `20`? where does it come from

spec/weather_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'weather'
+
+describe Weather do
+  subject(:weather) {described_class.new}
+  it { is_expected.to respond_to(:stormy?) }
+
+  it "should return true or false" do
```
dearshrewdwit: Thinking about what object the `#rand` method is being called on will help us understand how to set up our tests. What info can you find about how rand works?

PR #839, opened 2017-07-09
Airport challenge

artemisxen:


lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    raise "Maximum capacity of the airport is reached!" if full?
+    raise "This plane has already landed!" if planes.include? plane
+    raise "The plane can't land due to stormy weather!" if stormy?
+    planes << plane
+  end
+
+  def take_off(plane)
+    raise "The plane can't be nil!" if plane == nil
+    raise "There are no planes at the airport for take off!" if empty?
+    raise "This plane is already flying!" unless planes.include? plane
```
dearshrewdwit: This is interesting. the error message is assuming it's in the air, but the logic of the check is only if it's in the planes array

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  private
+
+  def full?
+    planes.length == capacity
+  end
+
+  def empty?
+    planes.length.zero?
```
dearshrewdwit: https://ruby-doc.org/core-2.2.0/Array.html#method-i-empty-3F

:)

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def stormy?
+    @weather.stormy == true
```
dearshrewdwit: I would find a way to delegate any and all logic to the weather object.

lib/plane.rb
```diff
@@ -0,0 +1,13 @@
+class Plane
+
+  attr_accessor :landed
+
+  def landed?
+    landed == true
```
dearshrewdwit: what's the purpose of these methods? are we comparing or assigning?

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
+require 'airport'
```
dearshrewdwit: This makes me suspicious. Usually in a unit test you should only require the subject under test file. Here we requiring three classes for one test.

dearshrewdwit: Update: there are indeed 3 classes used in this 1 spec. How would you isolate them?

spec/plane_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'plane'
+
+describe Plane do
+  subject(:plane) { described_class.new }
```
dearshrewdwit: How could you better organise this using context blocks?

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    before do
+      allow(weather).to receive(:stormy?).and_return(true)
+    end
+
+    it "does not allow plane to land if it is stormy weather" do
+      expect { airport.land(plane) }.to raise_error("The plane can't land due to stormy weather!")
+    end
+
+    it "does not allow plane to take off if it is stormy weather" do
+      airport.land(plane)
```
dearshrewdwit: Line 83 will probably raise an error because we haven't organised this test well enough. All the pieces are here! How would you use context blocks to set up different tests?

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    it "raises an error if plane = nil" do
+      plane1 = nil
+      expect { airport.land(plane1) }.to raise_error("The plane can't be nil!")
+    end
+
+    it "raises an error if the max capacity of the airport is reached" do
+      airport1 = Airport.new(5)
+      airport1.capacity.times { airport1.land Plane.new }
```
dearshrewdwit: Let's try to use fake objects to isolate our tests. How would you do that here?

lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+class Weather
+
+  def stormy
+    rand(100) > 90
```
dearshrewdwit: what object receives the rand message? If we understand that, then we can test it :)

lib/plane.rb
dearshrewdwit: Where are they used?

PR #941, opened 2018-01-06
Pull request

Leigan0:


lib/airport.rb
```diff
@@ -0,0 +1,47 @@
+class Airport
+  require_relative 'weather'
+  require_relative 'plane'
+
+  attr_reader :hangar, :weather
+  attr_accessor :capacity
```
mdsomerfield: Try and avoid attr_accessor wherever possible. A really good rule of thumb is if you need to change the internal state of a class from outside the class, it's probably in the wrong place - either the state should be somewhere else, or you shouldn't need to change it in this way. Here I would say the second - you can pass in capacity as a parameter in initialize e.g. def initialize(weather = Weather.new, capacity = DEFAULT_CAPACITY)

spec/features/multiple_land_and_take_offs_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'airport'
+describe 'multiple planes land at airport then take off' do
+  let(:plane1) { Plane.new }
+  let(:plane2) { Plane.new }
```
mdsomerfield: Is this a necessary test? 

I like that you have comprehensively tested this, that's great - but another principle of testing is that it should be efficient. If we have tested landing a plane and we have tested a plane taking off and we have tested the airport capacity (and I believe you have tested all those things) I am not sure that this test adds anything.;

PR #949, opened 2018-01-07
Complete Airport Challenge

rcvink:


lib/control_tower.rb
```diff
@@ -0,0 +1,22 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class ControlTower
+  @weather = Weather.new
```
dearshrewdwit: I don't want to see instance variables hanging around like this
- either initialize the object with it, or wrap this in a method.

spec/plane_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'plane'
+
+describe Plane do
+  subject(:plane) { described_class.new }
+  let(:airport) { double(:airport) }
+
+  it 'can create an instance of Plane class' do
+    expect(plane).to be_instance_of(Plane)
```
dearshrewdwit: redundant test

spec/plane_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  context 'when landing' do
+    it 'lands successfully' do
+      plane.land(airport)
```
dearshrewdwit: You can use `before` to help DRY out test setup

lib/plane.rb
```diff
@@ -0,0 +1,21 @@
...
+  def take_off
+    raise 'Plane is already airborne' unless landed?
+    @airport = nil
+  end
+
+  def landed?
+    !@airport.nil?
+  end
+
+  def grounded_at?(airport)
```
dearshrewdwit: this feels weird to me: i'd probably get the control tower to be responsible for this

The assumption here is that each airport has a control tower, but perhaps this is not accurate in your implementation? And the control tower is some omniscient omnipotent God version?
```ruby
# control tower
def check_presence?(plane)
  plane.airport == airport
end
```

PR #950, opened 2018-01-07
Airport_Challenge PR

lewmoore:
initial PR for airport_challenge


lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 20
+
+  attr_accessor :weather, :capacity, :plane
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @weather = Weather.new
```
dearshrewdwit: This dependency (something that is relied on by this Airport class) is tightly coupled to Airport.

I'd like to see this injected instead as an argument to the initializer.

spec/weather_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'weather.rb'
+
+describe Weather do
+  subject(:weather) { described_class.new }
+
+  describe '#initialize' do
+    it 'should set the weather to clear or stormy' do
+      expect(Weather::CURRENT_CONDITIONS).to include subject.weather
```
dearshrewdwit: I want test coverage of the stormy method ⚡️ ⛈

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
+class Plane
+  def initialize
+    @flying = true
+  end
+
+  def take_off
+    raise 'Plane cannot take off: plane already flying' if flying
+  end
+
+  def land(plane)
```
dearshrewdwit: LEWIS.

what is this

PR #951, opened 2018-01-07
Pull no 2

bpourian:


lib/Airport.rb
```diff
@@ -0,0 +1,43 @@
+class Airport
+  require_relative 'plane'
+  require_relative 'weather'
+
+  attr_reader :airport
+  attr_reader :weather
+  attr_accessor :capacity
```
mdsomerfield: attr_accessor is usually a code smell. It's rare that letting other classes control internal state is the right design decision. I'd consider making capacity an initialize parameter. 

Also, do you need attr_readers for :airport and :weather? You might feel like you do for test purposes, but if what you're really interested in is 'is the airport full' than rather than exposing :airport it's better to have a method 'isAirportFull' - it communicates better and achieves stronger encapsulation.

lib/Airport.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def take_off(plane)
+    raise 'Unable to take off - stormy weather.' if weather.stormy?
+    take_off_confirmation(plane)
+  end
+
+private
+  def take_off_confirmation(plane)
+    @airport.delete(plane)
+    "#{plane} no longer in airport."
```
mdsomerfield: It feels like you've added this for test purposes (line 61 of your airport_spec). You have a method below 'plane_in_hanger' - I'd prefer to use this. All you are really asserting when you test the output is that the method is outputting a string. When you assert against 'plane_in_hangar' you are actually testing the logic.

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    it "should update capacity as required" do
+      airport = Airport.new
+      airport.capacity = 10
+      expect(airport.capacity).to eq 10
+    end
+  end
+
+  describe '#land' do
+    it "should land at an airport" do
+      plane = Plane.new
```
mdsomerfield: I'd probably move some of this setup to a before(:each) method. It makes the test less cluttered and a bit more DRY. It's not always possible, but if you find yourself writing the same code a lot (plane = Plane.new, airport = Airport.new(weather)) it's worth thinking about.

spec/plane_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'plane'
+require 'airport'
+
+describe Plane do
+
+  it "should create a new instance" do
+    expect(subject).to be_a Plane
```
mdsomerfield: This is an unnecessary test. When learning to write good unit tests sometimes we feel like we have to test everything or there's an itch where we feel like we haven't tested enough even though there doesn't seem like much to test. 

This is really just testing state. The plane 'is'. It isn't 'doing'. We should focus on 'doing' - on behaviour, not state. Because your plane class is empty, you don't need to test it.

lib/Plane.rb
```diff
@@ -0,0 +1,4 @@
+class Plane
+  attr_reader :plane
```
mdsomerfield: It feels odd to have an attribute called plane it a class called plane. What does this do? I would expect my plane here to probably have some state (is it on the ground or in the air) but not much more than that.

PR #961, opened 2018-01-08
Airport Challenge

reenz:


lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def take_off(plane)
+    fail "Airport is empty." if @planes.empty?
+    fail "Cannot take off due to stormy weather" if @weather.stormy?
+    if @planes.include? plane
+      @planes.delete(plane)
+      @planes
+    else
+      fail "This plane is not at the airport"
```
mdsomerfield: Minor point, but it's good to be consistent - your other validation (airport is empty or stormy weather) is assessed with a fail... if at the start of the method. I'd do the same here i.e.

```
fail "This plane is not at the airport" if !@planes.include? plane
@planes.delete(plane)
```

Also, on line 31, you return planes - why? It feels like the intent of the method has been achieved by line 30, anything else is probably unnecessary (and allowing external code access to the @planes array also violates the encapsulation of state within the class)

lib/plane.rb
```diff
@@ -0,0 +1,15 @@
...
+  def initialize
+    @flying = true
+  end
+
+  def flying?
+    @flying
+  end
+
+  def land
+    @flying = false
```
mdsomerfield: It looks like the plane can land but can't take off.

lib/plane.rb
```diff
@@ -0,0 +1,15 @@
+class Plane
+
+  def initialize
+    @flying = true
```
mdsomerfield: I'd want this to be a parameter e.g.

```
def initialize(flying = true)
  @flying = flying
end
```

That way you can initialise state more easily when testing

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  it 'lands a plane at airport' do
+    plane = Plane.new
+    expect(airport.land(plane)).to eq(plane)
+  end
+
+  it 'take off a plane from airport' do
+    plane = Plane.new
+    airport.land(plane)
+    expect(airport.take_off(plane)).to eq []
```
mdsomerfield: Ah, now I understand why you return the @planes array from take_off - to permit easy testing?

It feels like if the number of the planes in the airport is something we want to know, then there should be a method or attribute accessor for this. So this test would then be:

// Arrange
plane = Plane.new
airport.land(plane)

// Act
airport.take_off(plane))

// Assert
expect(airport.numberOfDockedPlanes).to eq 0

(If you want to know more about Arrange/Act/Assert for testing, then try reading http://defragdev.com/blog/?p=783 - I find that it's a really good way to think about unit tests)

PR #971, opened 2018-01-28
Marcus Gardiner Pull Request

marcusfgardiner:
Marcus' README Edits - My approach and mini code review
=================

My approach
---------
* First thing I did was read all the user stories and map out a domain model of the different elements in the model and drew a simple diagram of the situation to contextualise it in my head
* I then made a list other elements to consider, such as the weather aspects and edge cases
* I broke down the problem into a list of features for me to drive using BDD
* Then I put these features in priority order, easiest first, for BDD
* As I went stage by stage completing each feature using BDD and RGR, I noted down nice to have elements that I would do if I had time at the very end
* After all core features were complete, I did some of the 'nice to have' elements e.g. splitting out methods into modules that I includes
* An important learning and application was using a solid Debugging process throughout (Digest error message, tighten the loop on where the problem was, fix the bug)

Mini personal code review: My appraisal of my attempt at the weekend challenge
---------

* I found writing the core code was quite an enjoyable process. I did not get stuck anywhere as I built up the core functionality through BDD, which took c.3 hours. I think this was due to the level of similarity with Boris bikes and due to applying my learnings from the week, particularly to very carefully read each and every error message (which meant I didn't write erroneous tests)
* Next steps if I had the time would be to simplify the test code (particularly the 'let' examples, which are long and take too much focus to keep track of) and spend more time ensuring the code was of good quality and refactored (e.g. fulling applying XP principles).
* There is also of course a lot of additional functionality that could be built, but I enjoyed applying the YAGNI principles in this case (You aren't gonna need it!)
* I was not sure if I should build direct tests of the boolean methods, which currently don't have tests as they are private methods
* I would like to review the quality of the design/structure of my code and tests, and verify with an experienced developer what I should focus on to better apply XP principles
* I made a first attempt at a feature test using a spec file, but don't know how to do this in practice, so keen to learn!


README.md
```diff
@@ -1,6 +1,31 @@
+Marcus' README Edits - My approach and mini code review
+=================
+
+My approach
```
EdwardAndress: Great approach and it's useful for us to see your process documented as part of the PR.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  attr_accessor :hangar, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    raise 'Cannot land, airport full' if full?(hangar)
```
EdwardAndress: It's good that you're trying to abstract some functionality into modules.  Since, however, you have access to `#hangar` in the Airport class, you don't need to pass in `hangar` as an argument.

lib/full_empty_module.rb
```diff
@@ -0,0 +1,7 @@
+module Full_Or_Empty
+  private
+
+  def full?(storage)
+    storage.count >= capacity
+  end
+end
```
EdwardAndress: This allows you to share functionality between different classes which all want to check whether a thing is full.  It could be a bike rack or a hangar.  Nice.  I think, seeing this and thinking about potentially re-using the module, I understand why you're passing `hangar` as an argument.  It's because you can assume that the storage item will have a `#capacity` attribute, but you cannot assume it will always be referred to in the same way on each class (i.e. a bike rack will not be called a hangar).

It would be cool if you could pass an argument to the Module (there is no simple way of doing that but there are some messy ways).

lib/plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
+end
```
EdwardAndress: I'd have expected the plane to do more than this.

EdwardAndress: It could, for example, store some state.

lib/weather_module.rb
```diff
@@ -0,0 +1,7 @@
+module Weather
+  private
+
+  def stormy?
+    rand(10) > 8
+  end
+end
```
EdwardAndress: It's interesting that you chose to use a module again.  I like it!

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+  let(:airport_full) do
+    not_stormy
+    described_class::DEFAULT_CAPACITY.times { subject.land(double('A plane')) }
+    subject
+  end
+  let(:custom_capacity) { 50 }
+  let(:airport_custom_capacity) { Airport.new(custom_capacity) }
+
+  let(:not_stormy) { allow(subject).to receive(:stormy?).and_return(false) }
+  let(:stormy) { allow(subject).to receive(:stormy?).and_return(true) }
```
EdwardAndress: "'let examples, which are long and take too much focus to keep track of" – you've already seen the thing I was going to point out, which is great.

There are a few solutions that I can think of.  I'll try to steer you towards then rather than just write an essay.
1. Dependency injection.
2. Stubbing '#rand()'
3. Investigate how you can just restructure (more or less) the exact same code.  Specifically, think about when something like 'airport_full' comes into existence and what that means in terms of where you can do `allow(subject).to receive(:stormy?).and_return(false)`

If this is unhelpful drop me a DM on slack and I will elaborate.

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  context '#land(plane)' do
+    it { is_expected.to respond_to(:land).with(1).argument }
+
+    context '#stormy? = false' do
```
EdwardAndress: ==

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      context '#full'
+
+      it 'Raises an error if airport is full and try to land plane' do
+        expect { airport_full.land(plane) }.to raise_error('Cannot land, airport full')
+      end
+    end
+  end
+
+  context '#take_off(plane)' do
```
EdwardAndress: this should be 'describe'

The use of context, below, is good although I'd probably choose to have a plain english description.  `context 'the weather is stormy' do ....`

As you'll have found context and describe will, functionally, work the same but they can be used to convey different meaning and that's the important part.  Context blocks communicate some specific conditions in which the tests are being carried out.

EdwardAndress:
Nice work overall.  I like the fact that you're using modules.  There are some minor comments in-line.  Please do drop me a line on Slack if you have any questions.

spec/feature_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+#
+#   airport.land(plane)
+#
+#   15.times { airport.land(Plane.new) }
+#
+#   4.times { airport.take_off(hangar[rand(8)]) }
+#
+#   8.times { airport.land(Plane.new) }
+#
+# end
```
EdwardAndress: Did you get stuck or run out of time?

I find it a bit tricky to extract features from these challenges because the functionality is pretty limited so there's a big overlap between features and basic functionality.  I'd say that features here would be things like:

1. A plane cannot land if the airport is full.
- This differs from unit tests which just individually test the return values of things like `Airport#stormy?` and `Airport#land(plane)`
2. A plane cannot land if the hangar is full.

PR #972, opened 2018-01-28

Jestfer:


lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def take_off(plane)
+    before_take_off(plane)
+
+    hangar.delete(plane)
+
+    weather.change
+
+    "#{plane} has just taken off"
```
mdsomerfield: Minor point: presentation logic in a business logic context is usually a violation seen as poor separation of concerns. The responsibility of this method is to process the take off of a plane - _reporting_ on the take off such that the user can see it is a different responsibility. 

It's a subtle distinction, but my sense is that you have included this simply so that you can test it, and since you raise concerns about your tests in your self-review I thought it was worth pointing out. When you are including output simply for the purpose of testing it's usually either a sign of a design flaw or a sign that your approach to testing needs additional thought.

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  private
+
+  def before_land(plane)
+    raise "Already in hangar" if plane?(plane)
+    raise "Hangar is at its full capacity" if full?
+    weather.change & raise("Can't land due to stormy weather") if weather.stormy?
```
mdsomerfield: It doesn't feel like the responsibility of the airport to change the weather - the weather is an external factor. I would expect this could simply to be weather.get_weather (or something like that). It probably does the same thing in this context, but the difference is that when you call the method 'change' I expect the weather to change, whereas get_weather simply tells me that 'there is weather', and I don't need to worry about whether it's going to change or not or even weather it is random or calculated by some algorithm.

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+
+class Airport
+  attr_reader :hangar, :capacity, :weather
+
+  DEFAULT_CAPACITY = 6
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+    @weather = Weather.new
```
mdsomerfield: You mention in your self-review that you have trouble mocking Weather. I think this is probably the root of your probably - you instantiate weather within your Airport class. 

Consider:

```
def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
  @hanger = []
  @capacity = capacity
  @weather = Weather.new
end
```

Now your test can look something like:

```
it 'should prevent landing when weather is stormy' do
  allow(mockWeather).to receive(:get_weather).and_return(:stormy);
  airport = new Airport(10, mockWeather);
  expect { subject.land(@plane) }.to raise_error("Can't land due to stormy weather")
end
```

This may not exactly reflect your code in its current state, but the idea is that you have a mock object which you inject into your airport and that allows you to control the output. 

More generally, this is known as dependency injection and it's a technique which, among other things, allows you to isolate code for testing.

spec/airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    it 'should raise error if plane takes off from wrong airport' do
+      another_airport = Airport.new
+
+      expect { another_airport.take_off(@plane) }.to raise_error("Plane not in hangar")
+    end
+
+    it 'should prevent takeoff when weather is stormy' do
+      subject.weather.condition = 'stormy'
```
mdsomerfield: I've left a comment about this above and I think you're already aware from your self review, but modifying internal state like this is generally a sign that you haven't encapsulated your state as well as you could.

PR #995, opened 2018-02-25
Submitting my airport_challenge

chiakimz:


spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    message = "It's stormy! You can't land."
+    expect { airport.land(plane) }.to raise_error message
+  end
+
+  it "raises an error when full" do 
+    allow(airport).to receive(:stormy?).and_return false
+    airport.land(plane)
+    message = "The airport is full!"
+    expect { 
+      loop do airport.land(plane) end
```
EdwardAndress: This is quite a clever solution but perhaps it would be clearer to write it another way where the capacity of the airport is specifically referred to..?  I'll leave you to think about that one :)

PR #1049, opened 2018-04-23
pull request for weekend challenge

SiobhanGi:


SiobhanGi:
I struggled doing the feature tests, took me a bit to find an easy way to get #stormy_conditions? from the weather class to pass to Airport class as an argument.

I think I did well naming the methods, it reads like a story. I would have liked to have changed in the Weather class the 0.2 float value to a constant

I want to improve on writing unit tests and keeping my methods short. I don't think I achieved that very well.

PR #1099, opened 2018-08-12
Airport Challenge Complete

LondonJim:


spec/airport_spec.rb
```diff
@@ -0,0 +1,91 @@
+require 'airport'
+
+describe Airport do
+
+  class Plane
+  end
+
+  let(:plane) { Plane.new }
```
samjones1001: While this line, combined with your dummy place class included above, allow you to isolate tests correctly, RSpec provides some syntax to allow us to do the same thing more succinctly which is a bit more conventional

```ruby
let(:plane) { double :plane }
```

Using this would mean that you could remove lines 5-6

spec/airport_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    it 'does not let a plane already in the airport land again' do
+      allow(subject).to receive(:stormy?) { false }
+      subject.land(plane)
+      expect { subject.land(plane) }.to raise_error 'plane is already in the airport'
+    end
+
+    it 'adds a named plane to @planes_on_ground array when it lands and #stormy? == false' do
+      allow(subject).to receive(:stormy?) { false }
```
samjones1001: You're repeating this line in a number of your tests which violates the idea of keeping code DRY a little bit.

RSpec provides a couple of handy helper methods - context and before - which would allow you to get around this by doing something like

```ruby
describe Airport do 
  context 'When the weather is good' do
    before(:each) do
      allow(subject).to receive(:stormy?)  { false }
    end

    //all your tests which require good weather go here

  end
end
```

samjones1001: You can even nest further describe blocks within the context if you want to - this can really help with keeping tests organised

spec/weather_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'weather'
```
samjones1001: I really like the idea of including Weather into Airport as a module - it feels like a sensible design choice.

That said, the way you're currently testing your module is a bit unconventional 🙂 

Really, you want to test it as *part of the class which includes the module* rather than by creating a dummy class for the purpose of testing as you've done here.

Take a look at the Boris Bikes challenge step called 'using modules as mixins' for more information on how this works practically.

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
```
samjones1001: I think we mentioned in yesterday's workshop that a class should encompass both state and behaviour - if either of these things is missing then it might not be a valid class.

Given that in this case both are missing I have my doubts as to whether this is valid 🙂 

If the way in which you've designed your system means that you don't really need a defined plane object, that's potentially absolutely fine, but I wouldn't include an empty one like this just for the sake of it.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  DEFAULT_CAPACITY = 8
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes_on_ground = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail 'not an identifiable plane' if plane.class != Plane
```
samjones1001: This is really interesting - I can see the thinking behind it I think, but in practice this is setting up a really tight dependency between your Airport and Plane classes.

Right now, there's no way your Airport can accept anything but a Plane.  What if the brief later changes and we also want the airport to accept helicopters?

Ideally, your objects should know as little about each other as possible - your airport probably shouldn't actually care what is landing, just that it can be put into the planes_on_ground array

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    fail 'plane is already in the airport' if at_airport?(plane)
+    fail 'landing denied, airport full' if @planes_on_ground.count >= @capacity
```
samjones1001: There's an opportunity to refactor some stuff here out into a small private method.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    fail 'landing denied, weather is stormy' if weather_check == true
```
samjones1001: weather_check = true feels slightly obtuse to me.  If I didn't know the domain, I think I'd be confused by what it mean.  Does true = stormy?  Good weather?

You could probably make it clearer by renaming the method to something more communicative.

PR #1106, opened 2018-08-12
Samir: Airport Challenge (Week1)

LazySamir:


lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+require_relative "./plane.rb"
+
+class Airport
+  DEFAULT_CAPACITY = 20
+  attr_accessor :hanger
+  attr_accessor :capacity
```
samjones1001: From what I can see, these could both be attr_reader rather than attr_accessor - if you're not clear on the difference you might want to do some research.

Generally speaking, you  want to provide the lowest level off access to your instance variable as is required by you application.

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
...
+    @hanger = []
+    @capacity = capacity
+  end
+
+  def good_weather?
+    [false, true].sample
+  end
+
+  def empty?
+    @hanger.empty?
```
samjones1001: Try getting rid of the @ here and running your tests / code.  

What happens?  Any sense of why?

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
+class Plane
+
+  def land(airport, weather)
+    raise "plane has already landed" if airport.hanger.include? self
+    raise "hanger is full" if airport.full?
+    raise "bad weather: can't land" unless weather == true
```
samjones1001: weather == true feels a little bit obtuse to me : )

I think if someone wasn't familiar with the airport challenge, they would be likely to have trouble understanding what it means.  Try to aim for code which really communicates its intent.

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+    airport.hanger << self
```
samjones1001: This line involves directly accessing an instance variable of another class - this is generally something that you want to avoid.  

The good news is this is easily done.  In your Airport.rb file, you'd need to add something like
```ruby
def add_to_hangar(plane)
  hangar << plane
end
```  

Then you could change this line to
```ruby
ariport.add_to_hangar(plane)
```

samjones1001: You've actually already done this in certain places with your Airport's .empty? and .full? methods

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
+require "airport"
+
+describe Airport do
+  let(:cap)         { 30 }
+  let(:sub2)        { Airport.new(cap) }
+
+  context "when first initialized" do
+    it { is_expected.not_to eql(nil) }
```
samjones1001: You mentioned wanting to dry out your tests in your self reflection - these first three tests feel like a good place to start.

I tend to think of .to respond_to type tests as being a great stepping stone in the TDD process, but not as something which should stick around long term.

If you think about it, once you add more in depth tests like those below, you are implicitly testing that you can create instances of the class under test - if you weren't able to then all of the subsequent tests would inevitably fail - so you have effectively rendered these tests redundant.

spec/plane_spec.rb
```diff
@@ -0,0 +1,68 @@
+require "plane"
+
+describe Plane do
+  let(:airport)         { Airport.new }
```
samjones1001: My guess is that this long list of let blocks is one of the things which is bothering you about your tests.  I'd agree that some of them are probably not strictly necessary, and that the way that they are organised makes them a little difficult to follow.

If you're interested in finding better ways to organise your tests I suggest that you have a look at both 'context' and 'before' blocks - by combining the two you should find you can cut down on some repetition and generally make your tests easier to follow.

Happy to talk this through with you in person if you'd like.

lib/plane.rb
LazySamir: I can definitely see that, will avoid it in future. Not just that, 'unless weather' would have sufficed ha.

lib/airport.rb
LazySamir: Ah, you're right, thanks.

spec/airport_spec.rb
LazySamir: I'll keep an eye on these in future. What do you think about: 
```context "when first initialized" ``` ?

I was reading that context is generally used when there is a counter context to contrast it with (this would also make the rspec test output layout more logical) and was thinking maybe this makes more sense as a describe.

spec/plane_spec.rb
LazySamir: I spent so long trying to get those 'before' working. The mistake I made (repeatedly) was not understanding that you wrap the code in the before but the context around the tests. It didn't occur to me that before was needed to be ended before the tests. On reflection, it's in the name ha.

lib/airport.rb
LazySamir: I can't access the code right now but my initial feeling is that it will run because the attr reader is making it accessible as is. Like it would be from another class.

LazySamir:
Thank you for the feedback Sam. Much appreciated.

lib/airport.rb
samjones1001: Correct!  Given that you have the readers, you'll probably want to avoid accessing the instance variable directly in this way by removing the @ signs.

spec/airport_spec.rb
samjones1001: Agreed - I tend to use context in the way you're describing here.  For example, a context for good weather and a context for stormy.

I don't think there's anything wrong with your usage here, but it's probably unnecessary.

PR #1109, opened 2018-08-13
airport_challenge

Alastair2D:


lib/airport.rb
```diff
@@ -0,0 +1,56 @@
...
+    @hangar = []
+    @weather_station = weather_station 
+  end  
+  
+  def status(plane)
+    hangar.include?(plane) ? "#{plane} is in the hangar" : "#{plane} is in the sky" 
+  end  
+  
+  def check_weather
+    @safe = weather_station.conditions_safe?
```
samjones1001: I'm not entirely convinced that you need to use an instance variable here.  I think this could be simplified to

```ruby
def check_weather
  weather_station.conditions_safe?
end
```

so that the method just returns true/false.

Then, in the place where the method is being called, on line 33 for example, you could have

```ruby
fail 'Error - weather is stormy' if !check_weather
```

Arguably, that line doesn't communicate intention particularly well, but that could be easily fixed with a small change to the method name.

samjones1001: It also feels as though the check_weather method should probably be private.  Do you ever plan on calling it from outside of the Airport class?

lib/airport.rb
```diff
@@ -0,0 +1,56 @@
...
+  end  
+  
+  def take_off(plane)
+    clear_to_launch?(plane)
+    plane.fly
+    hangar.delete(plane)
+    "#{plane} has left the airport"
+  end  
+  
+  def clear_to_launch?(plane)
```
samjones1001: This method and clear_to_land? could probably also be made private too - try to only make methods public if they absolutely need to be i.e. if you expect another object or a user to interface with them directly.

lib/plane.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def initialize(flying = false)
+    @flying = flying
+  end  
+
+  def flying?
+    @flying
+  end  
+
+  def grounded?
```
samjones1001: This is probably superfluous - you can tell if a plane is grounded by checking whether flying? returns false.

You could argue that it improves readability however.

spec/airport_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  let(:mockPlane) { double :plane }                              # let(:plane) { instance_double("Plane") }
+  let(:mockPlane2) { double :plane }                  
+  let(:mockWeatherStation) { double :weatherstation }            # If we don't initialize Airports with Weather.new would this be valid: et(calmWeather) { double :weather, :conditions_safe? => true }
+  
+  describe '#initialize ', :initialize do 
+    it 'with an empty @hangar' do 
+      expect(subject.hangar).to be_empty           
+    end
+    it 'with a default capacity of 1' do
+      expect(subject.capacity).to eq(Airport::DEFAULT_CAPACITY) # expect(subject.capacity).to eq 1
```
samjones1001: There are a couple of instances in this file of commented out code.  In this context it really doesn't matter, but if you were submitting a real tech test for review you would want to remove all of these first.

I think the explanatory notes you've left to yourself are great, but the same rule applies.

spec/airport_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  
+  describe '#status ', :status do 
+    before { subject.hangar << mockPlane }
+    it 'provides plane location' do 
+      expect(subject.status(mockPlane)).to eq "#{mockPlane} is in the hangar"
+    end
+  end 
+
+  describe '#take_off', :take_off do 
+    it { is_expected.to respond_to(:take_off).with(1).argument }
```
samjones1001: I tend to delete these 'respond_to' tests once I've got some more rigorous tests in place.  By testing the behaviour of take_off below you implicitly test that Airport can respond to take_off, as the tests would otherwise fail.  These simple tests become redundant.

spec/airport_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+    end
+  
+    context 'when weather is stormy' do
+      it 'does not allow planes to land' do 
+        allow(mockWeatherStation).to receive(:conditions_safe?).and_return(false)
+        expect { subject.land(mockPlane) }.to raise_error 'Error - Weather is stormy'
+      end
+    end
+  end
```
samjones1001: This spec file seems to be very busy with setting up mocks and stubbing out methods in multiple places.  Without looking at exactly where this could be tidied up, this suggests to me that you could probably simplify things a bit.  

I'd have a look at making use of your context blocks, and using before blocks within them in order to cut down on the need to repeatedly set up tests.

spec/airport_spec.rb
Alastair2D: Sure

spec/airport_spec.rb
Alastair2D: Agreed, I am kind of keeping the notes / questions / syntax in these programs just to be able to refer back to them

lib/plane.rb
Alastair2D: Take your point.  Actually spoke to Sophie about this.  I'm still getting my head around the practices involving query methods.  I like them a lot for readability.  The alternative to having a grounded? method (and I do recognise that it is superfluos logically as it is just !flying so the same users can call flying? and get an equivalent response.  However, I kind of like - and you may tell me otherwise - the line in #take_off(plane) and #land(plane) that says plane.fly and plane.touch_down as this explicitly links them to plane state

lib/airport.rb
Alastair2D: I understand what you are saying - that methods should only be made public if they need to be, because another object or user requires access.  Are you saying that #clear_to_launch/land - by virtue of the fact they are only called by the airport (and not needed to be called by any other object outside the Airport class) - should be private?

What if we introduced an AirTrafficController who did need to call this method at other times?  Would it then need to be public?

In terms of 'not testing private methods' - how does that work?  Literally don't bother TDD-ing them?

PR #1121, opened 2018-09-09
First pull request for Airport challenge

afaraone:
I have managed to complete all of the user stories.
I've attempted to use mocking in all of my user tests, however I have had some issues mocking the Weather class for Airport classes.
All of my tests are passing.
There are some aspects of the code that I am unhappy with. Looking back, I would like store a Weather object as an instance variable for an Airport object. However, this would require rewriting quite a lot of the code. Since I am still a bit unfamiliar with the correct process for mocking I have avoided doing this for now.


lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  attr_reader :hangar, :capacity
+
+  def initialize(capacity = Airport::DEFAULT_CAPACITY)
+    @capacity = capacity
+    @hangar = []
+  end
+
+  # #land and #takeoff use Weather class in order generate value for 
+  # stormy whenever called
+  def land(plane, weather = Weather)
```
samjones1001: You could probably inject a reference to you Weather class at initialisation rather than having to do it in both your take_off and land methods, although this is a very minor refactor.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    raise "#{plane.name} has already landed!" if @hangar.include?(plane)
```
samjones1001: it looks as though you're directly accessing the @hangar variable, rather than making use of the attr_reader you've set up.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    raise "No room in hangar!" if @hangar.size == @capacity
```
samjones1001: I generally like to refactor these kind of guard clauses out into their own private methods for readability - something like this:
```ruby
raise "No room in hangar!" if hangar_full?
```

and then
```ruby 
private
def hangar_full?
  hangar.size == capacity
end
```
Where the guard clauses are relatively small and simple to follow as is the case here it doesn't make a huge difference, but I think its worth bearing in mind.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    @hangar.delete(plane) 
+    puts confirmation("taken off", plane), display_hangar
+  end
+
+  private
+  def storm_check(weather)
+    raise "It is too stormy to do that!" if weather.new.stormy
+  end
+
+  def confirmation(action, plane)
```
samjones1001: I actually quite like this ad the display_hangar method - I think they provide some nice extra feedback to the user.

That said, I think you're right to identify that they potentially add some extra complexity that wasn't called for. As a general rule, you should only write the code that your user stories call for - try to avoid making assumptions about what extra features could be useful.

lib/weather.rb
```diff
@@ -0,0 +1,12 @@
+# Weather objects gain a randomised @number on instantation
+# #stormy is based value of @number
+
+class Weather
```
samjones1001: You could refactor this class to not need the @number instance variable

```ruby
class Weather
  def stormy
    rand(100) >= 90
  end
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    it 'adds a plane object to @hangar and puts a confirmation message' do
+      # Set up
+      mockplane1, mockplane2 = double(name: "Plane 1"), double(name: "Plane 2")
+      
+      # Assert
+      expect { subject.land(mockplane1, MockNotStormy) }.to output("Plane 1 has landed.\nThe hangar contains Plane 1.\n").to_stdout
+      expect(subject.hangar).to eq([mockplane1])
+      expect { subject.land(mockplane2, MockNotStormy) }.to output("Plane 2 has landed.\nThe hangar contains Plane 1, Plane 2.\n").to_stdout
+      expect(subject.hangar).to eq([mockplane1, mockplane2])
```
samjones1001: As a general rule, I would avoid having multiple expects in a single unit test.

PR #1152, opened 2018-10-06
Airport Challenge

aimeesophia:


lib/airport_challenge.rb
```diff
@@ -0,0 +1,34 @@
+class Airport
```
samjones1001: Overall, this looks like a pretty tidy class.  The methods are concise and well named, so it's easy to see exactly what each part of the code is intended to do.

The only obvious problem is that, right now, it's not particularly cohesive - it manages a list of landed planes AND determines the current weather.  Based on what you know now, can you think of a sensible extraction you might make?

lib/airport_challenge.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_accessor :hanger
+  attr_reader :capacity
+
+  def initialize(capacity = 20)
+    @capacity = capacity
+    @hanger = []
+  end
+
+  def land(plane)
+    raise "Cannot land, weather is stormy." unless weather != "stormy"
```
samjones1001: This is a personal preference as much as anything, but I like to extract little checks like this out into private methods for readability.  So, something like:

```ruby
def land(plane)
  raise "Cannot land, weather is stormy." unless good_weather?
end

private
def good_weather?
  weather != "stormy"
end
```

lib/airport_challenge.rb
```diff
@@ -0,0 +1,34 @@
...
+    @hanger << plane
+  end
+
+  def take_off(plane)
+    raise "Cannot take off, plane is not in airport." unless check_hanger(plane)
+    raise "Cannot take off, weather is stormy." unless weather != "stormy"
+    @hanger.delete(plane)
+  end
+
+  def check_hanger(plane)
```
samjones1001: Feels to me as though this could be a private method.  As a rule, you should expose as little of a class as reasonably possible.  If a method is unlikely to be called on directly by a user or another object, make it private by default.

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'airport_challenge'
+
+describe Airport do
```
samjones1001: This looks as though your testing has been pretty thorough, which feels like good evidence of a solid TDD process.

The only issue is that the tests aren't currently isolated.  Again, based on what you know now, do you think you could fix this?

PR #1158, opened 2018-10-07
Finished Airport challenge

imtiyazzaman1:


hnryjmes:
Really great work. Your code is clear and I can understand the intention of most lines without needing comments. You have shown that you understand guard conditions, using doubles, and object-oriented design. Your method names are readable. You have used method stubbing to handle randomness in your test. You have shown that you understand the single responsibility principle. You have made sure that variables are appropriately scoped, i.e. that classes cannot directly change data inside other classes.

PR #1169, opened 2018-10-07
Finished airport_challenge

runticle:


README.md
```diff
@@ -88,3 +88,7 @@ Finally, don’t overcomplicate things. This task isn’t as hard as it may seem
 * **Submit a pull request early.**  There are various checks that happen automatically when you send a pull request.  **Fix these issues if you can**.  Green is good.
 
 * Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+
+## My Approach
```
imogenmisso: Maybe include here how your file is used from a users perspective. You could do this using a code example or an irb transcript.

spec/plane_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "plane"
+
+describe Plane do
+  let(:plane) { Plane.new }
+  let(:airport) { double(:airport, hanger: [], full?: false, stormy?: false) }
+  
+  describe "#land" do
```
imogenmisso: Good use of describe to group your tests. You could also use #context

lib/airport.rb
```diff
@@ -0,0 +1,17 @@
...
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hanger = []
+    @capacity = capacity
+  end
+
+  def full?
+    @hanger.length == @capacity
+  end
+
+  def stormy?
```
imogenmisso: This should be put in a separate 'Weather' class. The airport needs to be able to access weather information but it shouldn't be responsible for determining the weather. Use require 'weather' to call the weather class in your airport file.

PR #1170, opened 2018-10-07
Mostly complete, basic features + capacity override, no weather features, no doubles

hnryjmes:


lib/airport.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def request_landing(plane)
+    if @planes.count < @capacity
+      @planes << plane
+      plane.landed = true
+    else
+      raise "ERROR: Airport is full. Request rejected."
+    end
+  end
```
imtiyazzaman1: Method could've been refactored as below to improve readability 
```
raise "ERROR: Airport is full. Request rejected."  if @planes.count >= @capacity
@planes << plane
plane.landed = true
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  it 'can have a different capacity if this is was specified at creation' do
+    big_airport = Airport.new("big_airport", 100)
+    i = 0
+    100.times do
+      big_airport.request_landing(Plane.new("plane_#{i}"))
+      i += 1
+    end
+    expect { big_airport.request_landing(Plane.new("plane_#{i}")) }.to raise_error
```
imtiyazzaman1: error message could've been specified to match the guard clause in the class

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
...
+  def initialize(name = self, capacity = DEFAULT_CAPACITY)
+    @name = name
+    @planes = []
+    @capacity = capacity
+  end
+
+  def request_landing(plane)
+    if @planes.count < @capacity
+      @planes << plane
+      plane.landed = true
```
imtiyazzaman1: A method within the plane class could have been used to change the value of @landed so the plane class has more control over it

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
...
+    else
+      raise "ERROR: Airport is full. Request rejected."
+    end
+  end
+
+  def request_take_off(plane)
+    @planes.delete(plane)
+    plane.landed = false
```
imtiyazzaman1: A method within the plane class could have been used to change the value of @landed so the plane class has more control over it

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+end
```
imtiyazzaman1: Method names match the class

spec/airport_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+end
```
imtiyazzaman1: error messages are tested but the functionality of `request_landing(plane)` is not

lib/controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+  end
+
+  def instruct_plane_to_take_off(plane, airport)
+    airport.request_take_off(plane)
+    if !plane.landed?
+      "#{plane.name} taken off from #{airport.name}!"
+    else
+      "ERROR: Plane didn't take off!"
+    end
```
imtiyazzaman1: method could be refactored to use a guard clause instead of an `if` block

lib/controller.rb
```diff
@@ -0,0 +1,19 @@
+class Controller
+  def instruct_plane_to_land(plane, airport)
+    airport.request_landing(plane)
+    if plane.landed?
+      "Landed #{plane.name} at #{airport.name}!"
+    else
+      "ERROR: Plane didn't land!"
```
imtiyazzaman1: method could be refactored to use a guard clause instead of an `if` block

spec/controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+describe Controller do
+  describe "#instruct_plane_to_land" do
+    it "can instruct a plane to land" do
+      plane_1 = Plane.new("BA178")
+      expect(plane_1.landed?).to eq false
+      airport_1 = Airport.new("Heathrow")
+      expect(airport_1.planes).to_not include plane_1
+      expect(subject.instruct_plane_to_land(plane_1, airport_1)).to eq "Landed BA178 at Heathrow!"
+      expect(airport_1.planes).to include plane_1
+    end
```
imtiyazzaman1: Test could be broken up so multiple `expects` are not used

spec/controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+  describe "#instruct_plane_to_take_off" do
+    it "can instruct planes to take off" do
+      airport_1 = Airport.new("Heathrow")
+      plane_1 = Plane.new("BA178")
+      airport_1.request_landing(plane_1)
+      expect(plane_1.landed?).to eq true
+      expect(subject.instruct_plane_to_take_off(plane_1, airport_1)).to eq "BA178 taken off from Heathrow!"
+      expect(airport_1.planes).to_not include plane_1
+    end
```
imtiyazzaman1: Test could be broken up so multiple `expects` are not used

spec/controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+end
```
imtiyazzaman1: Test could have been isolated with doubles

spec/controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+require 'airport'
+require 'plane'
+
+describe Controller do
+  describe "#instruct_plane_to_land" do
+    it "can instruct a plane to land" do
+      plane_1 = Plane.new("BA178")
+      expect(plane_1.landed?).to eq false
+      airport_1 = Airport.new("Heathrow")
+      expect(airport_1.planes).to_not include plane_1
```
imtiyazzaman1: `allow` could have been used instead of `expect`

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+class Airport
+  attr_reader :name, :planes, :capacity
```
imtiyazzaman1: private method could have been used to call @planes so it doesn't expose the array

lib/plane.rb
```diff
@@ -0,0 +1,14 @@
+class Plane
+  attr_reader :name
+  attr_accessor :landed
```
imtiyazzaman1: Use attr_reader instead attr_accessor so other classes cannot change it

lib/controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def instruct_plane_to_take_off(plane, airport)
+    airport.request_take_off(plane)
+    if !plane.landed?
+      "#{plane.name} taken off from #{airport.name}!"
+    else
+      "ERROR: Plane didn't take off!"
+    end
+  end
+end
```
imtiyazzaman1: Fulfils SRP

PR #1174, opened 2018-10-07
Airport Challenge 

Br0ckers:
Storied 1, 2, 5 & 6 completed, struggled to get the mock to work although created the code for the test I had to roll this back as I was unable to pass the double test.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,9 +1,17 @@
+
+require './lib/weather'
+require './lib/plane'
+
+
+
+
 class Airport
 
-  #def initialize
-    #@plane_inv = []
-  #  @max_cap = 6
-  #end
+  def initialize
+  @weather = Weather.new.met_report
+  #@plane_inv = []
+  #@max_cap = 1
+  end
 
   def land(plane)
     @plane = plane
@@ -11,9 +19,8 @@ class Airport
   end
 
   def take_off(plane)
-    #weather? = stormy
-    #if weather? == stormy return "Can't take off bad weather!"
-    @plane = plane
+    #raise "Unable to take off due to bad weather!" if @weather == "stormy"
+    return "Plane has taken off"
 
   end
 end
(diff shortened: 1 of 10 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+require './lib/weather'
+require './lib/plane'
+
+
+
+
+class Airport
+
+  def initialize
+  @weather = Weather.new.met_report
```
samjones1001: By calling the `met_report` method here in the initialize you're setting the weather for an Airport object straight away, and you're then stuck with the same weather forever - perhaps this is intentional, but it feels a little odd to me.

samjones1001: A nice example of delegating responsibility to another object though!

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+  #@plane_inv = []
+  @max_cap = []
+  end
+
+  def land(plane)
+    raise "Airport full" unless @max_cap.length < 1
```
samjones1001: This is a personal preference more than anything, but I like extracting out little checks like this out into names private methods for readability. So, something like:

```ruby
def land(plane)
  raise "Airport full" unless space_at_airport
end

private
def space_at_airport
  @max_cap.length < 1
end
```

lib/plane.rb
```diff
@@ -0,0 +1,6 @@
+class Plane
+  def location?
+    return "air"
```
samjones1001: Looks as though the plane is always in the air right now, even when it's landed 🙂

spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'airport'
+require 'weather'
+
+describe Airport do
```
samjones1001: Looks as though you application is pretty well tested overall, which suggests a strong TDD process. Do you feel as though you'd be able to go back and isolate your tests now based on what you've learned this week?

PR #1179, opened 2018-10-07
Submission

sacullezzar:
Features covered: landing and departing of planes.
Maximum capacity can be defined by user
planes reports status (flying or not)
Weather class created but yet to integrate through TDD

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -3,21 +3,24 @@ require_relative 'weather.rb'
 
 class Airport
   DEFAULT_CAPACITY = 20
-  def initialize
+  attr_accessor :plane, :capacity, :flying
+  def initialize(capacity = DEFAULT_CAPACITY)
     @hangar = []
+    @capacity = capacity
   end
 
   def depart_plane(plane)
     fail "No Planes in the hangar!" if @hangar.empty?
-    @hangar.pop
-    plane(@@flying = true)
+    @hangar.delete(plane)
+    $plane_status = true
+    plane
   end
 
   def land_plane(plane)
-    fail "Hangar is full!" if @hangar.count >= 20
+    fail "Hangar is full!" if @hangar.count >= @capacity
     @hangar << plane
-    plane(@@flying = false)
+    $plane_status = false
+    plane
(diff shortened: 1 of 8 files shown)


lib/plane.rb
```diff
@@ -0,0 +1,10 @@
+require 'airport.rb'
+class Plane
+  def initialize
+    $plane_status = true
```
samjones1001: Ruby developers *really* don't like global variables - try not to use them!

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+require_relative 'plane.rb'
+require_relative 'weather.rb'
+
+class Airport
```
samjones1001: This looks like a pretty tidy class and feels pretty cohesive, but I can see that it doesn't make use of your Weather object - I'm guessing because it made things difficult to test?

Might be worth thinking about what you would do in order to introduce it here and test it well based on what you've learned this week.

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def depart_plane(plane)
+    fail "No Planes in the hangar!" if @hangar.empty?
+    @hangar.delete(plane)
+    $plane_status = true
+    plane
+  end
+
+  def land_plane(plane)
+    fail "Hangar is full!" if @hangar.count >= @capacity
```
samjones1001: It's a stylistic thing more than anything, but I like extracting little checks like this out into private methods, as I think it helps readability.  So, you could have something like:

```ruby
def land_plane(plane)
  fail "Hangar is full!" if hangar_full?
end

private
def hangar_full?
  @hangar.count >= @capacity
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "./lib/airport.rb"
+require "./lib/plane.rb"
+describe Airport do
+  before(:each) do
+    @airport = Airport.new
+    @plane = Plane.new
```
samjones1001: Do you feel as though you could create fully isolated tests at this point?

PR #1181, opened 2018-10-08
Airport challenge

hurc87:

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,14 +1,15 @@
 require_relative "plane"
 
 class Airport
+  attr_reader :plane
+
   def release_plane
-    Plane.new
+    fail "No planes at the airport to take off" unless @plane
+    @plane
   end
 
   def land(plane)
     @plane = plane
   end
-
-  attr_reader :plane
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -5,4 +5,10 @@ class Airport
     Plane.new
   end
 
+  def land(plane)
+    @plane = plane
+  end
+
+  attr_reader :plane
+
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,8 @@
+require_relative "plane"
+
+class Airport
+  def release_plane
+    Plane.new
+  end
+
+end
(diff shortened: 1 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,47 @@
+require_relative "plane"
+
+class Airport
```
samjones1001: This looks like a very good class overall.  I like that the methods are all short and clearly named, so it's very east to read and clear to me what each element of the code is doing.  I also really like the fact that you've broken some checks out into little private methods to further increase readability.

The only problem is that right now the class is not *cohesive* - it's responsible for maintaining a list of landed planes _*and*_ determining the weather.

Given what you've learned this week, can you think of a sensible class extraction you could make here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,72 @@
+require 'airport'
+
+describe Airport do
+  it { is_expected.to respond_to :release_plane }
```
samjones1001: These type of respond_to tests are a great first step in the TDD process, but I think they can be deleted once you have better tests in place.

For example, you have a test below to ensure that the release_place method does what it's supposed to do.  If your Airport object couldn't respond to release_plane, that test would fail - you're _*implicitly*_ testing that the object can respond to the method.

spec/airport_spec.rb
```diff
@@ -0,0 +1,72 @@
+require 'airport'
+
+describe Airport do
```
samjones1001: Looks like this class is really well tested, which suggests a good TDD process, although they're not fully isolated. Do you feel like you could fix this based on what you've learned this week?

spec/airport_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  it { is_expected.to respond_to :release_plane }
+
+  it { is_expected.to respond_to(:land).with(1).argument }
+
+  it { is_expected.to respond_to :weather_selector }
+
+  it { is_expected.to respond_to :stormy? }
+
+#  it "lands the plane" do
```
samjones1001: This is a small thing, but try to delete commented out code before you push to github - it helps keep your codebase looking tidy and professional 🙂

PR #1184, opened 2018-10-08
David Best airport challenge

davidbest1882:


lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+class Weather
+
+  def check
```
samjones1001: looks like you had this method working as expected, then changed it to always return stormy 🙂 

I'm guessing that's because you were having trouble with stubbing out the randomness involved - do you feel more comfortable with how this would work now? Think about going back and fixing this if you have the time.

spec/plane_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'plane'
+
+describe Plane do
```
samjones1001: I'm not seeing a plane class 🙂

spec/weather_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'weather'
+
+describe Weather do
+
+  it "returns that the weather is sunny" do
```
samjones1001: Again, see if you're now able to go back and figure out how to make these tests work.

PR #1186, opened 2018-10-08
Moishe Airport Challenge

bmoishe:
I have reached User Story 6.


lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+class Airport
```
samjones1001: Overall, this looks like a well written class - all your methods are short and very clear about what they're intended to do, so it's very clean, readable code.

My only problem is that, right now, it's not particularly cohesive - it's responsible for maintaining a list planes _*and*_ determining the weather. Based on what you've learned this week, maybe you can think of a sensible class extraction you could make here to improve this?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+    raise "You cannot takeoff. It is too stormy" unless weather != "stormy"
+    @hanger.delete(plane)
+  end
+
+  def weather
+   ["stormy","sunny"].sample
+  end
+end
+
+def reached_capacity
```
samjones1001: I like that you've broken out this check into it's own method, but I feel like it could be made private - I don't think it will ever need to be called from outside the class.

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
```
samjones1001: Generally, I'd advise against having completely empty objects like this.

samjones1001: Unless you were planning to put something in here but just ran out of time, in which case don't worry 🙂

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    allow(airport).to receive(:weather) { "stormy" }
+    expect { airport.land(plane) }.to raise_error "You cannot land. It is too stormy"
+  end
+
+ # User Story 5
+  it 'I want to prevent landing when airport is full' do
+    airport = Airport.new
+    plane = Plane.new
+    allow(airport).to receive(:weather) { "sunny" }
+    allow(airport).to receive(:reached_capacity) { "full" }
```
samjones1001: stubbing out methods on the class under test like this is generally not a good idea - we want to make sure that the class actually works, and doing this makes it harder to do that.

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'airport.rb'
+
+# User Story 1
+describe Plane do
+
+end
+
+describe Airport do
```
samjones1001: Your tests here look pretty good - nice work on figuring out stubbing!

At the moment, though they're not isolated. Do you feel like you could improve this given what you've learned this week?

PR #1187, opened 2018-10-08
airport challenge

Captom4:


airport.rb
```diff
@@ -0,0 +1,17 @@
+class Airport
+  attr_reader :planes
+  attr_accessor :capacity
+  def initialize(capacity: DEFAULT_CAPACITY, weather: weather.new)
```
EdwardAndress: Nice use of named parameters and dependency injection (for weather).  Does it work though?

airport.rb
```diff
@@ -0,0 +1,17 @@
...
+    @planes = Array.new
+  end
+  def land_plane(plane)
+
+  end
+  def takeoff_plane(plane)
+
+  end
+  private
+  DEFAULT_CAPACITY = 4
```
EdwardAndress: Nice use of a default constant.

plane.rb
```diff
@@ -0,0 +1,22 @@
+class Plane
+  attr_reader :landed
+  def initialize(landed = false)
```
EdwardAndress: So the plan starts off in the air?

spec/plane_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rspec'
+require_relative '../plane'
+
+describe Plane do
```
EdwardAndress: Instead of defining `@plane` you can use `subject` which is automatically created by RSpec when you use the `describe Plane do` syntax.  This is because `described_class` will be equal to `Plane` and RSpec does `subject = described_class.new` in the background.

This is only helpful in cases where you do not pass parameters to `.new`

weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
```
EdwardAndress: Nice implementation of `Weather`

spec/weather_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+require_relative '../weather'
+
+describe Weather do
+  subject(:weather){described_class.new}
+  describe '#new' do
+    it 'should create a new weather object' do
+      expect(described_class).to eq(Weather)
+    end
+  end
+  describe '#stormy?' do
```
EdwardAndress: In this case, the method could return `true` every single time and your test would still pass.  So testing for one thing `or` another isn't very helpful.

See if you can use some mocking here.

spec/plane_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      expect(@plane.landed).to eq(false)
+    end
+  end
+  describe '#land' do
+    describe 'with landed equal to false' do
+      it 'should return true' do
+        expect(@plane.land).to eq(true)
+      end
+    end
+    describe 'with landed equal to true' do
```
EdwardAndress: When you're testing a method there might be several things that you expect to happen.  In this case there's the return value, which you're testing for here, but there's also a change to the state of the plane (if the conditions are right it will land).  Can you add some tests to cover those bits of behaviour?

EdwardAndress:
Hey Tom, nice attempt. There's lots of good things in here.  The tests are structured well but you could probably achieve better coverage of the behaviour in places.  Please see the comments I made inline against your code.

PR #1188, opened 2018-10-08
Complete!

Coombszy:


lib/airport.rb
```diff
@@ -0,0 +1,39 @@
+require './lib/airtraffic'
+require './lib/plane'
+
+class Airport
+  attr_reader :at_control
+
+  def initialize
+    @capacity = 20
+    @at_control = Airtraffic.new
```
EdwardAndress: Nice.  The next step is to use Dependency Injection.  See if you can find out how to do that!

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
+class Plane
+  attr_reader :in_flight
+
+  def initialize(in_flight)
```
EdwardAndress: Can this default to `false`?

spec/airport_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'airport'
+require 'airtraffic'
+require 'plane'
+
```
EdwardAndress: The typical RSpec structure has a describe block for each method and then `it` blocks for each test relating to that method.  Something like this...

```
describe "#method" do
  it "does something" do
    expect(subject.method).to eq "HELLO WORLD"
  end
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+end
+
+def plane_do_before
+  Plane.new(true)
+end
+
+# behaviour tests
+describe Airport do
+  it 'full? returns false (no plans are landed)' do
+    expect(subject.full?).to eq(true).or(eq(false))
```
EdwardAndress: You should just be testing for one outcome here, rather than `true` or `false`.  You can create different contexts to test for behaviour in different circumstances.

EdwardAndress:
Nice work overall.  I think that if you structure your tests as I described, that will help you to focus on testing one method at a time, in different contexts.  Let me know if you have any questions.

PR #1190, opened 2018-10-08
CG - Pull Request

Charley-Gooch:
my pull request my dudes


EdwardAndress:
Hey Charley,

Nice work!

I'm going to leave a few comments inline against your code.  Let me know if you have any questions about them.

lib/Airport.rb
```diff
@@ -0,0 +1,45 @@
+class Airport
+  @airportName = nil
```
EdwardAndress: You don't need to assign nil to these instance vars here.

EdwardAndress: The names become slightly tautologous (you end up doing `airport.airportName` but `airport.name` would be equally clear).

lib/Airport.rb
```diff
@@ -0,0 +1,45 @@
...
+  @airportId = nil
+  @airportPlaneCapacity = nil
+  @airportPassangerCapacity = nil
+  @airportWeather = nil
+    
+  def initialize(airportName, airportId, airportPlaneCapacity, airportPassangerCapacity, airportWeather = "clear skies")
```
EdwardAndress: Once you have more than a couple of parameters, it's worth considering the use of named parameters.

```
def initialize(name: name, id: id, plane_capacity: plane_capacity, passenger_capacity: passenger_capacity, weather: 'clear skies')
```

The advantage is that when you call `.new` you don't need to remember in what order the parameters should provided.

lib/Main.rb
```diff
@@ -0,0 +1,37 @@
+require_relative "Plane"
```
EdwardAndress: This is useful but we'd normally use a feature test here.

lib/Airport.rb
Charley-Gooch: makes sense

lib/Airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    @airportName = airportName
+    @airportId = airportId
+    @airportPlaneCapacity = airportPlaneCapacity
+    @airportPassangerCapacity = airportPassangerCapacity
+    @airportWeather = airportWeather
+  end
+
```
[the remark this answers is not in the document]
Charley-Gooch: i'm simply used to using these from Java
a better language

lib/Airport.rb
```diff
@@ -0,0 +1,45 @@
...
+  def setAirportPlaneCapacity(capcity)
+    @airportPlaneCapacity = capcity
+  end
+  
+  def setAirportWeather(weather)
+    @airportWeather = weather
+  end
+  
+#Getters
+  def getAirportId()
```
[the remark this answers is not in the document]
Charley-Gooch: again java syntax

EdwardAndress:
There are some nice things in here and you've clearly picked up Ruby very quickly.  Each time you learn a new language, however, it's important to also learn the idioms of that language.  Using rubocop will help a little, as will reading other people's code.  In the list of pull requests you'll see solutions from a ton of other students, in different cohorts.

As you, yourself, inferred during the code review, there are some classes here which are perhaps not entirely necessary at this point in the application's development.

Finally... where are the tests?

PR #1191, opened 2018-10-08
Almost complete Airport Challenge

amarzook1:


EdwardAndress:
Hey Ahmed!  Nice effort.  I'm going to leave some comments inline against the code.  Watch this space 👀 !

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'plane.rb'
+require_relative 'weather.rb'
+
+class Airport
+  
+  def initialize
+    @planes = []
+    @parked = false
+    @weather = Weather.new
```
EdwardAndress: Good call to make a weather class.  The next step would be to use dependency injection.  See if you can find out what that means and how to use it.

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'plane.rb'
+require_relative 'weather.rb'
+
+class Airport
+  
+  def initialize
+    @planes = []
+    @parked = false
```
EdwardAndress: What does `@parked` refer to?  Whether or not there are planes in the hangar?

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+  def takeoff
+    fail 'Weather is Stormy' if stormy?
+    @parked = false
+    @planes.pop
+  end
+  
+  attr_reader :parked
+  
+  def stormy?
+    @weather == true
```
EdwardAndress: `@weather` refers to an instance of the `Weather` class.  So `@weather == true` is never going to return true.  How can you fix this by using one of the methods defined in the `Weather` class?

spec/weather_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "weather.rb"
+
```
EdwardAndress: We covered some stuff about stubbing in the code review, which should help you to improve your tests here.

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'airport'
+
+describe Airport do
+  
+  it 'responds to land' do
+    expect(subject).to  respond_to(:land).with(1).argument
+  end
+
```
EdwardAndress: The typical RSpec structure is to have a `describe` block for each method you're testing, then an `it` block for each test relating to that method.

Something along these lines...

```
describe '#land' do
  it 'lands a plane and adds it to the planes array' do
    subject.land(plane)
    expect(subject.planes).to eq [plane]
  end
end
```

spec/plane_spec.rb
```diff
@@ -0,0 +1,3 @@
+describe Plane do
```
EdwardAndress: If there's no behaviour to encapsulate, maybe you don't yet need a plane class.

lib/airport.rb
amarzook1: @weather.check

lib/airport.rb
amarzook1: yeah thinking of renaming it to location for better readability and putting it in the plane class instead to refactor better

spec/plane_spec.rb
amarzook1: going to add the location to the plane class instead to refactor my code a bit.

PR #1194, opened 2018-10-10
Pull Request - Phipax

phipax:


lib/airport.rb
```diff
@@ -0,0 +1,54 @@
+#require './lib/plane'
+class Airport
+ attr_reader :landedplanes_register
+ attr_reader :capacity
+
+
+ @@planes_inflight = Array.new
```
samjones1001: Holding this information in the Airport class feels a little odd to me - it implies that the airport is responsible for managing the planes which are currently in the air.

I think I would look at just having an instance variable on each Plane object which tracks if it's in flight or not - I think you could achieve the same effect this way.

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+ #@@airport_register = Hash.new
+
+
+
+ def initialize(capacity = 10)
+  @capacity = capacity
+  @landedplanes_register = Hash.new
+ end
+
+ def land_plane(plane, weather)
```
samjones1001: This and your take_off method are both very long - the general concensus in the Ruby community is that we should try to keep methods to 5 lines or less:

The reasoning for this is that as methods get longer, the likelihood that they have more than one responsibility increases - is there any behaviour in either of these methods that could be extracted out into a separate method?

lib/plane.rb
```diff
@@ -0,0 +1,12 @@
+class Plane
+ def initialize
```
samjones1001: This is a very small point, but unless you want to pass in some arguments, or set up some instance variables, it's not necessary to write the initialise method - Ruby will do it for you.

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+  subject(:airport) {described_class.new}
+  let(:plane) {double :plane}
```
samjones1001: looks like you were all set up to use a double to represent your plane, but then used real instances of planes in your tests anyway! 🙂 

how might you go about making your Airport tests completely isolated?

PR #1196, opened 2018-11-02
Early pull request to highlight errors

alittlecross:


alittlecross:
Newer versions of Gems have been installed when using Bundle.

PR #1201, opened 2018-11-02
Testing challenge - randomisation of weather with 'mocking' tests required but seeing where I am 

aarongill1:


lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  attr_reader :planes_in_airport
+  attr_accessor :weather
+  attr_accessor :plane_capacity
+
+  DEFAULT_PLANE_CAPACITY = 50
+
+  def initialize
+    @planes_in_airport = []
+    @weather = Weather.new
```
samjones1001: This is something you may we'll have realised by working through the materials last week, but creating another object within an initialise method like this is not the ideal way to do it because it creates a very hard dependency between the two objects - you can never create an Airport without also creating a Weather.

Imagine a scenario in which you want to be able to provide different weather objects depending on where the airport is located. So a plane in the Arctic might need an ArcticWeather object, whereas one neat the equator would take a TropicalWeather object.  Right now, there's no way for you to do this.

You could fix the problem using dependency injection:

```ruby
def initialize(weather = Weather)
  @planes_in_aiport = []
  @weather = weather.new
  @plane_capacity = DEFAULT_PLANE_CAPACITY
end
```

This pattern gives you extra flexibility.  You can still do
```ruby 
airport = Airport.new
```

and use the default argument to create an airport with a normal Weather object, but you could also do something like

```ruby
arctic_airport = Airport.new(ArcticWeather)
```

This also has the added benefit of making your tests easier to isolate

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @plane_capacity = DEFAULT_PLANE_CAPACITY
+  end
+
+  def plane_is_added_to_available_fleet(plane)
```
samjones1001: I really like that you've extracted all of these little checks out into clearly named methods - I think it helps with readability.  However, my guess is that some of them could probably be made private.

As a general rule, unless you expect that a method is every going to be called either by a user or by another object you should default to making it private.  This stops users from doing unexpected things with your objects and makes their interfaces clearer and easier to understand.

samjones1001: Actually, now I've looked at your plane class I can see that it seems to be making use of most of these, so maybe they can't be private after all 🙂

lib/plane.rb
```diff
@@ -0,0 +1,16 @@
+class Plane
+
+  def land(airport)
+    fail 'Plane cannot land as there is no room' if airport.full?
+    fail 'Stormy so the plane cannot land' unless airport.check_weather?
```
samjones1001: This is a very minor thing, but I think the check_weather? method could be more clearly named.  The question mark at the end indicates to me that it should return true or false, but its not obvious at first under what conditions it would return true.

Maybe `weather_is_clear?` or something similar would communicate better?

spec/airport_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'airport.rb'
+
+describe Airport do
```
samjones1001: This class feels a little bit under tested at the moment.  There are a bunch of publicly facing methods in Airport without a corresponding unit test.  I think you probably have tested them through your Place spec, but given that the behaviour is in your Airport class they should really be getting tested here.

Again, you may have learned more about this and about testing your classes in isolation over the past week, but if it's not clear why this is important feel free to come and grab me for a chat!

spec/weather_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'weather.rb'
+
+describe Weather do
+  let(:weather) { double :weather }
+  it 'checks that a Weather object exists' do
+    expect(subject).to be_instance_of Weather
+  end
+  it 'checks random weather choice' do
+    allow(subject).to receive(:weather).and_return("sunny")
```
samjones1001: Great to see that you're figuring out how to use stubs already but stubbing the class under test like this is generally not a great way to create meaningful tests.

If you think about what you're doing here, you're specifying that when you call .weather on the object that it should return "sunny", then checking that it returns "sunny" - rather than testing your object's behaviour, your effectively testing that Rspec's stubbing functionality works 🙂 

Stubbing is super useful when you're testing interactions between objects and keeping your tests isolated.  For this use case, I'd probably look at using srand to stub out the randomness.

PR #1203, opened 2018-11-02
Airport Challenge

Ajay233:
Wrote unit tests and code to satisfy the requirements identified in the user stories:
- Can create a new Plane
- Can create a new Airport
- Can land a plane which will add the plane to the landing strip (the airport's array)
- A plane can take off which will remove it from the landing strip
- Errors will be thrown a Plane tries to land or take off when the weather is stormy
- Planes can not land of the landing strip is full
- The max capacity for the airport can be increased if a developer needs to
- Added a weather constant, randomised weather, refactored code slightly to accommodate changes, fixed broken tests using stubbing and used rubocop to get code in correct format.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,2 @@
+class Plane
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,2 +1,6 @@
 class Plane
+
+  def land
+  end
+  
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,6 +1,7 @@
 class Plane
 
   def land
+    "Safely landed"
   end
-  
+
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -13,21 +13,26 @@ class Airport
   def land(plane)
     fail 'Sorry, no room' if @landing_strip.length == @max_capacity
     fail 'Sorry, too stormy' if WEATHER.sample == "Stormy"
-    # plane.status = "landed"
+    plane.flying = false
     @landing_strip << plane
     "Safely landed"
   end
 
   def take_off(plane = landing_strip.last)
-    fail 'Sorry, too stormy' if WEATHER.sample == "Stormy"
-    fail 'Already in the air' if plane.flying?
+    fail 'Sorry, too stormy' if stormy?
+    fail 'Can not take off: Plane is in the air' if plane.flying?
+    fail 'Can not take off: Plane is not at the airport' unless plane_at_terminal?(plane)
     @departure = @landing_strip.pop
-    departure.flying = true
+    @departure.flying = true
     "flight number #{departure} is no longer at the airport"
   end
 
-  def flying?
-    @flying
+  def stormy?
+    WEATHER.sample == "Stormy"
+  end
+
+  def plane_at_terminal?(plane)
+    @landing_strip.include?(plane)
   end
 
 end
@@ -35,8 +40,8 @@ end
 class Plane
   attr_accessor :flying
 
-  def initialize
-    @flying = false
+  def initialize(flying = false)
+    @flying = flying
   end
 
   def flying?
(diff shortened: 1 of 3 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,52 @@
+WEATHER = ["Sunny", "Stormy"]
```
samjones1001: Having an airport determine its own weather arguably feels reasonable, but it probably means that this class is violating the single responsibility principle.

Right now it's managing landed planes *and* determining the weather.  The suggests that a Weather class would probably be a good extraction - it would also have the added benefit of making testing the randomness a bit easier.

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+  attr_reader :departure
+  attr_accessor :weather, :max_capacity, :landing_strip
+
+  def initialize(landing_strip = [Plane.new], max_capacity = DEFAULT_CAPACITY)
+    @landing_strip = landing_strip
+    @max_capacity = max_capacity
+  end
+
+  def land(plane)
+    fail 'Sorry, no room' if @landing_strip.length == @max_capacity
```
samjones1001: This is more a personal preference than a hard and fast rule, but I like to extract little checks like this out into small private methods for readability.  So something like:

```ruby
fail 'Sorry, no room' if airport_full?
```

Then, at the bottom of the class

```ruby
private
def airport_full?
  @landing_strip.length == @max_capacity
end
```

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+    fail 'Sorry, too stormy' if stormy?
+    fail 'Can not land plane: Plane is already on the ground' unless plane.flying?
+    plane.flying = false
```
samjones1001: Directly accessing the instance variables of another object like this is generally best avoided.  A better solution would be to have a method on your plane class which you could call from here which then does the same thing.

So over in plane.rb you could have 
```ruby
def land
  @flying = false
end
```

then here you could just call `plane.land`

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+    @landing_strip << plane
+    "Safely landed"
+  end
+
+  def take_off(plane = landing_strip.last)
+    fail 'Sorry, too stormy' if stormy?
+    fail 'Can not take off: Plane is in the air' if plane.flying?
+    fail 'Can not take off: Plane is not at the airport' unless plane_at_terminal?(plane)
+    @departure = @landing_strip.pop
```
samjones1001: This line makes total sense so long as I'm using the default argument for the message, but imagine a situation in which the Airport's landing strip looks like this
```ruby
[plane_1, plane_2, plane_3]
```

If I run the method, passing in plane_1, plane_3 will still be the one which actually takes off 🙂

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+    @departure.flying = true
+    "flight number #{departure} is no longer at the airport"
+  end
+
+  def stormy?
```
samjones1001: This method and the one below could probably be made private.  Generally speaking, unless you explicitly expect that a method will be called by a user or another object, default to making it private.

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+    WEATHER.sample == "Stormy"
+  end
+
+  def plane_at_terminal?(plane)
+    @landing_strip.include?(plane)
+  end
+
+end
+
+class Plane
```
samjones1001: Again, a very small thing, but I'd recommend splitting your classes into separate files.  When you get to building larger systems, keeping them in the same file in this way could lead to you having thousands of lines of code which is going to be pretty difficult to navigate.

spec/airport_spec.rb
```diff
@@ -0,0 +1,92 @@
+require 'airport'
+
+describe 'Airport' do
+  let(:airport) { Airport.new }
+  let(:plane) { Plane.new }
```
samjones1001: Based on what you did in week 2, do you think you could now properly isolate these tests using a double in place of your Plane object?

PR #1204, opened 2018-11-04
First pull request

topickering:
First pull request after initial steps

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -13,7 +13,7 @@ class Airport
   end
 
   def land(plane)
-    fail 'This Airport is full' if full?
+    fail 'This airport is full' if full?
     current_weather
     fail 'Too stormy to try landing' if stormy?
     @planes << plane.land(self)
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -13,13 +13,14 @@ class Airport
   end
 
   def land(plane)
+    fail 'This Airport is full' if full?
     current_weather
     fail 'Too stormy to try landing' if stormy?
-    fail 'This Airport is full' if full?
     @planes << plane.land(self)
   end
 
   def take_off(plane)
+    fail 'Plane not at this airport' unless @planes.include?(plane)
     current_weather
     fail 'Too stormy to try take off' if stormy?
     @planes.delete(plane.take_off)
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,17 @@
+class Airport
+
+  attr_accessor :weather, :capacity
+
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @weather = current_weather
+    @capacity = capacity
+  end
+
+  def current_weather
+    set = rand(10)
+    set > 0 ? @weather = 'fine' : @weather = 'stormy'
+  end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -16,14 +16,14 @@ class Airport
     fail 'This airport is full' if full?
     current_weather
     fail 'Too stormy to try landing' if stormy?
-    @planes << plane.land(self)
+    @planes << plane.landed(self)
   end
 
   def take_off(plane)
     fail 'Plane not at this airport' unless @planes.include?(plane)
     current_weather
     fail 'Too stormy to try take off' if stormy?
-    @planes.delete(plane.take_off)
+    @planes.delete(plane.taken_off)
   end
 
   private
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,17 +1,43 @@
+require_relative 'plane'
+
 class Airport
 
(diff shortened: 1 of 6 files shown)


topickering:
I have reviewed my tests and have fixed the bug which was preventing certain tests (which had to account for the randomisation of weather) from functioning as required.

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+    fail 'Plane not at this airport' unless @planes.include?(plane)
+    current_weather
+    fail 'Too stormy to try take off' if stormy?
+    @planes.delete(plane.taken_off)
+  end
+
+  private
+
+  def current_weather
+    set = rand(10)
```
Katerina-codes: What does "set" mean - could this have a better name? :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
+require 'airport'
+
+describe Airport do
+
+  let(:plane) { double :plane }
+  let(:plane_two) { double :plane }
+
+  it 'sets a default capacity' do
```
Katerina-codes: Is this test needed? We know that the capacity will be set to the default capacity because that's how you set the class up :)
This is an example of testing state not behaviour. We know that Airports's default capacity (state) will always be 20. 
If you wanted to test "behaviour" then a method (with some specific behaviour like **reduce_capacity(capacity)** ) that affects that state would be the way of testing capacity is going up or down as we expect.

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+
+  it 'allows the setting of a capacity' do
```
Katerina-codes: Check my comment above - do you think the same applies here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    capacity = rand(1..35)
+    expect(Airport.new(capacity).capacity).to eq capacity
```
Katerina-codes: Check my comment above - do you think the same applies here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  describe '#land' do
+    it 'stores landed planes at the airport' do
+      allow(plane).to receive(:landed)
+      allow(subject).to receive(:stormy?) { false }
+      n = rand(1..5)
```
Katerina-codes: What is 'n'? Could we give it a better name :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      n.times { subject.land plane }
+      expect(subject.planes.length).to eq n
```
Katerina-codes: I think there's an example in the tests where a loop is used like you have here - so sorry about the truth bomb that's coming next but I would actually suggest avoiding using logic in your tests as much as possible. Tests should be as simple as possible and as specific as possible. 
You could test exactly the same thing here on a fewer number of planes (the behaviour is what's important - **landed planes are stored** - not the number of planes that land) and just set up the test in pretty much the same way just minus the loop.

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+    it 'changes the status of the landed plane' do
+      allow(subject).to receive(:stormy?) { false }
+      allow(plane).to receive(:landed)
+      allow(plane).to receive(:status).and_return(subject)
+      subject.land plane
+      expect(plane.status).to eq subject
```
Katerina-codes: Could you explain this test to me please? Maybe via slack :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      expect(subject.planes.include?(plane)).to eq false
+    end
+    it 'will only allow a plane to take off if at that airport' do
+      allow(plane).to receive(:landed).and_return(plane)
+      airport_1 = Airport.new
+      airport_2 = Airport.new
+      allow(airport_1).to receive(:stormy?) { false }
+      allow(airport_2).to receive(:stormy?) { false }
+      airport_1.land(plane)
+      expect { airport_2.take_off(plane) }.to raise_error 'Plane not at this airport'
```
Katerina-codes: Are two airports needed to test this behaviour? "will only allow a plane to take off if at that airport"

spec/plane_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  let(:airport) { double :airport }
+
+  it { is_expected.to respond_to(:landed).with(1).argument }
+
+  describe '#landed' do
+    it 'is only possible to land a flying plane' do
+      subject.landed(airport)
+      expect { subject.landed(airport) }.to raise_error 'This plane has already landed'
+    end
+    it 'is possible to land a plane at a specific airport' do
```
Katerina-codes: Correct the spacing here

PR #1207, opened 2018-11-04
Airport_challenge

leoncross:


lib/weather.rb
```diff
@@ -0,0 +1,14 @@
...
+  def conditions
+    randomizer == 1 ? "Stormy" : "Sunny"
+  end
+
+  def randomizer
+    return rand(1..5)
+  end
+
+end
+# weather = Weather.new
```
Katerina-codes: This is dead code - could you remove it please?

spec/airport_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+  end
+
+  it "confirms that the default landing capacity can be overwridden" do
+    airport = Airport.new(30)
+    expect(airport.capacity).to eq 30
+  end
+
+  it "confirms that the plane has landed at the airport (via name storage)" do
+    plane = Plane.new
+    airport = Airport.new
```
Katerina-codes: Do you need a new instance in each test? Can you DRY this up?

PR #1209, opened 2018-11-04
Airport Challenge

kirillzabrodin:
Still working on README but project complete otherwise

### DIFF
--- a/lib/aeroplane.rb
+++ b/lib/aeroplane.rb
@@ -1,5 +1,10 @@
 require_relative 'airport'
 
 class Aeroplane
-  
+  DEFAULT_STATE = 'flying'
+  attr_reader :status
+
+  def initialize(state = DEFAULT_STATE)
+    @status = state
+  end
 end
--- a/lib/aeroplane.rb
+++ b/lib/aeroplane.rb
@@ -14,4 +14,16 @@ class Aeroplane
     @status = where
   end
 
+  def takeoff
+    @status = 'flying'
+  end
+
+  def airport
+    @status != 'flying' ? @status : false
+  end
+
+  def flying
+    @status == flying
+  end
+
 end
--- a/lib/aeroplane.rb
+++ b/lib/aeroplane.rb
@@ -1,0 +1,5 @@
+require_relative 'airport'
+
+class Aeroplane
+  
+end
--- a/lib/aeroplane.rb
+++ b/lib/aeroplane.rb
@@ -4,13 +4,11 @@ class Aeroplane
 
   DEFAULT_STATE = 'flying'
 
-  attr_reader :status
-
   def initialize(state = DEFAULT_STATE)
     @status = state
   end
 
-  def land(where = 'airport')
+  def land(where)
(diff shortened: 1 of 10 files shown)


lib/aeroplane.rb
```diff
@@ -0,0 +1,27 @@
...
+  def initialize(state = DEFAULT_STATE)
+    @status = state
+  end
+
+  def land(where)
+    @status = where
+  end
+
+  def takeoff
+    @status = 'flying'
```
Katerina-codes: Can you DRY up all these strings please? :)

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    aeroplane.takeoff
+  end
+
+  private
+
+  def full?
+    @aeroplanes.count >= @capacity
+  end
+
+  def set_weather
```
Katerina-codes: Should the airport have the responsibility of setting the weather?

spec/aeroplane_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+describe Aeroplane do
+
+  it 'can tell it is in the air' do
+    expect(subject.flying).to eq true
+  end
+
+  it 'can tell it is at an airport' do
+    subject.land('An airport')
+    expect(subject.airport).to eq 'An airport'
```
Katerina-codes: Can you explain this please?

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      expect { subject.takeoff(aeroplane_other) }.to raise_error "Aeroplane not at airport"
+    end
+  end
+
+  describe ' #capacity' do
+
+    it "capacity can be changed" do
+      larger_airport = Airport.new(21)
+      larger_airport.storm = false
+      21.times { larger_airport.land(aeroplane) }
```
Katerina-codes: - I would suggest avoiding having logic (like this loop) in your tests where you can - I know that the Makers shows an example with a loop but including logic makes the test more complicated and they should be as simple as possible. You can test this same behaviour without a loop by setting the capacity to a small number (for example 2) and setting up the test with 2 planes being landed.

- This test is kind of doing two things (which is fine!), checking 'capacity can be changed' but also checking that an error is raised when the airport is full. I would suggest changing the test name to the something more about raising an error when capacity is reached as that seems to be more what the test does and the modifiable capacity is  tested as a by-product of that anyway.

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      expect { larger_airport.land(aeroplane_other) }.to raise_error "#{described_class.name} full"
+    end
+
+    it "defaults capacity of #{described_class::DEFAULT_CAPACITY}" do
```
Katerina-codes: This test seems unnecessary, we know that if we don't override the default capacity then it will be what we set the default to be. This test name seems to not be doing what you say it is and when you look at the expectation it is a duplication of the test above.

PR #1217, opened 2018-11-04
Airport_challenge

lucafrancesc:


lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY, status = 'safe')
+    @capacity = capacity
+    @hangar = []
+    @condition = status
+  end
+
+  def safe?(status = 'yes')
+    status == 'yes' ? (@condition = 'safe') : (@condition = 'not safe')
+    @condition == 'safe' ? true : false
```
jonurry: There's a couple of issues here:
@condition seems to be a bit confusing. In your initialise method it is set to 'safe'. Then it is set to either 'safe' or 'unsafe' in the safe? method before being transformed into a boolean. You should decide on what you want @condiotion to represent and then stick to it.
The line:
  @condition == 'safe' ? true : false
is the same as:
  @condition == 'safe'
Line 15 seems almost redundant...

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      expect(subject.safe?('yes')).to eq true
+    end
+
+    it 'is safe not to fly or land' do
+      expect(subject.safe?('no')).to eq false
+    end
+
+    it 'is not safe to land' do
+      subject.safe?('no')
+      allow(subject).to receive(:safe?).and_raise('It is not safe to land')
```
jonurry: I'm not sure what this test is doing? You set up the conditions to receive an error message but then you don't test anything? Allow is usually used when mocking. To test an assertion you use expect.

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    it 'is not safe to fly' do
+      plane = Plane.new
+      subject.landing(plane)
+      subject.safe?('no')
+      allow(subject).to receive(:safe?).and_raise('It is not safe to fly')
```
jonurry: Again, this isn't really testing anything. You need to use 'expect' instead of 'allow' to make an assertion. Take some time to look into what allow and expect do and when to use them.

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  describe '#hangar' do
+    it { is_expected.to respond_to :hangar }
+
+    it 'returns an empty hangar' do
+      expect(subject.hangar).to eq []
+    end
+
+    it 'returns the plane(s) in the hangar' do
+      plane = Plane.new
```
jonurry: Ideally, you would use a double (or mock) to represent the Plane. When testing the Airport it should not rely on the implementation of another object. Here, you are using an actual Plane. Use a double to ensure that you don't introduce errors by relying on the implementation of another object.

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+  end
+
+  describe '#landing' do
+    it { is_expected.to respond_to(:landing).with(1).argument }
+
+    it 'returns the plane landed' do
+      plane = Plane.new
+      subject.safe?('yes')
+      expect(plane.status).to eq('flying')
```
jonurry: You are making multiple assertions here. It is best to make one assertion per unit test. You could split this into 4 separate tests and setup each with a beforeEach() block.

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+      plane.taken_off
+      @hangar.include?(plane) ? @hangar.delete_at(@hangar.index(plane)) :
+      "Plane not in the hangar"
+      plane
+    else
+      fail 'It is not safe to fly'
+    end
+  end
+
+  private
```
jonurry: Nice use of private methods

lib/plane.rb
```diff
@@ -0,0 +1,17 @@
+class Plane
+
+  attr_reader :status
+
+  def initialize(status = 'flying')
+    @status = status
+  end
+
+  def landed(confirm = 'yes')
```
jonurry: It seems a bit weird that you pass in a parameter that determines whether the plane can land or not. I guess that this is an intermediate solution until you get the Weather working correctly?

PR #1221, opened 2018-11-05
Unfinished challenge

AsadK47:


MHUS25:
Feedback from code review:
- Commits show a good TDD process following the red-green-refactor loop
- Good refactoring with use of private methods
- Code is readable and method names are clear
- Plane, Airport & Weather are separate classes with single responsibilities

PR #1234, opened 2018-11-28
pass first test - planes land at airport

scass91:


README.md
```diff
@@ -13,6 +13,57 @@ Airport Challenge
...
+ => true
+2.5.0 :011 > airport.take_off(plane)
+Traceback (most recent call last):
+        3: from /Users/simon/.rvm/rubies/ruby-2.5.0/bin/irb:11:in `<main>'
+        2: from (irb):17
+        1: from /Users/simon/Projects/airport_challenge/lib/airport.rb:21:in `take_off'
+RuntimeError (Unable to take off)
+```
+
+
```
stonefarmer9: This is a really informative and well written readme!

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    plane.update_status
+    @planes << plane
+  end
+
+  def take_off(plane)
+    fail "Unable to take off" if no_planes == true || weather == true
+    fail "This plane is not at the airport!" unless @planes.include? plane
+    plane.update_status
+    @planes.pop
+  end
```
stonefarmer9: Could this method and the landing method be refactored to make the decision on weather outside of the method? Or as a new method, could it turn the methods into simpler control flow system drawing on other methods for information?

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  attr_accessor :status
+
+# start with false - links to update_status and intialises planes in air
+  def initialize(status = "In Air")
+    @status = status
+  end
+
+  def update_status
+    @status = (@status == "In Air" ? "Landed" : "In Air")
```
stonefarmer9: Why does the true value for @status return Landed? It reads like it should return In Air?

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def no_planes
+    @planes.length.zero?
+  end
+
+  def full_airport
+    @planes.length >= @capacity
+  end
+
+  def weather
```
CazaBelle: Should the airport be determining the weather?

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
...
+    @status
+  end
+
+  def landed
+    @status != "In Air"
```
stonefarmer9: Would 'Landed' be a clearer return value?

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'airport'
+
+class Plane
```
CazaBelle: Nice use of Plane class

lib/weather.rb
```diff
@@ -0,0 +1,6 @@
+class Weather
+  def stormy
+    return true if rand(100) < 40
```
stonefarmer9: Love that this is statistically accurate for the UK!

spec/airport_spec.rb
```diff
@@ -0,0 +1,83 @@
+require "airport"
+require "weather"
+
+describe Airport do
+
+  describe "#landing_plane" do
+    it "It allows planes to land" do
+      plane = Plane.new
+      @weather = allow(subject).to receive(:weather) { false }
```
stonefarmer9: I find the use of false unclear? could it be more descriptive in regards to stormy/clear?

spec/airport_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      expect(subject.landing_plane(plane)).to eq [plane]
+    end
+
+    it "it prevents landings at a full airport" do
+      airport = Airport.new
```
AlinaGoaga: I like how you organized the code around the two main functions - landing and take offs! I'd recommend perhaps using a before each to reduce the repetition of  airport = Airport.new and / or plane = Plane.new

spec/airport_spec.rb
CazaBelle: Do we use instance variables in testing?

spec/airport_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+
+    it "It initializes with default capacity given no capacity" do
+      airport = Airport.new
+      expect(airport.capacity).to eq Airport::DEFAULT_CAPACITY
+    end
+
+  end
+
+end
```
stonefarmer9: The tests look really nice and are wonderfully layed out. They seem to test almost all aspects of your program and your knowledge of testing comes across as really strong!

spec/airport_spec.rb
```diff
@@ -0,0 +1,83 @@
+require "airport"
```
CazaBelle: tests are thorough and easy to read

spec/airport_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    it "it prevents plane from taking off if the weather is bad" do
+      plane = Plane.new
+      airport = Airport.new
+      airport.landing_plane(plane)
+      allow(airport).to receive(:weather) { true }
+      expect { airport.take_off(plane) }.to raise_error "Unable to take off"
+    end
+
+    it "it raises an error when taking off a plane which is not at the airport" do
```
CazaBelle: Great test.

lib/weather.rb
AlinaGoaga: I think rand's a good option for this - it's short and meets the purpose! What I'd recommend would be just to leave it at def stormy 
          rand(100)<40 
        end 
If the expression evaluates true it will return true, without us having to use the return true bit

PR #1235, opened 2018-12-01
Completed airport_challenge

rednblack99:


lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+      @weather = "clear"
+    when 1
+      @weather = "stormy"
+    end
+  end
+
+  def land(plane)
+    fail "The weather is too rough." if @weather == "stormy"
+    fail "The hanger is full." if @hanger.length >= @capacity 
+    @hanger.push(plane)
```
Sindex42: Could '@hanger.length >= @capacity' be split into its own method?

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def weather(conditions = rand(2))
+    case conditions
+    when 0
+      @weather = "clear"
+    when 1
+      @weather = "stormy"
+    end
+  end
+
```
Sindex42: Could weather be split into its own class? Does Airport need to be responsible for determining the weather?

CelineKaslin: Does the airport the best place to determinate the weather ? 
Maybe a weather class will have been useful in this case ;)

spec/airport_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  it "prevents takeoff if weather is stormy" do
+    airport = Airport.new
+    airport.weather(1)
+    expect { airport.takeoff(Plane.new) }.to raise_error("The weather is too rough.")
+  end
+
+end
+
+describe Plane do
```
subomionanuga: Consider a separate test file for your plane class. Even if it is the only test in there

spec/airport_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  it "creates airport with default capacity of 5" do
+    Airport.new
+    expect(@capacity).equal? 5
+  end
+
+  it "prevents landing when airport is at default capacity" do
+    airport = Airport.new
+    5.times { airport.land(Plane.new) }
```
subomionanuga: Shoudn't you be able to sub '5' for DEFAULT_CAPACITY since you have that defined?

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+class Airport
+
+  DEFAULT_CAPACITY = 5
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hanger = []
+    @capacity = capacity
+  end
+
```
CelineKaslin: maybe a name for the airport to know in which one the plane is in, because plane can only take off from the airport they are in .

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
+
```
CelineKaslin: maybe a state of the pane can be add like flying and landed as flying plane cannot take off again and landed one cannot land again

spec/airport_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    plane = Plane.new
+    expect(subject.land(plane)).equal? plane
+  end
+
+  it "removes planes from hanger upon takeoff" do
+    plane = Plane.new
+    expect(subject.takeoff(plane)).equal? plane
+  end
+
+  it "removes specific plane requested from hanger upon takeoff" do
```
CazaBelle: Good test to check which plane is leaving.

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+  def weather(conditions = rand(2))
```
CazaBelle: How is weather getting the condition. Does it make sense for rand to be running in the argument?

PR #1236, opened 2018-12-02
my attempt at the weekend challenge

pbaker2018:


lib/airport.rb
```diff
@@ -0,0 +1,34 @@
+class Airport
+
+  def initialize(capacity = 1000)
```
sjmog: Did you think about extracting a constant for this magic number?

PR #1237, opened 2018-12-02
Airport Challenge

CazaBelle:
Weekend Challenge passing all RSpec tests

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -3,7 +3,7 @@ require_relative 'plane'
 class Airport
   attr_reader :capacity
 
-  DEFAULT_CAPACITY = 50
+  DEFAULT_CAPACITY = 1
   
   def initialize(capacity = DEFAULT_CAPACITY)
     @capacity = capacity 
@@ -18,7 +18,7 @@ class Airport
   end
 
   def take_off
-    #we need the plane to leave the airport 
+    @hanger.pop
   end
 
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,23 +1,34 @@
 require_relative 'plane'
+# require_relative 'weather'
 
 class Airport
   attr_reader :capacity
 
-  DEFAULT_CAPACITY = 1
+  DEFAULT_CAPACITY = 50
   
   def initialize(capacity = DEFAULT_CAPACITY)
     @capacity = capacity 
     @hanger = []
   end
-    
+  
+  def stormy?
+  x = Random.rand(1..2)
+    if x == 1
+      true
+    else
+      false
+    end
+  end
 
   def land(plane)
     @plane = plane
     @hanger.push @plane
+    @hanger
 
   end
 
   def take_off
+    raise "storm" if stormy? == true 
     @hanger.pop
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,5 +1,4 @@
 require_relative 'plane'
-# require_relative 'weather'
 
 class Airport
   attr_reader :capacity
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,27 @@
+require_relative 'plane'
+
+class Airport
+  attr_reader :capacity
+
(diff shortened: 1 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+  attr_reader :capacity
+
+  DEFAULT_CAPACITY = 50
+  
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity 
+    @hanger = []
+  end
+  
+  def stormy?
```
tomlovesgithub: Could the progam user change the chances of it being stormy easily?
How would your program work if the weather was passed in as a separate argument?

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    expect { airport.take_off(plane) }.to raise_error 'Planes grounded because of stormy weather'
+  end 
+
+  it 'allow takeoff when weather is clear' do
+    airport = Airport.new
+    plane = Plane.new
+    allow(airport).to receive(:stormy?).and_return(false)
+    expect { airport.take_off(plane) }.not_to raise_error 
+  end 
+
```
tomlovesgithub: Extensive tests which is great, consider if all of them are necessary, do they all relate to user stories? are there any overlaps?

lib/airport.rb
AlinaGoaga: Good choice on using rand to determine the weather! I'd recommend perhaps something like the below to make it a bit shorter:
def stormy?
   rand(0..1) == 1
end 
If it evaluates to true it will return that. If not you'd get false

AlinaGoaga: Sorry forgot to add (pretty much what we discussed this morning :) ). Since the airport isn't determining the weather i'd make a separate class for it

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+    x = rand(1..2)
+    if x == 1
+      true
+    elsif x == 2
+      false
+    end
+  end
+
+  def land(plane)
+    fail 'Airport full' if @hanger.length == @capacity
```
AlinaGoaga: I'd add the check on the weather in here as well, just like you did for the take offs. Perhaps it's also considering adding a check on whether the plane has already landed or not (if it tries to land again?)

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+  
+  def stormy?
+    x = rand(1..2)
+    if x == 1
+      true
+    elsif x == 2
+      false
+    end
+  end
+
```
stonefarmer9: Could the stormy weather be in another class?

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  it 'should instruct plane to land' do 
+    plane = subject.land(plane)
+    expect(plane).to eq plane
+  end
+
+  it 'should put a plane in the hanger when lands' do
+    plane = Plane.new
+    expect(subject.land(plane)).to eq [plane]
```
stonefarmer9: does this test still pass correctly? Or does it pass no matter what you expect?

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    airport = Airport.new
+    expect(airport.capacity).to eq 50
+  end
+
+  it 'should be able to set airports capacity to a desired capacity' do
+    airport = Airport.new(20)
+    expect(airport.capacity).to eq 20
+  end
+
+end
```
stonefarmer9: I really like these capacity tests, really clear and easy to read

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'airport'
```
AlinaGoaga: Good work on the test's coverage! 
I'd maybe add context / describe to organize them in groups either by behaviour or the functions that they relate to (say if you think about landings, I'd use a describe to group all of the ones that are related to that)

PR #1242, opened 2018-12-02
Airport challenge work in progress

stonefarmer9:
Work in progress, having completed 5 out of 6 user stories. All tests are passing. 
- Airport can land planes and store them
-Planes can take off from airport and are no longer present in airport
- Planes cannot land in full airport
- Airports are initialised with a capacity variable with a default of 1
-Weather is randomly generated
-Planes cannot land during stormy weather


README2.md
```diff
@@ -0,0 +1,45 @@
+# Airport Challenge
```
scass91: How does your code appear in irb? Would you be able to provide an example of this?

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+class Airport
+
+  def initialize(capacity = 1, weather = Weather.new)
+    @capacity = capacity
+    @hangar = []
+    @weather = weather
+  end
+
+  def land(plane)
+    weather = @weather.weather_condition
```
scass91: Would you be able to explain this? My understanding may be off, it looks like when the wether is stormy it returns an empty hangar - what is the reasoning for this?

scass91: Should it be a fail condition 'bad weather' instead?

README2.md
```diff
@@ -0,0 +1,45 @@
...
+
+Program for control of flow of planes through an airport, depending on whether weather is stormy or clear as gnenerated by random
```
CazaBelle: Beautiful Readme!

README2.md
```diff
@@ -0,0 +1,45 @@
...
+- The tests using doubles and mocks I feel could be made more eloquent. For now I feel they are easily readable but could be maintain that readability to be made simpler through more research.
+- I consider removing all let variables, as sometimes I feel like they do not make things easier to read, as discovered when my tests were peer reviewed and the AAA method above was introduced.
+- Further research is required to see if the second/sub describe block is necessary? It was based off of a similar test conducted previously by myself and while works for this scenario it could be made better.
+
+## Authors
+
+
+* **Georges Le Vaillant** - [GLV](https://github.com/stonefarmer9)
+
+
```
tomlovesgithub: Outstanding read me, perhaps consider adding a TDLR

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    if weather == 'stormy'
+      @hangar
+    elsif full?
+      fail 'Hangar full'
+    else
+      @hangar << plane
+    end
+  end
+
```
Sindex42: Does the 'hangar full' guard clause need to be inside a elsif statement?

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
```
tomlovesgithub: intelligent separation of weather class

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+  def initialize(capacity = 1, weather = Weather.new)
```
CazaBelle: Would setting a default_capacity (constant value) be useful in this case? http://rubylearning.com/satishtalim/ruby_constants.html

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @capacity = capacity
```
CazaBelle: Good use of instance variables

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
...
+  def weather_condition
+    rand(10).zero? ? 'stormy' : 'clear'
```
tomlovesgithub: Am i correct to say that this means its a once in ten that it could be stormy? 
how easy would it be for the chances to be changed to once in four or otherwise?

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
+require './lib/airport.rb'
```
scass91: Consider changing to require_relative?

lib/airport.rb
subomionanuga: Consider initialising a DEFAULT_CAPACITY constant. You would be able to use this in the spec file as well

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    weather = Weather.new
+    allow(weather).to receive(:weather_condition).and_return('clear', 'stormy')
+    stormy_airport = Airport.new(1, weather)
+    stormy_airport.land(plane)
+    hangar = stormy_airport.take_off
+    expect(hangar.count).to eq 1
+  end
+
+  describe 'land' do
+
```
scass91: Is this only for an airport with capacity 1? Consider removal of magic number and adding a DEFAULT_CAPACITY constant

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @hangar = []
+    @weather = weather
+  end
+
+  def land(plane)
+    weather = @weather.weather_condition
+    if weather == 'stormy'
+      @hangar
```
CazaBelle: If the weather is stormy why would it return the hanger?

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
```
scass91: Consider removal of magic number for capacity, change this to a constant DEFAULT_CAPACITY

README2.md
```diff
@@ -0,0 +1,45 @@
...
+
```
subomionanuga: This README is magnificent 👏

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @capacity = capacity
+    @hangar = []
+    @weather = weather
+  end
+
+  def land(plane)
+    weather = @weather.weather_condition
+    if weather == 'stormy'
+      @hangar
+    elsif full?
```
CazaBelle: How could you reduce the number of things that the method land is doing?

lib/airport.rb
tomlovesgithub: Will i break your tests if i suddenly make this capacity 25? 
perhaps this 1 counts as a "magic number"?

lib/airport.rb
Sindex42: Should this be another guard clause?

spec/weather_spec.rb
```diff
@@ -0,0 +1,8 @@
+require './lib/weather.rb'
```
CazaBelle: Good use of separate spec files

lib/plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
```
tomlovesgithub: is there any methods that your plane could contain?

lib/airport.rb
michaelnguyen974: If possible it would be good to lower the number of conditionals as this may introduce bugs later on.

lib/airport.rb
beacourage: I liked this!

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  let(:plane) { Plane.new }
+
+  it { is_expected.to respond_to(:land).with(1).arguments }
+
+  it 'Lands and keeps a plane' do
+    hangar = airport.land(plane)
+    expect(hangar.include?(plane)).to eq true
+  end
+
+  it 'Lets a plane take off' do
```
CazaBelle: Is there a why to make further tests to accommodate more then one plane?

PR #1243, opened 2018-12-02
Airport challenge so far

beacourage:


lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'plane'
+
+class Weather
+
+  def stormy?
+    weather = Weather.new
+    if weather == :stormy
```
rednblack99: A ternary operator could make this if/else statement (if weather) super concise. Have a little google if you get a chance :)

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+      true
+    else
+      false
+    end
+  end
+
+end
+
+class Airport
```
joaoag: Maybe separate the Weather and Airport classes into dif docs? Good having them as different classes though

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+end
+
+  def initialize (weather: Weather.new)
+     @planes = []
+     @weather = weather
+     @capacity = DEFAULT_CAPACITY
+  end
+
+  def land(plane)
+  fail 'Bad weather' if weather.stormy?
```
rednblack99: I love the way you made a .stormy? method. It makes your code extremely readable!

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def stormy?
+    weather = Weather.new
+    if weather == :stormy
+      true
+    else
+      false
+    end
+  end
+
```
Sindex42: Do true and false need to be explicitly returned? weather == :stormy already evaluates to true of false

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+end
+
+class Airport
+attr_reader :plane, :capacity, :weather
+DEFAULT_CAPACITY = 10
+
+
+def full?
```
stonefarmer9: Could full be private? As we only want the air traffic controller to be able to report that its hangar is full?

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def land(plane)
+  fail 'Bad weather' if weather.stormy?
+  fail 'Airport is full' if full?
+  @planes << plane
+  end
+
+
+  def takeoff(plane)
+  fail 'Bad weather' if weather.stormy?
```
rednblack99: "Bad weather" isn't a super descriptive fail message. It's good for this use but consider mentioning stormy specifically so the user knows a storm is at fault rather than snow or volcano if we added them as weathers

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    expect{airport.land}.to raise_error 'Cannot land due to bad weather'
+  end
+
+  it 'cannot land when airport is full' do
+  airport.land(plane)
+  error = 'Cannot land since airport is full'
+  expect { airport.land(double :plane) }.to raise_error error
+  end
+
+end
```
stonefarmer9: Good test coverage,could the indentation be standardised ?

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  @planes.pop
+  end
+
+  # private
```
rednblack99: Why did you comment this out? Not sure if it's official practice but maybe add a little note explaining why you used a hash rather than just deleting

CazaBelle: It is best practice to remove all commented out code before submitting :)

lib/airport.rb
jlaffbabs: Methods with question marks return booleans so you could just put weather == :stormy and it would return true or false without needing the if/else

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'plane'
+
```
scass91: Consider adding a weather.rb file

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+class Airport
+attr_reader :plane, :capacity, :weather
+DEFAULT_CAPACITY = 10
+
+
+def full?
+  planes.count >= @capacity
+end
+
```
Sindex42: Does 'full?' need to be a public method?

lib/plane.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
+  attr_accessor :landed
+
+  def land
```
rednblack99: Really interesting that you used an attr_accessor and defined the land method in your Plane class. In many ways it makes more sense since it is the plane that lands but does make things a little more complicated!

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    expect{airport.takeoff}.to raise_error 'Cannot take off due to bad weather'
+  end
+
+  it 'prevents landing when weather is stormy' do
+    weather = Weather.new
+    allow(weather).to receive(:stormy?).and_return(true)
+    airport = Airport.new
+    expect{airport.land}.to raise_error 'Cannot land due to bad weather'
+  end
+
```
CazaBelle: Should there be a test to check that the user can change capacity?

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'airport'
+
+describe Airport do
+
+  it 'will instruct a plane to land' do
+    airport = Airport.new
```
rednblack99: There are 'let' matchers in RSpec so you can just create airport and plane at the very beginning of your tests and they're exist throughout. Really helpful!

lib/plane.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
+  attr_accessor :landed
```
Sindex42: Does :landed need to be an attr_accessor if you have a method (land) that changes it?

lib/airport.rb
subomionanuga: I think this file will be a little easier to read if this method is defined a little lower in the file i.e after the initialize method

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    expect(airport).to respond_to(:land).with(1).argument
+  end
+
+it { is_expected.to respond_to(:takeoff).with(1).argument }
```
rednblack99: Good use of the one-line syntax. Just remember it still needs 2 space indent as it's within 'Describe Airport do'

lib/airport.rb
scass91: Consider initialising with a default capacity (capacity = DEFAULT_CAPACITY)

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    weather = Weather.new
+    allow(weather).to receive(:stormy?).and_return(true)
+    airport = Airport.new
+    expect{airport.land}.to raise_error 'Cannot land due to bad weather'
+  end
+
+  it 'cannot land when airport is full' do
+  airport.land(plane)
+  error = 'Cannot land since airport is full'
+  expect { airport.land(double :plane) }.to raise_error error
```
rednblack99: Could you a specific error message here rather than just 'error'

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    expect(airport).to respond_to(:land).with(1).argument
+  end
+
+it { is_expected.to respond_to(:takeoff).with(1).argument }
+
+  it 'prevents takeoff when weather is stormy' do
+    weather = Weather.new
+    allow(weather).to receive(:stormy?).and_return(true)
+    airport = Airport.new
+    expect{airport.takeoff}.to raise_error 'Cannot take off due to bad weather'
```
Sindex42: Does takeoff need an argument in order to work?

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  it 'prevents landing when weather is stormy' do
+    weather = Weather.new
+    allow(weather).to receive(:stormy?).and_return(true)
```
rednblack99: Love the stub matcher!

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'airport'
```
CazaBelle: I think a few extra test here would be useful here

PR #1244, opened 2018-12-02
Functioning Airport

jlaffbabs:


lib/weather.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def stormy?
+    type_of_weather == :stormy
+  end
+
+  private
+  POTENTIAL_WEATHERS = [:stormy, :cloudy, :rainy, :sunny, :dry]
+
+  def type_of_weather
+    POTENTIAL_WEATHERS[Random.rand(5)]
```
scass91: Why is a random number being randomised?

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  def take_off
+    fail "It is too stormy for take off" if stormy?
+
+    fail "There are no planes" if empty?
+
+    @planes.pop
+  end
+
+  private
+  def stormy?
```
joaoag: Nice one creating private methods for each of the weather conditions. Makes it easier to test / debug

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    weather == true
+  end
+
+  def full?
+    @planes.length >= @capacity
```
joaoag: Again good having a separate method for whether the plane capacity is full.

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  let(:airport) { Airport.new }
+  let(:plane) { Plane.new }
```
rednblack99: Love this! Really helpful way to avoid defining airport and plane in each test

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    it 'responds to take off' do
+      expect(airport).to respond_to(:take_off)
+    end
+
+  end
+
+  context '#non-stormy weather' do
+    before { allow(airport).to receive(:weather) { false } }
+
+    it 'take off removes the plane from the airport' do
```
scass91: Does this work if there is more than 1 plane already at the airport?

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def take_off
+    fail "It is too stormy for take off" if stormy?
+
+    fail "There are no planes" if empty?
+
+    @planes.pop
+  end
+
+  private
```
joaoag: Good use of private methods

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  let(:airport) { Airport.new }
+  let(:plane) { Plane.new }
+
+  context '#responds' do
```
Sindex42: Do we still need to test for responds_to if we have more detailed tests for land and take_off

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  def stormy?
+    weather == true
+  end
+
+  def full?
+    @planes.length >= @capacity
+  end
+
+  def empty?
+    @planes.length.zero?
```
rednblack99: Like the use of the zero? method here

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
+class Airport
+  attr_reader :planes, :weather
+  DEFAULT_CAPACITY = 150
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @weather = Weather.new.stormy?
```
stonefarmer9: Does instantiating weather here keep your code rigid against future changes? Rather than instantiate it as a parameter?

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    fail "There are no planes" if empty?
+
+    @planes.pop
+  end
+
+  private
+  def stormy?
+    weather == true
+  end
+
```
CelineKaslin: Do you think that the airport itself is the right place to determinate the weather ?
Maybe an other class will help for it :)

lib/plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
```
subomionanuga: Am I right in assuming this has been defined so that new instances of plane can be created?
The is no test for it though

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
+
+  def stormy?
+    type_of_weather == :stormy
+  end
+
+  private
+  POTENTIAL_WEATHERS = [:stormy, :cloudy, :rainy, :sunny, :dry]
```
stonefarmer9: I like the inclusion of extra weather modes as it seems like a natural evolution of the base program

lib/weather.rb
rednblack99: I don't believe you need to explicitly call on the Random class here. You should be able to simply run [rand(5)]

lib/airport.rb
tomlovesgithub: whats the purpose of making them private?

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      expect(airport.planes).to eql([])
+    end
+
+    it 'landing adds the plane to the airport' do
+      airport.land(plane)
+      expect(airport.planes).to eql([plane])
+    end
+
+    it 'cannot land if the airport is full' do
+      Airport::DEFAULT_CAPACITY.times { airport.land(plane) }
```
stonefarmer9: Does using Default capcity tell the reader how many times the plane will be landed? I.e is it an expressive test?

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
+
+  def stormy?
+    type_of_weather == :stormy
+  end
+
+  private
```
CazaBelle: Nice use of alternate weather conditions

lib/airport.rb
```diff
@@ -31,7 +34,7 @@ def full?
   end
 
   def empty?
-    @planes.length == 0
+    @planes.length.zero?
```
tomlovesgithub: would "@planes.empty?" do the same thing? 
is there a reason you have used ".length.zero?"?

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  let(:airport) { Airport.new }
+  let(:plane) { Plane.new }
+
+  context '#responds' do
+
```
CazaBelle: Too many whitespaces in between tests makes it harder for me to see what test is within what test block

lib/plane.rb
CelineKaslin: How do you know the state of the plane ? If it is flying or landed already ? As I think you have to know the stage because if it is flying it can't take off again and if it is landed already it cannot land again

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    it 'cannot land if the airport is full' do
+      Airport::DEFAULT_CAPACITY.times { airport.land(plane) }
+      expect { airport.land(plane) }.to raise_error 'Airport is full'
+    end
+
+  end
+
+  context '#stormy weather' do
```
CazaBelle: Nice organization with context

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
```
beacourage: I like that you added a new class and folder for weather

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
+class Airport
+  attr_reader :planes, :weather
+  DEFAULT_CAPACITY = 150
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @weather = Weather.new.stormy?
+    @capacity = capacity
+  end
+
```
CelineKaslin: Maybe an airport name will be useful when you initialize, like that you know later on, the origin and destination, for planes to be removed from the airport they are taking_off from and add to the airport they are landing in.

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def stormy?
+    type_of_weather == :stormy
+  end
+
```
CelineKaslin: I do not understand your intention in this line/method. Why do you have 2 stormy? methods (1 in Airport class and 1 in Weather) ?

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  let(:airport) { Airport.new }
+  let(:plane) { Plane.new }
+
```
CelineKaslin: I like the let use to not repeat yourself :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  context '#non-stormy weather' do
+    before { allow(airport).to receive(:weather) { false } }
+
+    it 'take off removes the plane from the airport' do
+      airport.land(plane)
+      airport.take_off
+      expect(airport.planes).to eql([])
+    end
+
```
CelineKaslin: I am confused on line 26, every time that a plane will take off does your airport will be empty ?

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    fail "It is too stormy to land" if stormy?
+
+    fail "Airport is full" if full?
+
+    @planes << plane
+  end
+
+  def take_off
+    fail "It is too stormy for take off" if stormy?
+
```
CelineKaslin: using the fail to throw an error if the airpot is full is well done

lib/airport.rb
jlaffbabs: Thanks for the feedback 🥇

spec/airport_spec.rb
jlaffbabs: I coded it in a way that the particular plane doesn't matter (my take off doesnt take an argument) - so this checks that a plane is leaving but not a specific plane - guess it could use some tampering to make it check it's the particular plane

lib/airport.rb
jlaffbabs: Means they can't be called outside the methods within the class

lib/airport.rb
jlaffbabs: I have another class this is just to abide by the SRP and keep my methods in airport simple

PR #1251, opened 2018-12-02
Weekend work week 1

tomlovesgithub:


tomlovesgithub:
great code

lib/Airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  DEFAULT_CAPACITY = 25
+
+  def initialize(weather = rand(100), airport_capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @weather = (weather > 75 ? "stormy" : "clear")
+    @capacity = airport_capacity
+  end
+
+  def takeoff(plane)
+    fail "#{plane} Cannot take off" if weather? || full?
```
subomionanuga: Ahhh I like how you've merged the bad weather and airport full bit of code here.
Not sure if it might be a low coupling issue but I think it is clever

PR #1253, opened 2018-12-03
Airport challenge 2 - Emanuele Gorga

emanuelegorga:
With a little better understanding I managed to review all my code and redo the challenge.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -10,17 +10,22 @@ class Airport
   def initialize(weather = Weather.new, capacity = DEFAULT_CAPACITY)
     @weather = weather
     @airport = []
-    @capactiy = capacity
+    @capacity = capacity
   end
 
   def land(plane)
     raise "No planes can land, stormy weather." if @weather.stormy?
+    raise "No more planes can land, airport full." if full?
     @airport << plane
   end
 
   def takeoff
     raise "No planes can take off, stormy weather." if @weather.stormy?
     @airport.pop
+  end
+
+  def full?
+    @capacity == @airport.count
   end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -14,12 +14,12 @@ class Airport
   end
 
   def land(plane)
-    raise "no planes can land" unless @weather.weather_condition == false
+    raise "No planes can land, stormy weather." if @weather.stormy?
     @airport << plane
   end
 
   def takeoff
-    raise "No planes can take off, stormy weather." unless @weather.weather_condition == false
+    raise "No planes can take off, stormy weather." if @weather.stormy?
     @airport.pop
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,11 @@
+require_relative 'airplane'
+
+class Airport
+
+  def land
+  end
+
+  def takeoff
+  end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,18 +1,25 @@
 require_relative 'airplane'
+require_relative 'weather'
 
 class Airport
 
   attr_reader :planes
 
-  def initialize
+  DEFAULT_CAPACITY = 20
+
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    @airport = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    raise "No planes can land, stormy weather." if @weather.stormy?
+    raise "No more planes can land, airport full." if full?
+    @airport << plane
+  end
+
```
CelineKaslin: great use of the raise 👍

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  it { is_expected.to respond_to(:land).with(1).argument }
+
+  it { is_expected.to respond_to(:takeoff) }
+
+  it { is_expected.to respond_to(:full?) }
+
+  describe '#take off' do
+    it "prevents to take off in case of stormy" do
+      forecast = create_sunny_weather
```
CelineKaslin: I do not understand your forecast variable. What are you trying to do with it

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  it "has capacity of 20" do
+    expect(Airport::DEFAULT_CAPACITY).to eq 20
+  end
+
+  def create_weather(stormy)
+    weather = double("Weather")
+    allow(weather).to receive(:stormy?).and_return(stormy)
+    weather
+  end
```
CelineKaslin: I am not sure to understand why you are passing stormy as an argument of `and_return` and then weather. Do you want to return "stormy" as a string ? 
i might just do not know a syntax and just not understand it

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def create_stormy_weather
+    create_weather(true)
+  end
+
+  def create_sunny_weather
+    create_weather(false)
+  end
+
```
CelineKaslin: for the 2 previous test, do you try to test or to create your method with a def ? 
Not sure that you are suppose to create a method in a spec.

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  it "has capacity of 20" do
+    expect(Airport::DEFAULT_CAPACITY).to eq 20
+  end
+
+  def create_weather(stormy)
+    weather = double("Weather")
+    allow(weather).to receive(:stormy?).and_return(stormy)
+    weather
+  end
+
```
scass91: I might be misreading this - is the spec file here creating 2 weather methods for stormy & sunny weather? Could these be stored within weather.rb?

spec/weather_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'weather'
+
+RSpec.describe Weather do
+
+  it { is_expected.to respond_to(:stormy?) }
+
+  it 'returns 1 or 2' do
+    expect(subject.stormy?).to be(1).or be(2)
+  end
```
CelineKaslin: great to test that it return what you want :) 
Maybe consider to stub to test if it return whatever you assign to one and whatever you assign to 2

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      allow(forecast).to receive(:stormy?).and_return(true)
+      expect { subject.land(airplane) }.to raise_error "No planes can land, stormy weather."
+    end
+  end
+
+  describe '#full?' do
+    it "prevents to land planes in case of full airport" do
+      capacity = 2
+      gatwick = Airport.new(create_sunny_weather, capacity)
+      gatwick.land(airplane)
```
scass91: Repeating here - consider changing this to 2 times.do - DRY

scass91: As in 2.times do Gatwick.land(airplane)

PR #1254, opened 2018-12-03
What I've done so far pull request

michaelnguyen974:


lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+require_relative 'plane'
+
+class Airport
+  attr_reader :hangar
+  def initialize
+    @hangar = []
+  end 
+
+	def land_plane(plane)
```
subomionanuga: Might be being too picky but the indentation here looks off

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "airport"
+  describe Airport do 
+    it { is_expected.to respond_to :land_plane }
+
```
subomionanuga: Consider using let at the top of your code to replace all instances of Plane.new

PR #1255, opened 2018-12-03
Pull request

joaoag:


lib/README.md
```diff
@@ -0,0 +1,147 @@
+=================
+Airport Challenge
+=================
+
+Summary
+
+After using Rspec for the first time this week, Makers set us a weekend challenge to get to grips with the program, the process of TDD as well as concepts such as high class cohesion and low coupling. We were asked to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we were given by Makers:
```
rednblack99: Great use of a README file. Very clear what you were working on. This is INCREDIBLY comprehensive.

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
+require_relative 'plane'
```
rednblack99: Have you tried this with just require as well? They're in the same directory so you may not need to use require_relative

lib/README.md
```diff
@@ -0,0 +1,147 @@
...
+* I started by breaking down the user stories into diagrams of objects and messages
+* Then I created the Airport and Plane classes
+* The methods were predictably #land and #take_off. I put both in the Airport class.
+* I also created #weather_reporter which generated random weather using #rand from inside the Airport class.
+* (Doing it again I would have a Weather class, with #weather_reporter inside, for higher coherance)
+* Following the models I created from the user stories, I used Rspec to test for expected object states and messages using doubles and stubs.
+* These tests drove the design of my program, as it emerged iteratively, test-by-test
+* I satisfied each of the user stories during feature tests, but have not been able to replicate all the outcomes in Rspec
+* I also did not have time to account for the edge cases
+
```
CelineKaslin: Great descriptive of the approach. Good break down of the problem

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+class Airport
+
+  attr_reader :plane, :capacity
+
+  def initialize
+    @plane = []
+    @capacity = 5
+  end
+
+  def weather_reporter(rand)
```
rednblack99: I could be horribly wrong here but I think you could just pass your rand(10) here in weather_reporter and make the weather variable in here to cut down your code :)

lib/README.md
```diff
@@ -0,0 +1,147 @@
...
+As an air traffic controller 
+To ensure safety 
+I want to prevent landing when the airport is full 
+
+As the system designer
+So that the software can be used for many different airports
+I would like a default airport capacity that can be overridden as appropriate
+
+```
```
stonefarmer9: Good inclusion of user stories

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+    if rand >= 3
+      p "Sunny"
+    else
+      p "Stormy"
+    end
+  end
+
+  def land(plane)
+    weather = weather_reporter(rand(10))
+    if weather == "Sunny" && @plane.count >= @capacity
```
rednblack99: There's a little bit of repetition here. I think you could just raise your weather error first and then check for capacity to meet DRY.

lib/README.md
```diff
@@ -0,0 +1,147 @@
...
+2.5.0 :016 > airport.change_capacity
+"What number would you like to set this airport's capacity to?"
+"Press enter to skip and use default capacity of 5"
+10
+10
+ => 10 
+2.5.0 :017 > airport.capacity
+ => 10 
+```
+
```
stonefarmer9: Nice readme, are there too many IRB examples? Do they make it easier or harder to read ?

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+    if weather == "Sunny"
+      p "The weather is sunny, safe to take off. The plane has successfully left the airport"
+      @plane.pop
+    else
+      raise "The weather is too stormy, you cannot take off"
+    end
+  end
+
+  def change_capacity
+    p "What number would you like to set this airport's capacity to?"
```
rednblack99: Love that you requested user input here. I haven't seen anyone else do that but if the engineer actually wanted to change capacity they would need an interface!

lib/README.md
subomionanuga: This README is magnificent 👏

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
+require_relative 'plane'
+class Airport
+
+  attr_reader :plane, :capacity
+
+  def initialize
+    @plane = []
+    @capacity = 5
+  end
+
```
CelineKaslin: Do you think that a name for airport might be useful, considering that the airport can take-off only from the airport from the airport they are in ?

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
+require_relative 'plane'
+class Airport
+
+  attr_reader :plane, :capacity
+
+  def initialize
+    @plane = []
```
Sindex42: Should @plane be more descriptive to avoid confusion?

lib/README.md
```diff
@@ -0,0 +1,147 @@
+=================
+Airport Challenge
```
AlinaGoaga: Really liked how descriptive the README is and that you've included examples on how to run this in irb

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def weather_reporter(rand)
+    if rand >= 3
+      p "Sunny"
+    else
+      p "Stormy"
+    end
+  end
+
```
Sindex42: Should the Airport class be responsible for determining weather?

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
+require_relative 'plane'
+class Airport
+
+  attr_reader :plane, :capacity
+
```
subomionanuga: Consider initialising a constant DEFAULT_CAPACITY so that you can change this if necessary without changing too much of your code. You should also be able to use it in your spec file

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'airport'
+require 'plane'
+
+describe Airport do 
+  let(:airport) { described_class.new }
```
AlinaGoaga: Well done on using let to prevent repetition!

lib/airport.rb
CelineKaslin: Is the airport the appropriate place to determinate the weather ?
Maybe a Weather separated class would have been good to create :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    expect(airport.land(plane)).to raise_error("This airport is full, you cannot land here")
+  end
+
+  it 'responds to #change_capacity' do
+    expect(airport).to respond_to(:change_capacity)
+  end
+
+  it '#change_capacity changes the airport capacity' do
+    
+    airport.land(plane)
```
AlinaGoaga: Have you thought about maybe using a loop (x.time do ... ) ? Might make it easier to generate the planes.

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+describe Airport do 
+  let(:airport) { described_class.new }    
+  let(:plane) { Plane.new }
+
+  it 'responds to #land' do
+    expect(airport).to respond_to(:land).with(1).argument
+  end
+
+  it 'receives a plane with #land' do
+    allow(airport.land(plane)).to receive(airport.weather_reporter(3)).and_return("Sunny")
```
rednblack99: Good use of the stub matchers!

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+  def land(plane)
+    weather = weather_reporter(rand(10))
+    if weather == "Sunny" && @plane.count >= @capacity
+      raise "This airport is full, you cannot land here"
+    elsif weather == "Sunny" && @plane.count < @capacity
+      @plane << plane
+    else raise "The weather is too stormy, you cannot land"
+    end
+  end
+
```
Sindex42: Does this need to be an else/if statement? 'raise :stormy' could be a guard clause at the beginning and simplify the method

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+  def take_off(plane)
+    weather = weather_reporter(rand(10))
+    if weather == "Sunny"
+      p "The weather is sunny, safe to take off. The plane has successfully left the airport"
+      @plane.pop
+    else
+      raise "The weather is too stormy, you cannot take off"
+    end
+  end
```
stonefarmer9: In both take_off and land could the weather determination be made in another method and that then feeds into an if/else statement? Basically do these methods have one responsibility.

lib/airport.rb
AlinaGoaga: Would recommend using a different class for this as the airport wouldn't really influence this bit

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    allow(airport.take_off(plane)).to receive(airport.weather_reporter(1)).and_return("Stormy")
+    expect { airport.take_off(plane) }.to raise_error("The weather is too stormy, you cannot take off")
+  end
+
+  it 'only allows #land in sunny weather' do
+    allow(airport.land(plane)).to receive(airport.weather_reporter(1)).and_return("Stormy")
+    expect { airport.land(plane) }.to raise_error("The weather is too stormy, you cannot land")
+  end
+
+  it 'only allows 5 planes to land' do
```
subomionanuga: I think the DEFAULT_CAPACITY constant would have either eliminated the need for this or allowed it to be refactored

lib/README.md
jlaffbabs: It's beautiful - love the feature tests!

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    airport.land(plane)
+    airport.land(plane)
+    airport.land(plane)
+    airport.land(plane)
+    airport.land(plane)
+    allow(airport).to receive(change_capacity).and_return(10)
+    airport.land(plane)
+    expect(airport.plane).to eql(6)
+  end
+
```
rednblack99: This is the most tests I've seen so far. Good work testing thoroughly.

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+    p "What number would you like to set this airport's capacity to?"
+    p "Press enter to skip and use default capacity of 5"
+    @capacity = gets.chomp.to_i
+    if @capacity.zero?
+      @capacity = 5
+    else
+      p @capacity
+    end
+  end
+
```
CelineKaslin: I would not have think to this method, nice :)

lib/airport.rb
joaoag: It definitely should! Thanks :thumbsup:

lib/airport.rb
joaoag: No! Thanks for the pointer :thumbsup:

lib/plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
+end
```
CelineKaslin: Maybe the plane state might have been added as the state of the plane maybe have  to be known from what  I understand from the challenge instructions

lib/airport.rb
jlaffbabs: Could also make separate methods for some of these guard clauses to clean up the code

spec/airport_spec.rb
joaoag: Thanks Alina! :thumbsup:

spec/airport_spec.rb
joaoag: Was desperate to use a loop, looked around a bit and couldn't see the syntax so prematurely gave up - will definitely do so in the future! Thanks

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    allow(airport.land(plane)).to receive(airport.weather_reporter(1)).and_return("Stormy")
+    expect { airport.land(plane) }.to raise_error("The weather is too stormy, you cannot land")
+  end
+
+  it 'only allows 5 planes to land' do
+    airport.land(plane)
+    airport.land(plane)
+    airport.land(plane)
+    airport.land(plane)
+    airport.land(plane)
```
CelineKaslin: Does it have a way to refactor that ? Maybe by doing a n.times

spec/airport_spec.rb
rednblack99: not seen described_class before. I'll need to look that up!

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  it '#change_capacity changes the airport capacity' do
+    
+    airport.land(plane)
+    airport.land(plane)
+    airport.land(plane)
+    airport.land(plane)
+    airport.land(plane)
+    allow(airport).to receive(change_capacity).and_return(10)
```
CelineKaslin: maybe a way to refactor here as well :)

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+  def weather_reporter(rand)
+    if rand >= 3
+      p "Sunny"
+    else
+      p "Stormy"
+    end
+  end
+
+  def land(plane)
```
AlinaGoaga: Good job on covering different scenarios here! Have you considered maybe initializing the weather when you create the airport and also maybe having a different method to check if the airport is full? Just so that we don't leave that much on  the land method's shoulders :)

lib/airport.rb
joaoag: Yeah I tried last minute to make it into a guard clause as a result of rubocop but it broke everything in a serious way. Will definitely look into how to use this in the future because it is messy as it stands :thumbsup:

lib/airport.rb
joaoag: Yes definitely, thanks Alina

lib/airport.rb
joaoag: Great shout Subomi, thank you

PR #1256, opened 2018-12-03
CKKH Nov Cohort Airport Challenge Pull Request

CKKH:


lib/airport.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'weather'
+require_relative 'plane'
+
+class Airport
+
+  attr_reader :plane, :weather
+
+  def initialize(plane = Plane.new, weather = Weather.new)
```
sjmog: Nice dependency injection!

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+    @plane = plane
+    @weather = weather
+  end
+
+  def land(plane)
+    @plane = plane
```
sjmog: This will only work with one plane. Is there a data structure we could use to work with more than one plane?

lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
+  def exists?
```
sjmog: I'm not sure what this method is meant to do.

lib/weather.rb
```diff
@@ -0,0 +1,8 @@
+class Weather
+
+  def weather_forecast
+    num = rand(1..5)
+    num == (1..4) ? true : false
+  end
```
sjmog: Nice! There is a shorter way: try just `rand(1..5) == (1..4)`

CKKH: Neat! Thanks Sam.

PR #1258, opened 2018-12-04
Airport Challenge (update/refactor) - Jethro Whaley

jaywayawyaj:
My (updated) attempt at the airport challenge. Code functions in IRB and passing all rspec tests. 100% coverage and 0 Rubocop offenses. Bonus (feature spec) begun but it's late and I want to be as fresh as possible for week 2!

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,7 @@
+class Airport
+
+  def land(plane)
+
+  end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -10,8 +10,9 @@ class Airport
     @hangar << plane
   end
 
-  def take_off(plane)
-    @plane = plane
+  def take_off
+    departed_plane = @hangar.pop
+    puts "#{departed_plane} has left the airport"
   end
 
 end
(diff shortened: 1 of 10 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,34 @@
+require './lib/weather'
+
+class Airport < Weather
```
tomlovesgithub: Clever use of inheritance, i attempted this myself, but was apprehensive and felt there may have been another way to do it, how else could we access Weather methods from inside Airport?

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_reader :plane, :hangar, :capacity
+  DEFAULT_CAPACITY = 50
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    raise "The plane cannot land" if stormy? || full?
```
tomlovesgithub: well written guard clause here, did you consider using the ternary operator? Is it possible to use it and achieve the same result?

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @hangar << plane
+  end
+
+  def take_off
+    raise "It is too stormy to fly" if stormy?
+    departed_plane = @hangar.pop
+    puts "#{departed_plane} has left the airport"
```
tomlovesgithub: nice confirmation message! Consider doing other messages using given args throughout!

PR #1261, opened 2019-01-11
First PR - checking for feedback

isabelcooper:


lib/airport.rb
```diff
@@ -0,0 +1,49 @@
+require_relative 'weather'
+require_relative 'plane'
+
+class Airport
```
samjones1001: This looks like a good class overall.  The methods are all reasonably concise and they're well named so it's clear to me exactly what each part of the code is intended to do.

It's also good that you've managed to keep the number of publicly facing methods to a minimum.  As a general rule, the smaller the public interface of an object the better.

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def initialize(capacity = CAPACITY)
+    @planes_in_airport = []
+    @capacity = capacity
+  end
+
+  def confirm_take_off(plane)
+    fail "This plane is not in the airport" if plane.current_airport != self
+
+    take_off(plane) if weather_ok?
```
samjones1001: This seems as though in the event that the weather is stormy, the plane will simply fail to take off without any feedback to the user.  From a usability perspective it might be nice to tie this condition to an error message.

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def take_off(plane)
+    @planes_in_airport.select! { |pia| pia == plane }
+    plane.park_in_airport(nil)
+    "#{plane} has now left airport"
+  end
+
+  def weather_ok?
+    Weather.new.random_weather != "stormy"
```
samjones1001: In this particular example the impact is minimal, but generally speaking I would avoid creating a new instance of an object every time you call a method like this unless you have an explicit reason to do so.  You can avoid this sort of multiple object creation by setting up your objects as part of your initialize method:

```ruby
class Airport
  attr_reader :capacity, :planes, :weather
  DEFAULT_CAPACITY = 10 

  def initialize(capacity = DEFAULT_CAPACITY)
    @capacity = capacity
    @planes = []
    @weather = Weather.new
  end
end
```

You could then alter this `weather_ok?` method to be:

```ruby
def weather_ok?
  weather.random_weather != "stormy"
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,94 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+  before(:each) do
+    Airport.send(:public, *Airport.private_instance_methods)
+    @plane = Plane.new
```
samjones1001: This works absolutely fine, but a more conventional way of setting up variables for use in your RSpec tests would be to use a `let` block.

PR #1263, opened 2019-01-11
Completed airport-challenge for review

alexh148:


lib/airport.rb
```diff
@@ -0,0 +1,18 @@
+class Airport
+  attr_reader :planes, :weather, :capacity
+  DEFAULT_CAPACITY = 5
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+    @weather = ['Cloudy', 'Sunny', 'Stormy', 'Rainy', 'Snowy'].sample
```
samjones1001: Based on what we've talked about this week perhaps you can now see that ideally this would be extracted out into a separate Weather class.  As it stands, Airport is probably responsible for both maintaining a record of parked planes AND determining the weather.

lib/airport.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def safe?
```
samjones1001: This and the one below feel to me like they should be private methods.

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'airport'
+
+class Plane
+  attr_reader :state
+
+  def initialize
+    @state = 'Newly constructed'
+  end
+
+  def land(airport)
```
samjones1001: There's a generally agreed upon rule in the Ruby community that if a method has more than 5 lines then it's too long which makes both this and your take_off methods borderline. Is there anything you could feasibly extract?

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'airport'
+
+describe Airport do
```
samjones1001: This looks like a good set of tests 🙂

spec/plane_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'plane'
+
+describe Plane do
+
+  before(:each) do
```
samjones1001: Again, a very strong set of tests.  A small point is that, conventionally, it's more common to use a let block in order to set up variables for use in your tests.  Before each blocks tend to be used for performing repeated set up actions required for multiple tests.

That said, there is no functional difference between the implementations (I think 🙂 )

alexh148: Totally agree. I didn't fully understand let blocks when we did this challenge, so I chose to use before blocks since I knew what they did. However, now I know let blocks better, I would most certainly refactor with them. Will add to my future improvements section of the readme 🙂

lib/plane.rb
alexh148: I only realised that today when it was mentioned in the oystercard challenge 😅 
I'm thinking I'll extract the raise errors into .can_land? and .can_take_off? private methods. 

Do you think that's the best way, or is there another, more efficient way of refactoring? I'm not sure how else I can shorten the methods, since all the lines are very specific to landing and taking off.

lib/airport.rb
alexh148: I did originally try to have these as private methods, but the Plane class needs to be able to check them. Since the planes array is already public, I might move the .full? method to the Plane class and make it private. Once I make the weather its own class, perhaps I can do something similar with .safe?.

What do you think? Would your approach have been similar? Interested to know, as I'd like to get into best practices as early as possible 😅

lib/airport.rb
alexh148: Already got this in my future improvements section of the readme! 👍

PR #1264, opened 2019-01-11
First pull request

ktcornish:


lib/airport.rb
```diff
@@ -0,0 +1,48 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
```
samjones1001: this is a well put together class.  Methods and variables are all clearly named, so its very clear what each part of the code is intended to do.

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+  attr_reader :planes, :capacity
+
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land_plane(plane, weather = Weather.new)
```
samjones1001: There's a generally agreed upon rule in the Ruby community that if a method has more than 5 lines then it's too long which makes both this and your `take_off_plane` methods borderline.  Is there anything you could feasibly extract?

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+    grounded?(plane)
+    airport_full?
+    check_weather(weather)
+    @planes << plane
+    plane.is_grounded
+    self
```
samjones1001: I'm not clear of the need for returning self at the end here

samjones1001: My guess is that you've done it for the purpose of testing?  If so, I'd avoid making changes to the way your code works solely to get tests passing more easily.

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def take_off_plane(plane, weather = Weather.new)
+    in_airport?(plane)
+    check_weather(weather)
+    @planes.delete(plane)
+    plane.is_flying
+    self
+  end
+
+  def grounded?(plane)
```
samjones1001: This feels to me like this and all of the following methods should be made private.

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+class Weather
```
samjones1001: Great that you have a separate class for weather - this feels like a good extraction

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'airport'
+
+describe Airport do
+
+  let(:flying_plane) { double(:state => "flying", :is_grounded => "grounded") }
+  let(:grounded_plane) { double(:state => "grounded", :is_flying => "flying") }
+  let(:test_plane) { double(:state => nil, :is_flying => nil, :is_grounded => nil) }
+  let(:stormy_weather) { double(:condition => "stormy") }
+  let(:clear_weather) { double(:condition => "clear") }
```
samjones1001: Nice use of doubles to isolate your tests.  Looking over them quickly, it looks as though at least one scenario is left untested - an airplane being instructed to take off from an airport where it is not located.  There may be more 🙂

samjones1001: Also, I like that you're organising your tests using nested describe blocks.  You might also want to take a look at context blocks to organise them further.

PR #1270, opened 2019-01-13
airport challenge

SzilviH:


lib/airport.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'plane'
+
+class Airport
```
samjones1001: Overall the code in this class looks good to me. The methods are all well named and small enough to have a single responsibility.  It's really clear to me what each part of the code is intended to do.

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_reader :plane, :planes, :capacity
+  DEFAULT_CAPACITY = 50
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def full?
```
samjones1001: This feels to me like it should be a private method.

The way to approach deciding whether something should be public or private is to think about whether you would ever expect a user to interact with the method directly.  If not - as I suspect is the case here - always default to private.  You should be aiming to keep the public facing interface of your objects as small as possible.

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+    planes.count >= @capacity
+  end
+
+  def stormy?
+    rand(1..5) == (1..5)
```
samjones1001: This line looks to me as though it would always return false?

If you did that intentionally because you were struggling with testing randomness then don't worry, otherwise have a think about exactly what this line is doing, and how you might be able to change it in order to get it working as intended.

Also, after the session this morning maybe you can see that having this method in your Airport class means that it violates the single responsibility principle? Right now it is responsible for maintaining a list of landed planes *AND* determining the weather.

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def land(plane)
+    fail 'Airport is full' if full?
+    fail 'Storm prevented landing' if stormy?
+
+    @planes << plane
```
samjones1001: As we talked about this morning, the fact that you've got an attr_reader for planes means that you can drop the `@` to avoid accessing the instance variable directly.

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+describe Airport do
+  it { is_expected.to respond_to(:land).with(1).argument }
+  it { is_expected.to respond_to(:plane) }
+  it { is_expected.to respond_to(:take_off).with(1).argument }
+
+  it 'has a default capacity' do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+
+  describe '#land' do
```
samjones1001: It's great that you're using nested describe blocks to keep your tests well organised.

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def take_off(plane)
+    fail 'Storm prevented take off' if stormy?
+
+    @plane = plane
```
samjones1001: I'm not sure it's necessary to assign the plane to an instance variable here - that sort of implies that you want to associate it with the instance of the airport object and use it for something later and I'm guessing you don't?

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'airport'
+
+describe Airport do
+  it { is_expected.to respond_to(:land).with(1).argument }
```
samjones1001: These sort of `respond_to` tests can probably be removed once you've created more rigorous tests for the same methods.

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
```
samjones1001: Generally speaking if you find yourself with an object that doesn't have both state (instance variables) and behaviour (methods) you should question whether its a valid object.  Right now, your plane class has neither 🙂 

I expect that if you'd got to the extension exercise of preventing planes from taking off if they're already in the air and so on your plane would have been a bit more fleshed out, but for now it's not very valid!

PR #1281, opened 2019-01-13
Week 1 airport_challenge by Jess NS Leung

jesslns:


jesslns:
need more edge cases.

PR #1287, opened 2019-01-13
Partial solution

RubyKobayashi:


RubyKobayashi:
My goals for the Airport Challenge was to further progress on using TDD as a process:

I can TDD anything:

airrport_spec.rb for evidence lines ...

PR #1290, opened 2019-01-13
pull request

david-mears:
"I can refactor anything" - simplying bad_weather? method in penultimate commit.
"I can TDD anything" - testing edge-cases on lines 55/59 in spec/features/airport_features_spec.rb
"I can program fluently" - test isolation with stubs and doubles
"I can write code that is easy to change" - capacity is not hard-coded but has a default value

### DIFF
--- a/spec/features/airport_features_spec.rb
+++ b/spec/features/airport_features_spec.rb
@@ -15,20 +15,14 @@ describe Airport do
 
   it "can be given a different initial capacity" do
     # expect(Airport.new(15).capacity).to eq 15
-    different_capacity = 10 + Airport::DEFAULT_CAPACITY
-    gatwick = Airport.new(different_capacity)
+    gatwick = Airport.new(0)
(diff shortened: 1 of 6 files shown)


spec/features/airport_features_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+  describe '#capacity' do
+    it "can change capacity after initialization" do
+      heathrow.capacity = 45
+      expect(heathrow.capacity).to eq 45
+    end
+  end
+
+  describe '#land' do
+    before do
+      allow(heathrow).to receive(:weather).and_return(false)
```
david-mears: stub the method, not the class under test!
(don't quite know how - see model code by dearshrewdwit)

david-mears: Another solution: (Kernel is the object on which .rand lives)
allow(Kernel).to receive(:rand).and_return 1

PR #1292, opened 2019-01-14
Partial Solution 

vaithschmitz:
Missing test for weather randomness and unsure whether using .self for method calls is best practice.


vaithschmitz:
This weekend challenge helped with being able to model a concept first. 
While it might not have been TDD throughout, I made sure to have a solid understanding of how my product should work. In doing this, it became much clearer how different functions and instances interact.

PR #1293, opened 2019-01-14
Airport '19

dippedincode:


dippedincode:
### I can TDD anything
I used the TDD Red-Green-Refactor process to perform single unit tests. I committed to Git after each unit test.
### I can refactor anything
I refactored one of the spec files with doubles and using the keyword subject in my tests.
### I can program fluently

PR #1299, opened 2019-01-14
Pull request for review

JamesRudrum:
Pull request for review

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,19 @@
+require_relative './plane'
+
+class Airport
+
+  attr_accessor :plane
+
+  def initialize
+    @plane = Plane.new
+  end
+
+  def land
+    @plane.land
+  end
+
+  def take_off
+    @plane.take_off
+  end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -3,17 +3,20 @@ require_relative './weather'
 
 class Airport
 
+  DEFAULT_CAPACITY = 1
+
   attr_accessor :plane
   attr_reader :num_of_planes
 
-  def initialize
+  def initialize(capacity = DEFAULT_CAPACITY)
     @plane = Plane.new
     @weather = Weather.new
     @num_of_planes = []
+    @capacity = capacity
   end
 
   def land(plane)
-    raise Exception.new("Too stormy to land") if stormy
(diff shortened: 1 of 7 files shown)


JamesRudrum:
------------------------------------------------------------------------------------------------------

This weekend challenge helped me with being able to TDD anything as I used the given process throughout. However as there are so few commits (one commit for failing test, next commit for passing test) I am unable to evidence this and will be working on this. 

------------------------------------------------------------------------------------------------------

This weekend challenge helped me with being able to debug as from the tests that I wrote I was able to deal with syntax or other errors that were given from spec or other files. This should also be evidenced through commits. This leads on to being able to refactor code.

------------------------------------------------------------------------------------------------------

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  DEFAULT_CAPACITY = 1
+
+  attr_accessor :plane
+  attr_reader :num_of_planes, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @plane = Plane.new
+    @weather = Weather.new
+    @num_of_planes = []
```
JamesRudrum: this name for array makes no sense in later user stories

spec/airport_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+    expect(airport2.capacity).to eq 50
+  end
+
+  describe '#land' do
+    it { is_expected.to respond_to(:land) }
+    it { is_expected.to respond_to(:land).with(1).argument }
+
+    it "instructs plane to land at airport" do
+      airport = Airport.new
+      allow(airport).to receive(:stormy).and_return(false)
```
JamesRudrum: Don't stub class that am testing. Therefor for this example would stub the weather class to receive stormy?

spec/airport_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      plane = Plane.new
+      airport.land(plane)
+      expect(airport.num_of_planes.include?(plane)).to eq(true)
+    end
+
+    it "prevents landing when weather is stormy" do
+      airport = Airport.new
+      plane = Plane.new
```
JamesRudrum: Use doubles: Currently this test will break if something wrong with Plane class. However this isn't what is being tested here and further down. Therefore use a stand in (double) to test if the class takes the method actually being tested.

PR #1300, opened 2019-01-14
Challenge sollution

dangroze:


dangroze:
I can TDD anything:
- airport.spec for evidence of doubles and stubbing
- commits  show that the tests are green

PR #1301, opened 2019-01-14
Airport challenge submission

sebastianlavender:


lib/airport.rb
```diff
@@ -0,0 +1,45 @@
+require_relative 'plane'
+
+class Airport
```
samjones1001: Overall this looks like a well put together class.  You methods are well named and reasonably concise and it's very clear to me what each part of the code is intended to do.

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  DEFAULT_CAPACITY = 10
+
+  attr_reader :planes
+  attr_accessor :capacity
```
samjones1001: Is an accessor necessary here?  It implies that you might want to change the capacity of the airport on the fly after creation which I'm not sure is the case.  Certainly you don't seem to be doing so at the moment 🙂

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    @weather = weather
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail("Unable to land as plane as isnt flying") if plane.location != "flying"
+    fail("Unable to land due to stormy weather") if weather
+    fail("Unable to land due airport capacity being full") if capacity_full
+
+    plane.location = self
```
samjones1001: As we discussed earlier, directly accessing an object's instance variables like this is generally best avoided.  A better solution might be something like:

```ruby
class Plane
  attr_accesor :location

  def initialize
     @location = 'flying'
  end

  def set_location(location)
    location= location
  end
end
```

Then, in your airport you could do: 
```ruby 
   plane.set_location(self)
```

samjones1001: In this case the change is pretty minimal and it still works in exactly the same way, but it's preferential not to reach into an object's instance variables from outside so I'd still recommend it.

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    @planes << plane
+  end
+
+  def capacity_full
+    @planes.length == @capacity
```
samjones1001: You could drop the `@` from `planes` here in order to make use of your attr_reader

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+  def land(plane)
+    fail("Unable to land as plane as isnt flying") if plane.location != "flying"
+    fail("Unable to land due to stormy weather") if weather
+    fail("Unable to land due airport capacity being full") if capacity_full
+
+    plane.location = self
+    @planes << plane
+  end
+
+  def capacity_full
```
samjones1001: This feels to me like it should be a private method.

spec/airport_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    allow(subject).to receive(:weather).and_return(false)
+  end
+
+  it { is_expected.to respond_to(:land).with(1).argument }
+  it { is_expected.to respond_to(:take_off).with(1).argument }
+
+  let(:plane) { Plane.new }
+  let(:plane2) { Plane.new }
+
+  describe '#land' do
```
samjones1001: Great that you're using nested describe blocks to keep your tests organised.

PR #1303, opened 2019-01-15
Work in progrees pull request.

jeddhopkins:


spec/weather_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'weather'
+
+describe Weather do
+  describe '#stormy' do
+    it 'can be stormy' do
+      expect(subject.stormy?).to be false
```
samjones1001: This test doesn't seem to be checking what it says it's going to 🙂

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'airport'
+
+describe Airport do
```
samjones1001: The tests in this class all look robust but it looks to me as though there are some tests missing.  For example, what happens when a plane successfully takes off or lands?

PR #1304, opened 2019-02-09
airport and plane

sarah-byrne:


spec/airport_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'airport'
+
+describe Airport do
+  it "can instruct a plane to land in good weather" do
```
steve-f-kahn: Have you considered using a before statement to reduce creating a new airport and weather double in each context block.

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def take_off(airport)
+    raise "already flying" if @flying
+    raise "plane not at airport" if @airport != airport
+
+    @airport = nil
+    @flying = true
+  end
+
+  def flying?
```
steve-f-kahn: Could this be done with just adding @flying to attr_reader instead of a new method?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  DEFAULT_CAPACITY = 100
+
+  def initialize(weather = Weather.new, capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+    @weather = weather
+  end
+
+  def land(plane)
+    raise "airport full" if full?
```
steve-f-kahn: Could these raise and guard statements be moved to a different method to reduce the length of land?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+    raise "bad weather: planes cannot land" if @weather.stormy?
+
+    plane.land(self)
+    @planes.push(plane)
+  end
+
+  def take_off(plane)
+    return "bad weather: plane has not left" if @weather.stormy?
```
steve-f-kahn: Why use return here instead of raise as you did in previous times when dealing with bad weather.

spec/plane_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'plane'
+require 'airport'
+
+describe Plane do
+  it "starts off flying" do
```
steve-f-kahn: Could a before block reduce the amount of repetition here?

spec/weather_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+describe Weather do
+  it "says that it is stormy if the weather matches current definition of bad weather" do
+    weather = Weather.new
+    allow(weather).to receive(:check_weather) { Weather::BAD_WEATHER }
+    expect(weather.stormy?).to eq(true)
+  end
+
+  it "generates random weather" do
+    expect(subject.check_weather).to be_between(0, 99)
+  end
```
steve-f-kahn: This it block only checks one case. As you use rand(100), it could output 100 and as such this test would fail 1 in 101 times even though it is achieving what you want it to do.

lib/airport.rb
sarah-byrne: oops. typo

lib/plane.rb
sarah-byrne: yep, ta

spec/airport_spec.rb
sarah-byrne: didn't know about before, thanks

PR #1305, opened 2019-02-09
Airport Challenge Submission

Alastair-Revell:


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    weather_check
+    @hangar << plane
+  end
+
+  def takeoff
+    weather_check
+    @hangar.pop
+  end
+
+  def weather_check
```
dory1318: Private method for' weather_check' and 'isfull?' methods might be a better idea as it should not be publicly accessible

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    # 0: "stormy", 1: "sunny"
+    @weather = [1, 1, 1, 1, 0].sample
```
dory1318: To follow the "1 class - 1 responsibility" rule, a separate class for 'Weather' should be created

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require_relative "airplane"
+
+class Airport
+
+  attr_reader :hangar
+  attr_writer :size
+
+  def initialize
+    default_size = 5
```
dory1318: Have you thought about using CONSTANT for default_size? Usually, it is better to avoid so called 'magic numbers' as in long codes can cause confusions and problems.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    @hangar = []
+    @size = default_size
+  end
+
+  def land(plane = Airplane.new)
+    isfull?
```
dory1318: I like that the errors are in the 'weather_check' and 'isfull?' methods instead of 'land' and 'takeoff'! It makes the code look more clear and readable

PR #1308, opened 2019-02-10
Airport challenge completed 

steve-f-kahn:
I have completed the airport challenge


lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    fail "This plane is not in the airport" unless here?(plane)
+
+    out_going_plane = @planes.select { |x| x == plane }.pop
+    @planes = @planes.reject { |x| x == plane }
+    out_going_plane
+  end
+
+  def here?(plane)
+    here = false
+    p plane
```
sarah-byrne: Should you delete or comment out p's?

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
+class Airport
+
+  attr_reader :capacity, :planes
+
+  def initialize(capacity = 1)
```
sarah-byrne: a constant could be used for capacity

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    here
+  end
+
+  def weather(conditions)
+    @conditions = conditions
+  end
+
+  private
+  def stormy?
+    fail "It is too stormy to land or takeoff" if @conditions == "stormy"
```
sarah-byrne: Is this method making an assumption about why it is being asked if it is stormy?  What if the airport wanted to know if it was stormy before doing another process rather than take-off/landing? Would the fail be better in the calling method so that the fail message is definitely relevant to the process?

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def full?
+    fail "The airport is full" if @planes.count >= @capacity
```
sarah-byrne: the rubric says to use the attribute reader (i.e. use planes and capacity instead of @planes and @capacity) but I'm not sure I agree - I find the @s clear that you are referring to the instance variable

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    out_going_plane = @planes.select { |x| x == plane }.pop
+    @planes = @planes.reject { |x| x == plane }
+    out_going_plane
+  end
+
+  def here?(plane)
+    here = false
+    p plane
+    @planes.each do |x|
```
sarah-byrne: can use array.include? method to save coding the loop

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    @planes << plane
+  end
+
+  def take_off(plane)
+    stormy?
+    fail "This plane is not in the airport" unless here?(plane)
+
+    out_going_plane = @planes.select { |x| x == plane }.pop
+    @planes = @planes.reject { |x| x == plane }
```
sarah-byrne: is this doing the same as the line above?

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    @planes.each do |x|
+      p x
+      return true if x == plane
+
+    end
+    here
+  end
+
+  def weather(conditions)
+    @conditions = conditions
```
sarah-byrne: use rand() to generate random weather

lib/airport.rb
steve-f-kahn: Yes

lib/airport.rb
steve-f-kahn: Yeah, say no to magic numbers.

PR #1309, opened 2019-02-10
Completion of the Challenge

SimonaGioe:


lib/plane.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_reader :flying
+
+  def initialize
+     @flying = DEFAULT_STATE
+  end
+
+  def land(airport)
+    if landed
+      raise "Plane already landed"
+    end
```
dbugsy: Tiny suggestion to improve readability :)

```suggestion
    raise "Plane already landed" if landed
```

spec/air_traffic_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
+require "air_traffic_controller"
+
+describe AirTrafficController do
+  context "#instruct_landing" do
+    it "instructs landing if not stormy" do
+      plane = Plane.new
+      airport = Airport.new
```
dbugsy: Isolate your subject.

Normally we use a double for every object other than the object under test. This is to ensure that the tests pass because of something in the subject, rather than a side effect from elsewhere.

lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'airport'
+require_relative 'weather'
+
+class AirTrafficController
+
```
dbugsy: A class with no state is a signal that the class has no reason to exist - an object would normally hold state and behaviour. 

In this class, it is a sign that you do not need this class at all. For instance, you could have injected the weather into the Airport class directly.

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+  def accept(plane)
+    plane.land(self)
+    @planes << plane
+  end
+
+  def release(plane)
+    plane.takeoff(self)
+    @planes.delete(plane)
+  end
+
```
dbugsy: everything below this point should be private (encapsulation)

lib/air_traffic_controller.rb
SimonaGioe: Hey Dan i think you're looking at the incompleted challenge. Please check the full one here 
https://github.com/SimonaGioe/airport_challenge/tree/master/lib

SimonaGioe: Hey Dan i think you're looking at the incompleted challenge. Please check the full one here 
https://github.com/SimonaGioe/airport_challenge/tree/master/lib

PR #1310, opened 2019-02-10
airport challenge

limavadakara:


lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def weathergood?
+    goodweather = ["sunny"]
+    goodweather.include?(weather) ? true : false
+  end
+
+  def weather
+    all_weathers = ["sunny", "stormy"]
+    all_weathers[rand(2)]
```
alicelieutier: You can also use the `sample` method to do the same thing.

PR #1312, opened 2019-02-10
James Airport Challenge Pull Request

tso402:
All User Stories Implemented along with Edge Case Protections.

All Tests passing and no Rubocop exceptions.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,6 +1,7 @@
 require_relative 'plane.rb'
 
 class Airport
+  CAPACITY = 1
   attr_reader :planes, :airport_name
   def initialize(airport)
     @planes = []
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    planes_for_takeoff = @planes.select { plane }
+    fail "That plane is not at this airport" if planes_for_takeoff.empty?
+    
+    plane_for_takeoff = planes_for_takeoff.pop
+    @planes.delete(plane_for_takeoff)
+    return plane_for_takeoff
+  end
+
+private
```
dafuloth: Making internal methods private! That's a good idea

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
+require_relative 'plane.rb'
+
+class Airport
+  DEFAULT_CAPACITY = 20
+  attr_reader :planes, :airport_name, :capacity
```
dafuloth: parallel assignment is good, I should look to use it more

PR #1313, opened 2019-02-10
First pull request

Brendao1:
Basic functionality with .land and .takeoff methods.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -4,5 +4,8 @@ class Airport
     @planes = []
   end
 
+def land
+end
+
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -8,12 +8,11 @@ attr_reader :planes
     @planes = []
   end
 
-def land(plane)
-  @planes << plane
-end
+  def land(plane)
+     @planes << plane
+  end
 
-def takeoff(plane)
-end
-
-
+  def takeoff(*)
+     @planes.pop()
+  end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,14 +2,27 @@ require_relative 'plane'
 
 class Airport
 
-attr_reader :planes
+attr_reader :planes, :capacity
 
-  def initialize
+  def initialize(capacity=3)
+    @capacity = capacity
     @planes = []
   end
 
+  def full?
+    if @planes.size == capacity
+      true
+    else
+      false
+    end
+  end
+
   def land(plane)
-     @planes << plane
+     if full?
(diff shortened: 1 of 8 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  def full?
+    if @planes.size == capacity
+      true
+    else
+      false
+    end
+  end
+
+  def land(plane)
+     if full?
```
kate102: Guard Clause would be...  return true if @planes.size == capacity

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  it 'lets a plane land, adds that plane to the array and returns it' do
+    a = Airport.new
+    p = Plane.new
+    a.land(p)
+    expect(subject.land(p)).to eq subject.planes
+  end
+
+  it 'raises an error if the airport is full' do
```
kate102: This is a problem because you are trying to land a plane

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  it { is_expected.to respond_to :takeoff }
+
+  describe '#takeoff' do
+    it 'lets a plane takeoff, removes that plane from the array and returns that plane' do
+     a = Airport.new
+     p = Plane.new
+     subject.land(p)
+     expect(subject.takeoff(p)).to eq p
```
kate102: Has the plane taken off? This should return true or false

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    5.times{ a.land(p) }
+    expect { a.land(p) }.to raise_error(ArgumentError, "Airport is full")
+  end
+
+  it { is_expected.to respond_to :takeoff }
+
+  describe '#takeoff' do
+    it 'lets a plane takeoff, removes that plane from the array and returns that plane' do
+     a = Airport.new
+     p = Plane.new
```
kate102: Plane creation would assume its no airborne, possibly?

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    p = Plane.new
+    5.times{ a.land(p) }
+    expect { a.land(p) }.to raise_error(ArgumentError, "Airport is full")
+  end
+
+  it { is_expected.to respond_to :takeoff }
+
+  describe '#takeoff' do
+    it 'lets a plane takeoff, removes that plane from the array and returns that plane' do
+     a = Airport.new
```
kate102: I think that you can get rid of this a = Airport.new

PR #1314, opened 2019-02-10
The airplanes can land and take_off now! passengers hurry up!

soroushh:


lib/plane.rb
```diff
@@ -0,0 +1,19 @@
+class Plane
+  def land(airport)
```
simon-norman: perhaps move these guard conditions into the airport
it may be more appropriate for the airport to have the responsibility
 of saying when it is full

spec/airport_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "airport"
+
+describe Airport do
+  it "Making an airport in the airport class." do
```
simon-norman: # Maybe this test can be removed? If kept in, I would maybe add an expect
  # statement to make it clearer what you are testing

spec/plane_spec.rb
```diff
@@ -0,0 +1,86 @@
+require "plane"
+describe Plane do
+  it "Making a plane in possible" do
```
simon-norman: # Maybe this test can be removed? If kept in, I would maybe add an expect
  # statement to make it clearer what you are testing

spec/plane_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    plane = Plane.new()
+  end
+
+  it "A plane is able to land in an airport,when the weather is sunny." do
```
simon-norman: The tests could be shortened by moving some of it into a shared set up
 EG the plane and airport instantiation, and setting the airport to receive
sunny could be moved into a before each
For each of these, maybe have an expect statement to make it clearer what is tested?

spec/plane_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    plane = Plane.new()
+    airport = Airport.new()
+    allow(airport).to receive(:weather_conditions) { "sunny" }
+    plane.land(airport)
+  end
+
+  it "A plane is able to take off from an airport when the weather is good." do
```
simon-norman: It might be good to separate blocks of code that are more than 3 lines 
into separate sections, to make them more readable

spec/plane_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    expect(plane.in?(airport)).to eq true
+
+  end
+
+  it "when a plane has taken off the airport, it is not in airport anymore." do
+    plane = Plane.new()
+    airport = Airport.new()
+    allow(airport).to receive(:weather_conditions) { "sunny" }
+    plane.land(airport)
+    allow(airport).to receive(:weather_conditions) { "sunny" }
```
simon-norman: I think this allow can be removed, it's a duplicate of the one above

lib/plane.rb
```diff
@@ -0,0 +1,19 @@
+class Plane
```
simon-norman: Currently the plane is the interface, possibly the airport could be the main interface. The user stories note the user is the air traffic controller, rather than the pilot, so possibly the airport class could be the input for when planes take off or land

PR #1315, opened 2019-02-10
Exercise completed

simon-norman:


spec/unit/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
+require './lib/airport.rb'
+
+describe Airport do
+  describe 'Given weather conditions are sunny, ' do
```
soroushh: a comment

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+class Airport
+  attr_reader :planes
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity: DEFAULT_CAPACITY, weather:)
```
soroushh: What is this "weather:" here?

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @capacity = capacity
+    @planes = []
+    @weather = weather
+  end
+
+  def land(plane)
+    raise 'Plane cannot land as airport is full' if full?
```
soroushh: This items could be wrapped in different methods I think. In this way, every method would do a single task.

lib/plane.rb
```diff
@@ -0,0 +1,19 @@
+class Plane
+  attr_reader :is_airborne
+  
+  def initialize(is_airborne = false)
+    @is_airborne = is_airborne
+  end
+
+  def land
+    raise 'Plane is already on the ground' unless @is_airborne
+
```
soroushh: what is this is_airporne? argument?

soroushh: In this way, you are making every airplane able to be flying or not? yeah?

PR #1317, opened 2019-02-10
Completed submission for the airport challenge

matthewwhitaker:
Full details can be found in the README.md


spec/airport_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    new_plane = double('plane')
+    expect { subject.take_off(new_plane) }.to raise_error 'cannot take off, plane not in airport!'
+  end
+
+  context 'when stormy' do
+
+    it 'prevents planes taking off' do
+      subject = Airport.new(20)
+      new_plane = double('plane')
+      allow(subject.weather).to receive(:stormy_currently?).and_return false
```
Harrison25000: Not 100% sure if lines 63 & 64 are necessary as we're just testing for stormy? I may be wrong though.

spec/airport_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      subject = Airport.new(3)
+      allow(subject.weather).to receive(:stormy_currently?).and_return false
+      plane1 = double('plane')
+      subject.land(plane1)
+      plane2 = double('plane')
+      subject.land(plane2)
+      plane3 = double('plane')
+      subject.land(plane3)
+      plane4 = double('plane')
+      expect { subject.land(plane4) }.to raise_error 'cannot land - airport full'
```
Harrison25000: Could this test be simplified? Maybe to something like where '>@capacity' is involved?

PR #1318, opened 2019-02-10
Completed airport challenge... kind of.

Harrison25000:
I mean it works. I may have not read the README entirely and forgot to add separate classes and now I'm out of time to change it, but I think it works.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,5 @@
+class Airport
+  def landing(plane)
+    plane
+  end
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,7 +1,10 @@
 class Airport
   attr_reader :at_airport
+  attr_reader :capacity
   def initialize
     @at_airport = []
+    puts "Enter capacity number:"
+    @capacity = gets.chomp.to_i
   end
 
   def stormy
@@ -11,8 +14,9 @@ class Airport
   def landing(plane)
     if @stormy == true
       "Too stormy!"
-    elsif @at_airport.length == 20
+    elsif @at_airport.length == @capacity
       "Airport is full"
+
     else
       @at_airport.push(plane)
       print @at_airport
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -11,6 +11,8 @@ class Airport
   def landing(plane)
     if @stormy == true
       "Too stormy!"
+    elsif @at_airport.length == 20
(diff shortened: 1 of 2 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    if airport.stormy == true
+      expect(airport.landing("plane123")).to eq("Too stormy!")
+    end
+    puts airport.landing("plane123")
+  end
+
+  it 'cannot land when full' do
+    puts "-----"
+    airport = Airport.new
+    if airport.at_airport.length == :capacity
```
matthewwhitaker: tests including conditionals may risk being bypassed by users specifying capacity. Could it be more reliable to make test context very specific rather than trying to cover everything in one using conditionals?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+class Airport
+  attr_reader :at_airport
+  attr_reader :capacity
+  def initialize
+    @at_airport = []
+    puts "Enter capacity number:"
+    @capacity = gets.chomp.to_i
+  end
+
+  def stormy
```
matthewwhitaker: according to README weather intended to be randomly generated, but here is observed by the airport, essentially defining if it is stormy or not. Could you put weather as a class to define this behaviour then respond to it?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+      plane
+    end
+  end
+
+  def takeoff(plane)
+    if @stormy == true
+      "Too stormy!"
+    else
+      @at_airport.length.positive?
+      @at_airport.shift[plane]
```
matthewwhitaker: I found I was able to take off planes which don't exist in @at_airport. Could you add a test for this, and raise an appropriate error?

It seems this may be because 'shift' removes the first element in the array rather than the one matching an argument. 'Delete' may do what you're intending.

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+class Airport
+  attr_reader :at_airport
+  attr_reader :capacity
+  def initialize
+    @at_airport = []
+    puts "Enter capacity number:"
+    @capacity = gets.chomp.to_i
```
matthewwhitaker: The final user story (i.e. having default value for capacity) seems to work here with a default of zero, if an integer is not passed in line 7, but is not explicit.

PR #1319, opened 2019-02-10
Pull Request Airport Challenge 

kate102:
The current status of this challenge is the following:

------------------------------------------------------------------------
The results of rspec
Finished in 0.00388 seconds (files took 0.19891 seconds to load)
12 examples, 0 failures
COVERAGE: 100.00% -- 83/83 lines in 6 files

------------------------------------------------------------------------
The results of rubocop
Inspecting 9 files
.........
9 files inspected, no offenses detected

------------------------------------------------------------------------
I got to the stage where I needed to write a stub to override random weather conditions.
I ran out of time looking up stubs.

To Do:
Test stub
Edge Cases
Doubles


lib/weather.rb
```diff
@@ -0,0 +1,15 @@
+# This is the weather class
+
+class Weather
+ 
+  attr_writer :condition
```
Brendao1: Could this be changed to attr_accessor?

PR #63, opened 2015-06-14
Complete airport challenge

jonathansayer:

PR #1320, opened 2019-02-10
Airport Challenge Pull req.

Schofield88:
For your own safety, please avoid flying Schofield Airways.

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,5 @@
+require 'airport'
+
+describe Airport do
+
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,12 +1,19 @@
 require 'airport'
+require 'plane'
 
 describe Airport do
+
+  let(:lear) { Plane.new }
+  let(:gatwick) { Airport.new }
 
   it 'is actually a real Airport' do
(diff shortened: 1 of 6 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+  let(:lear) { Plane.new }
+  let(:gatwick) { Airport.new }
+
+
+  it 'is actually a real Airport' do
+    expect { Airport.new }.to_not raise_error
+  end
+
+  it "registers a landed plane" do
+    allow(gatwick).to receive(:stormy?) { false }
```
acodeguy: these lines can be called once using before(:each) to DRY-up tests

spec/airport_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  it "won't let a plane take off in stormy conditions" do
+    allow(gatwick).to receive(:stormy?) { true }
+    lear.location = gatwick
+    lear.status = "Landed."
+    gatwick.hangar << lear
+    expect{lear.take_off}.to raise_error("Takeoff prohibited due to weather conditions.")
+  end
+
+  it { is_expected.to respond_to :full? }
```
acodeguy: this could be considered redundant when full? is later tested later in a different test (line #63)

PR #1321, opened 2019-02-11
Airport Challenge

mmar8:
Done as much as I could. Submitting final code.


spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  before :each do
```
MinaMinchy: I like this, I didn't use it , it shorter, so we don't need to put every time the same line  conditions

PR #1322, opened 2019-02-11
airport_challenge, Amina

MinaMinchy:
I will continue to work on airport_challenge during the week.

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,5 @@
+require 'airport'
+
+describe Airport do
+  
+end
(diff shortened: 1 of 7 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'airport'
+
+
+describe Airport do
+  subject(:airport) { described_class.new(Airport::DEFAULT_CAPACITY) }
+  let(:plane) { double :plane}
+
+
+ describe 'instruct_to_land' do
+   it ' instructs a plane to land' do
```
mmar8: Its possible to convert it to one liner expression

PR #1323, opened 2019-02-11
It was OK but plenty of room for improvement

dafuloth:


spec/plane_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+
+  it "should be possible to initialize a plane to a specified airport" do
+    airport = Airport.new
+    plane_1 = Plane.new(airport)
+
+    expect(plane_1.airport).to eq(airport)
+    expect(airport.landed_planes).to include(plane_1)
+  end
+
+  describe "take_off(airport, stormy)" do
```
tso402: I think you could use a context subset for Not Stormy to enclose the tests where not stormy is the main condition being tested

spec/plane_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+    it "already landed, must take off before it can land" do
+      plane = Plane.new
+      airport = Airport.new
+      stormy = false
+      
+      expect(plane.airport).not_to eq nil
+      expect { plane.land(airport, stormy) }.to raise_error("Error. Already landed. Take off first")  
+    end
+
+    it "land at airport because airport is not full, and it is not stormy" do
```
tso402: Not stormy context could be used to group the tests. Potentially full and not full could also be contexts.

spec/plane_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+      
+      # unnecessary:
+      # expect(!airport.full?)
+
+      expect { plane.land(airport, stormy) }.to raise_error("Landing aborted: Stormy.") 
+
+      expect(airport.landed_planes).not_to include(plane)
+    end
+
+    it "aborts landing because destination airport is full" do
```
tso402: Full context could be used

spec/plane_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+  describe "Bonus: lands and takes off a number of planes" do
+    heathrow = Airport.new(3)
+    frankfurt = Airport.new(1)
+    cdg = Airport.new(2)
+
+    gb_plane = Plane.new(heathrow)
+    de_plane = Plane.new(frankfurt)
+    fr_plane = Plane.new(cdg)
+
+    it "" do
```
tso402: Should there be text for the test here?

spec/plane_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+      expect(airport.full?).to be true
+
+      expect { plane.land(airport, stormy) }.to raise_error("Landing aborted: Stormy and Airport full") 
+
+      expect(airport.landed_planes).not_to include(plane)
+    end
+
+  end
+
+  describe "Bonus: lands and takes off a number of planes" do
```
tso402: Is this the correct description as the tests seem to be checking capacity full rather than landing or taking off?

spec/weather_spec.rb
```diff
@@ -0,0 +1,19 @@
+require_relative "../lib/weather"
+
+describe Weather do
+
+  it { is_expected.to respond_to :stormy? }
+  
+  it '#stormy returns true if the random number is 0' do
```
tso402: Could use doubles to test random behaviour rather than fixing values

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
```
tso402: Not clear what the purpose of this class is, looks like other classes are interacting directly with it's instance variables rather than via class methods?

lib/airport.rb
```diff
@@ -0,0 +1,17 @@
+class Airport
+  attr_accessor :capacity
+  attr_accessor :landed_planes
+
+  def initialize(capacity = 10)
```
tso402: Could use class constant instead of a number here

spec/plane_spec.rb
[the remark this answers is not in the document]
dafuloth: Thanks. I definitely need to improve my knowledge of rspec syntax/patterns

spec/plane_spec.rb
dafuloth: my bad... "Bonus" section was work in progress, should not have pushed that.

spec/weather_spec.rb
dafuloth: And now I will go and learn the rspec syntax so that I can use doubles

PR #1325, opened 2019-02-11
airport challenge

Anuluxmy:
Only halfway


lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+    plane = Plane.new
+    airport.weather("stormy")
+    plane.plane_landed
+    @planes << plane
+  end
+
+  def take_off(plane)
+    plane = Plane.new
+    airport = Airport.new
+    if plane.plane_flying?
```
limavadakara: The take_off method of Plane allows take_off of a Plane already in flight.

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+      plane.plane_takeoff
+    end
+  end
+
+  def weather(random)
```
limavadakara: The weather method probably doesn't need an input as random is being overwritten anyway.

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  def initialize
+    @capacity = 5
+    @planes = []
+  end
+
+  def land(plane)
+    full?
+    airport = Airport.new
+    plane = Plane.new
+    airport.weather("stormy")
```
limavadakara: airport object is passing an input to the weather method, which is not used in the weather method. Also, Airport's land method is not using the response back from the weather method to determine whether landing is allowed.

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+    airport = Airport.new
+    if plane.plane_flying?
+      plane.plane_takeoff
+    end
+  end
+
+  def weather(random)
+    random = rand(1..2)
+    if random == 1
+      return "stormy"
```
limavadakara: The return from the weather method "sunny" or "stormy". This leaves the airport object to determine which one is good weather; so perhaps there is a method missing to tell the airport object that it is either good or bad weather.

PR #1329, opened 2019-03-10
Implemented the given user stories

Hives:
Outstanding:

- Edge cases
- Refactor?
- Code quality review
- $BONUS task


spec/airport_spec.rb
```diff
@@ -0,0 +1,230 @@
+require 'airport'
+
+describe Airport do
```
samanthaixer: Good set of tests that seem to cover a range of scenarios. Perhaps because of the volume of tests, I am sometimes finding it hard to follow what is being tested at each point. I will highlight at least one example below

spec/airport_spec.rb
```diff
@@ -0,0 +1,230 @@
...
+      end
+
+      context 'when the plane is already at the airport' do
+        before :each do
+          allow(@weather).to receive(:stormy?).and_return false 
+          @airport.land @plane
+          allow(@weather).to receive(:stormy?).and_return true
+        end
+
+        it 'raises an error' do
```
samanthaixer: I have found this set of tests a little confusing, perhaps just the ordering of the stubs and context descriptions

README.md
```diff
@@ -1,90 +1,52 @@
...
-2. Run the command `gem install bundle` (if you don't have bundle already)
-3. When the installation completes, run `bundle`
-4. Complete the following task:
-
-Task
------
+## Instructions
 
-We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
+In root folder execute `irb -r ./lib/airport.rb -r ./lib/weather_station.rb -r ./lib/plane.rb`.
```
mattstockman: Clear instructions on how to run the programme

README.md
```diff
@@ -1,90 +1,52 @@
...
+After implementing the features from the user stories I remembered about RSpec `context` blocks and spent a while refactoring all my tests...
 
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this weekend.
+After that I started writing user stories for the edge cases, and using the same process to implement them. I ran out of time before finishing all of the edge cases, but here's how I would have continued:
 
-**BONUS**
+### Outstanding edge cases
 
-* Write an RSpec **feature** test that lands and takes off a number of planes
+- Planes that are flying cannot take off / planes that are not flying cannot land
```
mattstockman: all steps clearly outlined

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+class Airport
+  DEFAULT_CAPACITY = 3
+
+  def initialize(weather, capacity = DEFAULT_CAPACITY)
```
samanthaixer: Could we discuss the benefits to setting the default in the initialize arguments versus as a constant outside the initialize?

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+  
+  def land(plane)
+    raise "That plane is already on the ground!" if @planes.include? plane
+    raise "⛈ Too dangerous to land. ⛈" if bad_weather?
+    raise "Airport full" if @planes.size >= @capacity
+
+    @planes << plane
+    puts '☀️ Plane landed! ☀️'   
+    self
```
samanthaixer: What is self here? And is there a reason that the confirmation message is output to the screen rather than returned by the method?

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def include?(plane)
+    @planes.include? plane
+  end
+
+  private
+
+  def bad_weather?
```
samanthaixer: Good naming convention for method. It makes reading the code where this method is called very easy

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+class Airport
```
mattstockman: nicely indented making it easy to read

lib/plane.rb
```diff
@@ -0,0 +1,21 @@
+class Plane
+  def initialize
+    @flying = true
+  end
+  
+  def flying?
```
samanthaixer: Could this be an attribute reader instead of method?

lib/plane.rb
```diff
@@ -0,0 +1,21 @@
...
+    @flying
+  end
+
+  def land
```
samanthaixer: Good simple methods that are easy to understand

lib/plane.rb
Hives: yes!

spec/airport_spec.rb
Hives: yes me too

lib/airport.rb
Hives: done

lib/airport.rb
Hives: not sure if i've done the right thing here, i will raise at the code review

lib/airport.rb
Hives: ta!

PR #1331, opened 2019-03-10
James Palmer Airport Challenge

jpalmerr:
My Airport Challenge code, week 1.


lib/air_traffic_control.rb
```diff
@@ -4,7 +4,14 @@
 
 class Control
 
+  attr_reader :capcity
```
fo0s: There's a typo on :capcity

PR #1332, opened 2019-03-10
Submitting a pull request for Airport challenge

WardahTarar:


README.md
```diff
@@ -13,78 +12,19 @@ Airport Challenge
...
-1. Fork this repo, and clone to your local machine
-2. Run the command `gem install bundle` (if you don't have bundle already)
-3. When the installation completes, run `bundle`
-4. Complete the following task:
+## Motivation
+The motivation behind this program was that to:
+- write code that is easy to change
+- can test-drive my code
+- can build with objects
+- know my language works
```
kimdiep: What you did well:

Good to see your intentions / notes for the project :)
Nice to see list of tech / framework used

kimdiep: Improvements: 
- Would be good to see a list of stories and approach to solving the problem
- It's a great idea to show the full story of how your app is used (from a user's perspective) in the README too to show them how to use it

spec/airport_spec.rb
```diff
@@ -3,55 +3,49 @@
...
-    plane = Plane.new
-    expect(subject.land(plane, "Sunny")).to eq [plane]
-  end
-
-  it "checks if more than one plane can land on an airport when instructed" do
+  it "checks if one or more plane(s) can land on an airport when instructed" do
     plane1 = Plane.new
     subject.land(plane1, "Sunny")
     plane2 = Plane.new
     expect(subject.land(plane2, "Sunny")).to eq [plane1, plane2]
```
kimdiep: I liked the way you approached this test :)

kimdiep: You have isolated the weather factors from the landing well

spec/airport_spec.rb
```diff
@@ -3,55 +3,49 @@
...
-    10.times { subject.land(Plane.new, "Sunny") }
-    expect(subject.change_capacity(15)).to_not eq "Airport is full"
+  it "Prevents a plane from landing on an airport when capacity of 10 is full" do
+    11.times { subject.land(Plane.new, "Sunny") }
+    expect(subject.land(Plane.new, "Sunny")). to eq "Airport is full"
   end
   
+  it "Changes the default capacity of 10 of an airport" do
+    subject.change_capacity(20)
+    15.times { subject.land(Plane.new, "Sunny") }
```
kimdiep: Nice way to test the capacity :)

lib/airport.rb
```diff
@@ -15,6 +16,10 @@ def land(plane, weather)
     end
   end
 
+  def change_capacity(new_capacity)
```
kimdiep: Good to see there is functionality to change the capacity after the object has been instantiated :)

spec/airport_spec.rb
```diff
@@ -49,9 +49,9 @@
     expect(subject.land(plane, "Stormy")).to eq 'No landing allowed in stormy weather'
   end
 
-  it "Prevents plane from landing when an airport when capacity is full" do
-    10.times { subject.land(Plane.new, "Sunny" )}
```
kimdiep: Nice test

spec/airport_spec.rb
kimdiep: Improvements:

- Structuring your tests into context and describe blocks can help to itemise your test scopes a bit better :)

Here is a snippet from the code rubric to help:

"Use context and describe blocks to create test scopes
If a group of tests share the same setup or are related logically, group them in a context block or a describe block. Use describe when the tests are related by a subset of behaviour (e.g 'landing') and use context when the tests are related by program state (e.g. 'when it is stormy')."

lib/airport.rb
```diff
@@ -6,7 +6,7 @@ def initialize(default_capacity = 10)
     @capacity = default_capacity
   end
 
-  def land(plane, weather, capacity)
+  def land(plane, weather)
     @plane = plane
     if @runway.length >= @capacity
```
kimdiep: I found your approach to 'land' method interesting :) Though there are multiple ways to solve the same problem, I particularly liked the way you put the weather as an argument too :) 

Another way of solving the same user story is to delegate the responsibility of the weather outside into another class/method perhaps?

spec/airport_spec.rb
```diff
@@ -46,12 +46,12 @@
 
   it "Prevents plane from landing on an airport when weather is stormy" do
     plane = Plane.new
-    expect(subject.land(plane, "Stormy", @capacity)).to eq 'No landing allowed in stormy weather'
+    expect(subject.land(plane, "Stormy")).to eq 'No landing allowed in stormy weather'
```
kimdiep: Nice one! :)
Same as with my code as well, using 'before' blocks might help to set up objects before your test cases :)

Here is a snippet from the code rubric which might help:

Use before blocks to set up objects rather than repeat code
For example, to set up stubbing behaviour that is shared across a number of tests:

describe 'a group of tests that need to call #land on a plane double' do
  before do
    allow(plane).to receive(:land)
  end

  it ...
end

spec/airport_spec.rb
```diff
@@ -46,12 +46,12 @@
...
   end
 
   it "Prevents plane from landing when an airport when capacity is full" do
-    10.times { subject.land(Plane.new, "Sunny", @capacity) }
-    expect(subject.land(Plane.new, "Sunny", @capacity)). to eq "Airport is full"
+    10.times { subject.land(Plane.new, "Sunny" )}
+    expect(subject.land(Plane.new, "Sunny")). to eq "Airport is full"
```
kimdiep: Another way of doing this would be to raise an exception

lib/airport.rb
```diff
@@ -1,13 +1,18 @@
 class Airport
   attr_accessor :runway
 
-  def initialize 
+  def initialize(default_capacity = 10)
     @runway = []
+    @capacity = default_capacity
```
kimdiep: Good use of instance variable :)

lib/airport.rb
```diff
@@ -5,29 +5,19 @@ def initialize
     @runway = []
   end
 
-  def land(plane)
+  def land(plane,weather)
     @plane = plane
-    @runway << @plane
+    weather == "Stormy" ? 'No landing allowed in stormy weather' : @runway << @plane
```
kimdiep: Ooo....ternary operators :) Cool!

lib/airport.rb
```diff
@@ -5,29 +5,19 @@ def initialize
     @runway = []
   end
 
-  def land(plane)
+  def land(plane,weather)
     @plane = plane
-    @runway << @plane
```
kimdiep: Good use of the shovel operator :)

lib/airport.rb
kimdiep: Improvements:

- Probably delegate the weather checking responsibility to another method perhaps? :)

lib/weather.rb
```diff
@@ -1,9 +1,9 @@
 class Weather
   def weather_check
-    if (rand(100)) < 20
+    if rand(100) < 20
```
kimdiep: Good use of the random number generation to signify the percentage likelihood of the type of weather happening :)

Improvements:

- You can refactor this into an array ["stormy", "sunny", "sunny", "sunny"] and sample from that array.

Here is the information from the code rubric which can help:


It can be refactored to have only one responsibility. Although this introduces more code, the goal is readability and reducing cognitive overload when scanning the code:

```
def stormy?
  random_outlook == :stormy
end

private

OUTLOOKS = [:stormy, :fine, :fine, :fine]

def random_outlook
  OUTLOOKS.sample
end
```
Note: Ruby already handles the responsibility of choosing randomly from and array with the sample method.

spec/airport_spec.rb
```diff
@@ -3,35 +3,62 @@
...
+    plane2 = Plane.new
+    subject.land(plane2)
+    expect(subject.takeoff(plane2, "Stormy")).to eq 'No take off allowed in stormy weather'
+  end
+
 end
+
+#     airport = Airport.new [plane]
+#     airport.stub(:weather_rand).and_return(5)
+#     expect(lambda { airport.plane_take_off(plane) }).to raise_error(RuntimeError)
```
kimdiep: Lambdas ahoy! I haven't figured out how to use lambdas myself yet, but I like your implementation

spec/airport_spec.rb
```diff
@@ -27,5 +28,10 @@
     subject.instructions("land",plane2)
     expect(subject.instructions("takeoff",plane2)).to eq [plane1]
   end
-  
+
+  it "Checks if the weather is stormy or sunny" do
+    weather = Weather.new
+    expect(weather.weather_check).to eq("Sunny") | eq("Stormy")
```
kimdiep: Improvement:

Consider the use of a "sunny" / "stormy" array, then you can check if it is in the array :)

I do like the use of your test though :)

spec/weather_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "weather"
+
+describe Weather do
+  it "responds to weather condition" do
+    weather = Weather.new
+    expect(weather).to respond_to(:weather_check)
```
kimdiep: Good to see you're checking the weather response to a condition :)

Gemfile.lock
```diff
@@ -53,9 +53,9 @@ PLATFORMS
 DEPENDENCIES
   rake
   rspec
-  rubocop (= 0.56.0)
+  rubocop (= 0.60.0)
```
kimdiep: Yay! Rubocop!! :)

spec/airport_spec.rb
kimdiep: https://medium.freecodecamp.org/how-to-dry-out-your-rspec-tests-using-shared-examples-d5cc5d33fd76

PR #1338, opened 2019-03-10
Airport Challenge - end of Week 1

samanthaixer:
This is my weekend Airport Challenge for Week 1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,90 +1,47 @@
 Airport Challenge
 =================
 
-```
-        ______
-        _\____\___
-=  = ==(____MA____)
-          \_____\___________________,-~~~~~~~`-.._
-          /     o o o o o o o o o o o o o o o o  |\_
-          `~-.__       __..----..__                  )
-                `---~~\___________/------------`````
-                =  ===(_________)
 
-```
-
-Instructions
+a
+My approach
 ---------
 
-* Challenge time: rest of the day and weekend, until Monday 9am
+* Read the user stories and break them down into things I need the programme to already do
 * Feel free to use google, your notes, books, etc. but work on your own
 * If you refer to the solution of another coach or student, please put a link to that in your README
 * If you have a partial solution, **still check in a partial solution**
 * You must submit a pull request to this repo with your code by 9am Monday morning
 
-Steps
+Reflection
 -------
 
-1. Fork this repo, and clone to your local machine
-2. Run the command `gem install bundle` (if you don't have bundle already)
-3. When the installation completes, run `bundle`
-4. Complete the following task:
+1. I didn't feature test anything in ```irb``` as I went. Would this have changed my design/approach at all?
+2. I have written what I think is the cleanest, simplest code for the task that I have ever written. My concern - have I actually answered all of the brief?
+3. My head just doesn't want to think up new edge cases!
 
-Task
------
-
-We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
 
 ```
-As an air traffic controller 
-So I can get passengers to a destination 
+As an air traffic controller
+So I can get passengers to a destination
 I want to instruct a plane to land at an airport
 
-As an air traffic controller 
-So I can get passengers on the way to their destination 
+As an air traffic controller
+So I can get passengers on the way to their destination
 I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent takeoff when weather is stormy 
+As an air traffic controller
+To ensure safety
+I want to prevent takeoff when weather is stormy
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent landing when weather is stormy 
+As an air traffic controller
+To ensure safety
+I want to prevent landing when weather is stormy
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent landing when the airport is full 
+As an air traffic controller
+To ensure safety
+I want to prevent landing when the airport is full
 
 As the system designer
 So that the software can be used for many different airports
 I would like a default airport capacity that can be overridden as appropriate
 ```
-
-Your task is to test drive the creation of a set of classes/modules to satisfy all the above user stories. You will need to use a random number generator to set the weather (it is normally sunny but on rare occasions it may be stormy). In your tests, you'll need to use a stub to override random weather to ensure consistent test behaviour.
-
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -1,6 +1,35 @@
...
 =================
 
+My approach
+---------
+
+* I read the user stories and broke them down into things I need the programme to do
+* For each line of "I want to be able to do this", I wrote the simplest test, fixed it, and repeated. I refactored my code every time the test passed.
+* Every now and again I refactored tests to make them easier to read/find and to ensure they didn't do unnecessary steps
+* Every now and again I ran rubocop to lint my code
+* Once I had done enough tests to complete a particular thing the code should do, I check the user stories to make sure I was still following the brief
```
Hives: sounds good!

README.md
```diff
@@ -1,6 +1,35 @@
...
+1. I didn't feature test anything in ```irb``` as I went. Would this have changed my design/approach at all?
+2. I have written what I think is the cleanest, simplest code for the task that I have ever written. My concern - have I actually answered all of the brief?
+3. My head just doesn't want to think up new edge cases!
+
+How to run my code
+-------
+
+1. Run any of the test files to test the classes. ```airport_spec``` or ```weather_spec``` or ```plane_spec```
+2. Run any of the commands in ```irb``` or a ruby file ```Airport.new``` ```Plane.new``` ```Weather.new```
+
```
Hives: i think you need to include something like:
- run `gem install bundle` if you haven't already
- run `bundle`
or else the user won't be able to get the tests to run?

i think you should also give more explicit instructions about how to run the program, e.g.:
- open irb
- run `airport = Airport.new`
- run `plane = Plane.new`
- run `airport.land plane`
etc. (i haven't read any of your code yet, so i don't know how you've implemented it :) )

README.md
```diff
@@ -1,6 +1,35 @@
...
+* NOTE: Line 7 of the weather class is not covered by a test. I have not worked out how to cover it.
+* NOTE: I have started the feature test but haven't worked out how to make it work with RSpec feature yet.
+
+Reflection
+-------
+
+1. I didn't feature test anything in ```irb``` as I went. Would this have changed my design/approach at all?
+2. I have written what I think is the cleanest, simplest code for the task that I have ever written. My concern - have I actually answered all of the brief?
+3. My head just doesn't want to think up new edge cases!
```
Hives: i feel your pain

README.md
```diff
@@ -1,6 +1,35 @@
 Airport Challenge
 =================
```
Hives: code review checklist says you should have a description of what the program does at the top of the README

lib/Plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
+end
```
Hives: title of source files should be in snake_case (i.e. `plane.rb`). (rubocop picks this up)

lib/Airport.rb
```diff
@@ -0,0 +1,35 @@
...
+      return "Successful take off"
+    end
+
+    raise "That plane isn't at the airport"
+  end
+
+  def capacity
+    return @max_capacity - @plane_list.length
+  end
+end
```
Hives: title of source files should be in snake_case (i.e. `airport.rb`). (rubocop picks this up)

spec/airport_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      allow(subject).to receive(:current_weather) { "Sunny" }
+      expect(subject.land(Plane.new)).to eq "Plane landed"
+    end
+
+    it 'should stop a plane from landing when the weather is Stormy' do
+      allow(subject).to receive(:current_weather) { "Stormy" }
+      expect { subject.land(Plane.new) }.to raise_error "No landing allowed - Stormy weather"
+    end
+
+    it 'should stop a plan from landing when the airport is full' do
```
Hives: 'plan' -> 'plane'

spec/airport_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    it { is_expected.to respond_to(:take_off).with(1).argument }
+
+    it 'should return a confirmation that the plane has taken off' do
+      allow(subject).to receive(:current_weather) { "Sunny" }
+      my_plane = Plane.new
+      subject.land(my_plane)
+      expect(subject.take_off(my_plane)).to eq "Successful take off"
+    end
+
+    it 'should raise an error if I try to make a non-existent plane take off' do
```
Hives: the plane isn't non-existent, it's just not at the airport. you could make the description of this test more explicit.

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+class Weather
+  def current_weather
+    random = Random.rand(1..50)
+    p random
```
Hives: can delete this line

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+class Weather
+  def current_weather
```
Hives: if you renamed this method something like `stormy?` you coud have it return either true or false, which would allow you to write it more briefly. You'd have to rewrite the relevant parts of airport.rb too

lib/Airport.rb
```diff
@@ -0,0 +1,35 @@
+class Airport
+
+  attr_reader :plane_list
+  attr_reader :max_capacity
+
+  def initialize(max_capacity = 20)
```
Hives: 20 is a magic number here. refactor using a constant.

lib/Airport.rb
```diff
@@ -0,0 +1,35 @@
+class Airport
+
+  attr_reader :plane_list
+  attr_reader :max_capacity
```
Hives: i think both of these variables should be private...

do you actually need to access max_capacity from outside the object, since you've got the `capactity` method?

[the rubric for the exercise says the plane list should be private](https://github.com/makersacademy/airport_challenge/blob/master/docs/review.md#do-not-expose-internal-implementation), so you could make a `airport.include? plane` method which would allow you to check whether a particular plane is at the airport

lib/Airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  def take_off(plane)
+    raise "No take-off allowed - Stormy weather" if current_weather == "Stormy"
+    raise "There are no planes at the airport" if plane_list.empty?
+
+    if @plane_list.include?(plane)
+      @plane_list.delete(plane)
+      return "Successful take off"
+    end
+
+    raise "That plane isn't at the airport"
```
Hives: guard condition should go at the top of the method i think?

README.md
samanthaixer: Good point!

README.md
samanthaixer: Agreed

lib/Airport.rb
samanthaixer: Good to know. I would change this

lib/Airport.rb
samanthaixer: Sounds good

lib/Airport.rb
samanthaixer: I have made this change, good call

lib/Airport.rb
samanthaixer: On my machine all of the source files are in snake_case, not sure why this is showing with a capital on git! Will need to have a play with adding and removing files maybe

PR #1340, opened 2019-03-10
2

alexanderchen6:


ownREADME.md
```diff
@@ -0,0 +1,20 @@
+**OWN README**
```
amyj0rdan: Add to README explanation of the program, how the user can use it and how the user can run tests

ownREADME.md
```diff
@@ -0,0 +1,20 @@
...
+
+The last point is quite confusing, as I was thinking about changing capacity, rather than changing default capacity. So to test this, I change the default capacity in test, and the test would be to expect that airport would still be able to accommodate the number of increased planes landed.
+
+To make this test pass (which took a while), I made attribute configuration for the Airport class so that if default capacity is not set, its set to a default,(e.g. 20), if it is, it is set to default capacity. capacity is set to default to default capacity unless it is changed.
+
+To do the stub took quite a while(2- 3hours), as I experimented using a Double of airport, but it wouldn't respond to other methods in Airport class.
+
+
+Then, to pass the extra validation tests(edge cases), e.g. plane cannot takeoff if already flying, cannot land if already landed. I made test to raise error if these conditions are present. The code to pass the test would be incorporated into the takeoff and land methods.
+
```
amyj0rdan: Great to list the things that you didn't manage to get to. Maybe you could outline broadly how you would approach fixing these?

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
+require './lib/airport'
+require './lib/plane'
+
+describe Airport do
+	 # let(:default_capacity) { 20 }
+	 before { allow(subject).to receive(:stormy).and_return(false) }
```
amyj0rdan: I'm not clear if there are tests to check if the plane *doesn't* fly if it is stormy. Might be useful to group tests by context (eg, when stormy, plane doesn't take off, when not stormy, plane does take off)

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+	 describe '#land' do 
+ 		 it 'lands a plane' do 
+  			 plane = Plane.new
+  			 expect(subject.land(plane)).to eq([plane])
+  		end 
+ 		# do I replace this test?
+ 		# it 'initialize with no plane in airport' do 
+ 		# expect()
+ 		# end 
+ 		 it 'check if any plane in airport and returns all planes' do
```
amyj0rdan: Might be good to add a test for more that one plane.

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  			 expect { subject.takeoff(plane) }.to raise_error 'cannot takeoff'
+  		end 
+
+ 		 it 'plane flying or in queue cannot take off' do
+  			 plane = Plane.new
+  			 expect { subject.takeoff(plane) }.to raise_error "Plane is currently flying or in queue"
+  		end
+ 	end 
+
+	 describe '#stormy?' do
```
amyj0rdan: Might be good to have the weather as a separate class - at the moment the airport class does both the actions of the airport and determines the weather

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
+require_relative './plane.rb'
+
+class Airport
+	 attr_reader :planes
+	 attr_accessor :stormy
+	 attr_accessor :capacity
+	 attr_accessor :default_capacity
+
+	 def initialize(default_capacity = 20, capacity = default_capacity,
```
amyj0rdan: I find it a little confusing that the user would need to enter so many values at initializing an airport. Could use a constant for default_capacity that is set to @capacity unless another value is provided?

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+	 def land(plane)
+ 		# p @planes.count	
+ 		 raise 'cannot land' if stormy?
+ 		 raise 'Airport full' if full? 
+ 		 raise 'already landed' if @planes.include?(plane)
+
+ 		 @planes << plane 
+ 	end 
+
+	 def check
```
amyj0rdan: As @planes is already an attribute reader, maybe this method is no longer necessary?

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+ 	 # returns true or false
+ 	end 
+
+	 def full?
+ 		# p @planes
+ 		# p @planes.count
+ 		 true if @planes.count >= @capacity
+ 	end 
+
+	 def random_weather
```
amyj0rdan: Unclear to me how this interacts with your stormy? method - stormy seems to be set to false automatically. Does your airport ever experience stormy weather?

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+	 def check 
+ 		 @planes
+ 	end 
+
+	 def takeoff(plane)
+ 		# p @planes
+ 		 raise 'cannot takeoff' if stormy?
+ 		 raise 'Plane is currently flying or in queue' unless @planes[0] == plane
+
+ 		 @planes.pop
```
amyj0rdan: This will remove the last plane in the array rather than the specific plane given to the method - might be good to use the .delete method

lib/plane.rb
```diff
@@ -0,0 +1,12 @@
+class Plane
+  attr_reader :flying
+
+  def initialize
+  		@flying = true
+  end 
+
+  def flying?
```
amyj0rdan: Would be great to change this to false once the plane lands

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+ 		 @stormy = stormy
+ 		 @default_capacity = default_capacity
+ 		 @capacity = capacity
+ 		# if capacity != nil 
+ 		# @capacity = capacity  
+ 		# 	else
+ 		# @capacity = default_capacity
+ 		# end 
+ 	end 
+
```
amyj0rdan: This method is really clear

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+ 		 @stormy
+ 	 # returns true or false
+ 	end 
+
+	 def full?
+ 		# p @planes
+ 		# p @planes.count
+ 		 true if @planes.count >= @capacity
+ 	end 
+
```
amyj0rdan: Is there a test for random_weather?

spec/airport_spec.rb
alexanderchen6: used stub to override

spec/airport_spec.rb
alexanderchen6: added test to land 5 planes

lib/airport.rb
alexanderchen6: changed default capacity to constant

lib/airport.rb
alexanderchen6: removed

lib/airport.rb
alexanderchen6: my approach is that plane takes off in order, like in a real airport. if plane is not the first one to take off, it cannot take off.

lib/plane.rb
alexanderchen6: added plane.flying= false into plane method

PR #1341, opened 2019-03-10
Airport Challenge - Lauren

LaurenQurashi:
Week 1 Weekend Challenge


spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    before(:each) do
+      @airport = Airport.new
+      @plane = Plane.new
+      allow(@airport).to receive(:stormy?) { false }
+      # will run these functions before doing each test... pretty jazzy.
+      # mock enables me to fabricate data so that tests dependent on this
+      # won't mess up.
+    end
+
+    it 'responds to fly_plane' do
```
micosmin: What do you think about creating a separate test block for landing and a separate one for flying inside the clear weather and stormy weather?

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      expect { @airport.capacity.times { @airport.land_plane(@plane) } }.to raise_error("Too Full!")
+      # the airport.capacity.times thing is filling up our hangar to capacity,
+      # to check it won't add any more planes in.
+    end
+
+  end
+
+  context 'Stormy Weather' do
+
+    before(:each) do
```
micosmin: What do you think about not repeating the creation of plane and airport, as the only thing that changes is the weather?

spec/weather_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  before(:each) do
+    @weather = Weather.new
+  end
+
+  it 'responds to forecast' do
+    expect(subject).to respond_to :forecast
+  end
+
+  it 'returns either stormy or clear when forecast is called' do
+    expect(@weather.forecast).to eq("Stormy").or eq "Clear"
```
micosmin: I like this! I'll use in mine whenever i'll get the chance next!

spec/weather_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'weather.rb' # links the main code to the test file.
+
+describe Weather do
+  before(:each) do
```
micosmin: What do you think about using a let construct here instead of a before?

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def fly_plane
+    raise "Stop, no departures or arrivals due to stormy weather!" if stormy?
+    raise "No Planes!" if empty?
+
+    # throws an error if there isn't any planes.(potentially not needed)
+    puts "Skeet Skeet!"
+    # confirms the plane has left.
+    @hangar.pop
```
micosmin: What do you think about removing any plane, not the last one inserted in the array?

lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+class Weather
+
+  def forecast
```
micosmin: What about making forecast a class method?

PR #1349, opened 2019-03-11
Airport-challenge

harrygalliano:


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+require_relative 'planes'
+
+class Airport
+  attr_reader :planes
+
+DEFAULT_CAPACITY = 20
```
hybridbad: Perhaps add this constant to the argument in initialize to give user option to change capacity

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def initialize
```
hybridbad: Declare argument for capacity and let attribute be read by class

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    fail 'No planes available' if empty?
+    @planes.pop
+  end
+
+  def land(our_plane)
+    fail "Airport full" if full?
+    @planes.push(our_plane)
+    end
+
+
```
hybridbad: Shouldn't the predicate methods full and empty be in private?

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+DEFAULT_CAPACITY = 20
+
+  def initialize
+    @planes = []
+  end
+
+  def take_off
+    fail 'No planes available' if empty?
+    @planes.pop
```
hybridbad: this removes any plane in the array, how would you make this remove only the plane asking to take off, perhaps add an argument to take_off for a single plane object

PR #1350, opened 2019-03-22
Pull Request - my version of the airport challenge

DeannaGreen:


lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
```
samjones1001: Generally speaking, for a class to be valid it should contain both state (instance variables) and behaviour (methods).  Right now your plane class contains neither 🙂 

Given that that's the case, you might want to think about whether you actually need a plane class at all.

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
+  attr_reader :number
+
+  def initialize
+    @number = rand(1..4)
+    generate_weather
```
samjones1001: I think I'd tend to avoid calling methods from within the initialize like this - I see it as mostly for setting the initial values for an object's instance variables.

spec/airport_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { double :plane }
```
samjones1001: Great work on using doubles already to isolate your tests!

spec/airport_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  it { expect(subject).to respond_to(:land_plane) }
```
samjones1001: This kind of `respond_to` test is a great first step in your TDD process, but I tend to delete them once I have better tests in place.

If you look further down the file, you have a few other tests which look at the behaviour of the `land_plane` method.  If the Airport couldn't respond to that method then these would definitely fail, so this test is now redundant.

PR #1355, opened 2019-04-06
First commit and begining of test

stuartpet:
Start of the challenge. First commit and beginnings of tests.

### DIFF
--- a/lib/Airport.rb
+++ b/lib/Airport.rb
@@ -1,5 +1,5 @@
 class Airport
-  def plane_landed(plane)
-    plane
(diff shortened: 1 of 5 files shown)


lib/Airport.rb
```diff
@@ -1,5 +1,10 @@
+require_relative "plane"
 class Airport
-  def plane_landed
+  def land
     Plane.new
   end
+
+  def airport(plane)
```
NapperJLG: Method name could be a bit clearer - not easy to see what the method does

PR #1356, opened 2019-04-06
Add Airport land Plane function, create and update progress.md

SimonChenboYi:


spec/airport_spec.rb
```diff
@@ -0,0 +1,109 @@
+require 'airport'
+
+describe Airport do
+
+  before(:each) do
+    @weather = double :weather
+    @subject = described_class.new(weather)
+
+    def plane_example
+      plane = double :plane
```
alicelieutier: You could have all this function in one line, with something like this: 
```
plane = double(:plane, :land? => true, :taken_off? => true)
```
see https://relishapp.com/rspec/rspec-mocks/docs/basics/test-doubles

spec/airport_spec.rb
```diff
@@ -0,0 +1,109 @@
+require 'airport'
+
+describe Airport do
+
+  before(:each) do
```
alicelieutier: Overall, I don't think you are using before wrong, but there are a lot of things here, and you should keep in mind that all of them will be done before each test. Do they all need to run before each test?
This is where the difference between `before` and `let` is interesting. Here's a ressource about it:
https://dzone.com/articles/rspec-let-vs-before

SimonChenboYi: Thanks for sharing the resource.  Now, I understand why 'let' is preferred.

spec/airport_spec.rb
SimonChenboYi: Thanks, alicelieutier.  Use hash to allow the double is indeed much drier. I will refactor my code accordingly.

PR #1357, opened 2019-04-06
airport_challenge - callum marshall

callum-marshall:


lib/airport.rb
```diff
@@ -0,0 +1,39 @@
+class Airport
+
+  attr_reader :planes
```
ayelishaamoah: What is the purpose of 'attr_reader :planes' in your current program?

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'airport'
+
+describe Airport do
+  it { is_expected.to respond_to(:land).with(1).argument }
```
ayelishaamoah: I would remove this test as you are testing the land method's output later on.

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  it { is_expected.to respond_to(:take_off) }
```
ayelishaamoah: I would also remove this test as you are testing the take_off method's output later on.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+
+def full?
+  @planes.count >= CAPACITY
+end
+
+def empty?
+  @planes.empty?
+end
+
+def stormy?
```
ayelishaamoah: I really like how you structured this method, very concise and clever way of lowering the probability of the weather being stormy.

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  describe '#land' do
+    it 'lands plane' do
+      plane = Plane.new
+      allow(subject).to receive(:stormy?) { false }
+      expect(subject.land(plane)).to eq [plane]
```
ayelishaamoah: What does using square brackets around [plane] do?

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect { subject.land(Plane.new) }.to raise_error("airport is full")
+    end
+
+    it "raises an error when airport is stormy" do
+      allow(subject).to receive(:stormy?) { true }
+      expect { subject.land(Plane.new) }.to raise_error("dangerous weather!")
+    end
+  end
+
+  describe '#take_off' do
```
ayelishaamoah: What about using doubles when referencing the Plane class? e.g 'plane = double(:plane)' instead of 'plane = Plane.new' ?

myREADME.md
```diff
@@ -0,0 +1,38 @@
+# Airport Challenge #
```
ayelishaamoah: The README shows that you've used the user stories to define the program behaviour and your tests match these descriptions which shows that you've followed the TDD process. 

What about adding some information about how to install/ use the program?

ayelishaamoah:
I think your code is very readable - I found it easy to understand what each method was doing and preferred your implementation of the 'stormy?' method to my own. 

You also have good test coverage overall and all are passing. 

You could test for 'stormy?' by adding a stub for the random number in your land and take-off method tests.

Commits were frequent but could possibly be split out a bit more so that you are only committing changes to one method or test at a time and could mention what they change it.

PR #1358, opened 2019-04-07
Airport Challenge Fist Commit

guilhe0756:


lib/airport.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'planes'
+
+# I have defined an airport's maximum capacity to be 16 planes
+
+class Airport
+
+  attr_reader :airport, :capacity
+
+  def initialize
```
Xa1n: Good division of work for each file. Using both classes properly instead of bundling all methods into one file.

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+  def take_off(plane)
+    raise "Stormy weather, do not take off!" if stormy?
+
+    @airport.delete(plane)
+    plane
+  end
+
+# I have decided that the chance of stormy weather is 20%
+  def stormy?
+    rand(1..5) == 1
```
Xa1n: 'Rand(1..5)' reads slightly cryptic. Could you potentially create a separate method that randomly generates an outcome, and in your 'stormy?' method add that method? Mainly for readability purposes. The code appears to be functional otherwise.

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def override_capacity(new_capacity)
```
Xa1n: could you explain what this does in a commit? I'm having trouble understanding it.

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+    @capacity = new_capacity
+  end
+
+  private
+  def full?
+    @airport.count == @capacity
```
Xa1n: Would your test fail if '@airport.count' was greater than the '@capacity'?

lib/planes.rb
```diff
@@ -0,0 +1,35 @@
...
+    @plane.pop
+  end
+
+  private
+  # def wrong?
+  #
+  # end
+  
+  def landed?
+    @plane.count == 1
```
Xa1n: How could you expand this test if there are multiple planes that have landed?

lib/planes.rb
```diff
@@ -0,0 +1,35 @@
...
+    raise "Already landed." if landed?
+
+    @plane << airport
+    @plane[0]
+  end
+
+  def take_off(airport)
+    raise "Wrong airport." unless @plane.include?(airport)
+
+    raise "Already left the airport." if empty?
```
Xa1n: Line might be unnecessary as it tests for the same thing as the line above i.e. 'is the airport empty?' or specifically 'is the plane you are trying to take off in the specified airport?'

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'airport'
+
+RSpec.describe Airport do
+  describe '#land' do
+    it 'instructs a plane to land on an airport' do
+      plane = Planes.new
+      allow(subject).to receive(:rand) { 2 }
```
Xa1n: Same as comment about the actual code. The stub - ':rand' and the '2' is a bit cryptic.

Xa1n:
Overall, really readable code. Both classes have a clear responsibility, and seem to interact well with each other. Only a few methods that need refactoring for clarity's sake. Tests seem rigorous, although a use of doubles might have been useful to ensure you're isolating unit tests and testing individual methods. All round great though, thanks :)

lib/airport.rb
guilhe0756: Thanks!

lib/airport.rb
guilhe0756: I agree. As it is now, it is not obvious what rand is doing here, because it's just numbers. Another option would be to explain that in the readme, but then the code wouldn't be self-sufficient.

lib/airport.rb
guilhe0756: It's the implementation of the last user story: "As the system designer
I would like a default airport capacity that can be overridden as appropriate".

Basically, it overrides the value of @capacity.

lib/airport.rb
guilhe0756: I wouldn't think so, because it would never allow `@airport` to accept another plane if is full. However, to prevent that, I could replace `==` with `>=` which I assume is what you are getting at.

lib/planes.rb
guilhe0756: Fixed it, thanks.

spec/airport_spec.rb
guilhe0756: I'll have to work on that, thanks. This is great feedback.

PR #1359, opened 2019-04-07
Almost done ( still need to fix take_off error)

SarpongAbasimi:


spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  RSpec.configure do |config|
+    config.example_status_persistence_file_path = './spec/fail_test.txt'
+  end 
+
+  let(:plane) { Plane.new }
+
+  describe '#land_plane' do
+    it 'should be able to land plane' do
+      expect(subject.land(plane)).to eq([plane])
```
MaxRobertsDear: why does the line read ...to eq([plane])? why is '[plane]' in square brackets?

lib/airport.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'plane'
+
+class AirPort
+
+  def initialize
+    @track_landed_plane = []
```
MaxRobertsDear: i would reword this to '@track_landed_plane' to '@track_landed_planes' to make it obvious that you are expecting more than one plane to be stored

lib/airport.rb
```diff
@@ -2,16 +2,30 @@
...
+    delete_plane(plane)
+    confirm?(plane)
+  end
+
+  private
+  def confirm?(plane)
+    true unless @landed_plane.include?(plane)
   end
 
+  def delete_plane(plane)
```
MaxRobertsDear: by virtue of passing in the argument, you already imply that you are deleting a plane. therefore you could refactor to 'delete(plane)'

lib/airport.rb
```diff
@@ -2,16 +2,30 @@
...
+    plane_id = @landed_plane.index(plane)
```
MaxRobertsDear: LOVE the use of array here!

spec/airport_spec.rb
```diff
@@ -15,8 +15,14 @@
...
-    it 'should let plane to take off' do
-      expect(subject.take_off(plane)).to eq(plane)
+    it 'raises an error if there is no plane' do
+      expect { subject.take_off(plane) }.to raise_error 'There are currently no planes'
+    end
+  end
+
+  describe '#landed_plane' do
+    it 'should return an array' do
+      expect(subject.landed_plane).to eq(Array.new)
```
MaxRobertsDear: I'm a little confused by this. perhaps you could use a more descriptive class instead of 'Array'?

PR #1360, opened 2019-04-07
Weekend Challenge - Airport

domw30:
Add contributing file.


domw30:
Add contributing files.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def land(plane)
+    raise 'Plane unable to land, weather is stormy' if stormy?
+    raise 'Plane unable to land, airport full' if full?
+
+    @planes_landed.push(plane)
+  end
+
+  def take_off(plane)
```
wemsteral: In my code, I didn't use an argument in the take_off method because I wanted the take_off method to access the same instance of plane...?

PR #1361, opened 2019-04-07
Airport_challenge solutions

Xa1n:


lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def new_plane
+    Array.new(rand(1..capacity)) { Plane.new }
+  end
+
+  def full
+    @planes.count >= @capacity
+  end
+
+  FORECASTS = [:stormy, :normal, :normal, :normal]
```
guilhe0756: Why is this a constant? Why would it important for this not to ever change?

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+  CAPACITY = 5
```
guilhe0756: Similar question here: why is this a constant?

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  attr_accessor :capacity
```
guilhe0756: I don't understand what an `attr_accessor` does. How is it different from an `attr_reader`? Why would you need for `capacity`?

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  attr_reader :planes
+  attr_reader :weather
+
+  def initialize(capacity = CAPACITY)
+    @capacity = capacity
+    @planes = [].concat new_plane
```
guilhe0756: How is concat different than push?

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+    @weather = Weather.new
+  end
+
+  def takeoff(plane)
+    fail "Plane cannot take off in this weather" if storm
+
+    @planes.delete(plane)
```
guilhe0756: I like the use of delete here. To me, it shows that planes is really an object.

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def storm
+    random_forecast == :stormy
+  end
+
+  private
+
+  def new_plane
+    Array.new(rand(1..capacity)) { Plane.new }
```
guilhe0756: It is not obvious to me why the method new_plane in the singular creates a random number of planes (up to the airport's capacity). I understand that it is setting a random number of initial planes that are at a particular airport, but then the method could have been called "airport's current state" or "number of planes on patio" or something along those lines.

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'airport'
+
+RSpec.describe Airport do
+
+  describe 'initialization' do
+    subject(:airport) { described_class.new }
```
guilhe0756: I like this. It is very clear what you are doing.

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    let(:plane) { double :plane }
```
guilhe0756: I'm not sure what is happening here. Why are you using double and not something similar to what you did on line 6? Aren't you instantiating Plane?

guilhe0756:
Overall, I think your code's best attribute is its readability. Almost everything is clear and one can understand what you are doing. Regarding commits, I'd say you could have had a few more, and the last one kinda shows that: you have a big message for only one commit, i.e. you could have committed for each of those messages. Well done!

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def full
+    @planes.count >= @capacity
```
guilhe0756: Somehow, when I run rspec, this line is not being tested.

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  FORECASTS = [:stormy, :normal, :normal, :normal]
+
+  def random_forecast
+    FORECASTS.sample
```
guilhe0756: Somehow, when I run rspec, this line is not being tested.

lib/airport.rb
Xa1n: Here, because is under private I realise I don't need it to be a constant. Constants are important for preventing yourself or others from changing values that should not be changed, as well as improving readability. In this case, having FORECASTS adjacent to stormy/normal helps visually differentiate between the grouping and the elements. But it is not necessary in private.

lib/airport.rb
Xa1n: Here, I want capacity to not change, and improve readability for user by explaining that CAPACITY is an important object in this piece of code. However, I contradict this point elsewhere where I have used capacity as an instance variable. But I wanted to have an unchanging default capacity (ie never have an airport with a capacity of zero) as well as a way to forcibly change that capacity when needed.

lib/airport.rb
Xa1n: Here, 'attr_accessor' allows me to change the capacity from its initial default value. An example piece of code would be:
" :003 > airport = Airport.new
 => #<Airport:0x00007fce4f189910 @capacity=5, @planes=[#<Plane:0x00007fce4f189848>], @weather=#<Weather:0x00007fce4f189820>>
2.5.0 :004 > airport.capacity = 10
 => 10
2.5.0 :005 > airport
 => #<Airport:0x00007fce4f189910 @capacity=10, @planes=[#<Plane:0x00007fce4f189848>], @weather=#<Weather:0x00007fce4f189820>>
2.5.0 :006 > "

I am still unsure of other ways to do this, other than your way (instance_variable_set)

lib/airport.rb
Xa1n: push would have given me an array of new_planes **inside** the array of @planes. This is similar to flattening (which doesn't work here), in that it takes the elements inside new_plane and concatenates them to @planes

lib/airport.rb
Xa1n: Completely agree here.

spec/airport_spec.rb
Xa1n: I specified the double of the Airport class on line 6 as I was testing under the "describe Airport do" segment, but on line 7 I was testing the Plane instance within the Airport class. I am not sure if the line 6 way would work for plane here. I'll test it.

lib/airport.rb
Xa1n: Need to include one lol

lib/airport.rb
Xa1n: Yep, needed to include one for this too

PR #1364, opened 2019-04-07
Code so far 

jordantroberts:
My airport code so far. I need to:

- Update to defend against edge cases 
- Update plane.rb and plane_spec.rb to show status of the plane 
- Update README 
- Use doubles

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,60 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+  DEFAULT_CAPACITY = 10
+
+   attr_accessor :capacity
+   attr_accessor :weather
+   attr_accessor :planes
+
+
+  def initialize
+    @planes = []
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def land(plane)
+    if full?
+      "Airport full"
+    else
+    @planes << plane
+    'Plane landed'
+    end
+  end
+
+  def delete(*)
+    @planes.pop
+    "Plane is no longer in airport"
+  end
+
+  def empty?
+    @planes.empty?
+  end
+
+  def full?
+    @planes.count == @capacity
+  end
+
+  def control_landing(plane)
+    airport = Airport.new
+    return "It is too stormy to land" if weather == "Stormy"
+    return "Airport full" if full?
+
+    airport.land(plane)
+    end
+
+  def control_takeoff(plane)
+    airport = Airport.new
+    return "It is too stormy to take off" if weather == "Stormy"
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,60 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+  DEFAULT_CAPACITY = 10
+
+   attr_accessor :capacity
+   attr_accessor :weather
```
shektor: Would the `capacity`, `weather` and `planes` all need to be accessed from outside the `Airport` instance?

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+  def land(plane)
+    if full?
+      "Airport full"
+    else
+    @planes << plane
+    'Plane landed'
+    end
+  end
+
+  def delete(*)
```
shektor: What was the reason for using the `*` as the parameter? I am not sure how the method uses it.

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    @planes.pop
```
shektor: As I understand `.pop` will remove the last element in the array. Could this be updated to remove the `plane` object passed as an argument from the calling method `control_takeoff(plane)`

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def control_takeoff(plane)
+    airport = Airport.new
+    return "It is too stormy to take off" if weather == "Stormy"
+    return "The airport is empty" if empty?
+
+    airport.delete(plane)
+  end
+  
+  def weather
```
shektor: Could a `Weather` object be initialized when the `Airport` is created, instead of it creating a new object each time the method is called?

lib/airport.rb
```diff
@@ -0,0 +1,60 @@
...
+    airport = Airport.new
+    return "It is too stormy to land" if weather == "Stormy"
+    return "Airport full" if full?
+
+    airport.land(plane)
+    end
+
+  def control_takeoff(plane)
+    airport = Airport.new
+    return "It is too stormy to take off" if weather == "Stormy"
```
shektor: Its always sunny at your Airport?

spec/weather_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'weather'
+
+describe Weather do
+  let(:weather) { Weather.new }
+
+  it "is stormy outside" do
+    allow(weather).to receive(:forecast) { "Stormy" }
```
shektor: Should there be an expect statement after to check output?

PR #1366, opened 2019-04-07
Weekend airport challenge

Kintaro-Oe:


spec/weather_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'weather'
+
+describe Weather do
+  context "generates random weather" do
+    it "returns 'sunny' when rand is 1" do
+      allow(subject).to receive(:rand) { 1 }
```
CarlyJ88: I'm curious how the the random number generator test is ensuring that the weather is usually sunny. Has that been tested? I like your approach. I'm not sure how rand is working in the test but I can see it working well in your code.

spec/plane_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'plane'
+
+describe Plane do
+  context "when not stormy" do
+    let(:forecast) { "sunny" }
+    let(:airport) { Airport.new }
```
CarlyJ88: I'm not sure what this test is doing. Please explain this to me.

spec/weather_spec.rb
Kintaro-Oe: I separated the weather into its' own class due to its individual behaviour and functionality.
The tests above use 'allow to receive' is a stub used to set the outcome of the random generator so that I could test what it does in that particular situation.
The weather class wouldn't be used in the code unless the user wants to simulate random weather by passing Weather.randomiser as an argument during take off or landing

spec/plane_spec.rb
Kintaro-Oe: Context is a descriptor used to provide clarity to the following test e.g. under "x" conditions expect these test to pass.
I used let as a refactored way to set some variables. This link has more let examples http://www.betterspecs.org/#let

PR #1368, opened 2019-04-07
Partial solution to week 1 airport challenge

NapperJLG:
I got as far as making the Airport and Plane classes, writing methods for planes to take off and land and creating a Weather class that generated a random number.


stuartpet:
Great idea to use an array as the airport, and helped with the 'initialize' method at the beginning.

stuartpet:
Can we discuss the method you had for making weather random, as I want to sense check with my own ideas of how to build this feature?

PR #1370, opened 2019-04-07
Airport challenge update

tsankhalpara:
Please find my solution to this weekend challenge

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,16 +1,26 @@
 require_relative 'planes'
 
 class Airport
-  attr_reader :plane
+  attr_reader :planes
+  attr_reader :capacity
+  DEFAULT_CAPACITY = 20
 
-  def land
-    raise "Airport full" if full?
-    raise "Too stormy to land" if stormy?
-    @plane = 'plane'
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = DEFAULT_CAPACITY
+    @planes = []
   end
 
-  def take_off
+  def land(plane)
+    raise 'Airport full' if full?
+    raise 'Too stormy to land' if stormy?
+    raise 'Plane has already landed' if planes.include(plane)
+
+    planes << plane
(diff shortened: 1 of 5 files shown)


tsankhalpara:
README had not yet been updated

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    it "assumes the airport starts empty" do
+      airport = Airport.new
+      expect(airport.planes).to eq([])
+    end
+  end
+
+  describe '#land' do
+    plane = 'plane'
+    context 'when not stormy' do
+      before do
```
Hannah-Frost: I attempted to use `before do` in my own code, however I couldn't wrap my head around making it work. Would you be able to explain what the syntax should be, please?

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+      it 'instructs the plane to take off' do
+        expect(subject).to respond_to(:take_off)
+      end
+    end
+    context 'when stormy' do
+      before do
+        allow(subject).to receive(:stormy?).and_return true
+      end
+      it "Plane cannot land when stormy" do
```
Hannah-Frost: As the context is already `when stormy` should the it statement only say `"Plane cannot land"` rather than `"Plane cannot land when stormy"` (just to avoid repetition).

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+  def take_off
+    raise 'Too stormy to take off' if stormy?
+
+    plane_taken_off
+    @planes.pop
+  end
+
+private
+
+  def stormy?
```
Hannah-Frost: You could split weather into its own class for more clarification.

spec/airport_spec.rb
tsankhalpara: Great suggestion, it makes the tests much easier to read.

PR #1372, opened 2019-04-07
Week 1 Airport Challenge submission

ayelishaamoah:


README.md
```diff
@@ -61,30 +52,111 @@ So that the software can be used for many different airports
 I would like a default airport capacity that can be overridden as appropriate
 ```
 
+Approach
```
callum-marshall: Really good README.
Thorough and clear.

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'plane'
```
callum-marshall: I don't know 'require relative', only 'require', why did you use it?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  DEFAULT_CAPACITY = 15
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    raise "The plane is unable to land" if stormy?
```
callum-marshall: I don't know 'raise', is it different from 'fail', why did you choose it ?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'plane'
+
+class Airport
+
+  DEFAULT_CAPACITY = 15
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
```
callum-marshall: Could you explain why 'capacity' is initialised? How it it different from not initialising it?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def take_off
+    raise "Planes can't take off in stormy weather" if stormy?
+
+    @planes.pop
+    "The plane has taken off"
+  end
+
+private
+  def stormy?
```
callum-marshall: I don't fully understand how this method works, could you explain it to me please?

spec/airport_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'airport'
+
+describe Airport do
+
+  describe '#land' do
+
+    it 'does not allow landing in stormy weather' do
+      plane = double(:plane)
```
callum-marshall: what is double? what does it do?

spec/airport_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect { subject.land(plane) }.to raise_error "The plane is unable to land"
+    end
+
+    it 'does not allow planes to land if the airport is full' do
+      allow(subject).to receive(:stormy?).and_return(false)
+      15.times { subject.land(double(:plane)) }
+      expect { subject.land(double(:plane)) }.to raise_error "The airport is full"
+    end
+
+    it 'accepts different default capacities' do
```
callum-marshall: what does this test do? how does it do it?

spec/airport_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      allow(subject).to receive(:stormy?).and_return(false)
+      Airport::DEFAULT_CAPACITY.times { subject.land(double(:plane)) }
```
callum-marshall: I've never seen this syntax '::' what does it do? Why did you use it?

callum-marshall:
Very good, regular commits with clear messages.

callum-marshall:
Overall I think your code is really good. It has good readability and usability. Combined with a thorough readme which explains well both what you intended to do and did. 
100% test coverage, functional code and sudocop gives you full marks. Very impressive.

lib/airport.rb
ayelishaamoah: This is so that users can set a default capacity when creating a new instance of the Airport class e.g Airport.new(20) - this will set the capacity as 20.

lib/airport.rb
ayelishaamoah: I thought that require should be used for external files and require_relative is for files in your project directory - but I think both work.

lib/airport.rb
ayelishaamoah: I created an array with possible weather values.
Return a random weather string by using a random number between 0 and 1 that will be used as the index value.
If the string in random weather is "stormy" then it will return true.

spec/airport_spec.rb
ayelishaamoah: doubles allow you to pass in placeholder/ fake objects so that your tests have fewer dependencies on external classes.

spec/airport_spec.rb
ayelishaamoah: I started off using a stub for the stormy weather so that this won't stop the plane from landing.
15.times { subject.land(double(:plane)) } -- the default capacity is 15 so I landed 15 planes at the airport.
This then expects an error to be raised when I try to land a plane -- I've also used a double for the plane object.

spec/airport_spec.rb
ayelishaamoah: I saw this in the Boris bikes challenge - I think this is the only way to access the constant from within the test - although I'm not 100% sure why.

PR #1374, opened 2019-04-07
Airport Challenge Attempt

Hannah-Frost:


lib/airport.rb
```diff
@@ -8,4 +8,10 @@ def initialize
   def land_plane(plane)
     @planes << plane
   end
+
+  def plane_takeoff
+    @planes[0]
```
tsankhalpara: would using planes.pop be more useful to use to help refactor the code

lib/airport.rb
```diff
@@ -14,4 +14,8 @@ def plane_takeoff
     @planes.delete_at(0)
   end
 
+  def empty?
```
tsankhalpara: what was the reason for adding a definition for empty instead of full?

lib/airport.rb
```diff
@@ -1,26 +1,52 @@
...
   end
 
+  private
+
   def empty?
     @planes.empty?
   end
 
+  def full?
+    @planes.count.positive?
```
tsankhalpara: What was the reason for using this instead of comparing it to capacity?

lib/airport.rb
Hannah-Frost: It didn't occur to me to use `.pop` but I did end up using `.delete(plane)` instead. In the future I will use `.pop` when applicable though. :)

lib/airport.rb
Hannah-Frost: Good point! It was initially for clarification but it makes more sense to only have one (also makes it neater). I will change that now.

lib/airport.rb
Hannah-Frost: In my new code this is fixed:
```
  def full?
    @planes.count >= @capacity
  end
```
Thank you for your help!

PR #1376, opened 2019-04-08
Airport Challenge submission

shektor:


lib/airport.rb
```diff
@@ -0,0 +1,30 @@
+require_relative 'weather'
```
alicelieutier: I don't know what `require_relative` does, can you explain why you use it instead of `require`.

alicelieutier (commented):
Overall, I enjoyed your code.

jordantomiko (commented):
I really enjoyed reading your code :) What I thought was particularly good:

- Good, short descriptions in your spec files.
- Thorough testing using stubs.
- Lots of descriptive commits at each step. 
- Very usable code which worked in irb. Enabled me to set up a new airport, change the capacity and run the program.  
- You had zero rubocop errors! 🥇

PR #1377, opened 2019-04-08
Not complete solution

CarlyJ88:


README.md
```diff
@@ -13,78 +13,6 @@ Airport Challenge
 
 ```
 
-Instructions
----------
+For this challenge I will convert 6 user stories into test-driven code. A plane will be able to land in and take off from an airport, as long as the weather is sunny.
```
Kintaro-Oe: README could be expanded on a little e.g. What was the approach to taken? How much was achieved? How is the app used/instructions?

spec/plane_spec.rb
```diff
@@ -0,0 +1,3 @@
+require 'plane'
```
Kintaro-Oe: This file doesn't contain any tests. What does it do?

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
+require 'plane'
```
Kintaro-Oe: This line doesn't call the 'plane' file when feature testing. How could this be edited? Have you considered `require_relative`?

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def initialize(capacity = 1)
+    @capacity = capacity
+  end
+
+  def land_plane(plane)
+    @plane = plane
+  end
+
+  def plane_take_off(_plane)
```
Kintaro-Oe: I don't know what the underscore does in (_plane) could you tell me?

CarlyJ88: Because the variable isn't being used at this point. It is just showing intention. The _ hides the variable from rubocop but I wouldn't recommend it.

lib/airport.rb
CarlyJ88: I hadn't considered this. Thank you, I will look into it :)

spec/plane_spec.rb
CarlyJ88: I had the intention of writing tests in this file but I didn't manage to get to the point where I wrote any. I was mainly using Plane objects in my airport_spec tests.

README.md
CarlyJ88: Thank you, I agree. I will make sure I set aside more time to improve my README.

PR #1378, opened 2019-04-08
First User Story Satisfied

MaxRobertsDear:
Improved timetabling required this week ..

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,48 +1,16 @@
 require_relative 'plane'
 
-class Airport
-  attr_reader :identifier, :capacity, :planes_parked
-
-  # DEF_IDENTIFIER = "DEF"
-  DEF_CAPACITY = 50
-  STORM_PERCENTAGE = 0.80
-
-  def initialize(identifier = "DEF", capacity = DEF_CAPACITY)
-    raise error "Airport identifier must be a string" unless identifier.is_a?(String)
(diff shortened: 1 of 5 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    plane = Plane.new
+    subject.land(plane)
+    expect(subject.plane).to eq plane 
+  end 
+
+  describe '#take_off' do
+    it 'instructs plane to take_off' do
+      #assume that airport starts empty
+      plane = Plane.new
+      subject.land(plane)
```
SarpongAbasimi: I would like to understand why `subject.land(plane)` inside the `#take_off` method.

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+
+class Airport
+  attr_reader :identifier, :capacity, :planes_parked
+
+  # DEF_IDENTIFIER = "DEF"
+  DEF_CAPACITY = 50
+  STORM_PERCENTAGE = 0.80
+
+  def initialize(identifier = "DEF", capacity = DEF_CAPACITY)
+    raise error "Airport identifier must be a string" unless identifier.is_a?(String)
```
SarpongAbasimi: Could you please explain to me what this code is doing 
`raise error "Airport identifier must be a string" unless identifier.is_a?(String)`

PR #1381, opened 2019-04-08
Unfinished. Some tests completed.

ghost:


lib/airport.rb
```diff
@@ -4,6 +4,9 @@ def stormy
         (weather.new) == stormy?
     end    
 
+    def capacity
+    end    
+
```
yadlra: I really like the simplicity of the class and how you broke down the methods.

spec/airport_spec.rb
```diff
@@ -5,7 +5,10 @@
...
     
     describe 'allows the airport to take the weather into consideration'
-        it { is_expected.to respond_to(:stormy) }
+        it { is_expected.to respond_to :stormy  }
+        
+     describe 'can check airport capacity' 
+        it { is_expected.to respond_to :capacity }
 
     
     it 'allows planes to land' do
```
yadlra: I can understand and follow the spec file with the lib file.

lib/airport.rb
```diff
@@ -4,6 +4,9 @@ def stormy
...
     def land
```
yadlra: As we just agreed. While we could both condense everything, it makes sense that we have defined every activity the user wanted with a method.

PR #1383, opened 2019-05-05
Checking if solution passes checks

LaurenceTaylor:


spec/weather_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'weather'
+
+describe Weather do
```
samkitchen94: could there be more tests to assess that weather was picked randomly?

lib/airports.rb
```diff
@@ -0,0 +1,43 @@
+class Airport
+  DEFAULT_CAPACITY = 10
+
+  attr_reader :capacity, :weather
```
samkitchen94: do you need attr_reader :capacity if capacity is within the initialize method as well?

lib/planes.rb
```diff
@@ -0,0 +1,15 @@
+class Plane
+  attr_reader :status
```
samkitchen94: are both attr_reader and initialize necessary?

PR #1384, opened 2019-05-05
Completed Airport Challenge

EdinaBMakers:


README.md
```diff
@@ -1,90 +1,220 @@
...
+To ensure safety
+I want to prevent landing when the airport is full
 
 As the system designer
 So that the software can be used for many different airports
 I would like a default airport capacity that can be overridden as appropriate
 ```
 
+## Approach
```
n3ttl3t: I didn't include this here, is it to remind yourself?

README.md
```diff
@@ -1,90 +1,220 @@
...
+Objects and Messages:
 
-In code review we'll be hoping to see:
+* Airport
+  * land
+* Plane
+
+Test Cases:
+
+* Plane object can be initialised
```
n3ttl3t: This is very specific compared to Objects and Messages, was there a step inbetween?

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    raise "Cannot take off: Plane is not at Airport" unless at_airport?(plane)
+    raise "Cannot take off in stormy weather" if Weather.stormy?
+
+    @planes.delete(plane)
+  end
+
+  private
+
+  def full?
+    @planes.size >= capacity
```
n3ttl3t: I think you can just use ==, because the airport should never go above capacity

README.md
```diff
@@ -1,90 +1,220 @@
...
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+## Usage
+1. Run `irb` in the terminal
+2. require `./lib/airport.rb`
+3. require `./lib/plane.rb`
+4. require `./lib/weather.rb`
```
n3ttl3t: Could you include more instructions than this?

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
+  def self.stormy?
+    rand(1..10) == 10
```
n3ttl3t: I like how concise this is, just returning true of false

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
+
```
n3ttl3t: Could this be given some of the behaviour from airplane.rb?

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    raise "Cannot land: Plane is already at Airport" if at_airport?(plane)
+    raise "Cannot land: Airport is full" if full?
+    raise "Cannot land in stormy weather" if Weather.stormy?
+
+    @planes << plane
```
n3ttl3t: Are there a few things happening here that could be split into different methods to follow the SRP?

PR #1385, opened 2019-05-05
Initial request

n3ttl3t:
Maybe I'll look at the pull result to inform changes and do another. Or maybe I won't, it is Sunday night.


lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+class Airport
+  attr_accessor :planes
+  attr_accessor :capacity
+
+  def initialize(capacity = 1)
```
EdinaBMakers: Maybe you can use a default capacity as a constant

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
...
+    @planes = 0
+    @capacity = capacity
+  end
+
+  def check_capacity
```
EdinaBMakers: Ternary operator (@planes == @capacity ? 'full' : 'not full') can save you a few lines

EdinaBMakers: Why do you need to use 'self' if you have the instance variables?

lib/system_designer.rb
```diff
@@ -0,0 +1,3 @@
+class SystemDesigner
```
EdinaBMakers: When you initialize an airport you have already had a chance to pass a different size of capacity, so I think this class is unnecessary right now.

PR #1388, opened 2019-05-06
First pull request

breycarr:
User story 1 complete = plane is able to land at airport


breycarr:
User story 2 complete

PR #1389, opened 2019-05-06
Airport challenge

dafplant:


lib/airport.rb
```diff
@@ -2,20 +2,33 @@
 require_relative './weather'
 
 class Airport
+  DEFAULT_CAPACITY = 20
```
breycarr: why did you choose to make this the default capacity?

lib/airport.rb
```diff
@@ -0,0 +1,6 @@
+
+class Airport
```
breycarr: Why did you choose to create the Airport class before your tests?

lib/plane.rb
```diff
@@ -0,0 +1,6 @@
+
+class Plane
+  def landed?
```
breycarr: which test led you to create this method?

spec/airport_spec.rb
```diff
@@ -1,9 +1,12 @@
...
 describe Airport do
 
-it 'allows planes to land' do
-  expect(subject).to respond_to :land_plane
-end
+  it 'allows planes to land' do
+    expect(subject).to respond_to :land_plane
+  end
+
+  it { is_expected.to respond_to(:take_off).with(1).argument }
```
breycarr: does this not repeat the test "expect(subject).to respond_to :land_plane"?
Could you have expanded the original test, rather than write an entirely new one?

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+require_relative './weather'
+
+class Airport
+  DEFAULT_CAPACITY = 20
+
+  attr_reader :plane, :weather, :stormy
+
+  def initialize
+    @planes = []
+    @weather = weather
```
breycarr: Why does the weather attribute does not appear to be used after initialisation?

spec/airport_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      expect(subject).to respond_to(:land_plane).with(1).argument
+    end
+
+    it 'allows planes to take off' do
+      expect(subject).to respond_to(:take_off).with(1).argument
+    end
+
+    it 'gives error when airport is full' do
+      airport = Airport.new
+      plane = Plane.new
```
breycarr: did you need to intialise ```plane``` as a new instance of Plane, as you use ```Plane.new``` in all arguments?

This is also flagged up by rubocop:
```spec/airport_spec.rb:17:7: W: Lint/UselessAssignment: Useless assignment to variable - airport.```
```spec/airport_spec.rb:18:7: W: Lint/UselessAssignment: Useless assignment to variable - plane.```

File: spec/airport_spec.rb [hunk 7]
```diff
@@ -0,0 +1,33 @@
...
+      Airport::DEFAULT_CAPACITY.times { subject.land_plane(Plane.new) }
+      expect { subject.land_plane(Plane.new) }.to raise_error "airport is full"
+    end
+
+  end
+
+  describe '#weather' do
```
breycarr: Why are these methods not within weather_spec.rb?

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
+
+  def random_weather
+    rand(1..10)
```
breycarr: why did you choose this range for your random number generation?

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
+  def stormy
+    if random_weather >= 8
```
breycarr: why did you choose this set to create random weather?

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
...
+      "stormy"
+    else
+      "sunny"
+    end
```
breycarr: could you shorten this if statement to one line?

spec/weather_spec.rb
```diff
@@ -0,0 +1,9 @@
+require './lib/weather.rb'
+
+describe Weather do
+
+  it 'should test for stormy weather' do
+    expect(subject).to respond_to(:stormy)
```
breycarr: could you use another word for this method name, eg "conditions", or make it a status check, eg "stormy?"

spec/airport_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    it 'allows planes to take off' do
+      expect(subject).to respond_to(:take_off).with(1).argument
+    end
+
+    it 'gives error when airport is full' do
+      airport = Airport.new
+      plane = Plane.new
+      Airport::DEFAULT_CAPACITY.times { subject.land_plane(Plane.new) }
+      expect { subject.land_plane(Plane.new) }.to raise_error "airport is full"
```
breycarr: This fails when rspec is run, error message: "expected Exception with "airport is full", got #<RuntimeError: Airport is full> with backtrace:"

PR #1390, opened 2019-05-06
LM airport challenge

LisaMcCormack:


spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'airport'
+
+describe Airport do
+
+  it 'will instruct a plane to land if sunny' do
```
rayswilling: With the stud examples online, I read quite a bit on adding the 'context' block. Was this something you read about but skipped - i.e. does the test work and print out the logical break down of each test without the context block?

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    allow(subject.weather).to receive(:stormy?).and_return(false)
+    expect(subject.land(Plane.new)).to eq('Cleared to land')
```
rayswilling: to eq 'string' is a valid test?

PR #1393, opened 2019-05-06
Magnus' submission. airport_challenge

marbuthnott:


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def take_off
+    raise 'Cannot take off due to stormy weather' if weather.stormy?
+    @plane_list.pop
+  end
+
+  def full?
+    @plane_list.length >= @capacity
+  end
```
VladC24: The full? method should be private.

spec/plane_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'plane'
+
+describe Plane do
+  it 'creates new plane object' do
+    expect(Plane.new.is_a? Plane).to eq true
+  end
+
```
VladC24: describe Plane do
let(:plane) { Plane.new }
end
Would be smoother maybe.

spec/weather_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'weather'
+
+describe Weather do
+
+  it 'is expected to create new weather object' do
+    expect(subject.is_a? Weather).to eq true
+  end
+
```
VladC24: let(:weather) { Weather.new }

PR #1396, opened 2019-05-06
adding spec and lib file to repo for weekend airport challenge

Carlton7:


spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'airport'
+require 'weather'
+
+describe Airport do
+
+    describe '#land_plane' do
+        it 'should land planes at airport' do
+            weather = instance_double("Weather", :state => "sunny")
+            plane = Plane.new
```
marbuthnott: Using `let (:plane) { double(:plane) }` immediately after `describe Airport do` would negate the use `plane = Plane.new` for all other test cases.

PR #1398, opened 2019-05-06
Ray Swilling - 1st pull request

rayswilling:


lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @planes = []
+    @capacity = capacity
+    @weather = rand(11)
+  end
+
+  def full?
+    @planes.length > @capacity
+  end
+
+  def stormy?
```
LisaMcCormack: I think weather should have been a separate object because it is not strictly part of the airport class.

spec/plane_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        allow(airport).to receive(:stormy?). and_return false
+      end
+      it "instructs plane to land" do
+        expect(airport).to respond_to(:land).with(1).argument
+      end
+
+      context 'when full' do
+        it "raises an error when airport is full" do
+          airport = Airport.new(20)
+          if airport.full?
```
LisaMcCormack: using if - this is good standard practice for rspec?

spec/plane_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'plane'
+require 'airport'
+
+describe Airport do
```
LisaMcCormack: What does this do?

spec/plane_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  subject(:airport) { described_class.new(20) }
+  let(:plane) { double :plane }
+
+  describe '#land' do
+    context 'when not stormy' do
+      before do
+        allow(airport).to receive(:stormy?). and_return false
+      end
+      it "instructs plane to land" do
+        expect(airport).to respond_to(:land).with(1).argument
```
LisaMcCormack: could be wrong but rather than saying airport should respond to land with one argument you should say what actually happens, ie it will equal the plane array?

PR #1399, opened 2019-05-07
Pull request

olliebrownlow:


lib/weather.rb
```diff
@@ -1,6 +1,6 @@
 class Weather
-  WEATHER_PATTERNS = ["stormy", "sunny", "sunny",
-     "sunny", "sunny", "sunny", "sunny"
+  WEATHER_PATTERNS = [:stormy, :sunny, :sunny,
```
yem82: Weather patterns are either stormy or sunny here. Could you find a way of setting the weather pattens using DRY?

PR #1401, opened 2019-05-07
First Pull Request

yem82:


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require_relative 'plane'
+
+class Airport
+  DEFAULT_CAPACITY = 10
+
+  attr_accessor :planes, :capacity, :weather
```
olliebrownlow: Is there any need for this to be attr_accessor? Would attr_reader have done?

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
+require "airport"
+
+describe Airport do
+  plane = Plane.new
+
+  it { is_expected.to respond_to(:planes) }
+  it { is_expected.to respond_to(:take_off) }
+  it { is_expected.to respond_to(:land).with(1).argument }
+
```
olliebrownlow: Are these vacuous tests? Could be deleted.. If these methods are tested elsewhere then we get for free that Airport instances respond to them.

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      allow(subject.land(plane)).to receive(:weather) { 20 }
+      expect { subject.land(plane) }.to raise_error "No landing during storm"
+    end
+  end
+
+  describe "#take_off" do
+    it "a plane can take off" do
+      plane = subject.take_off
+    end
+
```
olliebrownlow: There is no expect statement in this test!

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    it "confirm plane has taken off" do
+      plane = Plane.new
+      subject.land(plane)
+      subject.take_off
+      expect(subject.planes.count).to eq(0)
+    end
+
```
olliebrownlow: This test works because there is only one plane in the airport. If you land 2 planes and then one takes off the test might fail as we'd expect the planes array to equal 1, not 0.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
```
olliebrownlow: Is there any need for this to be attr_accessor? Would attr_reader have done?

spec/airport_spec.rb
olliebrownlow: Are these vacuous tests? Could be deleted.. If these methods are tested elsewhere then we get for free that Airport instances respond to them.

PR #1403, opened 2019-05-07
Challenge partially complete. Weather generator missing

Cosmin-Croitoriu:


spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  describe "#take_off" do
+    it "allows a plane to take off" do
+      luton = Airport.new
+      plane = Plane.new
+      luton.land(false, plane)
+      expect(luton.take_off(false, plane)).to eq('You can now take off!')
+    end
+
+    it "doesn't allow the planes to take off due the weather" do
+      luton = Airport.new
```
hughbric: You call 'luton' and 'plane' for each test. There must be a way to DRY out the code and only assign these once at the top.

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'plane'
+require 'airport'
+describe Airport do
+  describe "#land" do
+    it "instructs plane to land" do
+      luton = Airport.new
+      plane = Plane.new
+      expect(luton.land(false, plane)).to eq('You have permission to land!')
```
hughbric: Passing a boolean to '#land' isn't very readable. What is false doing; do you need to pass it as an argument; can you not use a helper method within the airport class?

spec/plane_spec.rb
```diff
@@ -0,0 +1,4 @@
+require "airport"
```
hughbric: If you're not using either of these requires in the file, I would delete them.

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
+class Airport
```
hughbric: If you use attr_reader's you won't need to put an '@' before each instance variable.

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  def initialize(capacity = 50)
+    @planes = []
+    @weather
```
hughbric: @weather is not assigned to anything. If you create a new class for weather, you can acquire it from there.

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    @capacity = capacity
+    @number_of_planes = 0
+  end
+
+  def land(forecast, plane)
+    @weather = forecast
+
```
hughbric: You guard conditions I would put at the top of each method.

Run Rubocop and see what it says.

PR #1405, opened 2019-05-17
submission of airport challenge

LaurenOM:


lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+class Weather
+
+  def weather_number
+    rand(1..6)
+  end
```
lindseydew: Small point, but could have a method here which describes stormy to be when the random number is 1. eg:
``` 
   def stormy? 
        weather_number == 1
  end
```
So that the client can decide whether it's stormy by saying:
```
   weather = Weather.new
   if weather.stormy?
      ....
   end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'airport'
+
+describe Airport do 
+  let(:airport) { Airport.new }
```
lindseydew: Lovely use of doubles

PR #1406, opened 2019-05-17
Friday airport challenge

AnnaWeakclaw:


spec/airport_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+  it "knows that a plane has landed" do
+    allow(fake_plane).to receive(:land) { true }
+
+    expect { airport.hello(fake_plane) }.not_to raise_error
+    expect(airport.base).to include(fake_plane)
+  end
+
+  it "only registers a plan if it has landed" do
+    allow(fake_plane).to receive(:land) { false }
+    expect { airport.hello(fake_plane) }.to raise_error(RuntimeError)
```
lindseydew: Could be nice to assert on the message of the error as well as the type so that the test can be specific about which error it expects

spec/plane_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  it "cannot land if the weather is stormy" do
+    allow(fake_weather).to receive(:stormy) { true }
+    expect(plane.land(fake_weather)).to be false
+  end
+
+  it "cannot land if the TRUE weather is stormy" do
+    srand(5) # stormy is true => chosen_state = 4
+    weather = Weather.new
```
lindseydew: nice use of a seed value!

PR #1407, opened 2019-05-17
Airport challenge submission

JaydaKad:


spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  end
+
+  it 'can report the weather' do
+    expect(Weather.new).to respond_to(:weather_generator)
+  end
+
+  it 'can report weather as stormy' do
+    weather = Weather.new
+    allow(weather).to receive(:weather_generator).and_return(1)
```
lindseydew: Small point, but the tests relating to weather could be in a separate file.

lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+class Weather
+
+  def weather_generator
+    @outlook = rand(10)
+  end
+
```
lindseydew: Small point, but weather could own a method that returns a boolean to show the state of the weather.
Eg:
```
   def stormy? 
     @outlook == 1 
  end
```
then clients can use this as so:
```
    weather = Weather.new
    if weather.stormy? 
      ...
    end
```

PR #1410, opened 2019-05-18
Challenge complete

peterjcole:
See readme.MD for info

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,14 @@ Airport Challenge
 =================
 
 ```
-        ______
+                  __   _              __   _ 
+                _(  )_( )_          _(  )_( )_
+               (_   _    _)        (_   _    _)
+              / /(_) (__)         / /(_) (__) 
+             / / / / / /         / / / / / /
+            / / / / / /         / / / / / /
+
+        ______                          
         _\____\___
 =  = ==(____MA____)
           \_____\___________________,-~~~~~~~`-.._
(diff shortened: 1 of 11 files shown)


spec/feature/airport_planes_weather_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    expect(air_force_1.land(lsx_airport)).to eq("Plane landed at #{lsx_airport.name} (#{lsx_airport.code})")
+    expect(lsx_airport.planes).to include(air_force_1)
+    expect(lsx_airport.planes).to include(plane)
+    expect(plane.land(lsx_airport)).to eq(already_at_airport_message)
+    expect(plane.take_off(dhs_airport)).to eq(not_at_airport_message)
+    expect(plane.take_off(lsx_airport)).to eq("Plane took off from #{lsx_airport.name} (#{lsx_airport.code})")
+    expect(lsx_airport.planes).not_to include(plane)
+    expect(lsx_airport.contains?(plane)).to be false
+    expect(plane.take_off(lsx_airport)).to eq(already_in_air_message)
+  end
```
lindseydew: It can be good practise to aim to have one assertion statement per unit test, so that if one piece of behaviour change this can be easily tracked down

peterjcole: > It can be good practise to aim to have one assertion statement per unit test, so that if one piece of behaviour change this can be easily tracked down

Hi Lindsey,

The idea of that test was to be the bonus rspec feature test, but I wasn't too sure the best way to do it. 

Would you suggest having each line of the feature test in a separate 'if' - and have one instance of each class which runs through the whole test suite, rather than setting up a new class for each test like the normal unit tests?

lindseydew: Perhaps we can chat about this face to face - would you like to touch base tomorrow?

peterjcole: > Perhaps we can chat about this face to face - would you like to touch base tomorrow?

Sure! I'll try and catch you at some point tomorrow 🙂

README.md
```diff
@@ -2,7 +2,14 @@ Airport Challenge
 =================
```
samjones1001: README looks good. One thing you might want to add is some brief instructions on how a user could run the application.

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
+class Airport
+  CAPACITY_ERROR_MESSAGE = "Plane landed on top of another plane, everyone died"
+  attr_reader :name, :code, :planes, :capacity
+
+  def initialize(weather, name, code, capacity = 5)
+    @weather = weather
```
samjones1001: It's good that you're not initialising an instance of Weather within your initialize method here - that reduces coupling between your objects and allows you to properly isolate your unit tests.

A small usability improvement would be to make use of dependency injection and provide the weather class as a default parameter - this would prevent the user from having to manually create the Weather object.

Something like: 
```ruby
def initialize(name, code, weather = Weather, capacity = 5)
  @name = name
  @code = code
  @weather = weather
  @planes = planes
  @capacity = capacity
end
```

This implementation gives you you new problem of having to provide some argument for weather if you want to override the default capacity. There is a way around this that you might want to investigate 🙂

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def cleared_for_landing?
+    return :weather unless weather_ok?
+    return :capacity unless capacity_ok?
+
+    return true
+  end
+
+  def contains?(plane)
```
samjones1001: I feel like ruby has a built in array method that would do this for you 🤔

lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+    @planes.any? { |plane_in_airport| plane_in_airport == plane }
+  end
+
+  def weather_ok?
```
samjones1001: I really like these little extracted helper methods as I think they improve readability in the places that they're used.

That said, I feel like they should probably be private. As a general rule, unless you explicitly intend for a method to be called upon directly either by a user or by another object you should keep it private.

lib/plane.rb
```diff
@@ -0,0 +1,51 @@
...
+  NOT_AT_AIRPORT_MESSAGE = "I'm not at that airport!"
+  ALREADY_AT_AIRPORT_MESSAGE = "I'm already at an airport!"
+  ALREADY_IN_AIR_MESSAGE = "I'm already in the air!!"
+  CAPACITY_MESSAGE = "I can't land there, the airport is full!"
+
+  def initialize(state = :flying)
+    @state = state
+  end
+
+  def land(airport)
```
samjones1001: It feel like there are quite a lot of messages being passed back and forth between your objects when a plane lands or takes off.

- The plane sends a call to the airport to check if it's clear to land/takeoff
- The airport sends back some confirmation
- The plane sends another call to the airport in order to actually land/takeoff
- The plane sends some more calls to the airport in order to retrieve the name code

This isn't necessarily a problem, but the more back and forth we have between objects the more potential points of failure we're faced with - I wonder if the process could be simplified a little?

samjones1001: It's also worth noting that the implementation for the `land` and `takeoff` methods is very similar. So is the implementation for `landing_proceedure` and `takeoff_proceedure` methods.  Again, I wonder if there's some possible simplification here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,106 @@
+require 'airport'
+
+describe 'airport' do
+  let(:lsx_airport) { Airport.new(good_weather, 'Los Santos International', 'LSX') }
```
samjones1001: There's a lot of initial setup happening at the top of the file here. None of what you're doing seems unreasonable, but you might be able to reduce it a touch by using contexts for stormy and clear weather.

```ruby
let(:lsx_airport) { Airport.new(weather, , 'Los Santos International', 'LSX') }
let(:weather} .  { double(:weather) }

context 'When weather is clear' do
  before(:each) do
    allow(weather).to receive(:stormy?).and_return(false)
  end
  
  #all tests for clear weather go here.
end 
```

It's a pretty small change, so I'll let you decide whether or not it's worth it 🙂

spec/feature/airport_planes_weather_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'airport'
+require 'plane'
+require 'weather'
+
+describe 'feature test' do
+  let(:weather) { double(:weather) }
```
samjones1001: I notice you're using a double in your feature tests here - presumably to overcome the problem of randomness. The problem with this is that feature tests are intended to test that all of the objects in your system actually work together as intended - if you're using doubles you're not actually doing this. Your tests could still pass even if the weather class was completely broken.

I think a preferable way of achieving what you need here would be to use `srand` - that way you can control for the weather you need while still using real instances of all your classes and the behaviour they encapsulate.

PR #1411, opened 2019-05-19
Ed's Airport Challenge submission

edwardreeve:


lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @hangar = []
+    @capacity = capacity
+  end
+
+  def land(plane, weather = Weather.new)
+    raise 'Airport currently closed due to stormy weather' if weather.stormy?
+    raise 'Airport is full' if full?
+    raise 'Plane is grounded elsewhere' if plane.grounded == true
+    
+    plane.grounded = true
```
samjones1001: I think we covered this last week, but generally reaching directly into the state of another object in this way should be avoided.

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
+require_relative '../lib/weather'
+
+class Airport
```
samjones1001: This class looks pretty solid to me - methods are concise and well named, and it's clear what each part of the code is intended to do.

One (very minor) improvement might be to remove the necessity to create a brand new Weather object every time a plane lands or takes off by setting it up in the initialize method:

```ruby
def iniitialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
  @hangar = []
  @capacity = capacity
  @weather = weather 
end
```

assuming you set up an attr_reader for weather you could still make calls to weather.stormy? where required by your `land` and `take_off` methods.

samjones1001: The downside of this approach - which may or may not be obvious - is that by having two default arguments passed into your method you create a situation where if you ever want to override the second argument, you also have to provide a value for the first. For example, a call to

```ruby 
Weather.new(SomeOtherWeatherObject.new)
```

would be interpreted by ruby as meaning that you want the value of capacity to be the passed weather object, and the weather variable to still be assigned the default argument provided in your method declaration.

There is a way around this, but I'll leave it up to you to decide if you'd like to investigate further 🙂

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
```
samjones1001: I think Weather is a very sensible extraction, but as a rule, I see a class as a blueprint for an object containing both state (instance variables) and behaviour (methods)

Where one or the other of these two things aren't required, you might want to think about using some other structure - in ruby a module might be a good choice.

spec/airport_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'airport'
+
+describe Airport do
+  let(:airport) { Airport.new }
+  let(:plane) { instance_double("Plane", :grounded= => true, :grounded => false) }
```
samjones1001: Your double here seems to have two contradictory return values for the same method call - does this work? Presumably the second value overrides the first?

spec/plane_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'plane'
+
+describe Plane do
```
samjones1001: As I think you've already identified, the tests in this file don't really do anything. Again, this is a consequence of the way you've reached into the `Plane` class' internal state from outside - this makes it very difficult to test effectively.

spec/weather_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'weather'
+
+describe Weather do
+  it { should respond_to(:stormy?) }
```
samjones1001: This is a good starting point, but I feel like the class could benefit from additional testing. 

It could be something very simple using srand to force a particular result and ensuring that it returns true and false in the correct situations.

README.md
```diff
@@ -1,90 +1,76 @@
...
 ```
-        ______
-        _\____\___
-=  = ==(____MA____)
+                               ._
```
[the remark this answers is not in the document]
edwardreeve: oh Sam, wait until you see my takeaway challenge

lib/weather.rb
edwardreeve: Modules were referenced in the Code Review Rubric for takeaway challenge, but I don't think I've come across them yet. Is there an exercise/challenge which introduced these that I can revisit, or shall I just have a Google...?

spec/airport_spec.rb
edwardreeve: This is the attr_accessor thing of writing (grounded=) and reading (grounded) being two separate methods

samjones1001: Ah yes - I missed the '=' !

PR #1413, opened 2019-05-19
Lin airport challenge submission

LinTrieu:
See README.md for notes on challenge.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,7 +2,7 @@ require_relative 'weather'
 
 class Airport
 
-  attr_reader :planes, :weather
+  attr_reader :planes, :weather, :capacity
 
   DEFAULT_CAPACITY = 20
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,18 +1,24 @@
+require_relative 'weather'
+
 class Airport
 
-  attr_reader :planes
+  attr_reader :planes, :weather
 
-  def initialize
+  def initialize(weather: Weather.new)
     @planes = []
+    @weather = weather
   end
 
   def land(plane)
     @planes << plane
+    self
   end
 
   def plane_takeoff
     raise "There are no planes in the airport" if @planes.empty?
+    raise "Plane cannot take-off from airport due to stormy weather" if weather.stormy == true
     @planes.pop
+    self
   end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -14,7 +14,7 @@ class Airport
 
   def land(plane)
     fail 'Airport at full capacity' if full?
-    fail "Plane cannot land due to stormy weather" if weather.stormy == true
+    fail "Plane cannot land due to stormy weather" if weather.stormy == true #=> as 'weather' is in attr_reader, we do not need '@', otherwise we would need to @weather!!
     @planes << plane
     self
   end
@@ -31,6 +31,6 @@ class Airport
   end
 
   def empty?
-    @planes.count.empty?
+    @planes.empty?
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -4,9 +4,9 @@ class Airport
 
   attr_reader :planes, :weather
 
-  # DEFAULT_CAPACITY = 20
+  DEFAULT_CAPACITY = 20
 
-  def initialize(weather: Weather.new, capacity=20)
+  def initialize(weather=Weather.new, capacity=DEFAULT_CAPACITY)
     @weather = weather
     @capacity = capacity
     @planes = []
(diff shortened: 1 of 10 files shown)


LinTrieu:
See README.md for notes.

spec/plane_spec.rb
```diff
@@ -0,0 +1,11 @@
+require './lib/airport'
+require './lib/plane'
+
+describe Plane do
+
+  let(:airport) {Airport.new}
```
lindseydew: Small point, but I don't think airport is needed for this file

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    allow(weather).to receive(:stormy) {false}
+    airport.land(plane) # => num of planes  = 1
+    output = airport.planes.length
+    expect(output).to eq(1)
+  end
+
+  it 'can add planes to the planes array total' do
+    allow(weather).to receive(:stormy) {false}
+    5.times {airport.land(plane)}
+    output = airport.planes.length
```
lindseydew: Nice test design

lib/weather.rb
```diff
@@ -0,0 +1,23 @@
...
+  
+  def random_outlook
+    @outlook = [:bad, :sunny, :sunny, :sunny].sample # => when 'bad' outlook, planes cannot fly
+  end
+  
+  def stormy
+    random_outlook
+    # puts @outlook 
+    @outlook == :bad
+  end
```
lindseydew: Small point: a convention in ruby for methods that return true/false is to use a question mark at the end. Eg 
``` def stormy?
       end
```

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'weather'
+
+class Airport
+
+  attr_reader :planes, :weather, :capacity
+
+  DEFAULT_CAPACITY = 20
+
+  def initialize(weather=Weather.new, capacity=DEFAULT_CAPACITY)
```
samjones1001: Good use of dependency injection here, although passing two arguments with default values can lead to a problem.

If you wanted to create an instance of the airport class with the default weather object but an overriden capacity of 20, you might want to do something like:

```ruby 
airport = Airport.new(10)
```

However, because of the order of the arguments this would be interpreted by ruby as you setting up an instance where the weather argument should be equal to 10, and the capacity should retain the default of 20.

If you want to know about how you can get around this problem, take a look at a feature called keyword arguments.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def plane_takeoff
+    fail "There are no planes in the airport" if empty?
+    fail "Plane cannot take-off due to stormy weather" if weather.stormy == true
+    @planes.pop
+    self
+  end
+
+  def full?
```
samjones1001: I really like these little extracted methods - I think they improve readability in the places you're using them - but I think they should probably be made private.

As a general rule, unless you explicitly expect a method to be called either by a user or by another object, you should always make them private.

lib/plane.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
```
samjones1001: In order to be a valid class, I would normally expect to see both state (instance variables) and behaviour (methods). If one or the other of these is missing it can be a good sign that your system doesn't need a particular object.

lib/weather.rb
```diff
@@ -0,0 +1,23 @@
+class Weather
+
+  attr_reader :stormy
+
+  def initialize  #=> synonym for Class.new. Allows you to pass variables into a class method.
+   # weather = Weather.new #=> you would NOT create a new instance of the same class within the blueprint
+    @outlook #=> access only within class (if + in attr_reader: give access outside of class)
+  end
+  
+  def random_outlook
```
samjones1001: it looks as though this method only ever gets called from within your `stormy` method - this means it should probably be private.

lib/weather.rb
```diff
@@ -0,0 +1,23 @@
+class Weather
```
samjones1001: I don't see any tests for this class 😱

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  let(:weather) {double :weather}
+
+  it 'can respond to land' do
+    expect(airport).to respond_to(:land)
+  end
+
+  # 1. discussed with Lindsay on Mon 20 May for correct test (vs. incorrect @ the 2nd) 
+  it 'expects planes can land at the airport' do
+    airport_two = Airport.new(weather) # => declare a new instance of the Airport class, and insert double of weather as parameter
+    allow(weather).to receive(:stormy) {false} # => stub to ensure randomness
```
samjones1001: I notice that this line is repeated across several of your tests in order to cope with the randomness of your weather object.

Rather than duplicating lines, you could extract this out into the setup of your double

```ruby
let(:clear_weather)       { double(:weather, :stormy => false) }
let(:stormy_weather)    { double(:weather, :stormy => true) }
```

PR #1423, opened 2019-06-01
Lucy Hall Airport_Challenge

LucyMHall:
Here is my completed challenge.


spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
+require "airport"
+require "plane"
+require "weather"
+
+describe Airport do
+
+  before(:each) do
```
ruthmoog: You could use `let` as an alternative to `before :each` check out if that way you can reduce the upfront cost of creating each instance variable before every test?

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    expect(@airport).to respond_to(:take_off)
+  end
+
+  it "no longer contains a plane once it has taken off" do
+    @airport.land(@plane, @weather)
+    @airport.take_off(@plane, @weather)
+    expect(@airport.contents).not_to include(@plane)
+  end
+
+  it "confirms take off has occurred" do
```
ruthmoog: I like that you included this in your interpretation of the user stories - good idea!

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  it "has a default airport capacity" do
+    expect(@airport.capacity).to eq(5)
+  end
+
+  it "has a capacity which can be overridden" do
+    @airport.capacity = 3
+    expect(@airport.capacity).to eq(3)
+  end
+
+  it "responds true to full? when reached capacity" do
```
ruthmoog: This is perfectly readable, but you could also consider using context to group your test into chucks, eg
context airport full
test: plane can't land
test: everyone is squashed etc

context the rock is visiting the airport
test: long queue for selfies

There's more info on using context in the better specs webpage 👍

PR #1424, opened 2019-06-02
Alexis J weekend Airport challenge

ajosephides:


lib/airport.rb
```diff
@@ -0,0 +1,63 @@
+require_relative './weather.rb'
+
+class Airport
+
+  attr_reader :hanger
+
+  def initialize(capacity = 10)
```
alicelieutier: What does this "capacity'" do?

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+    @capacity = capacity
+    @hanger = []
```
alicelieutier: Why did you choose an array for the hangar

spec/weather_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'weather'
+
+describe Weather do
+  
+  it 'generate weather that is either sunny or stormy' do
```
bengscott2: Could use srand here to actually test random.

PR #1426, opened 2019-06-02
partial solution

chris-groves:


spec/airport_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'airport.rb'
+describe Airport do
+  it { is_expected.to respond_to :land }
```
ruthmoog: Nice one line test syntax

PR #1430, opened 2019-06-02
Celia From partial solution airport-challenge 

ghost:


Syed-A-Shah (commented):
Really simple and clean code.

ruthmoog (commented):
I like that you've used the `Plane` class to implement logic, as this was a different approach to mine.  What made you choose to use the plane over the airport?

I ran rubocop and got a couple of fails - have a look and see what you'd change.

Thanks for letting me take a look at your code :D Hope this helps 👍

PR #1431, opened 2019-06-02
Challenge completed and ready for review

aniasobo:


README.md
```diff
@@ -1,6 +1,22 @@
 Airport Challenge
 =================
 
+The program is divided into 3 classes: Airport, Plane and WeatherReport.
```
ajmpawlik: Great that you took time for README and wrote an extensive explanation on how your code works.

spec/plane_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+describe Plane do
+  it 'is in air by default' do
+    plane = Plane.new
+    expect(plane.status_report).to eq "in air"
+  end
+  it 'changes status when landed' do
+    plane = Plane.new
+    airport = Airport.new(1)
+    allow(airport).to receive(:permission_to_land_granted?) { true }
```
ajmpawlik: Interesting choice of test with "allow", learning from you here!

spec/weather_report_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'spec_helper'
+require 'weather_report'
+
+describe WeatherReport do
+  it 'can only be one of 10 options' do
+    weather = WeatherReport.new
+    expect(weather.forecast).to match(/clear|sunny|partly cloudy|light rain|light fog|overcast|showers|stormy|stormy|stormy/)
```
ajmpawlik: Is here .to match similar to .to eq? What is the meaning of | ?

lib/weather_report.rb
```diff
@@ -0,0 +1,17 @@
+class WeatherReport
+
+  attr_accessor :tower_call
+  
+  WEATHER = ["clear", "sunny", "partly cloudy", "light rain", "light fog", 
+            "overcast", "showers", "stormy", "stormy", "stormy"]
+
+  def initialize
```
ajmpawlik: Nice method!

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  def tower_permission_to_takeoff(plane)
+    if plane.status_report == "grounded" && permission_to_takeoff_granted?
+      @ground_control.delete[plane]
+      plane.update
+    else
+      p "Request denied"
+    end
+  end
+
+  private
```
ajmpawlik: What does it mean private here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'spec_helper'
+require 'airport'
+require 'weather_report'
+require 'plane'
+
+describe Airport do
+  context 'airport is not full and weather is not stormy' do
+    before(:each) do
```
ajmpawlik: What is the meaning of each here?

spec/plane_spec.rb
aniasobo: I'm not 100% sure that it worked in the test though

spec/weather_report_spec.rb
aniasobo: it's a regular expression, the | is an or so it matches the pattern in the weather report.

lib/airport.rb
aniasobo: private methods that return booleans to the public methods above.

spec/airport_spec.rb
aniasobo: Each test - each it block - so that the objects are all the same within the context block.

PR #1433, opened 2019-06-02
Weekend airport_challenge 

monikakaczan:


lib/airport.rb
```diff
@@ -0,0 +1,28 @@
+require_relative './plane.rb'
+require_relative './weather.rb'
+
+
+class AirportController #RZECZOWNIK
+
+  attr_reader :planes
+
+  def initialize #TWORZY SAMOLOTY
+    @planes = []
```
leonlevitate: Good that you've initialised an array!

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
+#As an air traffic controller
```
leonlevitate: Great that you've added comments to reference the user story

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+class Weather
+ attr_reader :weather
+
+ def initialize
+   @weather = ["stormy", "sunny"]
```
leonlevitate: Good that you've used sample method. I think the weather needed to be selected via a random number generator..

Gemfile.lock
```diff
@@ -53,9 +53,9 @@ PLATFORMS
 DEPENDENCIES
```
leonlevitate: Did you update the README file to briefly explain your approach to solving the challenge?

PR #1434, opened 2019-06-02
Rachel Griffiths Airport Challenge

rachjgriff:


feature_test.rb
```diff
@@ -0,0 +1,14 @@
...
+puts
+puts airport.capacity = 10
+puts
+print 5.times { airport.land(Plane.new) }
+puts
+print 6.times { airport.land(Plane.new) }
+puts
+print 10.times { airport.land(Plane.new) }
+puts
+print 11.times { airport.land(Plane.new) }
```
aniasobo: I can't tell what this test is testing for. Is it just to see the landed planes printed out? Is it missing an expect condition? What are all the putses putsing?

feature_test.rb
```diff
@@ -0,0 +1,14 @@
+# feature_test
```
lucianmot: Excellent ! you started debugging it! :+1:

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  attr_reader :capacity
+
+  DEFAULT_CAPACITY = 5
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
```
aniasobo: I like those methods, very succinct!

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @planes.pop
+  end
+
+  private
+
+  def full?
+    @planes.count >= @capacity
+  end
+
+  def empty?
```
lucianmot: Exhibit A1

spec/airport_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  it 'plane lands at the airport' do
+    plane = Plane.new
+    expect(subject.land(plane)).to eq [plane]
+  end
+
+  it 'plane is no longer at the airport' do
+    plane = Plane.new
+    subject.land(plane)
+    expect(subject.planes).to eq [plane]
```
aniasobo: Does this test pass like this, with .planes called like a method? I can't tell.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail 'Airport full' if full?
+
+    @planes << plane
+  end
+
+  def take_off
```
lucianmot: Is this suppose to raise an error a plane tries to take off but there aren't any planes in the airport? (ref Exhibit A1 line 33)

spec/airport_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    it 'raises an error when the airport is empty' do
+      expect { subject.take_off }.to raise_error 'Airport empty'
+    end
+  end
+
+  describe '#land' do
+
+    it 'raises an error when the airport is full' do
+      subject.capacity.times { subject.land(Plane.new) }
```
aniasobo: As far as I can tell, this tests the default capacity again? Maybe try adding a double with a different capacity to test it's set correctly with initialise?

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
```
lucianmot: I'm not sure if this is good practice, but I did this in mine, since class Plane is empty and does nothing at all, I mean it is probably gonna do something in a more complex exercise. But I think in this scenario you just can just add it at the end of lib/airport.rb and then you won't require require 'plane' in spec/airport_spec.rb

spec/plane_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'plane'
```
lucianmot: if you want to condense code you could copy all code in spec/airport_spec.rb and give you this entire file.

README.md
```diff
@@ -1,3 +1,25 @@
+## README: Rachel Griffiths
```
aniasobo: This is a great readme, I wish I'd remembered to keep track of my approach like you did! Good ideas for follow up too.

PR #1435, opened 2019-06-02
Completes Airport Challenge

rhysco8:


lib/airport.rb
```diff
@@ -0,0 +1,54 @@
+require './lib/plane.rb'
```
Danielandro: You can require files in the same directory using require_relative

e.g. require_relative 'plane'

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+class Airport
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @weather = Weather.new
+    @capacity = capacity
+  end
+
+  attr_reader :planes, :capacity, :weather
```
Danielandro: Not sure, but I think its a convention to define attr_reader's above the initialize definition

PR #1436, opened 2019-06-02
Airport Challenge - Luigi Palermo

l-palermo:
Thanks for reviewing the code.


README-Luigi
```diff
@@ -0,0 +1,90 @@
...
+* 7th requirement - TDD Prevent landing if airport is full
+* 8th requirement - TDD Overridable airport capacity
+* 9th requirement - TDD edge cases to prevent inconsistent states of the system
+
+
+## How to use the software
+
+Plaese see below a simple example on how the software runs:
+
+## Feature test
```
LucyMHall: Really nice feature test, reading this already gives me a pretty good idea about how your code works and is very helpful!

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require '/Users/student/Desktop/Desktop/projects/Week1/Weekend_1/airport_challenge/lib/airplane.rb'
```
LucyMHall: if the file you want to require is in the same directory as the current file you are in, you can write `require_relative "filename"` and you don't have to include the whole path

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+require '/Users/student/Desktop/Desktop/projects/Week1/Weekend_1/airport_challenge/lib/weather.rb'
+
+class Airport
+
+  attr_accessor :weather, :capacity, :hangar
```
LucyMHall: Do you want the user of the program to be able to alter the contents of the hangar and what the weather is? E.g. someone trying to land a plane overriding the weather that is actually present and saying that it is sunny instead, or say that a plane is or isn't in a hangar independent of whether you have landed it there?

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    @weather = :sunny
+    @capacity = 5
+    @hangar = []
+  end
+
+  def land(airplane, weather = Weather.new.status)
+    weather
+    raise 'The weather is stormy' if weather == :stormy
+
+    raise 'The airport is full' unless @hangar.length < @capacity
```
LucyMHall: Why did you choose to use 'unless'?

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require '/Users/student/Desktop/Desktop/projects/Week1/Weekend_1/airport_challenge/lib/airplane.rb'
+require '/Users/student/Desktop/Desktop/projects/Week1/Weekend_1/airport_challenge/lib/weather.rb'
+
+class Airport
+
+  attr_accessor :weather, :capacity, :hangar
+
+  def initialize
+    @weather = :sunny
```
LucyMHall: Why did you choose to make sunny the default weather condition?

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def take_off(airplane, weather = Weather.new.status)
+    weather
+    raise 'The weather is stormy' if weather == :stormy
+
+    raise 'The airplane is not in the airport' unless @hangar.include?(airplane)
+
+    airplane = @hangar.delete(airplane)
+    "The airplane #{airplane} has taken off"
```
LucyMHall: I like that you use string interpolation so that the confirmation that you give is personalised for each plane

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  attr_accessor :weather, :capacity, :hangar
+
+  def initialize
+    @weather = :sunny
+    @capacity = 5
+    @hangar = []
+  end
+
+  def land(airplane, weather = Weather.new.status)
```
LucyMHall: Why did you decide to include this `weather = Weather.new.status` up here?

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    weather
```
LucyMHall: What does this line do?

lib/weather.rb
```diff
@@ -0,0 +1,10 @@
+class Weather
+  def status
+    n = rand(1..10)
```
LucyMHall: It is interesting, and probably more accurate, that you haven't given sunny and stormy equal chances of appearing :)

spec/airplane_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'airplane'
```
LucyMHall: Do you need an rspec file for airplane?

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    it 'raise an error if the weather is stormy' do
+      weather = :stormy
+      expect { airport.land(airplane, weather) }.to raise_error 'The weather is stormy'
+    end
+
+    it 'raise an error if the airport is full' do
+      weather = :sunny
+      airport.capacity = 2
+      airport.hangar = [1, 2, 3]
```
LucyMHall: If airport capacity is 2 then should this array only contain [1, 2]?

lib/airport.rb
LucyMHall: Maybe nice to extract the `hangar.length < capacity` into a separate method e.g. full?

lib/airport.rb
l-palermo: I have used attr_accessor to enable the test to override the @variable. I should have use default parameter for the method initialise to allow the test to work. Or run a method before the expect 
 syntax. Thanks for the hint.

lib/airport.rb
l-palermo: To vary! :)

lib/airport.rb
l-palermo: Well, it was just to have a weather status when I initialise the class. 
It could have been random but I though why do not make it always ☀️sunny.

lib/airport.rb
l-palermo: Well I needed to pass the test a weather value and keep it random as default.
It's wrong so I am changing it now.

lib/airport.rb
l-palermo: It's redundant

spec/airplane_spec.rb
l-palermo: I need to test the class Airplane to see if it responds

spec/airport_spec.rb
l-palermo: Yes, your suggestion is more accurate.

spec/airplane_spec.rb
LucyMHall: does the code in your airplane.rspec file do this?

lib/airport.rb
l-palermo: do you think?

they would be 3 more lines of code to avoid using one word in the condition.

PR #1438, opened 2019-06-03
Weekend Challenge: Airport Challenge

Syed-A-Shah:


spec/airport_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+  it 'tells a plane to land' do
+    airport = Airport.new
+    expect(airport).to respond_to(:land).with(1).argument
```
celfro: Do you need to write .with(1).argument? Is there a way to refactor the line?

celfro: try and use a different matcher than respond_to

spec/airport_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+  it "adds a new plane to hangar" do
```
celfro: very clear what you want the test to do!

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+require_relative 'plane'
+class Airport
+  attr_accessor :hangar
```
celfro: why did you use attr_accessor here?

spec/weather_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'weather'
+
+RSpec.describe Weather do
```
celfro: Why did you use Rspec here?

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
...
+  def initialize
+    @hangar = []
+  end
+
+  def land(plane)
+    @hangar << plane
+  end
+
+  def take_off(plane)
+    fail 'No planes available' if empty?
```
celfro: What does fail do?

PR #1439, opened 2019-06-03
created class airport and spec file

lucianmot:


lib/airport.rb
```diff
@@ -0,0 +1,34 @@
+
+class Airport
+  attr_reader :initialize
```
mattstockman: do you need an attr_reader for initialize method?

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+  def initialize
+  @hanger = []
+  end
+
+  def land(plane)
+  @hanger << Plane.new
```
mattstockman: @hanger = [] should be indented

spec/airport_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'airport'
+
+describe Airport do
+  it "test if method initialize exists" do
+  expect(subject).to respond_to :initialize
```
mattstockman: could be easier to read using an explicitly defined subject?

PR #1440, opened 2019-06-03
Submitting work so far...

carlfjones:
I was sick over the weekend and feel like my brain was not with me on this one. Will work on it with a clearer head and hopefully conquer it.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,25 +1,31 @@
+require './lib/weather.rb'
 
 class Airport
-
+@weather = Weather.new
   def initialize(capacity = 5)
     @capacity = capacity
-    @planes = [1]
+    @planes = []
   end
 
+ def full?
+   @planes.length >= @capacity
+ end
+
     def land(plane)
-      if @planes.length < @capacity
-          @planes << plane
-        true
-      else
-        false
-      end
+      return false if full?
+       @planes << plane
     end
+
       def take_off(plane)
-        if @planes.length > 0
-           @planes.pop
-            true
+        if @weather = "stormy"
+          false
         else
-         false
+          if @planes.length > 0
+            @planes.pop
+            true
+          else
+            false
+        end
        end
       end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,8 @@
+class Airport
+  def land(plane)
+    true
+  end
+  def take_off(plane)
+    true
+  end
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,8 +1,26 @@
+
 class Airport
-  def land(plane)
-    true
+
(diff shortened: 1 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,32 @@
+require './lib/weather.rb'
```
rachjgriff: Why don't you need to 'require' the Plane class?

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+
+class Airport
+@weather = Weather.new
```
rachjgriff: Why is @weather defined outside of the initialize method?

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+    @capacity = capacity
+    @planes = []
+  end
+
+ def full?
+   @planes.length >= @capacity
+ end
+
+    def land(plane)
+      return false if full?
```
rachjgriff: Don't think you need to actually write return - Ruby should do this automatically

lib/weather.rb
```diff
@@ -0,0 +1,10 @@
+class Weather
+
+  weather = rand(1..5)
+    if (1..4) === weather
```
rachjgriff: What does triple = do?

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  context 'plane is in the air' do
+      plane = Plane.new
+      it 'lands plane' do
+        expect(subject.land(plane)).to eq true
+      end
+      it 'if weather is stormy, planes can not land' do
+          allow(Weather.new).to receive("stormy")
+          expect(subject.land(plane)).to eq false
+      end
+      it 'if weather is stormy, planes can not land' do
```
rachjgriff: I'm a little confused by the name of the test - should it refer to the weather being sunny?

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
+require './lib/airport.rb'
+require './lib/plane.rb'
+require './lib/weather.rb'
+
+describe Airport do
+
+  context 'plane is in the air' do
```
rachjgriff: Why did you use 'context' rather than 'describes' - what is the difference?

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    plane = Plane.new
+    allow(subject).to receive(:full?) { true }
+    expect(subject.land(plane)).to eq false
+  end
+  it 'if the airport is full, planes can not land' do
+    plane = Plane.new
+    allow(subject).to receive(:full?) { false }
+    expect(subject.land(plane)).to eq true
+  end
+
```
rachjgriff: Would be good to see some raise_error tests in there.

PR #1442, opened 2019-06-03
Airport challenge - work in progress

ajmpawlik:


airport.rb
```diff
@@ -0,0 +1,20 @@
+class Airport
+  def get_hangar
```
aniasobo: What is the purpose of having a method assign a hangar? Are airports ever going to be initialised as objects that don't accept planes? I'd suggest moving the hangar to the initialize (currently missing) & adding an attribute reader of :hangar

airport_spec.rb
```diff
@@ -0,0 +1,36 @@
+require "airport"
+
+describe Airport do
+  it 'is available as described class' do
+    expect(described_class).to eq(Airport)
+  end
+
+  it { is_expected.to respond_to(:get_hangar).with(0).argument }
```
aniasobo: I might be wrong about this but I think the with argument might be redundant because the method doesn't take an argument, and it returns a value.

airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  it 'is checking if plane is allowed to land' do
+    airport = Airport.new
+    hangar = airport.get_hangar
+    expect(airport.allowed_to_land()).to eq true
+  end
+
+  it 'is checking if the plane landed' do
+  airport = Airport.new
+  hangar = airport.get_hangar
+  airport.land_plane('lll')
```
aniasobo: The .land_plane method doesn't take any arguments. Will this test pass?

PR #1445, opened 2019-06-28
rspec and rubocop passing

zi-codes:


lib/plane.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def land_at(airport)
+    airport.receive(self)
+    @location = airport.name
+  end
+
+  def take_off_from(airport)
+    airport.send(self)
+    @location = "airborne"
```
renatapop: Extra variables assigned

PR #1446, opened 2019-06-29
TB_Airport

TBrockster:
Finished user stories, but not refactored, using doubles or tested for edge-cases.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,5 @@
+class Airport
+  def land_plane(plane)
+    plane
+  end
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,5 +1,6 @@
 class Airport
   attr_reader :local_weather
+  attr_writer :local_weather
   def initialize
     @planes = []
     @airport_weather_object = Weather.new
@@ -7,14 +8,22 @@ class Airport
   end
 
   def land_plane(plane)
+    fail 'The Stormy weather prevents landing' if @local_weather == 'Stormy'
+    
     @planes << plane
     plane
   end
 
   def takeoff_plane
     fail 'No planes available for takeoff' if @planes.empty?
+    fail 'The Stormy weather prevents takeoff' if @local_weather == 'Stormy'
 
     @planes.pop
     'Plane has taken off'
+  end
+
+  def regenerate_weather
+    puts "Regenerating weather - it is currently #{@local_weather}!"
+    @local_weather = @airport_weather_object.generate_weather
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -30,6 +30,7 @@ class Airport
 
     @planes.pop
     'Plane has taken off'
+    # doesn't set plane to flying
   end
 
   def regenerate_weather
@@ -51,7 +52,7 @@ class Airport
     @local_weather == 'Stormy'
   end
 
-  def flying?(plane)
+  def flying?
     plane.state == 'Flying'
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,5 +1,17 @@
 class Airport
+  def initialize
+    @planes = []
+  end
+
   def land_plane(plane)
+    @planes << plane
     plane
+  end
+
+  def takeoff_plane
+    fail 'No planes available for takeoff' if @planes.empty?
+
+    @planes.pop
+    'Plane has taken off'
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,15 +1,19 @@
 class Airport
   attr_reader :local_weather
+  attr_reader :capacity
   attr_writer :local_weather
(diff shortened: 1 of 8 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,61 @@
...
+  attr_reader :local_weather
+  attr_reader :capacity
+  attr_reader :airport_id
+  attr_writer :local_weather
+  DEFAULT_CAPACITY = 5
+
+  def initialize(airport_id: "Toby's Airport", capacity: DEFAULT_CAPACITY)
+    @planes = []
+    @airport_weather_object = Weather.new
+    @local_weather = @airport_weather_object.weather
```
alicelieutier: What does this line do?

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
+  attr_reader :weather
+  def initialize
+    generate_weather
+  end
+  
+  def generate_weather
+    rand(5).zero? ? @weather = 'Stormy' : @weather = 'Sunny'
```
alicelieutier: Why are you using a ternary expression here? Could this be more readable in two lines?

lib/airport.rb
```diff
@@ -0,0 +1,61 @@
+class Airport
```
ckettell: I may be wrong, but I think you may need require relative at the top. I tried loading just airport.rb in IRB and meant I couldn't create an instance.

lib/airport.rb
```diff
@@ -0,0 +1,61 @@
...
+  attr_reader :local_weather
+  attr_reader :capacity
+  attr_reader :airport_id
+  attr_writer :local_weather
+  DEFAULT_CAPACITY = 5
+
+  def initialize(airport_id: "Toby's Airport", capacity: DEFAULT_CAPACITY)
+    @planes = []
+    @airport_weather_object = Weather.new
```
ted-scanlan: I get why you have your own class, but for me generating weather in the Airport class would be slightly more readable

lib/airport.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def full?
+    @planes.count >= capacity
+  end
+
+  def stormy?
+    @local_weather == 'Stormy'
+  end
+
+  def flying?(plane)
```
ckettell: This method I do not think works. On line 19, you call plane.flying? Whereas, this one requires an argument. Means that I can't land a plane

lib/airport.rb
```diff
@@ -0,0 +1,61 @@
...
+    @airport_weather_object = Weather.new
+    @local_weather = @airport_weather_object.weather
+    @capacity = capacity
+    @airport_id = airport_id
+  end
+
+  def land_plane(plane)
+    fail 'The Stormy weather prevents landing' if stormy?
+    fail 'The Airport is full' if full?
+    fail 'A plane cannot land if it is not flying' unless plane.flying?
```
ted-scanlan: would this create a more simple design if you were to use the state attr_reader in the Plane class instead?

lib/airport.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    plane.state = @airport_id
+    @planes << plane
+    plane
+  end
+
+  def takeoff_plane
```
ckettell: Without taking an argument, you wouldn't be able to find out if a specific plane has left the airport. By using pop, you can't keep track of which ones have left

spec/airport_spec.rb
```diff
@@ -0,0 +1,82 @@
+require 'airport'
+
+describe Airport do
+  it { is_expected.to respond_to(:land_plane).with(1).argument }
+  it { is_expected.to respond_to(:takeoff_plane) }
```
ted-scanlan: these "respond_to" are not that necessary

spec/airport_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  it { is_expected.to respond_to(:regenerate_weather) }
+  let(:flying_plane) { double :plane, :state => 'Flying', :state= => :state, :flying? => :state }
+  let(:stored_plane) { double :plane, :state => subject.airport_id, :state= => :state, :flying? => :state }
```
ted-scanlan: I'm not sure what's going on in the second half of this line

spec/airport_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  it 'has a variable capacity' do
+    my_airport = Airport.new(capacity: 10)
+    my_airport.local_weather = 'Sunny'
+    10.times { my_airport.land_plane(flying_plane) }
+    expect { my_airport.land_plane(flying_plane) }.to raise_error 'The Airport is full'
+  end
+
+  describe '#land_plane' do
+    it 'stores planes' do
+      subject.local_weather = 'Sunny'
```
ted-scanlan: it is better to use a stub here to replace the random i.e. allow(subject.to receive(class method) {value}

PR #1447, opened 2019-06-30
Will's Airport Challenge Attempt

willhowes:


README.md
```diff
@@ -1,90 +1,19 @@
...
-        ______
-        _\____\___
-=  = ==(____MA____)
-          \_____\___________________,-~~~~~~~`-.._
-          /     o o o o o o o o o o o o o o o o  |\_
-          `~-.__       __..----..__                  )
-                `---~~\___________/------------`````
-                =  ===(_________)
+### NOTE TO CODE REVIEWER
+I realised at a late stage that it would have perhaps been better to have the ```Plane``` class have methods of ```.take_off``` and ```.land rather``` than the ```Airport``` class having methods of ```.plane_take_off``` and ```.land_plane```.
```
richieganney: Why do you think it would have been better to have the land and take off methods in the Plane class?

lib/Airport.rb
```diff
@@ -0,0 +1,40 @@
+require_relative './plane.rb'
+require_relative './weather.rb'
+
+class Airport
+  include Weather
```
richieganney: Do you need 'include weather' as you have already used require_relative above?

i may be wrong but i didn't think you needed this

lib/Airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    @planes = []
+    generate_weather
+    @capacity = capacity
+  end
+
+  def land_plane(plane)
+    fail "Weather is stormy - Cannot land plane" if stormy?
+    fail "Airport full" if full?
+    fail "Plane not flying" unless plane.flying?
+
```
richieganney: empty space?

lib/Airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    plane.flying = false if plane.flying?
+    @planes << plane
+  end
+
+  def plane_take_off(plane)
+    fail "Weather is stormy - cannot take off" if stormy?
+    fail "Plane already flying" if plane.flying?
+
+    @planes.delete(plane)
```
richieganney: do not need the '@' if you have included this array in the initialize method.

i think...

lib/Airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  private
+
+  def full?
+    return true if @planes.length >= @capacity
```
richieganney: you can refactor to leave out 'return true if' and just have 'planes.length >= capacity' because this will implicitly return true anyway

lib/Airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def stormy?
+    return true if @weather == 'stormy'
```
richieganney: same refactor comment as above

lib/Plane.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize
+    @flying = false
+  end
+
+  def fly
+    @flying = true
+  end
+
+  def flying?
+    true if @flying == true
```
richieganney: can refactor and leave out 'true if' (same as above)

lib/Plane.rb
```diff
@@ -0,0 +1,16 @@
+class Plane
+
+  attr_accessor :flying
+
+  def initialize
+    @flying = false
```
richieganney: would it be better to have '@flying = false' as a default argument in your initialize method? e.g.

def initialize(flying = false)
  @flying = flying
end

maybe not but a suggestion

PR #1448, opened 2019-06-30
Pull request

KajaMaria:


lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 50
+
+  attr_accessor :planes, :stormy, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @stromy = stormy
```
riannemcc: Spelling mistake here(stromy). Did it affect the running of the code?

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+    @capacity = capacity 
+  end
+
+  include Weather
+
+  def land(plane)
```
riannemcc: Could this be reduced into single responsibility methods? Different methods for landing plane in bad weather and then when airport is at capacity?
I can't remember whether I did this myself but no it seems like it might be a good option...

spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+describe Airport do
+  let(:plane) { double :plane }
+  subject(:airport) { described_class.new }
+
+  describe '#new' do
+    it "has no planes when inititialized" do
+      expect(airport.planes).to eq []
+    end
+  end
+  describe '#land' do
```
riannemcc: Should this method name be clearer, to specify that it's a plane landing?

spec/plane_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'plane'
```
riannemcc: Empty spec file. No testing require for plane class. Should it be deleted?
But not spec file for the weather. Should the random number generator be tested? A test to see what happens it the number generates specific numbers, both below and over 8?

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  private
+
+
+  def empty?
+    @planes.empty?
+  end
+
+  def full?
+    @planes.length >= capacity
```
riannemcc: You've set an instance variable (@capacity) and then not used it in the code, so is it needed?

I don't know the answer to this...
I didn't get as far as using capacity so I'm not sure how I would have done it.

PR #1449, opened 2019-06-30
added airport and plane classes

gennyallcroft:


lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+    rand(1)
+  end
+
+  def stormy?
+    weather_generator == 1
+  end
+
+  def plane_take_off(plane,stormy=false)
+    raise "The plane cannot take off due to storms" if stormy?
+    raise "The plane is not at this airport" if !at_airport?
```
hemser1: Could we have used empty? on @planes_at_airport - therefore wouldn't need at_airport.

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    allow(subject).to receive(:weather_generator) { 0 }
+    subject.receive_plane(:plane)
+    airport2 = Airport.new
+    expect{airport2.plane_take_off(:plane)}.to raise_error 'The plane is not at this airport'
+  end
+
+  it 'planes will not take off in stormy weather' do
+    allow(subject).to receive(:weather_generator) { 0 }
+    subject.receive_plane(:plane)
+    allow(subject).to receive(:weather_generator) { 01}
```
hemser1: Should this be 1? not 01?

lib/airport.rb
```diff
@@ -0,0 +1,52 @@
...
+    @planes_at_airport << plane
+    @at_airport = true
+    plane
+  end
+
+  private def full?
+    @planes_at_airport.length >= DEFAULT_CAPACITY
+  end
+
+  def at_airport?
```
hemser1: Use this as a setter / getter method - therefore you only need to refer to 'at_airport?' not '@at_airport?' this is part of the DRY principle.

lib/plane.rb
```diff
@@ -0,0 +1,23 @@
...
+    raise "The plane cannot take off as it is already in the air" if flying?
+    @flying = true
+  end
+
+  def land
+    raise "The plane cannot land as it has not taken off yet" if !flying?
+    @flying = false
+  end
+
+  def flying?
```
hemser1: Again this could be used as a setter method therefore you only need to refer to 'flying?' not '@flying?'

PR #1451, opened 2019-06-30
Completed Airport Challenge

hemser1:
Rubocop not running due to error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,52 +13,72 @@ Airport Challenge
 
 ```
 
-Instructions
----------
-
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Steps
--------
-
-1. Fork this repo, and clone to your local machine
-2. Run the command `gem install bundle` (if you don't have bundle already)
-3. When the installation completes, run `bundle`
-4. Complete the following task:
-
 Task
 -----
 
 We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
 
+
 ```
-As an air traffic controller 
-So I can get passengers to a destination 
+------------------------------------------------
+As an air traffic controller
+So I can get passengers to a destination
 I want to instruct a plane to land at an airport
-
-As an air traffic controller 
-So I can get passengers on the way to their destination 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Create Airport class
+Create Plane class
+Create Method for landing a Plane at Airport
+(Test - lands a plane)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -73,7 +95,7 @@ In code review we'll be hoping to see:
 
 * All tests passing
 * High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc. 
+* The code is elegant: every class has a clear responsibility, methods are short etc.
```
gennyallcroft: take a look at the README guidelines for what information should be included in a README file

lib/airport.rb
```diff
@@ -0,0 +1,57 @@
...
+class Airport
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+    @weather = weather_state
+  end
+
+  def planes
```
gennyallcroft: What do these methods do? Are they are necessary?

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
+#Test Document to Test -
+require 'airport'
+require 'plane'
+
+describe Airport do
+  subject(:airport) { Airport.new }
```
gennyallcroft: What does this line do?

lib/airport.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def full?
+    planes.count >= capacity
+  end
+
+  def empty?
+    planes.empty?
+  end
+
+  def stormy?
```
gennyallcroft: include the weather methods in a separate file?

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  let(:plane) { double :plane }
+
+    describe 'initialization' do
+      it 'defaults capacity' do
+        described_class::DEFAULT_CAPACITY.times do
```
gennyallcroft: repetitive? could this be written more succinctly?

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    it 'lands a plane' do
+      allow(airport).to receive(:weather_state) { 'sunny' }
+      expect(airport.land(plane)).to eq [plane]
+    end
+
+    it 'takes off a plane' do
+      allow(airport).to receive(:weather_state) { 'sunny' }
+      airport.land(plane)
+      expect(airport.take_off).to eq plane
```
gennyallcroft: is this behaviour correct?

lib/airport.rb
hemser1: these are setter / getter methods :)

hemser1: they should be in private my bad!

spec/airport_spec.rb
hemser1: sets airport as the variable - so instead of putting subject we put airport. Makes no real difference.

lib/airport.rb
hemser1: fair point.

spec/airport_spec.rb
hemser1: fair point again :)

spec/airport_spec.rb
hemser1: returns plane object - however should really return string to confirm action.

PR #1452, opened 2019-06-30
airport_challenge

ckettell:


lib/weather.rb
```diff
@@ -0,0 +1,19 @@
...
+  def initialize
+    @forecast = forecast?
+  end
+
+  def forecast?
+    rand(10) == 2 ? 'Stormy' : 'Sunny'
+  end
+
+  def stormy?
+    @forecast == 'Stormy'
```
loicdewit: I don't know if I would have included the method stormy? as you could just as well call forecast directly in your airport class. We should ask the coaches if in terms of code quality, it is better to include it or omit it.

PR #1453, opened 2019-06-30
Daniel Verrall - wasn't able to pass Error tests for stormy weather.

djvcom:


lib/airport.rb
```diff
@@ -0,0 +1,39 @@
+require_relative 'weather'
+
+class Airport
+  attr_accessor :current_aircraft, :capacity, :airport_code, :safe_to_fly
```
Kaymo1990: Reminder to use attr_reader

PR #1455, opened 2019-06-30
complete weekend airport challenge

Kaymo1990:


airport_dir/lib/plane.rb
```diff
@@ -0,0 +1,29 @@
...
+  def initialize(airport)
+    @airplane_state = "Landed"
+    airport.airport_hanger.push(self)
+    @current_airport = airport
+  end
+
+  def takeoff
+   if good_weather == true
+     @airplane_state = "Flying"
+     @current_airport.airport_hanger.delete(self)
```
jamiepal: I think it's very neat to delete the particular plane from the airport hangar in this way. It seems like your planes can do the takeoff method while they're already in flying state. I think challenge readme said to try and prevent this.

PR #1457, opened 2019-06-30
Airport Challenge - pull req

marvinhm:


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+  attr_reader :capacity
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+    @weather = ""
+  end
+
+
+  def weather_generator
```
BenjaminDarking: This way works well, the spec asked for a number generator version which would have been using the rand method.

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def land(plane)
+    fail "Can not land while weather is this bad" if safe?
+    fail "Airport is full!" if full?
+    @planes << plane
+  end
+
+  def safe?
+    if @weather == "sunny"
```
BenjaminDarking: On my code I believe similar example to this, rubocop offered a refactoring option of:
def safe?
if @weather == "sunny 
  true
end
end
As mentioned, I believe it would auto return a false but may be worth a quick test

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+      true
+    elsif @weather == "stormy"
+      false
+    end
+  end
+
+  def full?
```
BenjaminDarking: Refactoring option - you can use an implied return on this method.
def full?
@planes.count >= @capacity 
end

PR #1458, opened 2019-06-30
Rianne McCartney - Week One Airport Challenge 

riannemcc:


lib/airport_challenge.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize
+    @airport = []
+  end
+
+  def land_plane(plane)
+    @airport << plane
+  end
+
+  def take_off(plane)
+    @airport.pop
```
KajaMaria: i think that you might need to specify what is being removed in case you have new variables added in the future ? this is what i have done:
```rb
@airport.pop(plane)
```

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
+  attr_reader :stormy
+
+  def initialize
+    @stormy = false
+  end
+
+  def stormy_weather
+    @stormy = true
```
KajaMaria: I have added "random" to define what makes weather stormy. but i like the way you have used a method to in your airport class to define it.

spec/airport_challenge_spec.rb
```diff
@@ -0,0 +1,35 @@
+require './lib/airport_challenge.rb'
```
KajaMaria: i don't think you need "lib" here. try if it works just by requiring your folder. You also might not have to use rb. 
if your folders are not in separate folders you should be able to just use require_relative :)

spec/airport_challenge_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+require './lib/plane.rb'
+require './lib/weather.rb'
+
+describe Airport do
+
+  it 'lands a plane' do
+    plane = Plane.new
```
KajaMaria: You may want to use use mocking in case your code would change in the future. 
You can use method "let" to create doubles for your plane:
```rb
let(:plane) { double :plane }
```

spec/airport_challenge_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  it 'allows plane to take off from airport' do
+    plane = Plane.new
+    airport = Airport.new
+    subject.take_off(plane)
+    expect(subject.take_off(plane)).to eq []
+  end
+
+  it 'prevents plane taking off if weather is stormy' do
+    plane = Plane.new
```
KajaMaria: method "let" would allow you not to repeat yourself :)

spec/airport_challenge_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    airport = Airport.new
+    weather = Weather.new
+    weather.stormy_weather
+    expect(subject.take_off).to eq [plane]
+  end
+
+  it 'raises error when weather is stormy' do
+    plane = Plane.new
```
KajaMaria: Same as "let" you can define your subject on top of the code just under the described class and you wouldn't have to repeat yourself.  I did it this way:
```rb
subject(:airport) { described_class.new } or use { Airport.new } it is the same thing
```

lib/airport_challenge.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def land_plane(plane, weather)
+    @hangar << plane unless weather.stormy? || hangar_full?
+  end
+
+  def take_off(weather)
+    @hangar.pop unless weather.stormy?
+  end
+
+  private def hangar_full?
```
KajaMaria: it might read easier if you leave private on top of your method. You could reuse it for other methods too :)

PR #1463, opened 2019-07-01
Airport challenge 

ted-scanlan:


lib/Airport.rb
```diff
@@ -0,0 +1,54 @@
+class Airport
+  attr_reader :stormy
+  attr_reader :capacity
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY, is_at_airport = false)
+    @hangar = []
+    @stormy = stormy
+    @capacity = capacity
+    @is_at_airport = is_at_airport
```
TBrockster: Shouldn't this be a property of the plane, not of the airport? If you land Plane A, you can then takeoff Plane B, without landing Plane B first.

lib/Airport.rb
```diff
@@ -0,0 +1,54 @@
...
+    raise "too stormy to land" if @stormy
+    raise "plane has already landed" if plane.landed == true
+    @hangar.push(plane)
+    @is_at_airport = true
+    plane.is_landed
+
+
+  end
+
+  def take_off(plane=nil)
```
TBrockster: Why does this have an argument? And why does that argument default to nil?

lib/Airport.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def report_storm
+    @stormy = true
+  end
+
+  private
+
+  def rand
+    random_weather = ["sunny", "sunny", "sunny", "sunny", "stormy"].sample
```
TBrockster: This could be done in a cleaner way, perhaps allowing for the 'chance of storm' to be easily changed in future.

lib/Airport.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def full?
+    @hangar.count >= @capacity ? true : false
```
TBrockster: Why is this a ternary operator?

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
+class Plane
+
+  attr_reader :landed
+
+def initialize(landed=false)
+@landed = landed
+end
+
+def is_landed
```
TBrockster: It seems like there is no way to change a landed plane back to 'not landed'.

spec/Airport_spec.rb
```diff
@@ -0,0 +1,104 @@
+require 'Airport'
+require 'plane'
+
+
+describe Airport do
```
TBrockster: I like the structure and nesting of your tests.

lib/Airport.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  end
+
+  def check_weather
+    todays_weather = rand
+    report_storm if todays_weather == "stormy"
+
+  end
+
+  def report_storm
```
TBrockster: I think this should be private? also it seems there is no way to change the weather back from stormy?

lib/Airport.rb
```diff
@@ -0,0 +1,54 @@
...
+  def take_off(plane=nil)
+    raise "hangar is empty" if @hangar.empty? && plane == nil
+    raise "plane is already flying" if @is_at_airport == false
+    raise "too stormy for take-off" if @stormy
+    @hangar.pop
+    return "plane has left the airport"
+
+  end
+
+  def check_weather
```
TBrockster: This is only called on by tests?

PR #1464, opened 2019-07-01
airport challenge

renatapop:


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+      raise 'Cannot land plane: airport full' if full?
+      raise 'Cannot land plane: weather is stormy' if stormy?
+      @planes << plane
+    end
+  
+    def take_off(plane)
+      raise 'Cannot take off plane: weather is stormy' if stormy?
+    end
+  end
+
```
zi-codes: Don't forget to delete the plane off your plane array when it takes off

PR #1465, opened 2019-07-01
richieganney/airport challenge

richieganney:


Gemfile
```diff
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 gem 'rake'
 gem 'rubocop', '0.60.0'
+gem 'pry'
```
willhowes: Not used pry before. Is it useful?

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    if stormy?
+      raise("Permission to take off denied due to stormy weather")
+    else 
+      planes.empty? ? raise("Error: no planes in airport") : authorize
+    end
+  end
+
+ private
+
+  def authorize
```
willhowes: Not sure I understand why @planes is returned at the end of this method.

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    puts "#{planes.pop} has successfully taken off" unless planes.empty?
+    @planes
+  end
+
```
willhowes: Should the airport_full? and stormy? methods be made private? I'm only saying this because that is what they got us to do for Boris Bikes...

spec/airport_spec.rb
```diff
@@ -0,0 +1,77 @@
+require 'airport'
+
+describe Airport do
+
+  describe '#instruct_land' do
+    it 'instructs a plane to land at an airport' do
+        allow(subject.weather).to receive(:sample) { "sunny" }
```
willhowes: I think I tried this syntax in my code to force the weather to be sunny but didn't work until I made a mock for Airport class. Would be good to chat about this.

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    @weather = ["stormy", "sunny", "sunny", "sunny", "sunny", "sunny"]
+  end
+
+  def instruct_land(plane)
+    if stormy?
+      raise("Permission to land denied due to stormy weather")
+    elsif airport_full?
+      raise("Error: Airport is full. find somewhere else to land")
+    else
+      @planes.include?(plane) ? raise("Error: this plane has already landed") : planes << plane
```
jamiepal: I think because you're checking whether the plane has landed in this way, it only checks whether the plane is docked at that current airport. Whereas a plane could land at another airport and I think you would still be able to instruct that plane to land at this airport. but it works fine with only one airport.

PR #1474, opened 2019-07-28
Airport Weekend Challenge Pull Request (Russell Wind)

Stegosauruss:


Stegosauruss:
I had to repeat myself quite a lot with the doubles of my planes. I would be interested in knowing the best practice for this.
I also struggled for a while as I was trying to set the location of my dummy planes when their attributes were read only. I had to move a lot of methods over to my planes as a result (which was probably more elegant anyway) but I would like to know if that is possible.
Many Thanks,
Russ

PR #1489, opened 2019-08-11
Completed airport challenge

nataliehe:


spec/plane_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'plane'
+
+describe Plane do
+  it { should be_a(Plane) }
```
lindseydew: Nice use of the one-liner!

spec/weather_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'weather'
+
+describe Weather do
+  it 'can randomly generate stormy weather' do
+    expect(subject).to respond_to(:stormy?)
```
lindseydew: This is great unit testing for the `Weather` object - well done.

I think it makes sense to check the method can be called and it will respond with either value.

As discussed in the kick off this morning, if you wanted to make the `stormy?` method deterministic you can make use of `srand` which is a `seed` to start randomness: 
https://apidock.com/ruby/Kernel/srand

spec/airport_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'airport'
+
+describe Airport do
+  describe '#land' do
+    it 'can instruct a plane to land' do
+      allow(subject.weather).to receive(:stormy) { false }
```
lindseydew: Small point, but this condition:
`allow(subject.weather).to receive(:stormy) { false }` could be set above your unit tests so that you don't have to repeat it for each test.

spec/airport_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      plane = Plane.new
```
lindseydew: This could also be refactored slightly by writing:
`let(:plane) { Plane.new }` after `describe` and before the unit tests and then each unit test will have a `plane` in scope

spec/airport_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      subject.land(plane)
+      expect { subject.land(plane) }.to raise_error "Plane already landed"
+    end
+
+    it 'raises an error when the airport is full' do
+      allow(subject.weather).to receive(:stormy) { false }
+      plane = Plane.new
+      allow(subject).to receive(:full?) { true }
```
lindseydew: Great use of stubbing here!

spec/airport_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    it 'raises an error if plane is not in the airport' do
+      allow(subject.weather).to receive(:stormy) { false }
+      plane = Plane.new
+      expect { subject.takeoff(plane) }.to raise_error "Plane is not in the airport"
+    end
+
+    it 'raises error when weather is stormy' do
+      allow(subject.weather).to receive(:stormy) { false }
+      plane = Plane.new
+      subject.land(plane)
```
lindseydew: Great testing of edge cases here.

spec/airport_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+  end
+
+  describe '#capacity' do
+    it 'shows a default capacity if not overriden' do
+      expect(subject.capacity).to eq(Airport::DEFAULT_CAPACITY)
+    end
+
+    it 'can be overriden' do
+      subject.capacity = 200
```
lindseydew: Nice, there is another way of `overwriting` variables in the airport class by modifying the `initialize` method.
Eg
```
   def initialize(DEFAULT_CAPACITY = 100)
       @capacity = DEFAULT_CAPACITY
    end
```
Then calling 
```
  airport = Airport.new
```
will set the capacity to the default value of `100`, and it is also possible to call:
```
   airport = Airport.new(1)
```
which will set the capacity to 1.

This approach is slightly preferable to allowing `capacity` to be an `attr_accessor` because it means it can only be set on instantiation of the class, and cannot be modified later.

PR #1490, opened 2019-08-12
Completed airport challenge

sebastianchristopher:


lib/airport.rb
```diff
@@ -0,0 +1,44 @@
+# require 'pry'
+require_relative './weather'
+require_relative './plane'
+
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :planes
+
+  def initialize(capacity = DEFAULT_CAPACITY)
```
lindseydew: Great use of a default arg here.

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+    @capacity = capacity
+    @planes = []
+  end
+
+  def instruct_plane_to_land(plane, weather = Weather.new)
```
lindseydew: Great use of dependency injection here.

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
+class Plane
+
+  def initialize
+    @flying = true
+  end
```
lindseydew: Nice!

Small point, but I think this could be simplified slightly to:

```
class Plane
   attr_reader :flying?
   def initialize
      @flying? = true
   end
    
   def take_off
      flying? = true
   end
   
   def land
      flying? = false
    end
end

```

lib/weather.rb
```diff
@@ -0,0 +1,17 @@
+class Weather
+  attr_reader :weather_rating
+
+  def initialize
+    generate_weather
```
lindseydew: Small point, but could modify this class so that the random_numbers are generated each time the `stormy?` method is called. That way the `weather` object could be a dependency on the `Airport` class rather than a dependency on each of the `land` and `take_off` methods

spec/airport_spec.rb
```diff
@@ -0,0 +1,95 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { double(:plane, flying?: true) }
+  let(:weather) { double(:weather) }
+  before(:each) { allow(plane).to receive(:land) }
+
+  context 'airport capcity' do # this is contingent on instruct_plane_to_land as I don't want a getter method for capacity
+    let(:weather) { double(:weather, stormy?: false) }
```
lindseydew: Great use of a double here well done.

spec/airport_spec.rb
```diff
@@ -0,0 +1,95 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { double(:plane, flying?: true) }
+  let(:weather) { double(:weather) }
```
lindseydew: Very nice isolation of unit tests!

spec/weather_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'weather'
+
+describe Weather do
+  it 'has a weather rating from 1 to 10' do
+    expect(subject.weather_rating).to be_between(1, 10)
```
lindseydew: Nice testing of this class. 
As discussed in kick off this morning, can make the `weather_rating` method deterministic by making use of an `srand` https://apidock.com/ruby/Kernel/srand

PR #1491, opened 2019-08-12
Airport Challenge

StefEmp:
Would you be able to provide some feedback on my code,

Thanks,

Stephane

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -14,7 +14,6 @@ describe Airport do
   end
   it 'should only allow 5 planes' do
     new = Airport.new
-    new.land('croissant')
-    expect {new.airport}
+    expect{new.airport.full}.to raise_error('the runway is full')
   end
 end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,20 @@
+require 'Airport'
+describe Airport do
+  it 'lands a plane' do
+    new = Airport.new
+    new.land('croissant')
+    expect(new.airport).to include('croissant')
+  end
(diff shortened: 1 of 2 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'Airport'
+describe Airport do
+  it 'lands a plane' do
+    new = Airport.new
+    new.land('croissant')
+    expect(new.airport).to include('croissant')
```
lindseydew: Nice, this is a meaningful test on testing the behaviour of `land` well done!
A couple of small stylistic things:
when creating a new instance of a class it's convention to name that object the same as the class.
Eg instead of saying:
`new = Airport.new`
It is convention to say:
`airport = Airport.new`
so that the name of the variable is something meaningful in what it is.

It is also worth giving the `land` method the right `type` of input. Eg:
```
  airport = Airport.new
  plane = Plane.new
  airport.land(plane)
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+  it 'plane leaves the airport, airport has no plane' do
+    new = Airport.new
+    new.land('croissant')
+    new.leave_airport
+    expect(new.airport).to_not include('croissant')
+    # expect(new.airport).to be_empty
```
lindseydew: Nice this test looks good

spec/airport_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+  it 'should only allow 5 planes' do
```
lindseydew: I don't think this test is testing the code that is written?

PR #1492, opened 2019-08-12
airport

claredowson:
To make sure I was on the right path


lib/airport.rb
```diff
@@ -0,0 +1,19 @@
+class Airport
+
+  attr_reader :plane
```
lindseydew: Nice use of an `attr_reader`

lib/airport.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def initialize
+    @plane = []
+  end
+
+  def landed(plane)
+    @plane.push(plane)
```
lindseydew: This is a great implementation of `land` well done!

lib/airport.rb
```diff
@@ -0,0 +1,19 @@
...
+    return @plane
+  end
+
+  def takenoff(plane)
```
lindseydew: Very nice implementation of `takenoff` well done!

spec/airport_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'airport'
+describe Airport do
+  it 'can create an instance of a airport' do
+    airport = Airport.new
+    expect(airport).to be_kind_of(Airport)
```
lindseydew: I can see you are applying to process of TDD here - great job!

spec/airport_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  it 'can create an instance of a plane' do
+    airport = Airport.new
+    expect(airport).to include(:plane)
+  end
```
lindseydew: I'm not sure this is testing what you want here. The text is saying `can create an instance of a plane`, so I feel the body of the unit test should be similar to the test above eg:
``` 
  plane = Plane.new
  expect(plane).to be_kind_of(Plane)
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  it 'should show a plane has landed' do
+    airport = Airport.new
+    expect(airport.plane).to exist(:landed)
+  end
+
+  it 'should check planes at the airport' do
+    airport = Airport.new
+    expect(airport.plane.landed).to eq(:landed)
```
lindseydew: I can see what you are getting at here, but I think some of the syntax isn't quite right

spec/airport_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  it 'should remove a plane when it has taken off' do
+    airport = Airport.new
+    airport.plane.landed
+    ariport.takenoff(plane)
```
lindseydew: This is excellent set up for the test well done. It needs an `expect` statement to be a fully complete test

PR #1493, opened 2019-08-12
Eloise_airport_challenge

EloisePrasad:


docs/airport.rb
```diff
@@ -0,0 +1,22 @@
...
+class Airport
+
+  attr_reader :num_of_planes
+
+  def initialize
+    @planes = []
+  end
+
+  def land
+    plane = Plane.new
```
lindseydew: Small thing, but you could make `plane` an argument of the `land` method so that the `Airport` class is not responsible for creating a new plane (there could be a requirement down the line which restricts the number of planes available for example). The code for this would look like:

```
 def land(plane)
    @planes.push(plane)
end
```

and the test would looks like:
```
#arrange
   airport = Airport.new
   plane = Plane.new
#act
    airport.lane(plane)
#assert
   expect(airport.total_planes).to eq(1)
```

docs/airport.rb
```diff
@@ -0,0 +1,22 @@
+require './docs/plane.rb'
+class Airport
+
+  attr_reader :num_of_planes
```
lindseydew: I think this line of code is not needed, since the tests call the method 
`total_planes` to find out this total.

`attr_reader` is a way of making instance variables accessible. You could instead write:
`attr_reader` :planes

#arrange
   airport = Airport.new
   plane = Plane.new
#act
    airport.lane(plane)
#assert
   expect(airport.planes.length).to eq(1)

Please let me know if this doesn't make sense

PR #1554, opened 2019-10-02
Mark three stories as stretch

jamesjoshuahill:
## Why

Several apprentices in the self-assessed as amber for module 3 because they did not complete the Airport challenge. Their coach assessment was green, which indicates good progress on the learning objective to learn how to do TDD.

To address this completion anxiety:
- we have asked the last two cohorts to focus on three stories first and marked the last three stories as stretch goals.
- we marked the two weather stories as stretch, because they require test isolation with doubles.

## What

- Mark three stories as stretch goals.
- Move task instructions above stretch stories for readability.
- Add headings so it's easier to skim read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,26 +35,30 @@ Task
 
 We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
 
+**Please focus on these stories first:**
 ```
-As an air traffic controller 
-So I can get passengers to a destination 
+As an air traffic controller
+So I can get passengers to a destination
 I want to instruct a plane to land at an airport
 
-As an air traffic controller 
-So I can get passengers on the way to their destination 
+As an air traffic controller
+So I can get passengers on the way to their destination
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,55 +35,68 @@ Task
...
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent takeoff when weather is stormy 
+As an air traffic controller
+To ensure safety
+I want to prevent landing when the airport is full
+```
+
+Your task is to test drive the creation of a set of classes/modules to satisfy all the above user stories.
```
Katerina-codes: Hi @jamesjoshuahill - not sure if Makers already encourages this in the curriculum elsewhere for the early weeks but I normally suggest devs focus on classes only in the first few weeks of the course.

jamesjoshuahill: I agree that removing modules makes sense. This is some existing wording in the README that moved in this pull request.

jamesjoshuahill:
@Katerina-codes and I discussed this change a few weeks ago. Apologies for the delay recording it:
- _stretch_ is a term that is not used at Academy and is only used at Apprenticeships. We want to avoid introducing dissonance.
- we agreed that reordering the stories so that the weather stories that introduce random behaviour are last makes sense. This is the most valuable part of this change.

I'm going to close this pull request and open a new one with a smaller change that only reorders the stories.

PR #1555, opened 2019-10-04
just added heading

albie-b:


adamr27:
I really like your use of refactoring where you have reduced the line count

adamr27:
maybe change some of your variable/method names to make them more clear for other people to understand eg temp, capacity

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  def takeoff(plane)
+    fail 'bad weather' if weather?
+
+    for i in (0...@planes.count)
+      if @planes[i] == plane
+        plane_tempory = @planes[i]
+        @planes.delete_at(i)
+      end
+    end
+    plane_tempory
```
jamesjoshuahill: Could you refactor this for loop using `each`, or another method available on [Array](https://ruby-doc.org/core-2.6.5/Array.html) for deleting?

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+class Airport
+  DEFAULT_CAPACITY = 5
+  attr_accessor :capacity
```
jamesjoshuahill: There is no test that changes `@capacity`, could this be an `attr_reader`?

spec/airport_spec.rb
```diff
@@ -0,0 +1,41 @@
+require './lib/airplane'
+require './lib/airport'
+describe Airport do
+  it 'instructs plane to land' do
+    plane = Airplane.new
```
jamesjoshuahill: Could you use a test double instead of a real instance of `Airplane`?

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  def initialize(input = DEFAULT_CAPACITY)
+    @planes, @capacity = [], input
+  end
+
+  def conditions(input = 'sunny')
```
jamesjoshuahill: I can't see a test for the default value for `input`. Is a default needed?

spec/airport_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    subject.land(plane)
+    expect(subject.inventory).to include(plane)
+  end
+  it 'checks plane is gone' do
+    plane = Airplane.new
+    subject.land(plane)
+    subject.takeoff(plane)
+    expect(subject.inventory).to eq([])
```
jamesjoshuahill: You could use a more expressive RSpec matcher here if you prefer:
```ruby
expect(subject.inventory).to be_empty
```

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+    plane_tempory
+  end
+  private
+  def full?
+    @planes.count >= @capacity
+  end
+
+  def weather?
+    @weather == 'stormy'
```
jamesjoshuahill: It looks like the weather conditions stay the same. How could you make them random?

PR #1556, opened 2019-10-04
Airport challenge code

joshblackmoor:


lib/airport.rb
```diff
@@ -1,13 +1,17 @@
...
+  def initialize(capacity = DEFAULT_CAPACITY)
     @parked_planes = []
+    @capacity = capacity
   end
   #instruct a plane to land
   def land(plane)
     raise 'Plane has already landed' if parked_planes.any? { |all_planes| all_planes == plane }
+    raise 'No parking spaces, DO NOT LAND' if parked_planes.count >= capacity
     plane.onGround
     parked_planes << plane
```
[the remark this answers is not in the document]
joshblackmoor: Thank you for the feedback James. I am going to add comments and separate the code to make it clearer for other people to read :)

spec/airport_spec.rb
```diff
@@ -13,7 +13,6 @@
       expect{ subject.land plane}.to raise_error('Plane has already landed')
     end
 
-    #need to try land more than 10 greater than 10 all_planes
     it 'Will not let planes land if capacity is full' do
      airport = Airport.new(20)
     20.times { airport.land Plane.new }
```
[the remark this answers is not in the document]
joshblackmoor: Thanks James! I tried to keep it simple but effective and I think that I have met this with this code.

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+require_relative 'plane'
```
jamesjoshuahill: I can't see the `Plane` class used in this file. If it isn't needed then please remove this require.

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+  describe "#takeoff" do
+   it 'Will let the plane take off' do
+     plane = Plane.new
+     subject.land(plane)
+     expect(subject.takeoff).to eq plane
+   end
+   end
+end
```
jamesjoshuahill: The indentation in this file is inconsistent which makes it harder to read. Please could you fix it?

In Ruby the most common indentiation is 2 spaces:
```ruby
describe Class do
  describe '#method' do
    it 'does something' do
      # some test here
    end
  end
end
```

lib/plane.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+  def flying?
+    @flying
+  end
+  def inAir
+    @flying = true
+  end
+  def onGround
+    @flying = false
+  end
```
jamesjoshuahill: In Ruby method names are written in snake case. When I see names like this I wonder why they are different to the others.

Please could you change `inAir -> in_air` and `onGround -> on_ground`?

README.md
```diff
@@ -12,6 +12,12 @@ Airport Challenge
                 =  ===(_________)
 
 ```
+Stories completed
+---------
```
jamesjoshuahill: Please could you list the stories you completed here?

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+    parked_planes << plane
+  end
+  #instruct plane to takeoff
+  def takeoff
+    raise 'No Planes available for take off' if parked_planes.count == 0
+    parked_planes.each do |plane|
+      plane.inAir
+      return parked_planes.shift
+    end
+  end
```
jamesjoshuahill: This take off method works well. To make this method more useful, could you try changing it so that a specific plane can be instructed to take off rather than the first one that landed?

lib/plane.rb
jamesjoshuahill: Also, I can't see tests for these two methods. Please could you write tests for them?

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+  DEFAULT_CAPACITY = 10
+  #run when new instance of airport is created
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @parked_planes = []
+    @capacity = capacity
+  end
+  #instruct a plane to land
+  def land(plane)
+    raise 'Plane has already landed' if parked_planes.any? { |all_planes| all_planes == plane }
+    raise 'No parking spaces, DO NOT LAND' if parked_planes.count >= capacity
```
jamesjoshuahill: Great use of guard clauses at the start of this method 👍

PR #1557, opened 2019-10-04
--

josiepollard:


spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'airport'
+
+describe Airport do
+  it { is_expected.to respond_to(:take_off) }
+  it { is_expected.to respond_to(:land).with(1).argument }
+
+  describe 'capacity' do
+    subject { Airport.new}
+    let(:plane) { double :plane }
```
Adam1824: I like your use of doubles, rather than making a new Plane object

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    it 'raises error when no planes in airport' do
+      expect { subject.take_off }.to raise_error 'no planes in airport'
+    end
+
+    it 'airport capacity is full' do
+      described_class::DEFAULT_CAPACITY.times do
+        subject.land(plane)
+      end
+      expect { subject.land(plane) }.to raise_error 'airport is full'
+    end
```
Adam1824: This is a good test but you should do a separate test to see if the plane you land is in the @airport instance variable

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
+require 'plane.rb'
+
+class Airport
+  DEFAULT_CAPACITY = 20
+  attr_accessor :airport, :capacity
```
jamesjoshuahill: I can't see a test that changes the values of `@airport` or `@capacity`, so could these be `attr_reader`?

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
+require 'plane.rb'
```
jamesjoshuahill: I can't see the `Plane` class being used in this file. If it's not needed please remove this require'

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+
+class Airport
+  DEFAULT_CAPACITY = 20
+  attr_accessor :airport, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airport = []
```
jamesjoshuahill: I think `@airport` is an uncomfortable choice for this instance variable. We end up with `airport.airport`, which is odd. What would be a more accurate name for this list?

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail 'airport is full' if full?
```
jamesjoshuahill: Great use of a guard clause :+1:

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+    airport << plane
+  end
+
+  def take_off(plane)
+    raise 'no planes in airport' if empty?
+    airport.pop
+    puts "#{plane} has left the airport"
```
jamesjoshuahill: This looks like a debugging print. Please remove it after you've done your debugging.

spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    it 'has default capacity' do
+      expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
```
jamesjoshuahill: Great testing of the default value for capacity.

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def land(plane)
+    fail 'airport is full' if full?
+    airport << plane
+  end
+
+  def take_off(plane)
+    raise 'no planes in airport' if empty?
+    airport.pop
```
jamesjoshuahill: What if there was more than one plane at the airport? Which plane would take off?

README.md
```diff
@@ -12,6 +12,7 @@ Airport Challenge
                 =  ===(_________)
 
 ```
+stories completed
```
jamesjoshuahill: Please could you list the stories you have completed here?

PR #1558, opened 2019-10-04
First commit

Will-Bath:


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+  attr_accessor :capacity
+
+  def initialize
+    @planes = []
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def print_planes
+    @planes
+  end
```
luke-jones-1: attr_reader :planes

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'airport'
+
+describe Airport do
+
+  it 'instructs plane to land at airport' do
+    plane = Plane.new
```
luke-jones-1: (double :plane) to make test independant, so changes to 'Plane' class wont affect outcomes of 'Airport' class tests

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    airport = Airport.new
```
luke-jones-1: instead of Airport.new you could use subject or described_class.new, so that changing the name of the class wont cause tests to fail.

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    takeoff = airport.take_off_plane
+
+    expect(airport.print_planes).to eq([])
+    expect(airport.print_planes).not_to include(takeoff)
+  end
+
+  it 'prevents landing when airport is full' do
+    airport = Airport.new
+    10.times { airport.land_plane Plane.new }
+    p airport.print_planes
```
luke-jones-1: comment this out at some point if not done already

lib/airport.rb
```diff
@@ -8,17 +8,47 @@ def initialize
...
+  #   if var = 0
+  #     weather = 'sunny'
+  #   else
+  #     weather = 'stormy'
+  #   end
+  # end
+  #
+  # def bad_weather?(weather)
+  #   weather == 'stormy'
+  # end
```
luke-jones-1: when creating a predicate method they wont usually take an argument, instead an instance variable may be better

spec/airport_spec.rb
```diff
@@ -25,8 +25,22 @@
   it 'prevents landing when airport is full' do
     airport = Airport.new
     10.times { airport.land_plane Plane.new }
-    p airport.print_planes
```
luke-jones-1: :-bd

lib/airport.rb
Will-Bath: It's got no argument in the final version of this predicate method

lib/airport.rb
Will-Bath: I've added this in, and it works a treat!

spec/airport_spec.rb
Will-Bath: With the double, I've successfully removed the dependency on the Plane class - and it's less work!

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'airport'
+
+describe Airport do
+
+  let (:plane) { double :plane }
+
+  it 'instructs plane to land at airport' do
+    # plane = Plane.new
+    # airport = Airport.new
```
jamesjoshuahill: I like how you've used `subject` and `let(:plane)` instead of these lines. Now you're done with them, please remove these comments.

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  it 'instructs plane to take off from airport' do
+    # plane = Plane.new
+    # airport = Airport.new
+
+    subject.land_plane(plane)
+    takeoff = subject.take_off_plane
+
+    expect(subject.planes).to eq([])
+    expect(subject.planes).not_to include(takeoff)
```
jamesjoshuahill: I think both of these expectations make sense in this scenario. They also overlap: if `subject.planes` is empty then it cannot include `takeoff`.

I recommend picking the expectation that gives you enough confidence in `take_off_plane()` and remove the other.

PR #1559, opened 2019-10-04
added stories

2Blagden:


lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    if @planes.count < @capacity
+      @planes.push(plane)
+      p @planes
```
lukegittins01: I like the way you have gotten visual aid from debugging your code

spec/airport_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'airport'
+describe Airport do
+
+  it "makes a plane land at an airport" do
```
lukegittins01: I like the way you have made sure the programme responds to your methods before making tests in the spec

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    if @planes.count < @capacity
+      @planes.push(plane)
```
jamesjoshuahill: I can't see a test that landing a plane adds it to the list of `airport.planes_here`. For example:
```ruby
it 'can instruct a plane to land' do
  plane = Plane.new

  subject.land(plane)

  expect(subject.planes_here).to include(plane)
end
```

lib/airport.rb
jamesjoshuahill: Please remove debugging prints after you have finished with them.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class Plane
+  def initialize
+  end
+end
+
+plane1 = Plane.new
+airport1 = Airport.new
+airport1.land(plane1)
+airport1.take_off(plane1)
```
jamesjoshuahill: Please remove this debugging code. If you would like to keep it somewhere you could create a separate file.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def planes_here
+    @planes
+  end
+end
+
+class Plane
+  def initialize
+  end
+end
```
jamesjoshuahill: Please put each class in a separate file.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+class Airport
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
```
jamesjoshuahill: I can't see a test for the default capacity of 10 planes. Please could you add one?
```ruby
it 'has a default capacity' do
  10.times do { subject.land(Plane.new) }
  
  expect { subject.land(Plane.new) }.to raise_error "airport full!"
end
```

README.md
```diff
@@ -12,7 +12,15 @@ Airport Challenge
                 =  ===(_________)
 
 ```
+Stories Completed
 
+As an air traffic controller
```
jamesjoshuahill: I think you have completed more than one story 👀

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    if @planes.count < @capacity
+      @planes.push(plane)
+      p @planes
+    else
+      raise "airport full!"
+    end
+  end
+
+  def take_off(plane)
+    @planes.delete(plane)
```
jamesjoshuahill: Great use of the `delete` method on Array. What if the plane was not at the airport?

PR #1560, opened 2019-10-04
readme

zero201926:
friday challenge

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Airport Challenge
                 =  ===(_________)
 
 ```
-
+## Stories completed
 Instructions
 ---------
 
@@ -36,25 +36,25 @@ Task
 We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
 
 ```
-As an air traffic controller 
-So I can get passengers to a destination 
+As an air traffic controller
+So I can get passengers to a destination
 I want to instruct a plane to land at an airport
 
-As an air traffic controller 
-So I can get passengers on the way to their destination 
+As an air traffic controller
+So I can get passengers on the way to their destination
 I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent takeoff when weather is stormy 
+As an air traffic controller
+To ensure safety
+I want to prevent takeoff when weather is stormy
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent landing when weather is stormy 
+As an air traffic controller
+To ensure safety
+I want to prevent landing when weather is stormy
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent landing when the airport is full 
+As an air traffic controller
+To ensure safety
+I want to prevent landing when the airport is full
 
 As the system designer
 So that the software can be used for many different airports
@@ -73,7 +73,7 @@ In code review we'll be hoping to see:
 
 * All tests passing
 * High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc. 
+* The code is elegant: every class has a clear responsibility, methods are short etc.
 
 Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this weekend.
 


lib/air_control.rb
```diff
@@ -3,20 +3,24 @@ class Airport
...
     @plane_storage = []
+    @weather = Weather.new
   end
 
   def take_off
-    fail 'BAD WEATHER, ALL PLANES GROUNDED' if @condition == :stormy
+    # weather = false
+    # fail 'BAD WEATHER, ALL PLANES GROUNDED' if weather
 
     plane_storage.pop
```
eagle310: could have done it so you can take pff a specific plane rather than just the last one

lib/air_control.rb
```diff
@@ -3,20 +3,24 @@ class Airport
...
   end
 
   def land(plane_id)
-    fail 'BAD WEATHER, ALL PLANES GROUNDED' if @condition == :stormy
+    # weather = true
+    # fail 'BAD WEATHER, ALL PLANES GROUNDED' if weather
 
     fail 'Airport is at max capacity!!!ABORT!' if full?
```
eagle310: I like that you made a separate method to check if the airport is full

lib/air_control.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+end
+
+class Weather
+  attr_accessor :condition
+  @@conditions = [:good, :stormy, :good, :good, :good]
+  def initialize
+    @condition = @@conditions.sample
+  end
+end
```
jamesjoshuahill: Please could you put each class and the tests for each class in separate files:
```
.
├── lib
│   ├── airport.rb
│   ├── plane.rb
│   └── weather.rb
└── spec
    ├── airport_spec.rb
    ├── plane_spec.rb
    ├── spec_helper.rb
    └── weather_spec.rb
```

spec/air_control_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'air_control.rb'
+describe Airport do
+  subject(:airport) { described_class.new }
+  subject(:weather) { Weather.new }
```
jamesjoshuahill: The `subject` is the object under test, so it's confusing to see two subjects. You could use `let` instead.

Also, I think this `weather` is shadowed by a local `weather` variable in every test, so I don't think it's being used.

spec/air_control_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  let(:plane) { double :plane }
+  let(:plane_unqie) { double :"8457" }
```
jamesjoshuahill: What does `unqie` mean?

spec/air_control_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    airport.plane_storage << plane_unqie
+    airport.take_off(plane_unqie)
+    expect(airport.plane_storage).not_to include(plane_unqie)
+  end
+
+  it 'confirm a plane that has taken off is no longer at the airport.' do
+    weather = double(:weather, :condition => 'stormy')
+    airport.weather = weather.condition
+    airport.plane_status(plane)
+  end
```
jamesjoshuahill: There is no expectation in this `it`. What do you expect `airport.plane_status(plane)` to return?

lib/air_control.rb
```diff
@@ -0,0 +1,54 @@
...
+    plane_storage << plane_id
+  end
+
+  def full?
+    plane_storage.count >= capacity
+  end
+
+  def plane_status(plane_id)
+    puts "Is plane #{plane_id} still in the airport:"
+    print plane_storage.include?(plane_id)
```
jamesjoshuahill: This looks like prints used for debugging. Please remove them.

spec/air_control_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  it 'prevents takeoff when weather is stormy' do
+    weather = double(:weather, :condition => 'stormy')
+    airport.weather = weather.condition
+    expect { airport.take_off(plane) }.to raise_error("BAD WEATHER, ALL PLANES GROUNDED")
+  end
+
+  it 'prevents landing when weather is stormy' do
+    weather = double(:weather, :condition => 'stormy')
+    airport.weather = weather.condition
+    expect { airport.land(plane) }.to raise_error("BAD WEATHER, ALL PLANES GROUNDED")
```
jamesjoshuahill: This works exactly as the story says "to prevent landing when the weather is stormy". However, I wonder if the wording of the error message makes sense. Has the plane that's trying to land been grounded?

spec/air_control_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  subject(:airport) { described_class.new }
+  subject(:weather) { Weather.new }
+  let(:plane) { double :plane }
+  let(:plane_unqie) { double :"8457" }
+  before do
+    airport.plane_storage << plane
+  end
+  it 'instructs a plane to land at an airport.' do
+    weather = double(:weather, :condition => 'good')
+    airport.weather = weather.condition
```
jamesjoshuahill: Could you inject the weather test double instead of using an accessor to set it?
```ruby
weather = double(:weather, :condition => 'good')
airport = Airport.new(20, weather)
```

lib/air_control.rb
```diff
@@ -0,0 +1,54 @@
+class Airport
+  DEFAULT_CAPACITY = 20
+  attr_accessor :plane_storage, :capacity, :weather
```
jamesjoshuahill: I can't see a test that changes `plane_storage` and `capacity`. Could they be `attr_reader`?

lib/air_control.rb
```diff
@@ -0,0 +1,54 @@
...
+    plane_storage.count >= capacity
+  end
+
+  def plane_status(plane_id)
+    puts "Is plane #{plane_id} still in the airport:"
+    print plane_storage.include?(plane_id)
+  end
+
+  def stormy?
+    @weather == 'stormy'
```
jamesjoshuahill: In the constructor, `@weather` is an instance of `Weather` and here it's compared to a string. That doesn't make sense. How could you use the weather object here?

lib/air_control.rb
```diff
@@ -0,0 +1,54 @@
...
+  def initialize
+    @plane = rand(1000)
+  end
+end
+
+class Weather
+  attr_accessor :condition
+  @@conditions = [:good, :stormy, :good, :good, :good]
+  def initialize
+    @condition = @@conditions.sample
```
jamesjoshuahill: This is a good way to select a random condition 👍 But it looks like the weather never changes. How could you make the condition random each time it's checked?

jamesjoshuahill: I can't see any tests for this class. Please could you write one?

lib/air_control.rb
```diff
@@ -0,0 +1,54 @@
...
+    @weather == 'stormy'
+  end
+end
+
+class Plane
+
+  attr_accessor :plane
+
+  def initialize
+    @plane = rand(1000)
```
jamesjoshuahill: Does this need to be random?

spec/air_control_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    airport = Airport.new(50)
+    airport.capacity.times { airport.plane_storage << plane }
+    expect { airport.land(plane) }.to raise_error("Airport is at max capacity!!!ABORT!")
+  end
+end
+
+describe Plane do
+  it 'an instance of plane is any integer' do
+    expect(subject.plane).to be_an(Integer)
```
jamesjoshuahill: I think `plane` is a poor choice of name for this method. For example:
```ruby
plane = Plane.new
plane.plane  #=> 101
```
What could be a more accurate name for this number?

spec/air_control_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    airport.weather = weather.condition
+    airport.plane_storage << plane_unqie
+    airport.take_off(plane_unqie)
+    expect(airport.plane_storage).not_to include(plane_unqie)
+  end
+
+  it 'confirm a plane that has taken off is no longer at the airport.' do
+    weather = double(:weather, :condition => 'stormy')
+    airport.weather = weather.condition
+    airport.plane_status(plane)
```
jamesjoshuahill: Is `plane_status` supposed to be called with a plane or a plane ID?

PR #1561, opened 2019-10-04
Finished first run of task

Adam1824:


lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    hanger.count >= capacity
+  end
+
+  def stormy?
+    weather == 'stormy'
+  end
+end
+
+class Plane
+end
```
jamesjoshuahill: Please could you put each class in a separate file?

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
+class Airport
+  DEFAULT_CAPACITY = 20
+  DEFAULT_WEATHER = 'sunny'
+  attr_accessor :hanger, :capacity, :weather
```
jamesjoshuahill: Please could you try avoiding `attr_accessor` for attributes?
- Perhaps `hanger` could be read only?
- Perhaps `capacity` and `weather` could be private?

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'airport.rb'
+
+describe Airport do
+  it 'can land a plane' do
+    # Arrange
+    plane = Plane.new
```
jamesjoshuahill: Please could you try using a double for the plane? And create a separate spec file to describe the `Plane` class.

lib/airport.rb
Adam1824: done

spec/airport_spec.rb
Adam1824: done

lib/airport.rb
Adam1824: i tried to use reader instead but I need to write to them & if they arent private I cant call them in the tests

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+describe Airport do
+  it 'can land a plane when sunny' do
+    # Arrange
+    plane = double(:plane)
+    weather = double(:weather, :status => 'sunny')
+    # Act
+    subject.weather = weather.status
+    subject.land_plane(plane)
+    # Assert
+    expect(subject.hanger).to include(plane)
```
jamesjoshuahill: I really like the way you split out Arrange, Act, Assert

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'airport.rb'
+
+describe Airport do
+  it 'can land a plane when sunny' do
+    # Arrange
+    plane = double(:plane)
+    weather = double(:weather, :status => 'sunny')
```
josiepollard: I like the use of doubles for making the plane and setting the weather

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  it 'can prevent take off when weather is stormy' do
+    # Arrange
+    plane = double(:plane)
+    weather = double(:weather, :status => 'stormy')
+    # Act
+    subject.weather = weather.status
+    # Assert
+    expect { raise subject.take_off_plane(plane) }.to raise_error 'Unsuitable take off conditions.'
```
josiepollard: you have written this piece of code all one one line but still makes a lot of sense

PR #1562, opened 2019-10-04
Tobias Young Pull

TobiasBenjaminYoungCGI:
Toby

### DIFF
--- a/lib/airplane.rb
+++ b/lib/airplane.rb
@@ -1,0 +1,5 @@
+class Airplane
+  def stormy?
+    true
+  end
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,24 @@
+require_relative 'airplane'
+class Airport
+
+  DEFAULT_CAPACITY = 20
+  attr_reader :plane
+  attr_accessor :airport_capacity
+
+  def initialize(airport_capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = airport_capacity
+  end
+
+  def takeoff
+    fail 'No planes in airport' if @planes.empty?
+
+    @planes.pop
+  end
+
+  def land(plane)
+    fail 'airport is full' if @planes.count >= airport_capacity
+
+    @planes << plane
+  end
+end
--- a/spec/airplane_spec.rb
+++ b/spec/airplane_spec.rb
@@ -1,0 +1,4 @@
+require 'airplane'
+describe Airplane do
+  it { is_expected.to respond_to :stormy? }
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,23 @@
+require 'airport'
+
+describe Airport do
+
+  it { is_expected.to respond_to :takeoff }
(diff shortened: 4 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,24 @@
...
+  attr_reader :plane
+  attr_accessor :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def takeoff
+    fail 'No planes in airport' if @planes.empty?
```
EarthWormGym: Nice use of one line syntax.

spec/airport_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  describe 'takeoff' do
+    let(:plane) { double :plane }
+    it 'plane takes off correctly' do
+      plane = double(:flying? => true)
+      subject.land(plane)
+      takeoff_plane = subject.takeoff
+      expect(takeoff_plane).to be_flying
```
EarthWormGym: Good use of doubles and simple variable names making it easy to read for other developers.

spec/airport_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'airport'
+
+describe Airport do
+
+  it { is_expected.to respond_to :takeoff }
```
EarthWormGym: No way of telling which plane is which so can't tell that the plane that has taken off is now no longer in the hanger just knows that a plane has left.

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'airplane'
+class Airport
+
+  DEFAULT_CAPACITY = 20
+  attr_reader :plane
+  attr_accessor :capacity
```
jamesjoshuahill: I can't see a test that changes the value of capacity after an airport is created. Would `attr_reader` be enough?

lib/airport.rb
jamesjoshuahill: Great example of a guard clause 👍

spec/airport_spec.rb
jamesjoshuahill: This expectation tests the double you created on line 31, which is always flying. Please could you test that the plane has been removed from the list of planes in the airport?

jamesjoshuahill: Also, currently `takeoff` will remove the last plane that landed. How could you improve it so that you can instruct a specific plane to take off from the airport?

spec/airplane_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'airplane'
+describe Airplane do
+  it { is_expected.to respond_to :stormy? }
```
jamesjoshuahill: Do you think reporting the weather should be the Airplane's responsibility?

lib/airplane.rb
```diff
@@ -0,0 +1,9 @@
+class Airplane
+  def flying?
+  end
```
jamesjoshuahill: I can't see any tests for this method. Please could you add one?

PR #1564, opened 2019-10-04
pull request

luke-jones-1:


lib/plane.rb
```diff
@@ -0,0 +1,25 @@
+# require '/Users/student/projects/airport_challenge/lib/airport.rb'
+
+class Plane
+
+  attr_reader :flight_status
```
Will-Bath: I like the use of flight status - has good potential for satisfying new user stories

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  let(:plane) {double :plane}
+
+  describe '#add_plane' do
+
+    it {is_expected.to respond_to(:add_plane)}
+
+    # it {is_expected.to respond_to()}
+
+    it 'confirms a plane is in the airport after it lands' do
```
Will-Bath: Good descriptors for tests

spec/plane_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'spec_helper'
+require 'plane'
+
+describe Plane do
```
Will-Bath: It's a little difficult to follow what the tests themselves are doing, particularly with having the airport double return true for some methods - it would benefit from some comments and more lengthy descriptions.

spec/plane_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    it 'changes flight status when instructed to land' do
+      subject.land(airport)
+      expect(subject.flight_status).to eq('landed')
+    end
+
+    it 'changes flight status when instructed to land' do
+      subject.take_off(airport)
+      expect(subject.flight_status).to eq('flying')
+    end
```
Will-Bath: Having these test descriptions be different would be more helpful

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+    @weather_stormy = false
+  end
+
+  def add_plane(plane)
+    fail 'weather is too stormy for landing' if weather_stormy?
+    fail 'airport is full so plane cannot land' if full?
```
jamesjoshuahill: Great use of guard clauses in this method 👍

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    @weather_stormy = true
+  end
+
+  private
+
+  def weather_stormy?
+    @weather_stormy
+  end
+
+  def full?
```
jamesjoshuahill: Good use of a private method 👍

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
+require '/Users/student/projects/airport_challenge/lib/airport.rb'
```
jamesjoshuahill: Please avoid using absolute paths with `require` because it makes your code less portable. This throws an error on my machine because my user is not called `student`.

jamesjoshuahill: Also, is the `Airport` class used in this file? If not, then you should remove the unnecessary require.

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
...
+    airport.remove_plane(itself)
+    @flight_status = 'flying'
+  end
+
+end
+
+# plane = Plane.new
+# airport = Airport.new
+# plane.land(airport)
+# p plane.take_off(airport)
```
jamesjoshuahill: Please remove debugging code. If you want to keep this as a sort of feature test then move it into another file.

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+describe Airport do
+
+  let(:plane) {double :plane}
+
+  it '' do
+    expect(Airport).to receive(:new).with(10)
+    airport = Airport.new(10)
+    # expect(airport.planes).to eq(10)
+  end
```
jamesjoshuahill: Please name this `it`.

Also, I think this is testing Ruby. How could you test that the capacity is 10?

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  it 'defines the capacity, planes and weather stormy arrays and variables.' do
+     airport = described_class.new(6)
```
jamesjoshuahill: The indentation looks off on this line. Please keep it consistent so your code is easier to read.

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    expect(airport.capacity).to eq(6)
+    expect(airport.planes).to eq([])
```
jamesjoshuahill: It's a good rule of thumb to have one expect per `it`, so I recommend splitting these out into separate `it`s.

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    expect{ raise airport.weather_stormy?}.to raise_error NoMethodError
```
jamesjoshuahill: Testing for some method not to be there is a bit odd. You should only test public behaviour and there is no need to test that private methods are private.

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  describe '#add_plane' do
+
+    it {is_expected.to respond_to(:add_plane)}
+
+    it 'will receive plane as argument' do
+      is_expected.to receive(:add_plane).with(plane)
+      subject.add_plane(plane)
+    end
```
jamesjoshuahill: Again, I think this testing Ruby. If you want to test that the method accepts 1 argument you could use:
```ruby
it { is_expected.to respond_to(:add_plane).with(1).argument }
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      expect(subject.planes).to include(plane)
+    end
+
+    it 'wont allow a plane to land if weather is too stormy' do
+      subject.weather_is_stormy
+      expect { raise subject.add_plane(plane)}.to raise_error 'weather is too stormy for landing'
+    end
+
+    it 'wont allow a plane to land if airport is full' do
+      (subject.capacity).times { subject.add_plane(plane) }
```
jamesjoshuahill: These round brackets are not necessary. When chaining methods you can leave them out:
```ruby
subject.capacity.times { ... }
```

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  attr_accessor :flight_status
+
+  def initialize(status='undefined')
+    @flight_status = status
+  end
+
+  def land(airport)
+    fail 'this plane has aleady landed' if flight_status == 'landed'
+    airport.add_plane(itself)
```
jamesjoshuahill: This works. I'd never seen `itself` used before so I just had to look it up! No bad thing 😄 It is more common to see `self` used to refer to the instance of a class.

spec/plane_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      subject.land(airport)
+    end
+
+  end
+
+  describe '#flight_status' do
+
+    it {is_expected.to respond_to(:flight_status)}
+
+    it {expect(subject.flight_status).to eq('undefined')}
```
jamesjoshuahill: Starting a plane in an `'undefined'` status feels like something is missing or unclear. When a plane arrives onto the air traffic controller's screen what status do you think it would be in?

PR #1565, opened 2019-10-04
Committing project files

zt3h:


lib/airport.rb
```diff
@@ -1,28 +1,32 @@
...
+    @capacity = capacity
     @planes = []
   end
 
   def takeoff(plane)
     p "Clear for takeoff :)"
-    @planes.delete(plane) #delete plane from array
+    @planes.delete(plane) # delete plane from array
     plane.take_off && plane.flying
   end
```
joshblackmoor: Nice method but to improve maybe instead of p to output you could use puts as P is mainly used for debugging

lib/plane.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def take_off
+    @flight = true #
+  end
+
+  def landed
+    @flight = false #
+  end
+
```
joshblackmoor: Good use of an instance variable

lib/airport.rb
jamesbingham: Thanks for feedback I will implement this change now.

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
+require 'plane'
```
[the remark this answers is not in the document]
jamesbingham: Yes your right not sure why this is added but has been removed now!

lib/plane.rb
```diff
@@ -0,0 +1,20 @@
+class Plane
+  #This class has the planes status of flight
+
+  def initialize
+    @flying = true
```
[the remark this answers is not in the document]
jamesbingham: This was a typo as it should all be "flying", corrected.

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+    plane.landed
+  end
+
+  def plane_count
+    @planes.count
+  end
+
+  def airport_full?
+    if @planes.length >= capcity
+      p "Airport is full try again later"
```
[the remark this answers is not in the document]
jamesbingham: A raise statement should do which I have added once I've implemented all the changes I will push.

spec/airport_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+require 'plane'
+
+describe Airport do
+  it "can allow a plane to takeoff" do
+    airport = Airport.new
+    plane = Plane.new
+
+    airport.takeoff(plane)
+    expect(airport.plane_count).to eq 0
+  end
```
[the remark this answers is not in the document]
jamesbingham: I have added airport.land(plane) before the takeoff to make sure the plane has landed before I test that the plane has taken off.

lib/plane.rb
```diff
@@ -0,0 +1,20 @@
+class Plane
```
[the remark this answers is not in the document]
jamesbingham: Added some tests for the Plane spec!

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+    @planes << plane # add plane to array
+    plane.landed
+  end
+
+  def plane_count
+    @planes.count
+  end
+
+  def airport_full?
+    raise "The airport is currently full" if @planes.length >= capcity
```
jamesjoshuahill: This use of `raise` looks good. However, I can't see a test for this method.

I would like you to try test-driving this behaviour by:
- delete the `airport_full?` method
- write a failing test for landing a plane when the airport is full
- make it pass
- refactor your code

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+  attr_reader :capacity
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def takeoff(plane)
+    p "Clear for takeoff :)"
```
jamesjoshuahill: Please remove prints used for debugging.

spec/plane_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+describe Plane do
+
+  it "can check if a plane's status is flying" do
+    subject.flying
+    expect(subject.isFlying).to eq(true)
+  end
+
+  it "can check if a plane is taking off" do
+    subject.take_off
+    expect(subject.isFlying).to eq(true) #this will return true if the plane is already flying (need to fix)
```
jamesjoshuahill: Do you need both the `take_off` and `flying` methods?

jamesbingham: I do not I guess I can remove the take_off method in plane.rb and replace it with the flying method as they both return true.

lib/airport.rb
jamesbingham: I have done the above and run the new test without airport_full which failed then added it back and works if the .times method is below 10

spec/plane_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'plane'
+
+describe Plane do
+  it "can check if a plane's status is flying" do
+    subject.flying
+    expect(subject.isFlying).to eq(true)
```
jamesjoshuahill: This method `isFlying` works, but it doesn't follow the Ruby naming conventions:
- methods and variables are snake_case
- methods that return true or false end with `?`, for example: `[].empty?`

Please could you change this method name to follow these conventions?

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def takeoff(plane)
+    @planes.delete(plane) # delete plane from array
+    plane.flying
```
jamesjoshuahill: This looks like it works, but I can't see a test for this behaviour. How could you test that the plane is flying after take off?

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def land(plane)
+    airport_full?
+
+    @planes << plane # add plane to array
+    plane.landed
```
jamesjoshuahill: This also looks like it works, but I can't see a test for this behaviour. How could you test that the plane is not flying after landing?

spec/plane_spec.rb
jamesbingham: Hi I have added the above to my code

spec/airport_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    expect(airport.plane_count).to eq(1)
+  end
+
+  it "can check to see when a plane lands if the airport is full" do
+    airport = Airport.new
+    plane = Plane.new
+
+    10.times { airport.land(plane) }
+
+    expect(airport.land(plane))
```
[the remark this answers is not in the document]
jamesbingham: Added this method after the expect statement

lib/airport.rb
jamesbingham: Added this test to expect flying variable to equal true once plane has taken off.

lib/airport.rb
jamesbingham: I have also added this test to airport spec :)

jamesjoshuahill:
Hi @jamesbingham, this updates look promising, but `rspec` is failing with a syntax error. Please could you make sure all the tests pass?

jamesjoshuahill:
> Hi, I fixed these syntax errors and also missed adding a attr_reader on the plane class all of it works now 100% coverage. Do I have enough tests to have enough evidence for this module?

Hi @jamesbingham, yes this is good evidence of your testing skill with 100% coverage. Well done! 😄

PR #1566, opened 2019-10-04
doing user story 3

EarthWormGym:
currently completing 3rd user story.

### DIFF
--- a/lib/Plane.rb
+++ b/lib/Plane.rb
@@ -1,0 +1,5 @@
+require_relative "air_traffic_controller"
+
+class Plane
+  
+end
--- a/lib/air_traffic_controller.rb
+++ b/lib/air_traffic_controller.rb
@@ -1,0 +1,24 @@
+require_relative "Plane"
+
+class AirTrafficController
+  attr_accessor :plane , :hanger
+
+  def initialize
+    @hanger = []
+  end
+
+  def plane_landing(plane)
+    @hanger << plane
+  end
+
+  def plane_taking_off(plane)
+    @hanger.each do |e| if e == plane
+      @hanger.delete_at(e)
+    end
+    end
+  end
+
+  def hanger_full
+
+  end
+end
--- a/spec/air_traffic_controller_spec.rb
+++ b/spec/air_traffic_controller_spec.rb
@@ -1,0 +1,20 @@
+require "air_traffic_controller"
+require "Plane"
+
+describe AirTrafficController do
+  it "intructs a plane to land" do
+    plane = Plane.new
+
+    instruct_landing = subject.plane_landing(plane)
+
+    expect(subject.hanger).to include(plane)
+  end
+
+  it "plane taking off" do
+    plane = Plane.new
+
+    plane_leaving = subject.plane_taking_off(plane)
+
+    expect(subject.hanger).to_not include(plane)
+  end
+end
--- a/README.md
+++ b/README.md
@@ -13,6 +13,16 @@ Airport Challenge
 
 ```
 
+##Stories Completed
+
+As an air traffic controller
+So I can get passengers to a destination
+I want to instruct a plane to land at an airport
+
+As an air traffic controller
+So I can get passengers on the way to their destination
+I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
+
 Instructions
 ---------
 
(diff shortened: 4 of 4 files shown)


spec/air_traffic_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+require "Plane"
+
+describe AirTrafficController do
+  it "intructs a plane to land" do
+    subject = AirTrafficController.new(10)
+
+    plane = double(:plane, full?: true)
+    subject.capacity.times { subject.plane_landing(plane) }
+
+    expect{ subject.plane_landing(plane) }.to raise_error "hanger full"
```
TobiasBenjaminYoungCGI: I like how you separated Arrange, Assert and Action.

lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,27 @@
+require_relative "Plane"
+
+class AirTrafficController
+  attr_accessor :plane , :hanger , :capacity
```
TobiasBenjaminYoungCGI: Using Attr_Accessor is smart way of programming, initialising your variables for the whole class

lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def plane_landing(plane)
+    raise "hanger full" if full?
+    @hanger << plane
+  end
+
+  def plane_taking_off(plane)
+    @hanger.each do |e| if e == plane
+      @hanger.delete_at(e)
```
TobiasBenjaminYoungCGI: Textbook use of an each loop, great for tracking.

spec/air_traffic_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "air_traffic_controller"
+require "Plane"
+
+describe AirTrafficController do
+  it "intructs a plane to land" do
```
TobiasBenjaminYoungCGI: Spelling error in your it statement.

lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+    end
+  end
+
+  def full?
+    @hanger.count >= capacity
+  end
+end
```
TobiasBenjaminYoungCGI: Could have been a private method as it's only used within this class.

lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    @hanger = []
+  end
+
+  def plane_landing(plane)
+    raise "hanger full" if full?
+    @hanger << plane
+  end
+
+  def plane_taking_off(plane)
+    @hanger.each do |e| if e == plane
```
jamesjoshuahill: This line is harder to read. I suggest moving the `if` onto the next line.

spec/air_traffic_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect{ subject.plane_landing(plane) }.to raise_error "hanger full"
+  end
+
+  it "plane taking off" do
+    plane = Plane.new
+
+    plane_leaving = subject.plane_taking_off(plane)
+
+    expect(subject.hanger).to_not include(plane)
+  end
```
jamesjoshuahill: In this test scenario, the plane has not landed at the airport, so how can it take off?

spec/air_traffic_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "air_traffic_controller"
+require "Plane"
```
jamesjoshuahill: If you used a `double` for all the planes then you could remove this require.

lib/air_traffic_controller.rb
jamesjoshuahill: The array method [delete_at](https://ruby-doc.org/core-2.6.5/Array.html#method-i-delete_at) accepts the index of an element. What is `e` in this loop?

lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,27 @@
+require_relative "Plane"
```
jamesjoshuahill: I can't see anywhere that the `Plane` class is used in this file. Please remove this unused require.

lib/Plane.rb
```diff
@@ -0,0 +1,5 @@
+require_relative "air_traffic_controller"
```
jamesjoshuahill: Please remove requires that are not used.

lib/air_traffic_controller.rb
jamesjoshuahill: Do these need to be `attr_accessor` or would `attr_reader` be enough?

lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class AirTrafficController
+  attr_accessor :plane , :hanger , :capacity
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
```
jamesjoshuahill: I can't see a test for the default capacity. For example:
```ruby
it 'has a default capacity' do
  air_traffic_controller = AirTrafficController.new
  
  expect(air_traffic_controller.capacity).to eq(10)
end
```

lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,31 @@
+
+class AirTrafficController
+  attr_reader :plane, :hanger, :capacity
```
jamesjoshuahill: I can't see `@plane` instance variable used anywhere. Is it needed, or can it be removed?

lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def plane_taking_off(plane)
+    counter = -1
+    @hanger.each do |e|
+      counter += 1
+      if e == plane
+        @hanger.delete_at(counter)
+      end
+    end
```
jamesjoshuahill: This code works well. I wonder if there might be someway to refactor it? If you like the for loop style with an index then you might like to try using [.each_with_index](https://ruby-doc.org/core-2.6.5/Enumerable.html#method-i-each_with_index)

jamesjoshuahill:
Great work, well done @TobiasBenjaminYoungCGI

PR #1567, opened 2019-10-04
Airport Challenge

lukegittins01:
Completed tasks 1, 2 and 3

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,10 @@ Airport Challenge
 
 ```
 
+Stories Completed
+
+
+
 Instructions
 ---------
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,31 @@
+class Airport
+
+attr_reader :capasity
+
+  def initialize
+    @airport_planes = []
+    @capasity = 20
+  end
+
+  def land(plane)
+    fail 'Airport is full' if airport_full
+    @airport_planes << plane
+  end
+
+  def list_of_planes
+    return @airport_planes
+  end
+
+  def take_off(plane)
+    @airport_planes.pop
+  end
+end
+
+  def airport_full
+    @airport_planes.count >= capasity
+  end
+
+
+class Plane
+
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,28 @@
+require 'airport'
+
+describe Airport do
+  it { is_expected.to respond_to :land }
+  it { is_expected.to respond_to :list_of_planes }
+
+
+  it 'lands a plane' do
+    plane = Plane.new
+    airport = Airport.new
+    airport.land(plane)
+    expect(airport.list_of_planes).to include(plane)
+  end
+
+  it 'takes off' do
+    plane = Plane.new
+    airport = Airport.new
+    airport.take_off(plane)
+    expect(airport.list_of_planes).not_to include(plane)
+  end
+
+  it 'raises an error if airport is full' do
+    plane = Plane.new
+    airport = Airport.new
+    airport.capasity.times { airport.land Plane.new }
+    expect { airport.land Plane.new }.to raise_error 'Airport is full'
+  end
+end


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def initialize
+    @airport_planes = []
+    @capasity = 20
+  end
+
+  def land(plane)
+    fail 'Airport is full' if airport_full
+    @airport_planes << plane
+  end
```
2Blagden: nice shorthand (<<)

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  it { is_expected.to respond_to :land }
+  it { is_expected.to respond_to :list_of_planes }
+
+
+  it 'lands a plane' do
+    plane = Plane.new
+    airport = Airport.new
+    airport.land(plane)
+    expect(airport.list_of_planes).to include(plane)
+  end
```
2Blagden: I think you could've used doubles here (same as mine)

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
+class Airport
+
+attr_reader :capasity
```
jamesjoshuahill: Spelling: `capasity -> capacity`

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  #----- this was my attempt at stormy weather test -----
+  #def weather
+    #return "stormy"
+  #end
+
+class Plane
+
+end
```
jamesjoshuahill: Please put each class in a separate file.

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    plane1 = double(:plane)
+    airport = Airport.new
+    airport.take_off(plane1)
+    expect(airport.list_of_planes).not_to include(plane1)
+  end
+
+  it 'raises an error if airport is full' do
+    plane1 = double(:plane)
+    airport = Airport.new
+    airport.capasity.times { airport.land plane1 }
```
jamesjoshuahill: It looks like the same plane can land many times. How could you prevent that?

README.md
```diff
@@ -13,6 +13,30 @@ Airport Challenge
...
+As an air traffic controller
+To ensure safety
+I want to prevent landing when the airport is full
+
+Stretch challenges 
+
+----- ATTEMPTED -----
+As an air traffic controller
+To ensure safety
+I want to prevent takeoff when weather is stormy
```
jamesjoshuahill: Good documentation of the stories you have completed and attempted 👍

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+
+attr_reader :capasity
+
+  def initialize
+    @airport_planes = []
+    @capasity = 20
+  end
+
+  def land(plane)
+    fail 'Airport is full' if airport_full
```
jamesjoshuahill: Good use of a guard clause 👍

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def take_off(plane)
+    #----- this was my attempt at stormy weather test -----
+    #fail 'Weather is stormy' if weather == "stormy"
+    @airport_planes.pop
+  end
+end
+
+  def airport_full
```
jamesjoshuahill: Think about the name of this method:
- `airport.airport_full` has a stutter in the name
- in Ruby methods that return `true` or `false` usually end with `?`

What would be a better name?

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def list_of_planes
+    return @airport_planes
+  end
+
+  def take_off(plane)
+    #----- this was my attempt at stormy weather test -----
+    #fail 'Weather is stormy' if weather == "stormy"
+    @airport_planes.pop
```
jamesjoshuahill: What if there was more than one plane at the airport, which one would take off?

lib/airport.rb
lukegittins01: I have changed the variable name in the recent push to capacity.

lib/airport.rb
lukegittins01: I have made sure that each class in in a separate file, the plane class can now be seen under the plane.rb file.

spec/airport_spec.rb
lukegittins01: I have made a variable called 'flying' and this is set to true when the plane has taken off and when it has landed the variable is set to false as it is on the ground. I have then made another test for that to test if take-off is rejected when the plane is already flying.

README.md
lukegittins01: Thank you 👍

lib/airport.rb
lukegittins01: Thank you 👍

lib/airport.rb
lukegittins01: In the recent push I have changed the method name to "full?" It makes a lot more sense to have it like that instead of "airport_full" as I see what you mean by it repeating itself 👍

lib/airport.rb
lukegittins01: In my code the array is popping the last element in the array so it will be popping the most recent plane that came in

PR #1568, opened 2019-10-04
heading

eagle310:


spec/air_port_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  it 'lands a plane' do
+    new_plane = Plane.new
+    subject.land(new_plane)
+    expect(subject.return_plane 0).to eq new_plane
+  end
+
+  it 'doesnt land plane if capacity is full' do
+    4.times do
+      subject.land(Plane.new)
+    end
```
zero201926: could have done the time on one line.

spec/air_port_spec.rb
```diff
@@ -0,0 +1,30 @@
+require './air_port'
+require './air_plane'
```
zero201926: I like that you divided your code into separate files.

air_port.rb
```diff
@@ -0,0 +1,25 @@
+class Airport
+  def initialize
+    @limit = 4
+    @planes = []
```
zero201926: could attr_reader this.

spec/air_port_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+describe Airport do
+  it { should respond_to(:land) }
+
+  it 'lands a plane' do
+    new_plane = Plane.new
+    subject.land(new_plane)
+    expect(subject.return_plane 0).to eq new_plane
```
jamesjoshuahill: The `return_plane` method gives a way to check if a plane is at the airport or not, but it exposes the way the airport stores planes. To use `return_plane` I have to know the index, which limits its use.

Could you use the `planes` method instead of `return_plane`?

air_port.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def return_plane(num)
+    plane = planes[num]
+    return plane
+  end
+
+  def full?
+    planes.count >= @limit
+  end
```
jamesjoshuahill: I can't see a test that uses this method, so perhaps you could make this method private? Or write a test for it?

spec/air_port_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  it 'correct plane takes off' do
+    airport = Airport.new
+    new_planes = []
+    3.times do
+      new_plane = Plane.new
+      new_planes.push(new_plane)
+      subject.land(new_plane)
+    end
+    expect(subject.take_off new_planes[2]).to eq [new_planes[0], new_planes[1]]
```
jamesjoshuahill: This test is great because it tests that the plane has been removed from the airport. Could you refactor it using the include matcher, so that it's easier to read?
```ruby
expect(list).not_to include(element)
```

air_port.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  attr_reader :planes
+
+  def initialize
+    @limit = 4
+    @planes = []
+  end
+
+  def land(plane)
+    fail 'Airport at capacity' if full?
```
jamesjoshuahill: Great use of a guard clause 👍

jamesjoshuahill:
The convention for Ruby source code is to be in a `lib` directory. Please could you move `air_plane.rb` and `air_port.rb` into a new directory called `lib`?

jamesjoshuahill:
Thank you for addressing the comments I made. This looks great 😄

PR #1571, opened 2019-10-18
Initial PR

khan2597:


lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+        @planes_arr = []
+    end
+
+    def land_at_airport(plane)
+        @planes_arr << plane
+    end
+
+    def take_off(plane)
+        @planes_arr.pop
+    end
```
david-mears: Rubocop had a helpful message for this method:

> lib/airport.rb:14:18: W: Lint/UnusedMethodArgument: Unused method argument - plane. If it's necessary, use _ or _plane as an argument name to indicate that it won't be used. You can also write as take_off(*) if you want the method to accept any arguments but don't care about them.
>     def take_off(plane)
>                  ^^^^^

PR #1573, opened 2019-10-18
Airport Challenge - TDD

Stupot83:
<br>Airport


lib/plane.rb
```diff
@@ -0,0 +1,23 @@
...
+class Plane
+  def initialize
+    landed!
+  end
+
+  def flying?
+    @flying
+  end
+
+  def landed!
```
david-mears: Great idiomatic use of `!` and `?` for Ruby method names. Maybe `land!` is more idiomatic than `landed!` since you use the method to land the plane.

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      expect(plane).to receive(:landed!)
+      airport.instruct_landing(plane)
+      expect(plane).to receive(:take_off!)
+      airport.instruct_take_off(plane)
+      expect(airport.hangar).to be_empty
+    end
+
+    it "if the hangar is empty a plane cannot take off and there is a message" do
+      allow(airport).to receive(:good_weather?).and_return(true)
+      expect(lambda { airport.instruct_take_off(plane) }).to raise_error(RuntimeError, "You ain't getting on no plane fool")
```
david-mears: you don't need to use a lambda - just use curlies:
```expect{ airport.instruct_take_off }.to raise_error etc```

File: lib/plane.rb [hunk 3]
```diff
@@ -0,0 +1,23 @@
...
+    @flying = false
+  end
+
+  def take_off!
+    @flying = true
+  end
+
+  def airborne?
```
david-mears: I'd query whether the difference between `flying?` and `airborne?` is clear from the method name

PR #1584, opened 2019-10-19
First pull request for airport challenge

topherb99:


lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+  attr_reader :capacity
+  def initialize
+    @planes = []
+    @stormy = [false, true, false, false].sample
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def confirm_departure(plane)
+    return "#{plane} is still grounded" if plane.grounded_in?(self)
+
```
robertwoolley99: Different approach to me - aircraft pushes and pulls to airport array within aircraft class.

lib/plane.rb
```diff
@@ -0,0 +1,31 @@
...
+class Plane
+
+  attr_reader :grounded
+  def initialize
+    @grounded = false
+  end
+
+  def land(airport)
+    return 'Too stormy to land' if airport.stormy == true
+    return 'Airport is full' if airport.full?
```
robertwoolley99: Should the plane make a decision  on landing re full airport or should the airport reject it ?

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { double(:plane) }
+  it 'Stores planes' do
+    subject.planes << plane
+    expect(subject.planes[0]).to eq plane
+  end
+
+  it 'returns cannot depart if plane has not departed' do
```
robertwoolley99: "Still present" rather than "Cannot depart" ?

PR #1585, opened 2019-10-19
Updated Pull request.

robertwoolley99:
Contains an update of the README.md file as well as use of private method inside the airport.rb

### DIFF
--- a/lib/aircraft.rb
+++ b/lib/aircraft.rb
@@ -13,9 +13,12 @@ class Aircraft
     @registration = registration
   end
 
-  def land_aircraft(airport_code)
+  def land_aircraft(airport_code = "")
     return fail "aircraft landed - cannot land again"\
      unless @status == "in flight"
+
+    return fail "need to specify airport to land"\
+     unless airport_code != ""
 
     @status = "landed"
     @airport = airport_code
--- a/lib/aircraft.rb
+++ b/lib/aircraft.rb
@@ -1,13 +1,17 @@
 class Aircraft
   DEFAULT_STATUS = "landed"
+  DEFAULT_AIRPORT = "none"
 
-  def initialize(status = DEFAULT_STATUS)
+  def initialize(status = DEFAULT_STATUS, airport = DEFAULT_AIRPORT)
     @status = status
+    @airport = airport
   end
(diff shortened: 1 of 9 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+    @aircraft_present.delete(departing_plane)
+
+    return departing_plane
+  end
+end
+
+private
+
+  def full?
+    if @airport_capacity > @aircraft_present.length
```
topherb99: does this not mean that the airport will then hold 21 planes as opposed to 20?

lib/atc.rb
```diff
@@ -0,0 +1,22 @@
+class Atc
+
+  def auth_landing(aircraft, airport, airport_code)
+    met_office = Weather.new
+    return fail "Poor weather. Airport closed." \
+    unless met_office.forecast == "fine"
```
topherb99: I like the implementation of met_office but is it necessary after a refactor?

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'airport'
+require 'aircraft'
+
+describe Airport do
```
topherb99: Maybe try to have a context 'When stormy', 'When weather fine' in order to make rspec more readable from command line

lib/aircraft.rb
```diff
@@ -0,0 +1,39 @@
+class Aircraft
```
topherb99: require relevant files in order to run the files without rspec

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  it 'lands a plane - into array. Unit test' do
+    heathrow = Airport.new "LHR"
+    b747 = Aircraft.new("in flight")
+    expect(heathrow.aircraft_arriving(b747)).to be true
+  end
+  it 'departs a plane - out of array. Unit test' do
+    heathrow = Airport.new "LHR"
+    b747 = Aircraft.new("landed", "LHR", "G-BNLO")
```
topherb99: Try using doubles instead of instantiating new objects each test

PR #1586, opened 2019-10-20
First pull request

Liatmoss:


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'plane'
+
+class Airport
+
+  DEFAULT_CAPACITY = 5
+
+  attr_accessor :capacity
+  attr_reader :plane
+  attr_accessor :weather
```
emmavanoss: Are these accessible for the sake of tests? If the code doesn't need them, maybe reconsider your tests so you don't need to make these states accessible.

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'airport'
+
+describe Airport do
+
+  describe '#weather' do
+    context 'change take off and landing conditions depending on weather'
+    it 'should not allow take off and landing in stormy weather' do
+      allow(subject).to receive(:weather) { 2 }
```
emmavanoss: I'm not sure what this is testing. Does it just check the stubbing works? I think it might be more helpful to test the behaviour that happens after you stub the weather.

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'airport'
+
+describe Airport do
+
+  describe '#weather' do
```
emmavanoss: I like how you've organised the tests. It makes the rspec output much easier to understand.

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+    fail 'Cannot land in stormy weather' if @weather == 2
+    fail 'Airport is full' if @landed.count >= capacity
+    fail 'That plane is already in the airport' if two_planes(plane)
+
+    plane.in_airport = true
+    @landed << plane
+    return @landed
+  end
+
+  def two_planes(plane)
```
emmavanoss: I found this method name a bit confusing. Also, I'm not sure that you need the method - can you just evaluate @landed.include? as part of the guard clause in land?

emmavanoss: Having said that maybe it's a good use of the SRP!

spec/airport_spec.rb
emmavanoss: This applies to a few tests below as well

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  end
+
+  def take_off(plane)
+    fail 'Cannot take off in stormy weather' if @weather == 2
+    fail 'There are no planes in the airport' if @landed == []
+
+    @landed.pop
+    plane.in_airport = false
+    return plane
```
emmavanoss: Why do you need to return plane here?

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+  DEFAULT_CAPACITY = 5
+
+  attr_accessor :capacity
+  attr_reader :plane
+  attr_accessor :weather
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @landed = []
+    @weather = rand(1..2)
```
emmavanoss: I think the spec asked for sunny weather to be more likely than cloudy. There are a few ways to do this

lib/airport.rb
Katerina-codes: Hi @emmavanoss, this sounds like it might be useful feedback but your comment is ambiguous, could you elaborate on what you mean?

PR #1587, opened 2019-10-20
Pull request for weekend challenge

SevenSecrets:
hopefully this is all green

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,7 +1,12 @@
 # /lib/airport
 
+require './lib/plane'
+require './lib/weather'
+
 class Airport
-  
+
+  attr_reader :planes
+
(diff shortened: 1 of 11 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    raise "Cannot land in stormy weather" if @weather.stormy?
+    raise "Airport is full" if @planes.length >= @capacity
+
+    @planes << plane
+  end
+
+  def takeoff(plane)
+    raise "Cannot take off in stormy weather" if @weather.stormy?
+
+    @planes.pop # change so you can control which plane?
```
amybalmforth: Wasn't sure what the comment was about so maybe better to remove it

PR #1588, opened 2019-10-20
Elise Aston airport weekend project

eliseaston:


lib/weather.rb
```diff
@@ -0,0 +1,8 @@
+class Weather
+
+  def forecast
+    forecast = rand(1..5)
+    forecast == 5 ? "stormy" : "sunny"
```
mattfreeman-london: This is very neatly expressed.

spec/airport_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    it "a new airport can be created with a custom capacity" do
+      sml_airport = Airport.new(5)
+      expect(sml_airport.capacity).to eq(5)
+    end
+  end
+
+  context "landing a plane" do
+
+    it "lands a plane in an airport" do
+      expect(airport.land(plane, today)).to eq("The plane has safely landed")
```
mattfreeman-london: As per rspec result, this line isn't working correctly. The test doesn't always pass.

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+require './lib/plane.rb'
+require './lib/weather.rb'
+
+class Airport
+  attr_reader :grounded_planes, :capacity
+
+  def initialize(capacity=20)
```
mattfreeman-london: use of 'magic number' - should be a separate contant variable DEFAULT_CAPACITY = 20

PR #1589, opened 2019-10-20
Weekend airport challenge

emmavanoss:


spec/plane_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+describe Plane do
+  context 'when first initialized' do
+    it 'raises an error when take_off is called' do
+      plane = Plane.new
+      airport = Airport.new
+      allow(airport).to receive(:stormy?).and_return false
+      expect { plane.take_off(airport) }.to raise_error 'Plane is not in the airport'
+    end
+
+    context 'when not stormy' do
```
Liatmoss: Think about where to use 'context' within tests if it should be placed there or if a 'describe' or more detailed 'it' can be used instead.

README.md
```diff
@@ -13,24 +13,76 @@ Airport Challenge
...
-4. Complete the following task:
+How to use the program
+----
 
-Task
+The following irb transcript should give an idea of how the program is
+used.
+
+```
+$ irb
```
Liatmoss: This is really good. Very detailed instructions on the expected outcome.

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  it 'returns "false" when stormy? is called if rand = 0.4' do
+    airport = Airport.new
+    allow(airport).to receive(:rand).and_return 0.4
+    expect(airport).not_to be_stormy
+  end
+
+  context 'with default capacity' do
+    it 'does not raise error when 10 planes are landed' do
```
Liatmoss: Using a test such as:
expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
can simplify these tests so that you don't need to test for a change in capacity

PR #1590, opened 2019-10-20
Weekend Challenge 1 - Airport Challenge

fahus:


spec/weather_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'weather'
+
+describe Weather do
+  let(:weather) { Weather.new }
+
+  it "is stormy outside" do
+    allow(weather).to receive(:randomise) { 80 }
```
jonesandy: Reading the test files it's not obvious why you need to return those certain numbers. To understand this I had to read the rb code file. Suggest maybe returning a more verbose response from that randomise method?

spec/airport_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      subject.take_off(plane)
+      expect(plane).to eq(plane)
+    end
+
+    it 'does not allow planes to take off in stormy weather' do
+      allow(subject).to receive(:stormy?).and_return true
+      expect { subject.take_off(plane) }.to raise_error "Too stormy for take off"
+    end
+    it 'does not allow a plane to take off if not in airport' do
+      allow(subject).to receive(:stormy?).and_return false
```
jonesandy: This line seems to be repeated a lot in the tests. Could this be specified inside a context block?

jonesandy:
Good naming conventions. Code was easy to read and follow. SRP followed with no extra methods.

PR #1592, opened 2019-10-20
Robert Rosiji : Airport Challenge

bibbycodes:


README.md
```diff
@@ -13,28 +13,12 @@ Airport Challenge
```
CodeKrakken: Very through README. Hadn't considered including code extracts other than standard setup instructions. Not sure if this is something you should change or something I should adopt.

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+class Airport
+  DEFAULT_CAPACITY = 40
+  attr_reader :hangar, :capacity, :airport_id
+  attr_writer :full, :hangar
+  
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @stormy = rand(9) <= 2
+    @full = false
```
CodeKrakken: Separate indicator for @full rather than always checking number of present aircraft could open a loophole - however in some cases this sort of thing is obviously essential!

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def in_airport?(plane)
+    return @hangar.include?(plane)
+  end
+
+  def stormy?
+    return @stormy
+  end
+
+  def full?
```
CodeKrakken: These are nice neat methods - as far as I can see there are no wasted words.

lib/feature_test.rb
```diff
@@ -0,0 +1,22 @@
...
+boeing_747.land(Heathrow)
+puts boeing_747.airport # => <airport ID>
+
+puts Heathrow.hangar # => <Plane:0x00007fc738826890> , <Plane:0x00007fc738826840>
+puts boeing_747.flying # => false
+
+boeing_747.land(Heathrow) # => Plane already in airport! (RuntimeError)
+
+Heathrow.allow_take_off(boeing_737)
+puts Heathrow.hangar # => <Plane:0x00007fc738826890>
```
CodeKrakken: It is not immediately clear to me how this ties in with the rest of the code. If it is a test, should it not be in the spec file?

lib/plane.rb
```diff
@@ -0,0 +1,38 @@
+class Plane
+  attr_accessor :airport, :flying
+
+  def initialize
+    @flying = true
+    @airport = ""
+  end
+
+  def flying?
```
CodeKrakken: Is this method necessary if the attribute can be accessed anyway?

lib/plane.rb
```diff
@@ -0,0 +1,38 @@
...
+    return @flying
+  end
+
```
CodeKrakken: Is this method necessary if the attribute can be accessed anyway?

PR #1593, opened 2019-10-20
tests passed

elfiyang16:
Didn't update README yet.

- Problems during coding
## No consideration of encapsulation

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,39 @@
+# require 'plane'
+
+class Airport
+  attr_reader :capacity
+  DEFAULT_CAPACITY = 30
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @plane_collection = []
+    @capacity = capacity
+    @weather = ["stormy", "sunny", "sunny", "sunny"].sample
+  end
+
+  def land(plane)
+    raise "Caution! It's stormy!" if stormy?
+    raise "Caution! Airport is full!" if full?
+    raise "The same plane already landed" if @plane_collection.include?(plane)
+
+    @plane_collection.push(plane)
+    @plane_collection
+    # plane.switch_status ? cannot do...
+  end
+
+  def departure(plane)
+    raise "Caution! It's stormy!" if stormy?
+    raise "The plane is not landed, so can't take off" unless @plane_collection.include?(plane)
+
+    @plane_collection.delete(plane)
+    @plane_collection
+  end
+
+  private
+  def stormy?
+    @weather == "stormy" ? true : false
+  end 
+
+  def full?
+    @plane_collection.count >= @capacity ? true : false
+  end
+end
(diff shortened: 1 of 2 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { double :plane }
+
+  context "0. initial set-up" do
```
munakh: Really good use of descriptive test names for easier readability.

spec/airport_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      expect(new_capacity.capacity).to eq 30
+    end
+  end
+  #======================================================
+  context "1. test landing function" do
+    it "a. let a plane land at the airport" do
+      expect(subject).to respond_to(:land).with(1).argument
+    end
+
+    it "b. prevent a plane from landing in stormy weather" do
```
munakh: Could do with use of context to differentiate tests in good and bad weather to save you from repeating the state of the weather in every test.

spec/airport_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    it "a. prevent a plane from taking off in stormy weather" do
+      allow_any_instance_of(Airport).to receive(:stormy?).and_return true
+      expect { subject.departure(plane) }.to raise_error "Caution! It's stormy!"
+    end
+
+    it "c. (when not stormy), confirm the plane is no longer in the airport " do
+      allow_any_instance_of(Airport).to receive(:stormy?).and_return false
+      subject.land(plane)
+      expect(subject.departure(plane)).not_to include(plane)
```
munakh: Same as above, think this should be subject.departue(plane) and then expect(subject.planes_collection).not_to include(plane)?

spec/airport_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    it "d. (when not stormy), the plane cannot take off unless previously landed " do
+      allow_any_instance_of(Airport).to receive(:stormy?).and_return false
+      expect { subject.departure(plane) }.to raise_error "The plane is not landed, so can't take off"
```
munakh: Needs extra line to say subject.departure(plane) and then the expect... line to test that a departed plane cannot depart again. This test is currently raising error every time a plane is departed, I think.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    raise "Caution! It's stormy!" if stormy?
+    raise "The plane is not landed, so can't take off" unless @plane_collection.include?(plane)
+
+    @plane_collection.delete(plane)
+    @plane_collection
+  end
+
+  private
+  def stormy?
+    @weather == "stormy" ? true : false
```
munakh: The 'true : false' is redundant, i.e. it doesn't need to be written out. The '?' includes this code in itself.

PR #1596, opened 2019-10-20
First pull request

CodeKrakken:
Build complete as per instructions but no chance to run Rubocop yet

### DIFF
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -15,7 +15,8 @@ class Plane
 # I want to prevent landing when the airport is full 
 
   def land(airport)
-    airport.hangar << self unless airport.stormy == true || airport.hangar.count >= airport.capacity
+    airport.hangar << self unless airport.stormy == true || 
+    airport.hangar.count >= airport.capacity
   end
   
 # As an air traffic controller 
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,43 @@
+require_relative 'airport.rb'
+
+class Plane
+
+# As an air traffic controller 
+# So I can get passengers to a destination 
+# I want to instruct a plane to land at an airport
+
+# As an air traffic controller 
+# To ensure safety 
+# I want to prevent landing when weather is stormy
+
+# As an air traffic controller 
+# To ensure safety 
+# I want to prevent landing when the airport is full 
+
+  def land(airport)
+    airport.hangar << self unless airport.stormy == true || airport.hangar.count >= airport.capacity
+  end
+  
+# As an air traffic controller 
+# So I can get passengers on the way to their destination 
+# I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
+
+# As an air traffic controller 
+# To ensure safety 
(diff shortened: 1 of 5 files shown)


lib/plane.rb
```diff
@@ -0,0 +1,54 @@
...
+# As an air traffic controller 
+# To ensure safety 
+# I want to prevent landing when the airport is full 
+
+  def land(airport)
+    return "Already landed" if @flying == false
+    airport.hangar << self unless airport.stormy == true || 
+    airport.hangar.count >= airport.capacity
+    @flying = false
+  end
```
bibbycodes: Could have used Error throwing here with `fail` statements

lib/plane.rb
```diff
@@ -0,0 +1,54 @@
...
+    return "Already in flight" if @flying == true
+    return "Plane not in hangar" if !airport.hangar.include? self 
+    airport.hangar.pop unless airport.stormy == true
+    @flying = true
+    confirm(airport)
+  end
+  
+  def confirm(airport)
+    "Take off successful"
+  end
```
Katerina-codes: What is the purpose of this method? This seems like a mix of concerns:

1. Communicating to the user
2. The actual logic for your program to run

Let's chat about this in person during coaching hours :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  
+# As an air traffic controller 
+# To ensure safety 
+# I want to prevent landing when the airport is full 
+
+  context "When hangar is not full"
+  
+  it "allows planes to land when hangar is not full" do
+    san_jose = Airport.new
+    san_jose.stormy = false
```
Katerina-codes: Though I am enjoying the quirkyness of the names for these variables, I would argue that this is not helpful naming and it would make your code more confusing then it should be in the long term. In short, I would stick to explict, clear and conventional names (plane, airport etc.)

spec/plane_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+    it "can only take off from airport it is in" do
+      jimbo = Plane.new(true)
+      san_jose = Airport.new
+      peninsula_de_osa = Airport.new
+      san_jose.stormy = false
+      peninsula_de_osa.stormy = false
+      jimbo.land(san_jose)
+      expect(jimbo.take_off(peninsula_de_osa)).to eq "Plane not in hangar"
+    end
+end
```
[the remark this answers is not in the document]
Katerina-codes: @bibbycodes can you be more specific with this feedback? How can you be clearer with regards to what you mean by "long-winded"?

spec/plane_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+    it "takes off from airport" do
+      jimbo = Plane.new
+      san_jose = Airport.new
+      san_jose.stormy = false
+      jimbo.land(san_jose)
+      jimbo.take_off(san_jose)
+      expect(san_jose.hangar.include? jimbo).to eq false
+    end
+    
+    it "confirms" do
```
Katerina-codes: What does this test confirm?

PR #1599, opened 2019-10-20
Submitting pull request for Airport Challenge

munakh:


spec/airport_spec.rb
```diff
@@ -3,46 +3,86 @@
...
-  describe 'landing planes' do 
-    # Testing plane landing
-    it 'lets a plane land' do
-      expect(subject).to receive(:land)
-      subject.land plane
+  subject(:airport) { described_class.new(weather, 20) }
+  let(:plane) { double :plane, landed: nil, taken_off: nil }
+  let(:weather) { double :weather }
+
+  context 'when weather is good' do
```
elfiyang16: Really good practice to using #let, #subject(){} and #before clause to set up the  prerequisite conditions, so that you avoided repeating the same assertion in  each individual test.

lib/airport.rb
```diff
@@ -1,30 +1,53 @@
+# frozen_string_literal: true
+
 require_relative 'plane'
+require_relative 'weather'
```
elfiyang16: like you create independent class weather & plane

lib/airport.rb
```diff
@@ -1,30 +1,53 @@
...
-  attr_reader :capacity, :planes
-  DEFAULT_CAPACITY = 1
-  
-  def initialize(capacity = DEFAULT_CAPACITY)
-    @planes = []
+  attr_accessor :capacity, :planes
+  DEFAULT_CAPACITY = 20
+
+  def initialize(weather, capacity = DEFAULT_CAPACITY)
+    @weather = weather
```
elfiyang16: If there's a weather class already, maybe you want to move @weather out of the airport class and initialise it in weather class, and using attr_reader to access it?

lib/airport.rb
```diff
@@ -1,30 +1,53 @@
...
-    
-  def take_off(_plane)
-    planes.pop
-    puts "#{_plane} has taken off" if !@planes.include?(_plane)
-  end       
-    
-  def stormy?
+
+  def take_off(plane)
+    taking_off_checks(plane)
```
elfiyang16: Really good practice to move all guard clause in another independent method to keep this method with single responsibility.

lib/airport.rb
```diff
@@ -1,30 +1,53 @@
...
+
+  private
+
   def full?
     planes.count >= capacity
   end
-  
+
+  def stormy?
+    @weather.stormy?
```
elfiyang16: maybe reprioritise this with the weather class

spec/airport_spec.rb
```diff
@@ -3,46 +3,86 @@
...
-    it 'gives an error when trying to land at full airport' do
-      subject.capacity.times { subject.land(plane) }
-      expect{subject.land(plane)}.to raise_error("Airport is full")
+  end
+
+  context 'when airport is full' do
+    subject(:default_airport) { described_class.new(weather) }
+    it 'gives an error when trying to land' do
+      allow(weather).to receive(:stormy?).and_return(false)
+      described_class::DEFAULT_CAPACITY.times { default_airport.land(Plane.new) }
```
elfiyang16: Good practice of not exposing real numbers but referencing the constant from the test class

PR #1600, opened 2019-10-20
Airport Challenge - Alex Clifford

Clifford2910:


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  attr_reader :planes
+  attr_accessor :capacity
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
```
ingridbjarman: Perhaps a slightly more obvious method name would be beneficial?

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    raise "cannot land the plane: airport at capacity!" if full?
+    raise "cannot land the plane: weather is stormy!" if stormy?
+    @planes << plane
+  end
+
+  def take_off(plane)
```
ingridbjarman: (Same as above)

PR #1602, opened 2019-10-20
Ingrid Bjarman - Airport Challenge

ingridbjarman:


spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'airport'
+
+describe 'airport' do
+  describe '#plane_landing' do
+    it 'responds to a method allowing planes to land' do
+      expect(Airport.new).to respond_to :plane_landing
+    end
+    it 'prevents landing when the weather is stormy' do
+      airport = Airport.new
```
Clifford2910: maybe look at using 'subject' to refactor airport = Airport.new which has been used multiple times.

lib/weather.rb
```diff
@@ -0,0 +1,8 @@
+class Weather
+  def stormy?
+    # weather = ["stormy", "clear skies", "light rain"]
+    # if weather.sample == "stormy"
+    # end
+    stormy = true
+  end
+end
```
Clifford2910: is the Weather class necessary if you have the same method 'stormy?' in the Airport class

PR #1603, opened 2019-10-20
Weekend challenge week 1

sarar0:


lib/plane.rb
```diff
@@ -0,0 +1,14 @@
...
+require_relative 'weather'
+
+class Plane
+  attr_reader :flying
+
+  def left_airport
+    @flying = true
+  end
+
+  def flying?
```
Ben230: This method is the same as attr_reader :flying

lib/plane.rb
```diff
@@ -0,0 +1,14 @@
+require_relative 'airport'
```
Ben230: Not using airport.rb or weather.rb in this class so no need to require them

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+require_relative 'plane'
```
Ben230: no need to require

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'airport'
+require 'plane'
+require 'weather'
+
+describe Airport do
+  let(:plane1) { Plane.new }
+  let(:today) { Weather }
```
Ben230: what does this mean

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  subject(:airport) { Airport.new(10) }
```
Ben230: airport already has a default capacity of 10 so there is no need to call it here

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  it 'has a default capacity' do
+    allow_any_instance_of(today).to receive(:sunny).and_return true
```
Ben230: .

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    Airport::DEFAULT_CAPACITY.times do
+      subject.land(plane1)
+    end
+    expect { subject.land(plane1) }.to raise_error 'Airport is full'
+  end
+
+  describe '#land' do
+    it { is_expected.to respond_to(:land).with(1).argument }
+    it 'returns the plane that has landed' do
+      allow_any_instance_of(today).to receive(:sunny).and_return true
```
Ben230: repeated code

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  it 'has a default capacity' do
+    allow_any_instance_of(today).to receive(:sunny).and_return true
+    Airport::DEFAULT_CAPACITY.times do
+      subject.land(plane1)
+    end
+    expect { subject.land(plane1) }.to raise_error 'Airport is full'
+  end
+
+  describe '#land' do
+    it { is_expected.to respond_to(:land).with(1).argument }
```
Ben230: this test is redundant

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      allow_any_instance_of(today).to receive(:sunny).and_return true
+      expect(subject.land(plane1)).to eq(plane1)
+    end
+    it 'raises an error when weather is stormy' do
+      allow_any_instance_of(today).to receive(:sunny).and_return false
+      expect { subject.land(plane1) }.to raise_error 'Weather is stormy'
+    end
+    it 'raises an error when airport is full' do
+      allow_any_instance_of(today).to receive(:sunny).and_return true
+      10.times do
```
Ben230: use the constant for default capacity instead of the number 10, in case you change the capacity and then this test would fail

PR #1606, opened 2019-10-21
Weekend challenge 1 - airport

Ben230:


spec/airport_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+describe Airport do
+  let(:plane) { Plane.new }
+  let(:airport) { Airport.new }
+  let(:weather) { double :weather, stormy?: false }
+
+  context '#land' do
+    it 'lands a plane to an airport' do
+      airport.land(plane, weather)
+      expect(airport.planes[0]).to eq plane
```
sarar0: Testing using a variable from the code (array planes is being used through the attribute reader).

spec/airport_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    it 'doesnt land if its stormy' do
+      weather = double(:weather, stormy?: true)
+      expect { airport.land(plane, weather) }.to raise_error 'cannot land while stormy'
+    end
+
+    it 'doesnt land if airport is full' do
+      Airport::CAPACITY.times { airport.land(Plane.new, weather) }
```
sarar0: Plane.new different from :plane defined above?

lib/plane.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
+  def initialize
+    @flying = false
+  end
+  attr_accessor :flying
```
sarar0: Is this an accessor because it needs to be modified?

spec/airport_spec.rb
sarar0: Yes!

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'weather'
+require_relative 'plane'
+
+class Airport
+  CAPACITY = 3
+  attr_reader :planes
+  attr_accessor :capacity
+  def initialize
```
Katerina-codes: Some space would be nice here :)

PR #1609, opened 2019-10-23
Move weather stories to the end

jamesjoshuahill:
## Why

This follows https://github.com/makersacademy/airport_challenge/pull/1554.

We want to reduce the completion anxiety experienced by apprentices on this challenge at the end of the TDD module.

We noticed that many of them get stuck on the first weather story that introduces random behaviour and the README suggests using a stub to override the random weather.

## What

This change moves the two weather stories to the end of the challenge. We think this will help learners consolidate the skill of test driving objects and methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,19 +46,19 @@ I want to instruct a plane to take off from an airport and confirm that it is no
 
 As an air traffic controller 
 To ensure safety 
-I want to prevent takeoff when weather is stormy 
-
-As an air traffic controller 
-To ensure safety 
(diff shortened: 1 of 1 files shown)


jamesjoshuahill:
> I think I sometimes use poorly ordered user stories as a lesson that the world will not always give you requirements in the easiest or best order.

I agree there is value in this. It's something I've noticed apprentices work out for themselves on later challenges, but haven't seen it on the first.

jamesjoshuahill:
Thank you for reviewing @alicelieutier 😄

PR #1611, opened 2019-11-04
first commit stories completed

tomhowarth76:


lib/airport.rb
```diff
@@ -0,0 +1,20 @@
+class Airport
+ def initialize
+   @allplanes = []
+ end
+ def planes
+   @allplanes
+ end
```
david-mears: It's slightly counterintuitive to have a method `.planes` return the attribute `@allplanes`. Newcomers to the codebase would expect these to be aligned.

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+ def capacity
+   capacity = 10
```
david-mears: How could you make the capacity specific to the individual airports, such that two airports could have different capacities?

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+ end
+ def plane_land
+   fail 'Airport Full' if @allplanes.count >= capacity
+   @allplanes << Plane.new
+ end
+
+ def plane_take_off
+   @allplanes.slice!(0)
+   @allplanes
```
david-mears: This method always takes off the first plane in the array. How would you let the method take off any plane specified?

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
+require './lib/airport'
+describe Airport do
+  it 'recognises the existence of an Airport' do
+    airport = Airport.new
+  end
+  it 'instruct a plane to land at an airport' do
+    is_expected.to respond_to :plane_land
```
david-mears: For both these `respond_to` tests, they're asking merely whether the method exists for the object you're testing. Here for example you're testing if `airport` knows of `plane_land`. That's a bit different from testing what the method does ('instruct a plane to land at an airport').

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+  it 'instructs a plane to take off' do
+    is_expected.to respond_to :plane_take_off
+  end
+  it 'knows number of planes in an airport' do
+    airport = Airport.new
+    airport.plane_land
+    expect(airport.planes.count).to equal(1)
```
david-mears: Good test.

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+  it 'confirms a plane is no longer in the Airport after take off' do
+    airport = Airport.new
+    airport.plane_land
+    expect(airport.plane_take_off.count).to eq(0)
```
david-mears: This test relies on the fact that `plane_take_off` returns you `@allplanes` - which can be counted. Since that return behaviour isn't the main function of `plane_take_off` (taking off is), someone could want to change it in the future, which would break this test. A more robust test would separate out the taking off and the counting of planes.

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+  it 'It prevents landing when the airport is full #ten planes for now' do
+    airport = Airport.new
+    10.times { airport.plane_land }
+    expect { airport.plane_land}.to raise_error 'Airport Full'
+  end
+  it 'has a recognised capacity that is specific to a case of the airport' do
+   airport = Airport.new
+   expect(airport.capacity).to be_a_kind_of(Integer)
```
david-mears: This test shows that `.capacity` gives an integer - but not that the capacity can be tailored to any number for any particular airport.

PR #1612, opened 2019-11-04
Creating Pull Request

selinaerhabor:


spec/airport_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    airport = Airport.new
+    expect(airport).to respond_to :take_off
+  end
+  it 'confirms plane has left the airport' do
+    airport = Airport.new
+    expect(airport).to respond_to :confirm_plane_left
+  end
+  it 'prevents landing when airport is full' do
+    plane = Plane.new
+    expect(plane).to raise_error 'Prevent Landing - Airport is full'
```
david-mears: Close - the error I get from rspec says
```
expected Exception with "Prevent Landing - Airport is full" but was not given a block
```
What's a 'block' in Ruby? Check the documentation for the `raise_error` matcher and see what's different.

spec/airport_spec.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/airport.rb'
+describe Airport do
+  it 'instructs a plane to land at airport' do
+    plane = Plane.new
+    airport = Airport.new
+    expect(airport).to respond_to :land
```
david-mears: This line tests whether `airport` has a method called `land`. That's a bit different from testing whether the plane is instructed to land. The same applies to your other `respond_to` tests: they're asking merely whether the method exists for the object you're testing.

PR #1613, opened 2019-11-04
add stories completed

LeighNLeal:


david-mears:
rspec was broken when I ran it - it complained of a syntax error. This suggests that you forgot to run rspec before committing the new test.
Here's some relevant documentation for the error: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/respond-to-matcher

Looking at your methods, quite a lot of them return something - a new Plane, an instance variable, a boolean - but it doesn't like these return values were test-driven, because there aren't tests which expect e.g. `true` when you call `Plane.new.landed?`

PR #1615, opened 2019-11-04
Have updated the README file with heading Stories Completed

Riz1702:


README.md
```diff
@@ -12,6 +12,8 @@ Airport Challenge
                 =  ===(_________)
 
 ```
+""😀Stories Completed😀""
+
```
LucyMatejka: Heading added successfully to READ.ME

spec/plane_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+    plane = Plane.new
+    expect(plane).to be_kind_of(Plane)
+  end
+  context "when instructed" do
+    it 'can land a plane safely at the airport' do
+      plane = Plane.new
+      airport = AirPort.new
+      expect(plane).to receive(:land).with(airport)
+      plane.land(airport)
+    end
```
david-mears: This test probably doesn't really test for whether a plane "can land safely at the airport", because the test passes, but you haven't written the `land` method. You can prove this by seeing that this test still passes even if you delete the `land` method.
So I wonder if you can tell what this test is actually doing.
In TDD, you write the test, and then watch it fail. In this case, I think you may have forgotten to test whether it was possible for the test to fail, since it always passes no matter how bare the Plane class is.

lib/plane.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
+  def plane
```
david-mears: Methods are conventionally a verb that describes the action of the method. So I'm not sure what this method is for.

README.md
```diff
@@ -13,6 +13,14 @@ Airport Challenge
...
 ```
 ""😀Stories Completed😀""
+1.1 USER STORY 1 :
+
+As an air traffic controller
+
+NOUNS: AIR TRAFFIC CONTROLLER, PASSENGERS, DESTINATION, PLANE, AIRPORT
+VERBS : Can_Get_Passengers, To_A_Destination, Instruct_A_Plane, To_Land
```
david-mears: I'd say 'instruct' and 'land' were the key verbs here.

PR #1616, opened 2019-11-04
Readme changes

JNuckley:


docs/airport.rb
```diff
@@ -1,21 +1,22 @@
+
+
```
savagevince: Use 'plane' rather than airplane for reading clarity when 'airport' is another variable?

docs/airport.rb
```diff
@@ -1,21 +1,22 @@
...
+attr_accessor :capacity 
 
-  def initialize
+  def initialize(capacity = DEFAULT_CAPACITY)
     @airplanes = []
+    @capacity = capacity
   end
+end
+
+
```
savagevince: I like that you have framed the expected code in comments

docs/airport.rb
```diff
@@ -1,21 +1,22 @@
...
 
-  # def capacity
-  # end
-  #
   # def land_plane
```
savagevince: Again, for clarity, the method should not necessarily include '_plane', as the act of landing or take-off is implicitly going to refer to the instance variable 'airplane' or 'plane'

spec/airport_spec.rb
```diff
@@ -6,17 +6,25 @@
   airport = Airport.new
   expect(airport).to be_instance_of(Airport)
 end
+
 it 'is expected to respond to :airplanes' do
   airport = Airport.new
-  expect(airport).to respond_to(airplanes)
+  expect(airport).to respond_to(:airplanes)
```
savagevince: I think rubocop advises not to put the ':airplanes' in parentheses

david-mears:
You're right to mention in [this commit](https://github.com/makersacademy/airport_challenge/pull/1616/commits/a563eac3347754eaa32a9c3bf0172d25246809cf)'s message that the capacity should be tested for, otherwise it's not test-driven ;)

The test you included but commented out was certainly on the way there - it was testing whether the attribute called 'capacity' existed and was the right number. You could improve that test by trying out the capacity by landing more planes at the airport than are allowed.

Perhaps you'd then need to add more code to pass that test, not least including the functionality of landing and taking off planes from `@planes`. So I think it would have made more sense to start by implementing the first few user stories *before* doing the capacity story which depends on the earlier stories.

PR #1619, opened 2019-11-04
Updates README.md to include 'Stories Completed'.

SoniaQ:


spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  # it 'can respond to plane takes off' do
+  #   airport = Airport.new
+  #   expect(airport).to respond_to(:plane_takes_off)
+  # end
+
+  # it 'can confirm plane has taken off' do
+  #   airport = Airport.new
+  #   airport.plane_takes_off
+  #   expect(airport.plane_takes_off).to respond_to(:take_off_confirmed)
```
david-mears: This isn't quite what `respond_to` does. Objects 'respond' to methods (if they know of the method). But `airport.plane_takes_off` looks like a method call, not an object.

PR #1620, opened 2019-11-04
Add "Stories completed" heading

rahulshah7:


spec/airport_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'airport'
+
+describe Airport do
+  it { is_expected.to be_an_instance_of(Airport) }
+
+  describe '#land' do
+    it { is_expected.to respond_to(:land).with(1).argument }
+
+    context 'if plane is already landed' do
+      let(:landed_plane) { double("landed_plane", :landed => true) }
```
david-mears: great use of doubles here.

david-mears:
There are some great tests and use of doubles here.
It seems you skipped over the first couple of user stories? Your `land` method doesn't land anything. It seems more logical to start with the basic functionality before testing for errors and edge cases.

PR #1625, opened 2019-11-05
Pull request - SW

SimoneW2:


david-mears:
It's a shame you were slowed down by changing your mind on what class to lead with. Perhaps it doesn't particularly matter which class you start with, as you'll need both. It's good to see from your commit history that you are following TDD and leaving informative commit messages.

spec/plane_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "./lib/plane.rb"
+describe Plane do
+  it 'makes an instance of the plane class' do
+    plane = Plane.new
+    expect(plane).to be_instance_of(Plane)
+  end
+
+  it 'land_plane method respondes to plane class' do
```
david-mears: Really it's the Plane class that's responding to the 'land_plane' method. In other words, a plane is being sent the message 'land_plane' and we want to know that it can recognise that message.

lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
+  def land_plane
```
david-mears: You could rename this to `land` and not lose any readability.

spec/plane_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    plane = Plane.new
+    expect(plane).to respond_to(:land_plane)
+  end
+
+  it 'lands plane' do
+    plane = Plane.new
+    expect(plane.land_plane).to eql("land")
```
david-mears: I guess this test was at an early stage of development? A call to `land_plane` should probably produce a lasting effect on the plane, or on something, rather than returning some text.

PR #1629, opened 2019-11-07
Adding my airport challenge files

docela:


lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+  def capacity_change(capacity)
+    @cap = capacity
+  end
+
+  def land(plane)
+    if @airport.length < @cap
+      @airport << plane
+    else
+      raise RuntimeError, "The airport has reached capacity. This plane will be unable to land."
+    end
```
david-mears: Very clear. You could make this method even clearer by putting the fail condition on one line at the start of the method:
```ruby
raise RuntimeError "The airport has reached capacity. This plane will be unable to land." if @airport.length < @cap
@airport << plane
```
That way you get a shorter method and it's easier for the eye to jump to the main thing that the method is for (@airport << plane).

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def takeoff(plane)
+    @airport.slice!(-1)
+    puts "This plane has taken off"
+  end
```
david-mears: Why does `takeoff` take an argument `plane`?

lib/planes.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
+  attr_accessor :plane
```
david-mears: I'm not sure what this line is for?

lib/planes.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  def initialize
+    @plane = plane
```
david-mears: What error do you get when you run `Plane.new` and comment out the `attr_accessor`? Perhaps this class ought to have been test-driven... ;) If you inspect your new `Plane` by using `p` or irb, you can see that `@plane=nil` - so `plane` on line 5 is calling the attr_accessor.

I find your intentions here a little confusing - maybe this would have been helped by having some tests as documentation. So, when you create a new `Plane`, it has an instance variable `@plane` which stores something called `plane` on the new `Plane` (that word sounds weird now). I don't know what this is for - can you explain this choice?

spec/airport_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'airport'
+
+describe Airport do
+  it "creates an airport and lands planes" do
+    plane = Plane.new
+    expect(subject.land(plane)).to eq [plane]
```
david-mears: For both this test and the test of `takeoff`, I think Arrange Act Assert would make for better tests. You're acting on the world (by landing or taking off a plane), and then we want to separately check (assert) that the world changed. E.g. if I put some rubbish in a bin, I could say:
```ruby
describe Bin do
  it "takes in rubbish" do
    rubbish = Rubbish.new
    subject.put_in(rubbish)
    expect(subject.contents).to include(rubbish)
  end
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  it "instructs plane to take off and confirm it`s no longer there" do
+    plane = Plane.new
+    expect(subject.takeoff(plane)).to eq(nil)
+  end
+
+  it "prevents landing of planes when capacity is full" do
+    airport = Airport.new
+    plane = Plane.new
+    # expect { (subject.land(plane)) > airport.capacity }.to raise_error { RuntimeError }
+    expect { airport.land(plane) > :cap }.to raise_error { RuntimeError }
```
david-mears: What does `x > y` do? It compares `x` with `y` and then returns `true` or `false` depending on whether `x` is indeed `>` than `y`.

spec/airport_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  it 'allows the default airport capacity to be overridden' do
+    airport = Airport.new
+    # expect(:cap).to eq(:capacity_change)
+    # expect(:cap).to be(:capacity_change)
+    expect {
+      subject.capacity_change(capacity)
```
david-mears: How does ruby know what you mean by the argument `capacity` here?

david-mears: (Hint: it doesn't.)

PR #1630, opened 2019-11-16
My own solution

AndreaDiotallevi:


MY_README.md
```diff
@@ -0,0 +1,294 @@
...
+ => #<Airport:0x00007fcb1a8b0d20 @capacity=1, @weather=#<Weather:0x00007fcb1a8b0cf8>, @planes=[#<Plane:0x00007fcb1a900078 @landed=true>]> 
+2.6.0 :017 > airport_2.land(Plane.new)
+Traceback (most recent call last):
+        6: from /Users/student/.rvm/rubies/ruby-2.6.0/bin/irb:23:in `<main>'
+        5: from /Users/student/.rvm/rubies/ruby-2.6.0/bin/irb:23:in `load'
+        4: from /Users/student/.rvm/rubies/ruby-2.6.0/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
+        3: from (irb):71
+        2: from (irb):71:in `rescue in irb_binding'
+        1: from /Users/student/Code/makersacademy/airport_challenge/lib/airport.rb:15:in `land'
+RuntimeError (Cannot land as the airport is full)
```
dbacall: Good information, but not what README is meant for. Maybe include in another file.

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    @planes = []
+  end
+
+  def land(plane)
+    fail "Cannot land due to stormy weather" if weather.stormy?
+    fail "Cannot land as the airport is full" if full?
+    fail "Cannot land as the plane has already landed" if plane.landed?
+
+    plane.land
+    @planes.push(plane)
```
dbacall: Maybe @planes could be called something like @planes_landed, so a bit clearer to read.

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
+class Airport
+
+  DEFAULT_CAPACITY = 50
+
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
```
dbacall: Is there a way to lock the weather initialization? So no one can create something like Airport.new(30, "random string")

lib/weather.rb
```diff
@@ -0,0 +1,8 @@
+class Weather
+  OUTCOMES = [:sunny, :sunny, :sunny, :stormy]
```
dbacall: Would be hard to implement this method realistically. As it would be more realistic to have maybe 99 :sunny and 1 :stormy. And also in that case it wouldn't be readable.

spec/plane_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "plane"
+
+describe Plane do
+
+  describe "#landed?" do
+
+    it "should return a boolean" do
+      plane = Plane.new
+      expect([true, false]).to include plane.landed?
```
alicelieutier: Don't you initialise the plane to always start in the sky? You should be testing that landed is false when you have just created the class.

And then that it is true when you have called `land`
And then that it is false when you have called `land` and then `take_off`

spec/weather_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "weather"
+
+describe Weather do
+
+  describe "#stormy?" do
+
+    it "should return a boolean" do
+      weather = Weather.new
+      expect([true, false]).to include weather.stormy?
```
alicelieutier: You should fix the environment so that you can test both cases (search "stubbing randomness")

spec/plane_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "plane"
+
+describe Plane do
```
alicelieutier: In this whole class there is only one test per method.  Do these methods only ever do one thing?

PR #1631, opened 2019-11-16
Pull Request for Airport Challenge

samfolo:


lib/errors.rb
```diff
@@ -0,0 +1,11 @@
...
+  CURRENTLY_AIRBORNE = 'This plane is already airborne'
+  CURRENTLY_GROUNDED = 'This plane is already grounded'
+  NO_AIRPORT = 'This plane needs an airport to take off'
+
+  AT_CAPACITY = 'This airport is currently at capacity'
+  NOT_AT_AIRPORT = 'Target plane is not at this airport'
+  STORMY_WEATHER_ON_TAKEOFF = 'Takeoff forbidden due to bad weather conditions'
+  STORMY_WEATHER_ON_LANDING = 'Landing forbidden due to bad weather conditions'
+  NOT_A_PLANE = 'Foreign object attempting to land at airport'
+end
```
HarryMumford: Possibly store these in a module instead of a class? A module would do the same job

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
...
+  def take_off
+    raise Errors::NO_AIRPORT unless accounted_for
+    raise Errors::CURRENTLY_AIRBORNE if airborne 
+
+    @airborne = true
+    'plane has taken off'
+  end
+
+  def land
+    raise Errors::CURRENTLY_GROUNDED unless airborne or !accounted_for
```
HarryMumford: Just be wary of the difference in precedence when using <code> ||</code> or <code>or</code>

lib/sky.rb
```diff
@@ -0,0 +1,7 @@
+class Sky
+  attr_reader :planes_in_transit
+
+  def initialize
+    @planes_in_transit = []
+  end
+end
```
HarryMumford: Maybe have name your class "routes" instead of "sky" to better match a real life situation. Although at this stage the difference may not be apparent. If you want to add features based on routes later this would make "sky" redundant

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  it 'should commission flights to a known airspace' do
+    expect(subject.airspace).to be_an_instance_of Sky
+  end
+
+  context '#hangar' do
+    it 'should not contain airborne planes' do
+      expect(subject.hangar).to all_be_grounded_planes
+    end
+  end
```
HarryMumford: Could this situation actually occur?

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY, airspace = Sky.new)
+    @hangar = [] 
+    @capacity = capacity
+    @weather = weather_report
+    @airspace = airspace
+  end
+
+  def harbour_plane(plane)
+    raise Errors::NOT_A_PLANE unless plane.is_a? Plane
```
alicelieutier: What you are doing here is protective programming where you are checking that the argument is of the expected type. This is common in some programming languages (like C#) but is not a common or recommended practice in Ruby, which typically uses duck typing and polymorphism.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
+require 'weather'
+require 'plane'
+require 'sky'
```
alicelieutier: Why do you need to require these files and not the error file?

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
+require 'errors'
+
+class Plane
+  attr_accessor :airborne, :accounted_for
```
alicelieutier: You only use these attr_accessors in your tests. To me that means that you are changing the code in order to test it, which is usually a sign of testing state instead of testing behaviour.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def harbour_plane(plane)
+    raise Errors::NOT_A_PLANE unless plane.is_a? Plane
+    raise Errors::STORMY_WEATHER_ON_LANDING if weather == :stormy
+    raise Errors::AT_CAPACITY if hangar.length == capacity
+
+    plane.land
+    hangar.push plane
+    airspace.planes_in_transit.delete plane
```
alicelieutier: Does this actually work? 
There's an `attr_reader` on that variable in `Sky`, but no `attr_accessor`.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
+require 'weather'
+require 'plane'
+require 'sky'
+
+class Airport
+  include Weather
+  attr_reader :hangar, :capacity, :airspace
+  attr_accessor :weather
```
alicelieutier: You are making this writable for your tests. You should probably be using dependency injection here instead.

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
+require 'airport'
```
alicelieutier: Many of your tests are about state and implementation. What you really want is to test behaviour.

lib/airport.rb
samfolo: Fair enough; I won't assume other files are requiring on behalf of any other file moving forward

lib/airport.rb
samfolo: will check this one

PR #1634, opened 2019-11-17
My solution to the airport challenge

ffgi-es:


.gitignore
```diff
@@ -7,3 +7,4 @@ CoverallsRakefile
 analysis/**/*
 .gitignore-analysis
 
+rough_feature_test
```
alicelieutier: You could totally have pushed this - it's nice to see drafts too :)

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 20
+  attr_reader :name, :capacity
+
+  def initialize(name, opts = {})
```
alicelieutier: Replace `opts` with `options`

lib/plane.rb
```diff
@@ -0,0 +1,30 @@
...
+    raise InvalidStateError, "Not in the air" unless flying?
+    return false unless airport.accept self
+
+    @airport = airport
+    true
+  end
+
+  def take_off
+    raise InvalidStateError, "Can't take off in the air" if flying?
+    return false unless @airport.take_off_slot self
```
alicelieutier: This is a small comment, but it is generally not best practice to have a method that has a side effect as a conditional. Making this into 2 steps would clarify what happens:
```rb
take_off_attempt_success = @airport.take_off_slot
return false if !take_off_attempt_success
@airport = nil
 true

```

The other pattern you can use here, that would probably be more idiomatic is using errors and try/catch block.

spec/airport_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+  let(:plane2) { double(:plane, flying?: false) }
+  let(:weather) { double(:weather, stormy?: false) }
+
+  def bad_weather
+    allow(weather).to receive(:stormy?).and_return(true)
+  end
+
+  subject { Airport.new("Paris", weather: weather) }
+
+  before(:each) { allow(subject).to receive(:weather_report).and_return("sunny") }
```
alicelieutier: You should never stub a method of the class under test. So `allow(subject)...` is always a red flag.
Instead, you should be using what you prepared above, something like:
`allow(weather).to receive(:stormy?).and_return(false)`

spec/airport_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  describe "#new" do
+    it "should accept an optional argument for custom capacity of 30" do
+      airport = Airport.new("Barcelona", capacity: 30)
+      expect(airport.capacity).to eq 30
+    end
+  end
+
+  describe "#count" do
```
alicelieutier: Nice!

spec/weather_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'weather'
+
+describe Weather do
+  describe "#stormy?" do
+    it "should return either true or false" do
+      expect(subject.stormy?).to eq(true).or eq(false)
```
alicelieutier: Instead of this, you may want to stub randomness and expect a specific result.

alicelieutier (commented):
Good solution!
I especially like the feature tests, and the code is quite readable.
Here are a few comments:

spec/airport_spec.rb
ffgi-es: Yeah, I can see how that could be problematic. I believe that was a hangover from before I separated the Weather out into its own class as that method was actually completely removed in the end.

PR #1636, opened 2019-11-17
airport_challenge pull request

ajbacon:


spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    it "should raise an error if trying to take-off when already in flight" do
+      expect { subject.take_off(plane_in_flight, fair_weather) }.to raise_error "Plane already in flight"
+    end
+  end
+
+  describe "#contains_plane?" do 
+
+    it "should return false after a plane has taken off" do
```
denriquem: Does this still apply when there are multiple planes at same airport?

spec/plane_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  let(:airport_name) { double(:airport_name, planes: []) }
+
+  describe "initialize" do 
+    
+    it "should have a location when first inititalsed" do
+      plane1 = Plane.new(airport_name)
+      expect(plane1.location).to eq(airport_name)
+    end
+
+    it "should update the location of the plane when landing at a new airport" do
```
denriquem: Nice!

spec/plane_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  describe "#location" do
+    it "should give the location of the plane when requested" do 
+      plane1 = Plane.new(airport_name)
+      expect(plane1.location).to eq(airport_name)
+    end
+  end
+
```
denriquem: Really clever  that the airport where the plane is at has been accounted for.

spec/plane_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  describe "#land" do
+    it "should give the location of the plane when requested" do 
+      plane1 = Plane.new(airport_name)
+      plane1.land
+      expect(plane1::in_flight?).to eq(false)
+    end
+  end
+
+  # describe "#in_flight?" do
```
denriquem: Really like the approach of passing in an argument specifying the airport for determining location.  Is there a way to still do this when plane in flight? Possibly setting an instance of Airport to equal in flight????

spec/weather_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "weather"
+
+describe Weather do
```
denriquem: Potential to be more precise with random whether generation using method stubs. This would ,ake sure that the output is something you have specified rather than just assessing response to a method

lib/errors.rb
```diff
@@ -0,0 +1,20 @@
+class Errors
```
denriquem: Really nice to have very specific error messages. Is this approach something which we would continue to use in the future, or should we be relying on error messages raised from within guard clauses? The approach of manually creating error messages in a seperate folder may become really painstaking in a large programme. On the other hand its super informative here!

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+    @planes.include?(plane)
+  end
+
+  private
+
+  def full?
+    @planes.count >= @capacity
+  end
+
+  def empty?
```
denriquem: empty already a method in ruby and will return true if contents is nil

lib/weather.rb
```diff
@@ -0,0 +1,8 @@
+class Weather
+
+  def stormy?
+    range = 1..100
+    true if range.to_a.sample == 1
```
denriquem: Stormy is true 1/100 times? A couple of questions about this, but easier face to face.

lib/airport.rb
alicelieutier: Here you are redefining the method on an object of type airport, though, which is fine.

lib/errors.rb
alicelieutier: I would have the text of the error close to the place where the error is thrown, unless it's an error that is thrown in many places across the program (readability vs DRY)

lib/plane.rb
```diff
@@ -0,0 +1,22 @@
+class Plane
+
+  attr_reader :location 
+
+  def initialize(location)
+    @location = location
+    location.planes << self
```
alicelieutier: Pushing directly into an instance variable of a different class is not good practice. You may want to create an `add_plane` method to Airport or something similar.

PR #1637, opened 2019-11-17
first attempt

Hyan18:
was told to do frequent pull requests

### DIFF
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,2 @@
+class Plane
+end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,2 +1,21 @@
 class Plane
+
+  def initialize
+    @flying = true
+    # we assume upon creation the plane is in flight, alternatively could have
+    # a default airport the planes are stored in upon creation
+  end
+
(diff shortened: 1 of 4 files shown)


.lib/weather.rb
```diff
@@ -0,0 +1,15 @@
+class Weather
```
alicelieutier: You have two weather classes 🤔 ?

lib/plane.rb
```diff
@@ -0,0 +1,25 @@
+class Plane
+  attr_reader :airport
+
+  def initialize
+    @flying = true
+    # we assume upon creation the plane is in flight, alternatively could have
+    # a default airport the planes are stored in upon creation
```
alicelieutier: This comment should be in the documentation (README) not in the code.

.lib/weather.rb
Hyan18: Sorry could you elaborate on this?
I think I only have one which is inside the weather.rb file

alicelieutier: You have a `.lib/weather.rb` and a `lib/weather.rb`.

PR #1639, opened 2019-11-17
Airport challenge attempt

dbacall:
Realised after about 4 hours that I hadn't been doing any commits, which is why there are so few.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,11 +2,14 @@ require "./lib/plane.rb"
 
 class Airport
   
-  attr_accessor :capacity
+  attr_accessor :capacity, :airport_name
 
   DEFAULT_CAPACITY = 30
 
-  def initialize( capacity = DEFAULT_CAPACITY )
+  DEFAULT_NAME = :airport
+
+  def initialize( airport_name = DEFAULT_NAME, capacity = DEFAULT_CAPACITY )
+    @airport_name = airport_name
     @capacity = capacity
     @planes_at_airport = []
   end
@@ -15,6 +18,7 @@ class Airport
     storm_warning
     raise "Airport full, plane cannot land yet" if full?
     raise "Plane already landed" if @planes_at_airport.include?(plane)
+    plane.land_at(@airport_name)
     @planes_at_airport << plane
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,44 @@
+require "./lib/plane.rb"
+
+class Airport
+  
+  attr_accessor :capacity
+
+  DEFAULT_CAPACITY = 30
+
+  def initialize( capacity = DEFAULT_CAPACITY )
+    @capacity = capacity
+    @planes_at_airport = []
+  end
+
+  def land(plane)
+    storm_warning
+    raise "Airport full, plane cannot land yet" if full?
+    raise "Plane already landed" if @planes_at_airport.include?(plane)
+    @planes_at_airport << plane
+  end
+
+  def take_off(plane)
+    storm_warning
+    raise "Plane not currently landed at this airport" if !@planes_at_airport.include?(plane)
+    @planes_at_airport.delete(plane)
+    "Plane has taken off"
+  end
+
+  private
+
+  def full?
+    @planes_at_airport.size >= @capacity ? true : false
+  end
+
+  def weather_conditions
+    rand(11)
+  end
+
+  def storm_warning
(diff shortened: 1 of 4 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
+require "./lib/airport"
+
+describe Airport do
+
+  describe "#land" do
+
+    it "raises an error if you try to land a plane that has already landed" do
+      plane = Plane.new
```
AndreaDiotallevi: The plane object should be a mock and not a real Plane object. This is why is the Plane class breaks, it will break this test also, which is about the Airport class.

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+    "Plane has taken off"
+  end
+
+  private
+
+  def full?
+    @planes_at_airport.size >= @capacity ? true : false
+  end
+
+  def weather_conditions
```
AndreaDiotallevi: I would create a separate class for Weather, because it is not a functionality that belongs to class Airport.

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
+require "./lib/plane.rb"
```
AndreaDiotallevi: You can just ```require "plane"```

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
+require "./lib/airport"
+
+describe Airport do
+
+  describe "#land" do
+
+    it "raises an error if you try to land a plane that has already landed" do
```
AndreaDiotallevi: It would be useful to use the keyword 'context' to structure the test within the describe block for instance "when stormy" and "when not stormy".

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      plane = Plane.new
+      allow(subject).to receive(:weather_conditions) { 6 }
+      subject.land(plane)
+      expect { subject.land(plane) }.to raise_error ("Plane already landed")
+      
+    end
+
+    it "gives an error message if capacity set to 20 and we try and land a 21st plane" do
+      airport = Airport.new(airport, 20)
+      allow(airport).to receive(:weather_conditions) { 6 }
```
AndreaDiotallevi: I think this is a smell that the weather conditions should be checked in a different class.

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+    storm_warning
+    raise "Plane not currently landed at this airport" if !@planes_at_airport.include?(plane)
+    @planes_at_airport.delete(plane)
+    "Plane has taken off"
+  end
+
+  private
+
+  def full?
+    @planes_at_airport.size >= @capacity ? true : false
```
AndreaDiotallevi: This line can be simplified without the ternary condition:
```@planes_at_airport.size >= @capacity```

File: lib/plane.rb [hunk 7]
```diff
@@ -0,0 +1,16 @@
+class Plane
+
+  attr_accessor :current_location
```
AndreaDiotallevi: I believe attr_accessor is not necessary here, as you don't want the instance variable to be changed from outside the object. It can be refectored to : ```attr_reader :current_location```

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    it "raises an error if you try to land a plane that has already landed" do
+      plane = Plane.new
+      allow(subject).to receive(:weather_conditions) { 6 }
+      subject.land(plane)
+      expect { subject.land(plane) }.to raise_error ("Plane already landed")
+      
+    end
+
+    it "gives an error message if capacity set to 20 and we try and land a 21st plane" do
+      airport = Airport.new(airport, 20)
```
AndreaDiotallevi: It might be useful to group initialised variables with ```let``` blocks, to avoid repetition.

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+    storm_warning
+    raise "Airport full, plane cannot land yet" if full?
+    raise "Plane already landed" if @planes_at_airport.include?(plane)
+    plane.land_at(@airport_name)
+    @planes_at_airport << plane
+  end
+
+  def take_off(plane)
+    storm_warning
+    raise "Plane not currently landed at this airport" if !@planes_at_airport.include?(plane)
```
AndreaDiotallevi: This line looks too long. Have you checked your code with the command ```rubocop```?

PR #1640, opened 2019-11-17
Pull request for my project

JKBero:


lib/plane.rb
```diff
@@ -0,0 +1,16 @@
+class Plane
+  attr_reader :grounded
```
alicelieutier: You don't need an `attr_reader` here

PR #1643, opened 2019-11-17
Attempt 1

HarryMumford:
Ran out of time to cover a couple of edge cases.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,10 @@
+class Airport
+  def land(plane)
+    "plane landed"
+  end
+  
+  def take_off(plane)
+    "plane took off"
+  end
+end
+
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,10 +1,32 @@
 class Airport
+  
+  DEFAULT_CAPACITY = 10
+
+  def initialize
+    @planes = []
+  end
+
   def land(plane)
-    "plane landed"
+    raise "airport is full" if airport_full?
+    @planes << plane
+    "plane landed" 
   end
   
   def take_off(plane)
-    "plane took off"
+    raise "no available planes" if airport_empty?
+    @planes.pop
+    "plane left airport"
   end
+
+  private
+
+  def airport_full?
+    true if @planes.size >= DEFAULT_CAPACITY
+  end
+
+  def airport_empty?
+    true if @planes == []
+  end
+ 
 end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,5 +1,7 @@
+require "weather.rb"
+
 class Airport
-  
+
   DEFAULT_CAPACITY = 10
 
   def initialize
@@ -8,12 +10,14 @@ class Airport
 
   def land(plane)
     raise "airport is full" if airport_full?
+    raise "too stormy for landing" if Weather.stormy?
     @planes << plane
     "plane landed" 
   end
   
   def take_off(plane)
     raise "no available planes" if airport_empty?
+    raise "too stormy for takeoff" if Weather.stormy?
     @planes.pop
     "plane left airport"
   end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,3 @@
+class Plane
+end
+
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,0 +1,8 @@
+class Weather
+  def weather(probability)
+    if rand(100) < probability
+      @weather = "stormy"
+    else
+      @weather = "normal conditions"
+  end
+end
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,4 +1,4 @@
-class Weather
+module Weather
   def weather(probability)
     if rand(100) < probability
       @weather = "stormy"
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,8 +1,7 @@
 module Weather
-  def weather(probability)
-    if rand(100) < probability
-      @weather = "stormy"
-    else
-      @weather = "normal conditions"
+  def self.stormy?
+    #true if lower than percentage chance of stormy weather
+    rand(100) < 50 ? true : false
   end
 end
+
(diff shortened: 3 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+require "weather.rb"
```
samfolo: maybe just 'weather'?

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    "plane left airport"
+  end
+
+  private
+
+  def airport_full?
+    true if @planes.size >= DEFAULT_CAPACITY
+  end
+
+  def airport_empty?
```
samfolo: you said you knew what you'd do here, so all good!

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    raise "airport is full" if airport_full?
+    raise "too stormy for landing" if Weather.stormy?
+    @planes << plane
+    "plane landed" 
+  end
+  
+  def take_off(plane)
+    raise "no available planes" if airport_empty?
+    raise "too stormy for takeoff" if Weather.stormy?
+    @planes.pop
```
samfolo: This may very well work, but what is you wanted a plane which isn't at the end of the array?

lib/airport.rb
samfolo: maybe <code>take_off</code> and <code>land</code> could be delegated to this class?

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
```
samfolo: maybe <code>take_off</code> and <code>land</code> could be delegated to this class?

lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+module Weather
+  def self.stormy?
```
samfolo: As a module, you could maybe keep <code>self.stormy?</code> as <code>stormy?</code>, so when you include it in a class it can be used as a standard method

lib/weather.rb
```diff
@@ -0,0 +1,7 @@
...
+    #true if lower than percentage chance of stormy weather
+    rand(100) < 50 ? true : false
```
samfolo: <code>rand(100) < 50</code> is good enough, but you already knew that

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
+require "airport"
+require "weather"
+
+describe Airport do
+  #landing a plane
```
samfolo: I read something in the rubric that due to comments being untestable, we should avoid using them to describe our code whenever possible; I can understand much of your code without these

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  describe "#land" do
+    it 'returns "plane landed" when a plane is landed' do
+      allow(Weather).to receive(:stormy?) {false}
+      plane = Plane.new
+      expect(subject.land(plane)).to eq("plane landed")
+    end
+
+    #full airport case
+    it "returns error if airport is full" do
+      plane = Plane.new
```
samfolo: Maybe use doubles instead? To isolate each test file

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      allow(Weather).to receive(:stormy?) {true}
+      expect{subject.land(plane)}.to raise_error("too stormy for landing")
+    end  
+  end
+
+  #taking off
+  describe "#take_off" do
+    it 'returns "plane took off" when an available plane takes off' do
+      plane = Plane.new
+      allow(Weather).to receive(:stormy?) {false}
```
samfolo: You could abstract these away into a <code>before(:each)</code> block, to avoid having to continuously type <code>allow(Weather).to receive(:stormy?) {false}</code> in the it-statements

spec/plane_spec.rb
```diff
@@ -0,0 +1 @@
+require "plane"
```
samfolo: similar comment to the <code>plane.rb</code> file

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    #full airport case
+    it "returns error if airport is full" do
+      plane = Plane.new
+      allow(Weather).to receive(:stormy?) {false}
+      Airport::DEFAULT_CAPACITY.times{ subject.land(plane) }
+      expect{subject.land(plane)}.to raise_error("airport is full")
+    end
+  
+    #stormy weather case
```
samfolo: you mentioned you'd prefer to use <code>context</code> rather than comments, so all good!

spec/weather_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "weather"
+
+describe Weather do
+  describe "#stormy?" do
+    it "returns true if stormy" do
+      allow(Weather).to receive(:stormy?) {true}
+      expect(Weather.stormy?).to eq(true)
```
samfolo: this seems to be only testing the double; I've found creating a dummy class whose sole purpose is to house the module is a good way to test its functionality in a vacuum

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class Airport
+
+  DEFAULT_CAPACITY = 10
+
+  def initialize
+    @planes = []
+  end
+
+  def land(plane)
```
samfolo: SRP - delegate to <code>Plane</class>? (mentioned below)

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  def take_off(plane)
+    raise "no available planes" if airport_empty?
+    raise "too stormy for takeoff" if Weather.stormy?
+    @planes.pop
+    "plane left airport"
+  end
+
+  private
+
+  def airport_full?
```
samfolo: this works, but when will <code>@planes.size</code> exceed <code>DEFAULT_CAPACITY</code>?

spec/weather_spec.rb
samfolo: maybe use: <code>expect(subject.stormy?).to be(true).or be false</code> to test for boolean result
or equally, <code>expect([true, false]).to include(subject.stormy?</code>

Eventually you'd want to write a test which implies <code>Weather.stormy?</code> is a boolean, rather than explicitly testing it

PR #1644, opened 2019-11-17
airport challenge

shafali03:


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+      raise 'Cannot land plane: weather is stormy' if stormy?
+      @planes << plane
+    end
+    
+    def take_off(plane)       
+    end
+    
+    private
+    
+    def full?
```
Hyan18: Good implementation of SRP

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+      if @planes.length >= @capacity
```
Hyan18: Unnecessary conditional (can get rid of if else since statement in itself returns true or false)

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'airport'
+
+describe Airport do
+  subject(:airport) { described_class.new(20)}
+  let(:plane) { double:plane }
+  
+  describe '#land' do
+    it 'instruct a plane to land' do
+      allow(airport).to receive(:stormy?).and_return false
+      expect(airport).to respond_to(:land).with(1).argument
```
Hyan18: might be better to have an equal to statement rather than respond to as it makes the test more rigorous.

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'airport'
+
+describe Airport do
+  subject(:airport) { described_class.new(20)}
+  let(:plane) { double:plane }
+  
+  describe '#land' do
+    it 'instruct a plane to land' do
+      allow(airport).to receive(:stormy?).and_return false
```
Hyan18: good use of method stub

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  
+  describe '#take_off' do
+    it 'instruct a plane to take off' do
+      expect(airport).to respond_to(:take_off).with(1).argument
+    end
+  end
+  
+  it 'stormy cannot land plane' do
+    allow(airport).to receive(:stormy?).and_return true
+    expect{ airport.land(plane) }.to raise_error 'Cannot land plane: weather is stormy'
```
Hyan18: good clear error message

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
+class Airport
+    def initialize(capacity)
+      @capacity = capacity 
+      @planes = []
+    end
+    
+    def land(plane)
+      raise 'Cannot land: Airport full' if full?
```
Hyan18: should leave spaces after guard clauses

alicelieutier: You should actually leave an empty line after the last guard clause, and before the rest of the function.

Hyan18: Sorry when I said space I meant an empty line

PR #1645, opened 2019-11-17
Pull request

jj49411:


Jamie95187:
Consider using rubocop for readability :)))

lib/plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
```
Jamie95187: Redundant file :)

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_reader :planes
+  attr_reader :weather
+  DEFAULT_CAPACITY = 50
+
+  def initialize
+    @planes = []
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def land(plane)
```
Jamie95187: Could consider SRP, refactor into individual methods. For example @planes.count >= @capacity could have changed to a full? method 👍

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+    fail "The airport is full" if @planes.count >= @capacity
+    fail "Cannot land because of the storm" if @weather == "stormy"
+    @planes << plane
```
Jamie95187: I don't think that the plane here is necessarily a newly created Plane object because you have not required the plane.rb file. Not sure 😕 ❓ ❓ ❓ ❔

Jamie95187: May encounter problems when considering whether the plane is flying or grounded 👍

alicelieutier: This is fine actually. Ruby relies on duck typing, and usually does not enforce type for arguments.

PR #1646, opened 2019-11-17
Airport challenge submission

denriquem:


spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    it "Does plane take off from airport" do
+      airport = Airport.new
+      allow(airport.weather).to receive(:generate) { "Sunny" }
+      expect { airport.take_off }. to output("Plane has taken off!").to_stdout
+    end
+
+    it "Airport full of planes. No room to land!" do
+      airport = Airport.new
```
ajbacon: can factor out airport = Airport.new from each test into a "before" block

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    it "Prevent take off when weather is stormy." do
+      airport = Airport.new
+      allow(airport.weather).to receive(:generate) { "Stormy weather!" }
+      expect { airport.take_off }.to raise_error("There's a storm. No flying.")
+    end
+
+    it "Prevent landing when weather is stormy" do
+      airport = Airport.new
+      plane = Plane.new
```
ajbacon: can "mock" a plane object so that the plane class doesnt have to exist

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    it "Does plane take off from airport" do
+      airport = Airport.new
+      allow(airport.weather).to receive(:generate) { "Sunny" }
+      expect { airport.take_off }. to output("Plane has taken off!").to_stdout
+    end
+
+    it "Airport full of planes. No room to land!" do
+      airport = Airport.new
+      capacity = 20
```
ajbacon: is this capacity variable used anywhere?

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @planes << plane
+  end
+
+  def take_off
+    # weather = Weather.new
+    todays_weather = @weather.generate
+    fail "There's a storm. No flying." if todays_weather == "Stormy weather!"
+
+    print "Plane has taken off!"
+    @planes.pop
```
ajbacon: the pop method will become troublesome later when doing the edge cases, it will want to delete a specific plane from the array (ie the specific plane that is instructed to take off)

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @planes = []
+    @weather = Weather.new
+    @capacity = capacity
+  end
+
+  def land(plane)
+    todays_weather = @weather.generate
+    fail "There's a storm. No landing." if todays_weather == "Stormy weather!"
+
+    fail "The airport is full." if @planes.size >= @capacity
```
ajbacon: as with boris bikes could refactor '@planes.size >= @capacity' to a method called 'full?'

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  attr_reader :planes
+  attr_reader :weather
+  attr_reader :capacity
+
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @weather = Weather.new
```
ajbacon: nice idea to initialize a weather object when creating an airport

PR #1647, opened 2019-11-17
Airport Challenge

JonathanPalma-code:


spec/plane_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'plane'
+
+describe Plane do
+
+  context 'does exist?' do
+    it { is_expected.to respond_to :yes }
```
ffgi-es: It could be possible to use `be_instance_of` instead of `responds_to` a method that you don't actually want as behaviour. It could be deleted later as the class and tests get flushed out.

spec/airport_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      allow(airport).to receive(:stormy?).and_return(false)
+      described_class::DEFAULT_CAPACITY.times { airport.to_land(plane) }
+      expect { airport.to_land(plane) }.to raise_error "Airport is full: Take off some planes."
+    end
+  end
+
+  describe '#to_land' do
+    before do # checks if weather is sunny before doing any actions
+      allow(airport).to receive(:stormy?).and_return(false)
+    end
```
ffgi-es: This is being repeated on lines 35-37. Maybe it could be moved out one context higher to avoid repetition.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+require_relative 'plane'
```
ffgi-es: Does the airport class really depend on the Plane class? It might not be necessary.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+    @planes_collected.pop unless denied_take_off
+  end
+
+  def stormy?
+    rand(1..10) >= 8
+  end
+
+  private 
+
+  attr_reader :plane
```
ffgi-es: I'm not completely clear where this is used?

spec/airport_spec.rb
ffgi-es: This stubbing can be overridden later when you want it to actually return true for `#stormy?`

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+require_relative 'plane'
+
+class Airport
+  DEFAULT_CAPACITY = 20
+
+  attr_accessor :capacity
```
ffgi-es: You might want to be careful about using an `attr_accessor` as changing the capacity once the airport is created and has some planes landed could cause unexpected problems.

PR #1650, opened 2019-11-18
airport challenge

TuanNguyen1010:


even13:
Please commit more often to improve readability of overall process :)

spec/airport_test_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+describe Airport do
+  describe '#land' do
+    it "lands at a aiport" do
+      eva_air = Airport.new
+      expect(eva_air.land(Plane.new)).to eq true
+    end
+    it "raise error if airport is full" do
+      eva_air = Airport.new
+      expect{ 11.times{eva_air.land(Plane.new)}}.to raise_error(fail)
```
even13: This test didn't pass, but it does now, nice discussion on this :) 👍

even13:
Maybe a README would improve readability?
Running rubocop might also help improve the code in that respect :) 
Thank you for the review, I think it was a very helpful one and very positive on both sides 👍
We don't see it because of the commits, but as we discussed the project together, you definitely followed the TDD process, and you have the appropriate tests for all the user stories you worked on. 👍

PR #1667, opened 2019-12-13
Solution of Carmen-Alina 

carmen-alina:
First 3 user stories built in a beginner's TDD approach.


carmen-alina:
Implementation in JavaScript of the first 5 user stories from the airport-challenge.

PR #1680, opened 2019-12-15
Airport Challenge

acmerlino1:


lib/airport.rb
```diff
@@ -1,38 +1,44 @@
...
-        return "stormy" if rand_number == 1
-        return "sunny"
-    end
-
-    def rand_number
-        rand(1..5)
-    end
-end
\ No newline at end of file
+  attr_reader :capacity, :planes
```
acmerlino1: Consider using attribute accessor so user can change capacity

lib/airport.rb
```diff
@@ -1,38 +1,44 @@
...
+    
+    raise "Cannot land because airport is full" if full?
+    
+    raise "I'm already landed" if @planes.include?(plane)
+    
+    @planes << plane
+  end
+
+  def take_off(plane)
+    raise "Cannot take off due to stormy weather" if weather == "stormy"
```
acmerlino1: add error messages to a hash so don't have to reuse the same error message string each time

PR #1681, opened 2019-12-15
added land, (confirm) take off

notriapatel:


spec/airport_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  describe '#land' do
+    it 'lands a plane' do
+      airport = Airport.new
+      expect(airport).to respond_to :land
+    end
+  end
+
+  describe '#take_off' do
+    it "let's a plane take off" do
+      airport = Airport.new
```
squareben1: Consider using subject to avoid issues later on

spec/airport_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  it 'confirms take off' do
+    airport = Airport.new
+    expect(airport).to respond_to :confirm_take_off
+  end
+
+  it 'raises error when the airport is full' do
+    airport = Airport.new
+    expect(airport.land).to raise_error 'Airport is full'
```
squareben1: Set up the conditions for the test to fail (i.e. land capacity.times planes)

PR #1689, opened 2019-12-16
airport_challenge 

jackwillee:


lib/plane.rb
```diff
@@ -0,0 +1,4 @@
+class Plane
+
```
junolouise: When trying to test in IRB and instantiate a new ```plane``` the error message 'uninitialized constant Plane' is highlighted. I think this is because you have an empty class here with no plane method inside. How did you plan on fixing this? Did you run manual tests throughout your process? Perhaps this is something to try and improve.

spec/plane_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'plane'
+describe Plane do
```
junolouise: This is an example of a test that unfortunately does not test for anything. Because your plane class is empty, other than checking that there is a class of Plane, this test will not check how it operates. In the materials provided with the weekend challenge, there are some really good videos on how to test and how to avoid this- I would reccomend checking them out. :)

spec/spec/spec_helper.rb
```diff
@@ -0,0 +1,100 @@
+# This file was generated by the `rspec --init` command. Conventionally, all
```
junolouise: Good to see that you used RSpec whilst testing. Do try and also apply the logic in my comment above to avoid redundant tests :)

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+  @array << plane
+  end 
+
+   def take_off(plane)
+   end 
+end 
+
+
+def full?
+@array.size >= @capacity
```
junolouise: I think that you need to explicitly state the capacity of the airport or else it will never be full? Also, there are inconsistent spacing and line breaks here. Please see my first comment and I suggest using rubocop to help fix these.

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
+class Airport
+  def initialize(capacity)
+    @capacity = capacity
+    @array = []
+  end 
+  def land(plane)
+  raise 'Capacity full cannot land another plane ' if full?
+  @array << plane
```
junolouise: Where else is plane mentioned?

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+  end 
+
+   def take_off(plane)
```
junolouise: What is happening to the plane inside of this method?

spec/airport_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'airport'
+describe Airport do 
+    subject(:airport) { described_class.new(20) }
+    let(:plane) {double :plane}
+it 'tells a plane to land' do 
+    expect(airport).to respond_to(:land).with(1).arguments
```
junolouise: Is there a way to refactor this so that the 1 argument instead explicitly shows the plane?

lib/plane.rb
junolouise: Sorry ignore the above I figured it out in the end and it did work however I don't think the Plane will operate functionally as currently, it is an empty class.

PR #1691, opened 2019-12-16
Shaun-airport-challenge

shaunlsy:
Shaun-airport-challenge

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,22 +1,27 @@
 require_relative 'plane'
+require_relative 'weather'
 
 class Airport
     attr_reader :planes
-    attr_accessor :capacity
+    attr_accessor :capacity, :weather
 
     DEFAULT_CAPACITY = 20
     
     def initialize(capacity = DEFAULT_CAPACITY)
         @capacity = capacity
         @planes = []
+        weather = Weather.new
+        @weather = weather.good
     end
 
     def land(plane)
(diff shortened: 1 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+class Airport
+    attr_reader :planes
+    attr_accessor :capacity, :weather
+
+    DEFAULT_CAPACITY = 20
+    
+    def initialize(capacity = DEFAULT_CAPACITY)
+        @capacity = capacity
+        @planes = []
+        weather = Weather.new
```
soph-g: Try using dependency injection for Weather :)

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+        @planes = []
+    end
+
+    def land(plane)
+        fail 'Airport is full' if planes.count == capacity
+        @planes << plane
+    end
+
+    def take_off(plane)
+        @planes.pop
```
basselalsayed: Use .delete(plane) for more specificity

lib/plane.rb
```diff
@@ -0,0 +1,12 @@
+class Plane 
+    
+    attr_accessor :landed
+    
+    def initialize(status = true)
+        @landed = status
+    end
+
+    def landed?
```
basselalsayed: landed method not necessary

PR #1696, opened 2020-01-25
Airport project complete except for a few edge cases

jlblumberg:


README.md
```diff
@@ -1,7 +1,6 @@
-Airport Challenge
-=================
+# Airport Challenge
```
Peter2-71828: Really clear README - Easily understandable

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
```
Peter2-71828: good encapsulation, weather contained in its own class

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'plane'
+require_relative 'weather'
+class Airport
+
```
Peter2-71828: could use attr_reader instead of @ attribute 
could use private then attr_writer instead o @atribute = attribute

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
...
+    @weather = weather
+  end
+
+  def tell_to_land(plane)
+    fail "Can't land; airport's full." if @parked_planes.size >= @capacity
+    fail "Can't land; stormy weather." if @weather.stormy?
+
+    @parked_planes << plane
+  end
+
```
Peter2-71828: _plane - could use plane

jlblumberg: Will remove, thanks!

PR #1697, opened 2020-01-26
Airport challenge

jasylwong:


README.md
```diff
@@ -1,39 +1,16 @@
 Airport Challenge
```
Mrtly: well done on updating the readme with a proper explanation for someone external (and a fancy pic!) 😊

spec/plane_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'plane'
+
+describe Plane do
```
Mrtly: unnecessary line/file since not testing for anything?

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
+  def self.stormy?
```
Mrtly: the self is so smart! ;)

PR #1700, opened 2020-01-26
Weekend challenge 1 - Airport Challenge - complete

lookupdaily:
Rubocop seems to be obsolete


lib/plane.rb
```diff
@@ -0,0 +1,24 @@
+# require_relative 'airport'
+
```
Pazoia: Your description for where the plane is at every step is very clear. Excelent method naming.

lib/Airport.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'plane'
+require_relative 'weather'
+
```
Pazoia: Code formatting is very clear, it's easy to find and read every method.

PR #1702, opened 2020-01-26
half-way done

Mrtly:
maybe 60% complete? I added as much as possible of the requirements, without breaking the rest. The commented out bits were breaking other tests, so I took them out. I got stuck at the stubs and didn't have enough time to read thru and add them.


feature_test.rb
```diff
@@ -0,0 +1,6 @@
+require './lib/plane.rb'
```
hansa-lallu: I like how you used this. I don't have an understanding of this just yet.

PR #1703, opened 2020-01-26
Weekend Challenge Week_1

Peter2-71828:


lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+  def add_plane
+    self.num_of_planes += 1
+  end
+
+  def minus_plane
+    self.num_of_planes -= 1
+  end
+
+  def weather
+    return rand(1..100) == 1 ? 'stormy' : 'sunny'
```
jlblumberg: Nice one liner! Easy to read and more succinct than an if/else.

spec/plane_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    airport = Airport.new
+    allow(airport).to receive(:weather) { 'stormy' }
+    plane = Plane.new(airport)
+    expect { plane.take_off(airport) }.to raise_error('stormy weather')
+  end
+
+  it 'Plane.landing should raise_error if weather is stormy' do
+    airport = Airport.new
+    allow(airport).to receive(:weather) { 'stormy' }
+    expect { subject.land(airport) }.to raise_error('stormy weather')
```
jlblumberg: Nice use of only one expects statement per it block. Apparently this is best practice - makes it easier to read and reduces the potential for conflated issues if a test fails.

spec/airport_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'airport.rb'
+require 'plane.rb'
+
+describe Airport do
+
+  it 'There should be airports' do
```
jlblumberg: Could you get away with missing this 'it' block out? Because (I think) RSpec 'tests' every 'it' block. If I'm reading this correctly, this block doesn't test anything. Could you instead swap it for a test that the class exists/has been defined, or even that it initializes with some set of variables?

lib/airport.rb
Peter2-71828: thanks

spec/plane_spec.rb
Peter2-71828: thanks. I will try and use before to make tests dry

PR #1704, opened 2020-01-26
Airport challenge - No edge cases

hansa-lallu:


lib/airport.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'weather'
+
+class Airport
+
+  DEFAULT_CAPACITY = 2
```
jasylwong: Good use of default capacity variable. I like how it is a small number, so you can test the 'full' cases easily.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @terminal = []
+    @capacity = capacity
+  end
+
+  def land(plane, weather)
+    fail "weather is stormy, do not land" if weather.stormy?
+    fail "No space to land" if full?
+
+    terminal << plane
```
jasylwong: Should this be pushing to the instance variable, @terminal ? Where and what is the local variable 'terminal'?

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  end 
+
+  private
+  attr_reader :terminal
+
+  def full?
+    terminal.count >= DEFAULT_CAPACITY
+  end 
+
+    # def empty?
```
jasylwong: This method looks good to me! Could be implemented.

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather 
+  def stormy?
+    rand(1..100) >= 75
```
jasylwong: Could shorten to: rand >= 0.75

spec/airport_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+
+# my third test - checks if the class Airport responds to the method land_plane
+  it 'responds to land_plane' do
+    expect(subject).to respond_to :land
+  end 
+# one line syntax: it {is_expected.to respond_to :land}
+
+  it { should respond_to(:land).with(2).arguments } 
+
+  it { should respond_to(:take_off).with(1).arguments }
```
jasylwong: Why does the land method take plane as an argument, but the take_off method doesn't?

spec/airport_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  it 'removes a plane when landed' do
+    plane = Plane.new
+    sunny_weather = double(Weather, stormy?: false)
+    subject.land(plane, sunny_weather)
+    expect(subject.take_off(sunny_weather)).to eq(plane)
+  end
+
+  it 'has a deafult capacity' do
```
jasylwong: deafult -> default

spec/airport_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    expect(subject.take_off(sunny_weather)).to eq(plane)
+  end
+
+  it 'has a deafult capacity' do 
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end 
+
+  describe '#land' do 
+    it 'raises an error to stop landing when airport is full' do
+      sunny_weather = double(subject, stormy?: false)
```
jasylwong: Since this assignment is used multiple times in the file, it could be defined once at the start to conform to DRY

spec/airport_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  describe "#land" do
+    it 'prevents landing when weather is stormy' do
+      plane = Plane.new
+      stormy_weather = double(subject, stormy?: true)
+      expect { subject.land(plane, stormy_weather) }.to raise_error "weather is stormy, do not land"
+    end
+
+    it 'allows landing when weather is sunny' do
```
jasylwong: Is this not the same as the test on lines 25+?

spec/airport_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    sunny_weather = double(Weather, stormy?: false)
+    expect(subject.land(plane, sunny_weather)).to eq([plane])
+  end
+
+  it 'removes a plane when landed' do
+    plane = Plane.new
+    sunny_weather = double(Weather, stormy?: false)
+    subject.land(plane, sunny_weather)
+    expect(subject.take_off(sunny_weather)).to eq(plane)
+  end
```
jasylwong: This test works when no other planes have landed after the plane in question. In such a scenario, the pop method in the take_off function will return the last plane rather than the one in question.
Could use is_an_instance_of

spec/plane_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'plane'
```
jasylwong: This test is good. I'm gonna steal it for my programme :)

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  DEFAULT_CAPACITY = 2
+   
+  attr_accessor :capacity
+    
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @terminal = []
+    @capacity = capacity
+  end
+
+  def land(plane, weather)
```
jasylwong: Do we want to allow the user to input the weather manually? He might input the wrong weather conditions, and allow take_off in stormy weather!

PR #1705, opened 2020-01-26
Airport Challenge - unfinished

EllieRichardsonJones:
Started on first two user stories

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,11 @@
+class Airport
+  attr_reader :plane
+
+  def land_plane(plane)
+    plane
+  end
+
+  def takeoff_plane(plane)
+
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,20 @@
+require 'airport.rb'
+describe Airport do
+# check method land_plane responds to Airport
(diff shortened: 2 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,11 @@
+class Airport
+  attr_reader :plane
```
Dlibmanw: So impressed that you're using attr_reader - I still have no idea how that works! :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'airport.rb'
+describe Airport do
+# check method land_plane responds to Airport
```
Dlibmanw: I think you meant Airport responds to land_plane? but in any case I think the example you wrote looks great, it's just to make to comment you added before the test, more clear

spec/plane_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'plane'
+describe Plane do
```
Dlibmanw: maybe try to use a test with the format of
expect(plane).to be_landed and the method on the plane.rb to return true if the plane landed? (I don't really know if its better than what you wrote - just forcing myself to come up with a suggestion point:)

PR #1706, opened 2020-01-26
Airplane Challenge

Kefuri:
Disclaimer - I had issues initially with getting gems to install, and so I removed the required ruby version which meant that the gems installed fine.


lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+  attr_reader :airport
+  def initialize(airport = true)
+    @airport = airport
+  end
+
+  def landed?
+    @airport
+  end
+
+  def planestatus
```
zsoobratty: Showing the planestatus as true or false for landed is efficient

spec/plane_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+    it 'should return true or false if a plane is in an airport' do
+      expect(plane.landed?).to eq plane.airport
+    end
+  end
+  describe '#plane_status' do
+    subject(:airport) { Airport.new }
+    it "should update a plane's airport status when it lands" do
+      plane = Plane.new(false)
+      allow(Weather).to receive(:stormy?) { false }
```
zsoobratty: It may be helpful and clearer to find a way to group the stubs showing the stormy status as true/false, rather than repetition

lib/plane.rb
zsoobratty: Could call this this something like isflying? to keep consistent with other boolean methods

PR #1707, opened 2020-01-26
Airport_Challenge Unfinished

Ovy95:


lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+    puts "Plane has taken off"
+            
+    end
+
+    def land(plane)
+        raise "Airport full can't land here" if full
+        @planes << plane
+    end
+
+    private
```
tobydawson1: Good use of private methods

lib/plane.rb
```diff
@@ -0,0 +1,4 @@
+
+class Plane
+
```
tobydawson1: could have method within the plane class for take off and landing.
 
def land
return self
end

plane.land will then return a instance of itself which you could call into your airport landing method. same goes for takeoff.

Plan.md
```diff
@@ -0,0 +1,36 @@
+First 3 storys are done just need to fix the Rspec errors on the current story.
+
+<!-- Jacks-Air:airport_challenge jack$ irb -->
```
tobydawson1: Getting inside the code with IRB! Good practice.

PR #1708, opened 2020-01-26
Airport challenge

BarriF13:
I have managed to do 3 steps of the challenge. More to add during the week hopefully

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -5,7 +5,7 @@ class Airport
   def planes_in_airport
     Plane.new
   end
-  def land
-    
+  def land(plane)
+
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,1 @@
+# airport challenge app
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -5,4 +5,7 @@ class Airport
   def planes_in_airport
     Plane.new
   end
+  def land
+    
+  end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,5 +1,8 @@
 # airport challenge app
+require_relative 'weather'
+
 class Airport
   def land
+    Weather.new
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,8 +1,8 @@
 # airport challenge app
-require_relative 'weather'
+require_relative 'plane'
 
 class Airport
   def land
-    Weather.new
+    Plane.new
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,7 +2,7 @@
 require_relative 'plane'
 
 class Airport
-  def land
+  def planes_in_airport
     Plane.new
   end
 end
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,18 @@
+# airport challenge app
```
DavidStewartLDN: Good opening comment!

lib/airport.rb
```diff
@@ -0,0 +1,18 @@
...
+require_relative 'plane'
+
+class Airport
+ # attr_accessor so everyone can access to this var
```
DavidStewartLDN: Is this a planned part of the code? If you add it in you will be able to view the planes in an airport using the method `airport_instance.planes`

lib/airport.rb
```diff
@@ -0,0 +1,18 @@
...
+  def initialize(capacity)
```
DavidStewartLDN: Well initialised so that capacity can be defined. If the user does not enter a capacity, what will the default capacity be?

lib/airport.rb
```diff
@@ -0,0 +1,18 @@
...
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land_plane(plane)
+    raise "cannot land: airport full" if @planes.length >= @capacity
```
DavidStewartLDN: Very nice use of raise, could put `.full?` method into separate code to follow SRP.

lib/plane.rb
```diff
@@ -0,0 +1,10 @@
+# making plane for my airports
+class Plane
+  def landed?
```
DavidStewartLDN: Very nice method idea, I didn't not think of this!

spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'airport'
+
+describe Airport do
+  subject (:airport){ described_class.new(20) }
+# this is for the landing_plane method in airport.rb
```
DavidStewartLDN: Well commented tests, I will follow your lead on this!

spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+# --below --expect the method land_plane so it fails then we need to write a method to airport file
+# --below-- first part of the story
+  it 'instruct a plane to lane' do
+    expect(airport).to respond_to(:land_plane).with(1).argument
+  end
+#--below-- expect method take_off_plane to fail first
+#--below-- second part of the story
+  it 'instruct the plane to take off' do
```
DavidStewartLDN: Could use context and it statements to break up test for landing, take_off and other methods

spec/plane_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'plane'
+
+describe Plane do
```
DavidStewartLDN: Again, it is week 1 so you have written good tests to test your code. It would be good to look into the RSpec recommended structure to understand how tests `context, describe` keywords are used. This is something I am finding challenging so we should definitely share notes on this.

PR #1709, opened 2020-01-26
Airport_Challenge

tobydawson1:


lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+require_relative 'plane'
```
Ovy95: Very easy to read the code, very well indented, as well as descriptions of tests.

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+module Weather
+
+    def self.forecast(seed = Random.new_seed)
+        rando = Random.new(seed)
+        num = rando.rand(2) 
+        array = ["stormy", "calm"]
```
Ovy95: Call the array, -  weather

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+module Weather
+
+    def self.forecast(seed = Random.new_seed)
```
Ovy95: I liked the way you used this method to always push out calm weather I wasn't sure on how to do this part.

PR #1712, opened 2020-01-27
Airport Challenge --> Not finished

Pazoia:
Got up to user story 2, project not passing rspec at the moment.


README.md
```diff
@@ -13,78 +14,47 @@ Airport Challenge
...
-As an air traffic controller 
-To ensure safety 
-I want to prevent landing when weather is stormy 
-```
-
-Your task is to test drive the creation of a set of classes/modules to satisfy all the above user stories. You will need to use a random number generator to set the weather (it is normally sunny but on rare occasions it may be stormy). In your tests, you'll need to use a stub to override random weather to ensure consistent test behaviour.
+## Table of contents
 
-Your code should defend against [edge cases](http://programmers.stackexchange.com/questions/125587/what-are-the-difference-between-an-edge-case-a-corner-case-a-base-case-and-a-b) such as inconsistent states of the system ensuring that planes can only take off from airports they are in; planes that are already flying cannot take off and/or be in an airport; planes that are landed cannot land again and must be in an airport, etc.
+* [Introduction](#introduction)
```
lookupdaily: Great to have links to different sections - I should do this as my README is quite long!

README.md
```diff
@@ -13,78 +14,47 @@ Airport Challenge
...
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc. 
+Ruby 2.6.0
 
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this weekend.
+## Launch
 
-**BONUS**
+Will need help filling out this field.  
+To be completed.
```
lookupdaily: You could start with just instructions to run irb and require the programme. A couple of follow up lines of example code could also be useful. e.g.

```
$ irb
> require './lib/airport.rb'
```
examples could be:
```
> airport = Airport.new
> airport.land_plane
> airport.take_off
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+describe Airport do
+
+  it { is_expected.to respond_to(:land_plane) }
+
+  it { is_expected.to respond_to(:take_off) }
+
+  it "should return plane" do
+    subject.land_plane
+  end
```
lookupdaily: Can you add an expect block here? What should happen when the plane lands? Your test on line 17-21 is perfect. How can you create a similar test for land, which will help you update the plane's location?

spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  it "instruct planes to take off" do
+    subject.take_off
+  end
+
+  it "checks if plane has taken off" do
+    plane = subject.take_off
+    expect(plane).to respond_to(:plane_flying?)
+    expect(plane.plane_flying?).to eq "Flying"
+  end
```
lookupdaily: This test is perfect for testing the expectation of the plane taking off - you should try using this for land! (see above comment). This should help you with the code for landing the plane

spec/plane_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'plane'
+
+describe Plane do
+
+  it { is_expected.to respond_to(:plane_flying?) }
+
```
lookupdaily: Can you add an additional test for expectations here - e.g. expect().to eq true/"flying"

lib/plane.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
+
+  def plane_flying?
+    "Flying"
+  end
```
lookupdaily: You could use your airplane methods to update this and perhaps add another method: location

spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    subject.land_plane
+  end
+
+  it "instruct planes to take off" do
+    subject.take_off
+  end
+
+  it "checks if plane has taken off" do
+    plane = subject.take_off
+    expect(plane).to respond_to(:plane_flying?)
```
lookupdaily: You don't need this line anymore - as below line does both!

PR #1713, opened 2020-01-27
Airport Challenge pull

zsoobratty:


lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    if planes.length.positive?
+      raise "Cannot take off while weather is stormy" if stormy?
+      
+      @planes.pop
+    else
+      raise "There are no planes"
+    end
+  end
+
+  def full?
```
Kefuri: great SRP applied here!

spec/airport_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  it 'can take off a plane' do
+    allow(subject).to receive(:stormy?).and_return(false)
+    plane = Plane.new
+    subject.land(plane)
+    expect(subject.take_off).to be_an_instance_of(Plane)
+  end
+
+  it 'prevents landing when airport is full' do
+    allow(subject).to receive(:stormy?).and_return(false)
```
Kefuri: Smart use of stubbing for the tests

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def land(plane)
+    raise "Cannot land while weather is stormy" if stormy?
+    raise "Plane capacity is full" if full?
+
+    @planes << plane
+  end
+
+  def take_off
+    if planes.length.positive?
```
Kefuri: I like this syntax! Smart way of checking for a plane in the array.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    else
+      raise "There are no planes"
+    end
+  end
+
+  def full?
+    @planes.length >= @capacity
+  end
+
+  def stormy?
```
Kefuri: To improve readability, this could be moved into its own class in its own file - and would allow for future improvements to the weather class.

spec/airport_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'airport'
+require 'plane'
+describe Airport do
+  it 'can land a plane' do
+    allow(subject).to receive(:stormy?).and_return(false)
```
Kefuri: Try to find a way to group tests by stormy and not stormy so that you only have to stub once?

PR #1714, opened 2020-01-27
Submission of weekend challenge

richardpattinson:


lib/airports.rb
```diff
@@ -0,0 +1,45 @@
...
+        @planes.pop
+    end 
+    def full? 
+        if @planes.length > @capacity 
+            return true
+        else 
+            return false
+        end
+    end
+    def clear?
```
blu3skies: I like how you have the weather method inside the airport class.  Would have saved me lots of time.

lib/airports.rb
```diff
@@ -0,0 +1,45 @@
...
+    def initialize(capacity = 20)
+    @capacity = capacity
+    @planes = []
+    @weather = self.forecast
+    end
+    def land(plane)
+        fail "Plane is already landed at a different station" unless plane.location == "Sky"
+        fail "Cannot land due to adverse weather conditions" unless self.clear?
+        fail "Airport is at capacity" if self.full?
+        plane.location = self
```
blu3skies: possibly separate into two methods to make SRP

lib/airports.rb
```diff
@@ -0,0 +1,45 @@
...
+        @planes.push(plane)
+    end
+    def take_off
+        fail "Cannot take off due to adverse weather conditions" unless self.clear?
+        @planes.last.location = "Sky"
+        @planes.pop
+    end 
+    def full?
```
blu3skies: could refactor these predicate methods into single line

spec/airports_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "airplanes"
```
blu3skies: I like the efficiency of your code. You did exactly what they asked to the bare minimum.

PR #1715, opened 2020-01-27
Pull for code review

saidbyced:
✈️

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,7 +2,7 @@ require_relative 'plane'
 
 class Airport
   DEFAULT_CAPACITY = 1
-  attr_accessor :capacity
+  attr_accessor :capacity, :stormy
   attr_reader :planes
   
   def initialize(capacity = DEFAULT_CAPACITY) 
@@ -13,11 +13,19 @@ class Airport
   def land(plane)
     fail "Airport is full" if full?
 
+    fail "Weather is stormy" if stormy?
+
     planes.push(plane)
   end
 
   def takeoff(plane)
+    fail "Weather is stormy" if stormy?
+    
     "#{plane} has left"
+  end
+
+  def stormy?
+    stormy
   end
 
   private
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,7 +2,7 @@ require_relative 'plane'
 
 class Airport
   DEFAULT_CAPACITY = 1
-  attr_accessor :capacity, :stormy
+  attr_accessor :capacity
   attr_reader :planes
   
   def initialize(capacity = DEFAULT_CAPACITY) 
@@ -25,7 +25,7 @@ class Airport
   end
 
   def stormy?
-    stormy
+    rand(2).zero?
   end
 
   private
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -23,5 +23,4 @@ class Airport
   def full?
     @planes.length >= capacity
   end
-
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -4,4 +4,8 @@ class Airport
   def land(plane)
     true
   end
+
+  def takeoff(plane)
+    "Plane has left"
+  end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,7 @@
+require_relative 'plane'
+
+class Airport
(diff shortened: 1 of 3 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require_relative 'plane'
```
EWright212: like how you used require_relative, much easier than writing full path

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def takeoff(plane)
+    fail "Weather is stormy" if stormy?
+    
+    "#{plane} has left"
+  end
+
+  def stormy?
+    rand(2).zero?
+  end
```
EWright212: I love how this is so simple

lib/plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
+end
```
EWright212: Perhaps this is no longer needed?

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+    it 'raises error when stormy' do
+      # subject.stormy = true
+      expect { subject.takeoff(plane) }.to raise_error(RuntimeError, "Weather is stormy")
+    end
+  end
+
+  context ':capacity' do
+    it 'has a default capacity' do
+      expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
```
EWright212: I like how you got this to work, I could not manage it - will be stealing

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'airport'
```
EWright212: love the layout of this file, easy to read :)

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require_relative 'plane'
+
+class Airport
+  DEFAULT_CAPACITY = 1
+  attr_accessor :capacity
```
EWright212: attractive_accessor is a new one I've not seen before, cool

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    fail "Weather is stormy" if stormy?
+
+    planes.push(plane)
+  end
+
+  def takeoff(plane)
+    fail "Weather is stormy" if stormy?
+    
+    "#{plane} has left"
```
EWright212: Could possibly remove the plane from the airport in this step

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def stormy?
+    rand(2).zero?
```
EWright212: Is it possible for zero to be returned here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect(subject.land(plane)).to be
+    end
+    it 'raises error when airport is full' do
+      Airport::DEFAULT_CAPACITY.times do
+        subject.land(plane)
+      end
+      expect { subject.land(plane) }.to raise_error(RuntimeError, "Airport is full")
+    end
+    it 'raises error when stormy' do
+      # subject.stormy = true
```
EWright212: a TODO comment here

lib/airport.rb
saidbyced: Thanks 🙂

lib/plane.rb
saidbyced: Future proofing for when planes have more attributes...? Maybe I could include this in another files till I really NEED another class? What do you think?

spec/airport_spec.rb
saidbyced: More than welcome! I stole it from the walkthrough of boris_bikes!

lib/airport.rb
saidbyced: It allows for reading and writing outside of the internal class methods - super handy!

lib/airport.rb
saidbyced: True...I was trying to code 'naively' and this was the most simple resolution for the user story I could see. I agree that this would be a much more accurate use of the 'takeoff' method.

lib/airport.rb
saidbyced: Yep; rand(5) returns a range between 0 and 4 🙂

PR #1716, opened 2020-01-27
Weekend challenge submitted

Dlibmanw:


lib/airport.rb
```diff
@@ -0,0 +1,11 @@
+require_relative 'plane'
+class Airport
+  def landing(plane)
+    fail "The airport is full" if @plane
```
EllieRichardsonJones: Nice descriptive string!

lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
+  def gone?
```
EllieRichardsonJones: House style suggestion - could have a more descriptive method name such as 'taken_off?'

spec/airport_spec.rb
```diff
@@ -0,0 +1,22 @@
+require './lib/airport.rb'
+describe Airport do
+  describe '#landing' do
+    it 'accept a landing plane' do
+      plane = Plane.new
+      expect(subject.landing(plane)).to eq plane
+    end
+
+    it 'raises an error when the airport is full' do
```
EllieRichardsonJones: Great description

lib/plane.rb
Dlibmanw: Great thanks!

PR #1718, opened 2020-02-03
First Pull Request

sophiewo:


spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  describe '#land' do
+    it 'raises an error if airport is full' do
+      plane = Plane.new
+      subject.land(plane)
+      expect { subject.land(plane) }.to raise_error "Airport full"
+    end
+  end
+end
+
```
lookupdaily: Tests all pass 💯

spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'airport'
+
+# subject(:airport)
+
+describe Airport do
+  it 'checks an airport exists' do
+    expect(Airport).to be_truthy 
+  end
+
+  it { is_expected.to respond_to :take_off }
```
lookupdaily: You could expand this test - by checking the expected value of `@amount_of_planes` after taking off 

e.g. 
`subject.take_off(plane)`
`expect(subject.amount_of_planes).to eq []`

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+require './lib/plane.rb'
+
+class Airport
+
+  attr_reader :plane
```
lookupdaily: Do you need an attribute reader for `@amount_of_planes` so you can test the outcome of land and take plane in your tests? (see airport_spec - line 10)

PR #1719, opened 2020-02-21
student version

lxierita:
my version of airport challenge without bonus feature implemented

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,2 @@
+class Airport
+end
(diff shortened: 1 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+	end
+
+	def fly_plane(plane)
+		if !has_plane?(plane)
+			raise StandardError, "This plane is not at the airport"
+		elsif @is_stormy
+			raise StandardError, "Too stormy to take off"
+		else 
+			plane.takes_off
+			@planes.delete_if { |landed_plane| landed_plane == plane }
```
tatsiana-makers: What's the motivation behind using `delete_if` here? Would `delete` work here instead?

PR #1720, opened 2020-02-21
Feb 2020 - Airport challenge submission

noravdh:


lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    @planes.delete(plane)
+    plane.starts_flying
+  end
+
+  def generate_number
+    rand(10)
+  end
+
+  def weather
+    if self.generate_number <= 1
```
Ion88: I think stormy and sunny weather should be half half

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'plane'
```
Ion88: Top 🥇 
I like your airport

spec/airport_spec.rb
```diff
@@ -0,0 +1,123 @@
+require_relative '../lib/airport'
+describe Airport do
+  context "landing planes" do
+    it "plane is in the airport once it has landed" do
+      plane = double("plane")
+      allow(subject).to receive(:weather) { "sunny" }
+      allow(plane).to receive(:stops_flying)
+      subject.land(plane)
+      expect(subject.planes).to eq [plane]
```
Ion88: Why you store plane in [ ] ? What does it mean?

README_solution.md
```diff
@@ -0,0 +1,31 @@
+This program emulates a working airport. It has the functionality to land planes and let planes take off, and responds to changes in weather.
```
tatsiana-makers: This is a great document - thank you for writing up your thoughts, it's really useful as a reviewer to get a sense of the code I'm about to read (:

README_solution.md
```diff
@@ -0,0 +1,31 @@
...
+   Methods that work on airport<br/>
+    - capacity<br/>
+      When a new airport is made, you can pass in the capacity of the airport.
+      If no capacity is passed in, it will default to a default capacity. Calling the capacity method on an airport will return the capacity.<br/>
+    - land<br/>
+      When you land a plane, you need to pass in the plane as an argument.
+      The method will then check to see if the airport is full, if the weather is stormy, and whether the plane is already in the airport.
+      If none of those are true, it will land the plane<br/>
+    - take_off<br/>
+      You can pass in a specific plane to take off. If you don't, the last plane to land will take off.
```
tatsiana-makers: Nice!

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  attr_reader :planes, :capacity
+  DEFAULT_CAPACITY = 50
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail "The airport is full" if full?
+    fail "It is too stormy to land" if weather == "stormy"
```
tatsiana-makers: One refactor we could do is to extract out the `if weather == "stormy"` into a separate method since it's used several times in different methods

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    fail "This plane is already in the airport" if @planes.include?(plane)
```
tatsiana-makers: This is interesting! What happens if the plane lands at Airport1, and then tries to land at Airport2?

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    @plane = plane
+    @plane.stops_flying
+    @planes << @plane
+  end
+
+  def take_off(plane = @plane)
```
tatsiana-makers: This is a cool use of the `@plane` instance variable! `@last_landed_plane` might be a clearer name for this variable

spec/airport_spec.rb
```diff
@@ -0,0 +1,123 @@
+require_relative '../lib/airport'
```
tatsiana-makers: Nice tests!

tatsiana-makers (commented):
Great work on the Friday challenge! I've left a few minor suggestions and opportunities for refactoring. I've also left a question around a plane landing in another airport while currently landed at an airport which might be something interesting to consider if you feel you want to keep going with this challenge

PR #1721, opened 2020-02-22
submit sollution

dariathompson:


sgneha:
**Appreciation**

* Firstly completed the whole challenge is appreciated.
* Really liked how descriptive your README.md is.
* You have written error messages first then used it in expect,is really clean.
  it 'raises an error if weather is stormy' do
      error = "Cannot land the plane because weather is stormy"
      allow(subject.weather).to receive(:stormy?).and_return(true)
      expect { subject.land(plane) }.to raise_error error
    end

**Suggestions**
* It would have been better if you checked the current status of plane in Plane class.

This plane has already landed
There are no planes to take off

* In class Weather it would have been better if its more descriptive how randomness of weather is choosen. 

def stormy?
    rand(6) == 5
  end

PR #1723, opened 2020-02-23
Pull request for Airport Challenge

jarasmar:


rafahg:
jara you are the best!!!

spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  it "allows to #land a plane in the airport" do
+    expect(subject).to respond_to{:land}
+  end
+
+  it "allows a plane to #take_off" do
+    expect(subject).to respond_to{:take_off}
+  end
+
+  it "confirms that a plane has take_off" do
+    subject.land(Plane.new)
```
plkujaw: Use `double` instead of actually creating new Plane instance to isolate tests for Airport class from other classes dependencies.
ie:
```
plane = double("plane")
subject.land(plane)
```
...

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def land(plane)
+    fail "Airport is full" if full?
+    @planes_in_airport << plane
+  end
+
+  def take_off
+    fail "Airport is empty" if empty?
+    @planes_in_airport.pop
+    print "Plane has taken off"
```
plkujaw: use `return` instead of `print` as we want this method to return certain message/status not just print it so we can use returned value somewhere to interact with

plkujaw:
Well done for doing it all by yourself! :)

PR #1724, opened 2020-02-23

stellenberger:
Pull request for Makers Academy Week 1 Weekend Challenge

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,3 +1,6 @@
+require_relative './plane.rb'
+require_relative './weather.rb'
+
 class Airport
 	DEFAULT_CAPACITY = 5
 	attr_reader :planes, :capacity, :weather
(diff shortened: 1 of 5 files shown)


poliver24:
**Pros:**

Your stormy method is nice and concise - (however could be rewritten to incorporate more standard english.)

Test "it....." statements well written and very clear to understand

Tests are neat and the logic is easy to see. 

I like that you included the user stories as comments in the _spec file, this makes reading your tests much clearer and more logical. 

_> it 'has a default capacity' do
> 		expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
> 	end
>_ 

I like how you have tested the DEFAULT_CAPACITY, allowing for future DEFAULT changes.

**Suggestions:**

Is separate Weather class necessary? Perhaps could be incorproated as a method?

_'The plane is no longer in the airport'_ - feel that this output could be reworded to be a little clearer

_

> def land(plane)
> 		fail 'this airport is full!' if @planes.length >= @capacity

_

perhaps for your guard conditions you could apply more SRP. 
e.g.

> _" fail 'this airport is full!' if full? "_

rafahg:
Stephen, I think that your code is brilliant and shows a great understanding on how classes interact with each other. I particularly like how you created the storm method within the Weather class. 
In one word, excellent.
- One suggestion ---- I dont really have any, i think that my code at the moment is similar to yours and even worse because i didnt really created more classes than Airport and Plane.
- One improvement ---- Just keep the work. Nothing to improve at this point of my knowledge, honestly.

PR #1725, opened 2020-02-23
Week 1 Weekend Challenge: Airport

plkujaw:


jarasmar:
Appreciation: I like how it covers all the extra considerations, like a plane cant takeoff from an airport it is not in currently.  Also seeing how you use the doubles has helped me understand how they work.
Suggestion: I would recommend trying coding it by yourself next time, at least the very basics, to help you fix those concepts in your mind.  And then upgrading them with other people code to help you also get new logics and procedures.

PR #1727, opened 2020-02-23
Karla's airport challenge

KG700:


lukedonov:
I like how concise your tests are, they have been refactored well and are clean and clear to read. 

I think in your take-off and landing methods you could refactor slightly using private methods.

PR #1728, opened 2020-02-23
Airport Challenge complete

WilliamJGrace:
First commit; added two files airport.rb and plane.rb; created tests to check methods work, takeoff and land, added a hanger instance variable, added a stormy method, tested that if stormy was true plane can't takeoff and land; refactored code

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,58 @@
+require 'airport.rb'
+
+describe Airport do
+  it "plane can land and takeoff" do
+    airport = Airport.new
+    plane = Plane.new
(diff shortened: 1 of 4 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'airport.rb'
+
+describe Airport do
+  it "plane can land and takeoff" do
+    airport = Airport.new
```
Thatguy560: Should work if you use let(:airport) { Airport.new } 
same with creating a new instance of plane let(:plane) { Plane.new }

PR #1729, opened 2020-02-23
Airport weekend challenge 

sgneha:


dariathompson:
I love your code! I like how you added state for Plane class, it makes it easy to track and read
```
class Plane
  attr_accessor :state
  def initialize
    @state = :landed
  end
end
```
I like using Module for Weather
```
module Weather
  def stormy?
    random_states == :stormy
  end
    private
  WEATHER_STATES = [:stormy, :sunny, :sunny, :sunny]
  def random_states
    WEATHER_STATES.sample
  end
end
```

Suggestions: 
add more description for README 
add error variables for tests, will make lines shorted and easier to read

PR #1730, opened 2020-02-23
Sophia Bell airport_challenge pull request

Kittaru87:


srikaanthtb:
code is easy to read and things are clearly defined, easy to understand, plane.rb could possibly be refactored, it's not immediately obvious what weather.rb is trying to do. Adding a comment will help.

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
+require_relative 'airport'
+
+class Plane
+    
+  attr_reader :location, :status
+
+  def initialize
```
srikaanthtb: file could be refactored into just the initialise method

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+require_relative 'airport'
+
+class Weather
+
+  def stormy?
```
srikaanthtb: not immediately obvious what this does

spec/plane_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'plane'
```
srikaanthtb: easy to read with well defined tests

lib/weather.rb
Kittaru87: now changed

srikaanthtb:
lol i completely forgot about this, doing some long overdue github cleanup are we?

PR #1733, opened 2020-02-23
Airport challenge. Suggested solution.

rafahg:


stellenberger:
I really like this clean code!!!! Especially the methods that follow the SRP, where each method follows one purpose. In particular, methods like full? and storm? 

One thing that could be improved is your storm method that is inside the Airport class. It picks a random number from 1 to ten and then returns true if the number is 4. This is a bit confusing as 4 is a particularly odd number to pick from 1 to 10, perhaps use the rand(6) == 5 method, or better, use an array with stormy and sunny and **sample** the array to get a random element.

poliver24:
**Pros:**

Really nicely written README file - really like that this shows your approach and the logic behind your thinking. For me this is super helpful in helping me to read your code and see your approach.

**Suggestions:** 

```
def storm?
    storm = rand(1..10)
    true if storm == 4

  end
```
I feel that your storm? method could be written differentl to incorporate more plain english and be easier to understand. 

`attr_reader :planes`

I THINK? that you might be able to use the attribute reader defined at the beginning of your code file to also incorporate the :plane variable.

`class Plane
  def plane?
    true
  end
end`

I am unsure where the "plane?" method is used within your code?

PR #1734, opened 2020-02-23
Build and test code for airport challenge

lilawalker:
The code in this pull request satisfies the requirements to build a REPL app that allows the user to land and take off planes from airports. Planes cannot land or take off if the weather is stormy. The airport's default capacity for planes is set to 100 but can be overwritten by the system administrator.


Swarbzz:
1. Incredibly readable and very well laid out.
2. I don't really have a suggestion to improve your code :)

PR #1735, opened 2020-02-24
weekend challenge attempt

srikaanthtb:


srikaanthtb:
user stories complete:

As an air traffic controller 

As an air traffic controller 

As an air traffic controller 
To ensure safety 

As the system designer

As an air traffic controller 
To ensure safety 

As an air traffic controller 
To ensure safety 


N.B for the last two user stories I couldn't get the Rspec tests working

lib/plane.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
+  def initialize
+    @flying = false
```
Kittaru87: When a new plane is called it isn't in an airport so I would suggest changing this initialize to true

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
+  def stormy?
+    if calculate == 3
+    return true
+  else
+    return false
+  end
+end
```
Kittaru87: This can be refactored

Kittaru87:
Code is easy to read *thumbs up*

Suggestion: Update README to say the problems you faced and what you would like to change in the future.

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'weather'
+require_relative 'plane'
+
+class Airport
+  CAPACITY = 3
+  attr_accessor :planes, :weather, :capacity
+  def initialize
+    @planes = []
+    @capacity = CAPACITY
+    @weather = weather
```
Kittaru87: I would suggest calling the weather class, eg @weather = Weather.new, it might help pass the stormy tests

srikaanthtb: change done

PR #1736, opened 2020-02-24
Ziyao Zheng Feb 2020 

educatedflea:


lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    @planes_in_the_airport = []
+  end
+
+  # User story 1 : the controller can instruct planes to land
+  # User story 3 : prevent landing when capacity has reached
+  def land(plane)
+    fail 'The airport has reached landing capacity' if @planes_in_the_airport.length >= @capacity
+    fail 'Cannot land when weather is stormy' if stormy?
+    @planes_in_the_airport << plane
+    @plane = plane
```
tatsiana-makers: This line seems to be the only place that we use the `@plane` instance variable. What is the purpose of this instance variable - do we still need it?

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  end 
+
+  # User story 2 : the controller can instruct planes to take off and confirm it is taken off
+  # to simplify the problem, assume that the one landed last will take off first although it is not necessarily realistic
```
tatsiana-makers: Thank you for this really helpful comment with your assumption - it might be interesting to consider what the code would look like if we didn't make the assumption that the last plane landed will be the one that takes off

tatsiana-makers: Thank you for this really helpful comment with your assumption! - it might be interesting to consider what the code would look like if we didn't make the assumption that the last plane landed will be the one that takes off

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  # User story 4 : archtect can override the default capacity 
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes_in_the_airport = []
+  end
+
+  # User story 1 : the controller can instruct planes to land
+  # User story 3 : prevent landing when capacity has reached
+  def land(plane)
+    fail 'The airport has reached landing capacity' if @planes_in_the_airport.length >= @capacity
```
tatsiana-makers: It might be nice to pull out the logic of `@planes_in_the_airport.length >= @capacity` into its own method

lib/plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
```
tatsiana-makers: What features might we care about in the plane class? 
Here are some questions which might be interesting to consider:
- can a plane that is already landed at an airport land at another airport without taking off?
- can a plane that has taken off but not landed take off again?

spec/airport_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  describe '#land' do
+    it 'can land airplanes' do 
+      expect(subject).to respond_to(:land).with(1).argument
+    end
+
+    it 'can instruct a plane to land' do 
+      plane = Plane.new
+      allow(subject).to receive(:stormy?).and_return(false)
+      expect(subject.land(plane)).to eq plane
```
tatsiana-makers: If you were to manually test the airport class in `irb`, what sort of steps would you take to convince yourself that a plane has successfully landed?

It might be useful to consider how we did this in the [Translating user stories into code workshop](https://github.com/makersacademy/skills-workshops/tree/master/week-1/translating_user_stories_into_code) and see if we can form an assertion based on our manual testing in `irb`

spec/airport_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      expect{subject.land plane}.to raise_error "The airport has reached landing capacity"
+    end 
+
+    it 'prevents landing when stormy' do
+      plane = Plane.new
+      allow(subject).to receive(:stormy?).and_return(true)
+      expect{subject.land(plane)}.to raise_error 'Cannot land when weather is stormy'
+    end 
+
+    it 'allows to land when sunny' do
```
tatsiana-makers: This test is identical to the one [here]( https://github.com/makersacademy/airport_challenge/pull/1736/files#diff-db8accee82ee18b6bcf0b34c6ecc8fb7R33). Do we need both test cases?

spec/airport_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      subject.land(plane)
+      expect(subject.take_off).to eq plane
+    end 
+
+    it 'confirms the plane has taken off' do
+      plane = Plane.new
+      allow(subject).to receive(:stormy?).and_return(false)
+      subject.land(plane)
+      subject.take_off 
+      expect(subject.planes_in_the_airport).not_to include(plane)
```
tatsiana-makers: This is a great assertion and might be something to inspire us when considering what assertion to write in the `it 'can instruct a plane to land' do` test case

PR #1737, opened 2020-02-24
airport challenge completed

poliver24:


rafahg:
Patrick i think that your code is brilliant. Well done

rafahg:
the only suggestion i have for you at this point is within the storm? method you can do it without an arrary, just for simplicity. Anyway, i think your way is super valid anyway.

rafahg:
To be honest i cannot improve your code in anyway. I did more or less the same than you with the same approach.

stellenberger:
This is some really amazing code, i like that you follow SRP strictly and your code is clear and consise.

Your tests are also laid out in a manner that is clear what you are trying to achieving, using context to describe weather tests and capacity tests. Nice!



```
    print 'The plane has left the airport'

    @airport.pop
  end
```
This is good, but to avoid confusion in future when adding more code, maybe remove the plane BEFORE you say 'The plane has left the airport'. so the following

```
    @airport.pop
    print 'The plane has left the airport'
```

Although I really like your method of sampling an array with stormy and sunny (it reads super well), the method itself could be in a separate class, Weather. If you wanted to added further weather functionality, itll be easier with a weather class!

PR #1738, opened 2020-02-24
weekend challenge submission

probablymenext:
- User stories covered
- Edge cases covered: planes that are already flying cannot take off and/or be in an airport; planes that are landed cannot land again


lukedonov:
I like how you created all of the different weather possibilities and wrote them into the tests. I think building a fleet and tracking each individual planes position is a good idea.

I think you could refactor your code to raise errors when planes can’t land or take-off instead of return false. It would avoid you having to use a long if else statement within your method and send the user a useful message.

PR #1739, opened 2020-02-24
Airport Challenge WIP

ghost:


lilawalker:
1. Appreciation: It is clear that after the first attempt at the tests that you learned a lot about simplifying the tests and the second attempt is much clearer. 
2. Suggestion: Each class and spec should be in a separate file. In the Plane class ensure to not pass any arguments as you you will not set these when you create a new instance of plane. Initialise status as opposed landed, so that status can be changed to landed or flying depending on where the plane is. In airport spec you don't need to check whether plane is working.

PR #1742, opened 2020-02-24
airport challenge

jamesoddy8:


lib/airplane.rb
```diff
@@ -0,0 +1,11 @@
+class Airplane
```
lroberts77: #99% of the tests, methods and classes are well named and described which makes the code readable

spec/aiport_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  # TEST: raise error if you try and takeoff a plane that is not at the airport
+  it "Raise error if plane already in sky and try takeoff" do
+    subject.sunny = true
+    allow(airplane).to receive(:landed).and_return(false)
+    error = "Airplane is already in the sky!"
+    expect { subject.takeoff_plane(airplane) }.to raise_error error
+  end
+
```
lroberts77: #improvement 
#put comments between each line so it explains what is happening on each line

PR #1743, opened 2020-02-24
airport task complete

Nimzyow:


Benjamaker (commented):
I really like the overall separation of methods. Mostly conforms to SRP.
I like the overall design as well. I think it accurately represents a real airport.

Benjamaker (changes requested):
You could refactor full? method as the ternary here is a bit unnecessary. I would also suggest encapsulation of both full? and empty? methods.

PR #1745, opened 2020-02-24
Airport Challenge - Peter

Thatguy560:


spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'airport.rb'
+require 'plane.rb'
+
+describe Airport do 
+  DEFAULT_CAPACITY = 10
```
WilliamJGrace: Don't need to set DEAFAULT_CAPACITY to 10 if you set it in the airport.rb file. Just reference it using Airport::DEFAULT_CAPACITY

PR #1746, opened 2020-02-24
Luke Donovan Airport Challenge

lukedonov:


KG700:
I like that you’ve initiated an airport with a default weather instance, and made an airport's capacity something that you can change.

Suggestions: 
- create a attr_writer or attr_accessor for @capacity rather than using the method def capacity(num) ; @capacity = num ; end
In your tests:
- you can remove weather = Weather.new as it’s not needed
- Rather than creating a plane instance for each test, you can create one at the top using let(:plane) { Plane.new }

PR #1748, opened 2020-02-24
Feb 2020 airport (avril)

AvrilHarding:
This is my airport challenge code. I wrote the tests and code for the user stories.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -4,4 +4,8 @@ class Airport
     Plane.new
     # creates a plane so it land at the airport
   end
+
+  def take_off
+    Plane.new
+  end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,6 +1,7 @@
 class Airport
 
-  def take_off
-
+  def land
+    Plane.new
+    # creates a plane so it land at the airport
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,11 +1,19 @@
 class Airport
 
+  def initialize
+    @planes = []
+  end
+
   def land
     Plane.new
+    fail 'Airport full, do not land' if @planes.count >= 5
+    @planes.pop
     # creates a plane so it land at the airport
   end
 
-  def take_off
+  def take_off(plane)
     Plane.new
+    fail 'No planes at the airport' unless @planes.empty?
+    @planes << plane
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,6 @@
+class Airport
+
+  def take_off
+
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -3,14 +3,24 @@ require 'airport'
 describe Airport do
  it  { is_expected.to respond_to :land }
 # the airport is expected to respond to the method take_off, take_off colons so this is not going to change
+ describe '#landed' do
+   it 'can not land when airport is full' do
+     5.times { subject.land Plane.new }
+     expect { subject.land Plane.new }.to raise_error "Airport full, do not land"
+   end
+ end
+
  it 'lands a flying plane' do
-  plane = subject.land
-  expect(plane).to be_landed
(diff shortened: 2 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,19 @@
+class Airport
+
+  def initialize
+    @planes = []
+  end
+
+  def land
+    Plane.new
+    fail 'Airport full, do not land' if @planes.count >= 5
+    @planes.pop
```
noravdh: this will take a plane out of the planes array, but is that what you want to happen during landing? Or would you want a plane to be added to the array?

lib/airport.rb
```diff
@@ -0,0 +1,19 @@
...
+    # creates a plane so it land at the airport
+  end
+
+  def take_off(plane)
+    Plane.new
+    fail 'No planes at the airport' unless @planes.empty?
```
noravdh: I think this should maybe be _if_, instead of _unless_?

spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'airport'
+# line 1 is to require the access to the lib airport ruby file
+describe Airport do
+ it  { is_expected.to respond_to :land }
+# the airport is expected to respond to the method take_off, take_off colons so this is not going to change
+ describe '#landed' do
+   it 'can not land when airport is full' do
+     5.times { subject.land Plane.new }
+     expect { subject.land Plane.new }.to raise_error "Airport full, do not land"
```
noravdh: great test set up!

lib/airport.rb
AvrilHarding: yes, that makes sense! :D

lib/airport.rb
```diff
@@ -0,0 +1,19 @@
+class Airport
+
+  def initialize
+    @planes = []
+  end
+
+  def land
```
tatsiana-makers: Should the land method take an input argument similar to the `take_off` method to allow us to land a specific plane?

lib/airport.rb
```diff
@@ -0,0 +1,19 @@
...
+    Plane.new
+    fail 'Airport full, do not land' if @planes.count >= 5
+    @planes.pop
+    # creates a plane so it land at the airport
+  end
+
+  def take_off(plane)
+    Plane.new
```
tatsiana-makers: We currently don't use the instance of a plane returned by `Plane.new`. Do we need this line here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+   end
+ end
+
+ it 'lands a flying plane' do
+  plane = Plane.new
+  subject.land(plane)
+  expect(subject.land).to eq plane
```
tatsiana-makers: If we ran the `Airport` class in `irb`, what would we expect to see if the `land` method was working correctly? Could we use this to write our assertion?

spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+ describe '#landed' do
+   it 'can not land when airport is full' do
+     5.times { subject.land Plane.new }
+     expect { subject.land Plane.new }.to raise_error "Airport full, do not land"
+   end
+ end
+
+ it 'lands a flying plane' do
+  plane = Plane.new
+  subject.land(plane)
```
tatsiana-makers: It looks like we're calling the `land` method with an argument, but looking at the method definition, it doesn't accept an argument. It looks like this is the cause of errors when running the tests with `rspec`.

spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  expect(subject.land).to eq plane
+  # the airport needs to land a flying plane, the plane needs to land, i expect the plane to be landing.
+end
+
+ it 'instructs a plane to take off' do
+   plane = Plane.new
+   subject.take_off(plane)
+  expect(subject.take_off).to eq plane
```
tatsiana-makers: This is a similar place where we can consider what our assertion should be (similar to the land method test above)

PR #1749, opened 2020-02-24
Airport

Nanariane:
This is the first commit that I have done

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,8 @@
+class Airport
+  def land
+  end
+
+  def take_off
+  end
+  
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,8 +1,13 @@
+require_relative 'plane'
+
 class Airport
   def land
+    # Plane.new
   end
 
   def take_off
+    Plane.new
+    return "Plane has left the airport"
   end
-  
+
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,26 +1,26 @@
 require_relative 'plane'
 
 class Airport
-
+  DEFAULT_CAPACITY = 10
   attr_reader :plane
+  attr_accessor :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @plane = []
+  end
 
 
-# def initialize
-#   @airport = []
-  #
   def land
     Plane.new
   end
 
 
   def store(plane)
-    @plane = plane
+    fail 'Airport full' if full?
+    @plane << plane
+    return plane
   end
-  #
-  # def plane
-  #   @plane
-  # end
-
 
 
   def take_off
@@ -28,4 +28,10 @@ class Airport
     return "Plane has left the airport"
   end
 
+end
+
+
+private
+def full?
+  @plane.count == @capacity
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,9 +1,24 @@
 require_relative 'plane'
 
 class Airport
+
+
+# def initialize
+#   @airport = []
+  #
   def land
-    # Plane.new
+    Plane.new
   end
+
+
+  def store(plane)
+    plane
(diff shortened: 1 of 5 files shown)


lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
```
victranfield: Do you need an attribute accessor here?

Nanariane: Yes I needed one .. but didn't put it in for some reasons ..

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    @plane = []
+  end
+
+
+  def land
+    Plane.new
+  end
+
+
+  def store(plane)
```
tatsiana-makers: What is the reason behind having both the `land` and the `store` methods?

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    fail 'Airport full' if full?
+    @plane << plane
+    return plane
+  end
+
+
+  def take_off
+    Plane.new
```
tatsiana-makers: Do we need this line of code? We're not using the instance of the Plane class that is created here

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    return "Plane has left the airport"
+  end
+
+end
+
+
+private
+def full?
+  @plane.count == @capacity
+end
```
tatsiana-makers: Should this block of code be within the code for the `Airport` class?

lib/plane.rb
tatsiana-makers: Interesting question! What attribute accessor would we define here and what would we read out using it?

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+describe Airport do
+  plane = Plane.new
+  it 'can land a plane' do
+    expect(subject).to respond_to(:store).with(1).argument
+  end
+
+  it 'stores something' do
+    plane = Plane.new
+    expect(subject.store(plane)).to eq plane
```
tatsiana-makers: If we used the `Airport` class in `irb`, what behaviour would we use to convince ourselves that the `store` method was working correctly? Could we use this to write our assertion here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  #     plane = Plane.new
+  #     subject.store(plane)
+  #     expect(subject.land).to eq plane
+  #   end
+  # end
+
+  it 'allows store planes' do
+    plane = Plane.new
+    subject.store(plane)
+    expect(subject.plane).to eq plane
```
tatsiana-makers: The assertion here looks like it might be actually almost be what we'd do in `irb` to convince ourselves that the `Airport.store` code was working (answering the question I posed in the comment above). 

This test is currently failing when we run `rspec` but I wonder if we can use the output to help us understand why it's failing and modify this test to a passing state

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+
+
+describe Airport do
+  it 'instruct a plan to take off' do
+    expect(subject).to respond_to :take_off
+  end
+
+    it 'confirms no longer in the airport' do
```
tatsiana-makers: How might a plane taking off impact the airport? For example, would the instance variables be impacted in any way? Let's capture this in a test

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    it 'has a default capacity' do
+      expect { subjecty.land }.to eq Airport::DEFAULT_CAPACITY
+    end
+
+
+    it 'shows when airport is full' do
+      # 20.times { subject.store Plane.new }
+      expect { subject.store Plane.new }.to raise_error 'Airport full'
+    end
```
tatsiana-makers: Having a look at the `rspec` output for these test might help us with updating the tests to get to a passing state.

PR #1750, opened 2020-02-24
Airport challenge

victranfield:
First commit - plane responds to landing - will raise airport full error
plane responds to takeoff

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,29 @@
+
+class Airport
+  DEFAULT_CAPACITY = 3
+  attr_accessor :capacity, :plane, :airport
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @airport = []
+  end
+
+  def landing(plane)
+    fail 'airport full' if full?
+    @airport << plane
+  end
+
+  def takeoff
+    fail 'unable to takeoff' if empty?
+    return "Taken off"
+  end
+
+  private
+    def full?
+      @airport.count >= DEFAULT_CAPACITY
+    end
+
+    def empty?
+      @airport.empty?
+    end
+
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,28 @@
+require "airport"
+
+describe Airport do
+  it { is_expected.to respond_to :landing}
+  it { is_expected.to respond_to :takeoff}
+
+
+
+  describe '#landing' do
+    it 'raises an error when full' do
+      # puts subject.capacity.times
+      subject.capacity.times {subject.landing double :plane}
+      # subject.landing double(:plane)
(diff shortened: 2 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+    @airport = []
+  end
+
+  def landing(plane)
+    fail 'airport full' if full?
+    @airport << plane
+  end
+
+  def takeoff
+    fail 'unable to takeoff' if empty?
```
Nanariane: Why did you use the empty method here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+describe Airport do
+  it { is_expected.to respond_to :landing}
+  it { is_expected.to respond_to :takeoff}
+
+
+
+  describe '#landing' do
+    it 'raises an error when full' do
+      # puts subject.capacity.times
+      subject.capacity.times {subject.landing double :plane}
```
Nanariane: What does double refers to here?

victranfield: https://relishapp.com/rspec/rspec-mocks/docs

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      # subject.landing double(:plane)
+      expect { subject.landing double(:plane) }.to raise_error 'airport full'
```
lindseydew: Excellent test set up! 
Do you need a double here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+    end
+end
+
+#describe '#takeoff' do
+
+#  it 'returns planes' do
```
lindseydew: Can see you are completely on the right lines with writing a test for the take_off method - great job! It looks like you're not quite sure what to assert on. 

If we ran the airport code in irb, what would you look at to confirm the take off method is working? Could you use that to help decide what would go in the expect? 

It might help to revisit this workshop: https://github.com/makersacademy/skills-workshops/tree/master/week-1/writing_tests

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+
+
+  describe '#landing' do
+    it 'raises an error when full' do
+      # puts subject.capacity.times
+      subject.capacity.times {subject.landing double :plane}
+      # subject.landing double(:plane)
+      expect { subject.landing double(:plane) }.to raise_error 'airport full'
+    end
```
lindseydew: How would you test your land method in the case where the airport is not full?

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
+
+class Airport
+  DEFAULT_CAPACITY = 3
+  attr_accessor :capacity, :plane, :airport
```
lindseydew: Do you need :plane in your attr_accessors here?

PR #1751, opened 2020-02-24
The best airport

Ion88:


spec/airport_spec.rb
```diff
@@ -12,10 +12,11 @@
     10.times do
       airport.land(plane)
     end
-    expect { airport.land(:plane) }.to raise_error 'Airport is full'
+    expect { airport.land(plane) }.to raise_error 'Airport is full'
```
AvrilHarding: good refactoring!

spec/airport_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    10.times do
+      airport.land(plane)
+    end
+    expect { airport.land(:plane) }.to raise_error 'Airport is full'
+  end
+end
+
+
+  it { is_expected.to respond_to(:take_off).with(1).argument }
+end
```
AvrilHarding: very clear syntax, easy to read and understand.

Ion88: It's clear because is almost nothing written, but thx 👍

spec/airport_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  subject(:airport) { described_class.new(10) }
+  let(:plane) { double :plane }
+
+  describe '#land' do
+  it { is_expected.to respond_to(:land).with(1).argument }
+
+  it 'not allow planes land when airport is full' do
+    10.times do
+      airport.land(plane)
+    end
```
lindseydew: What would the test look like in the case where we call the land method at the airport and the capacity wasn't full?

spec/airport_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect { airport.land(plane) }.to raise_error 'Airport is full'
+  end
+end
+
+describe '#take_off' do
+  it { is_expected.to respond_to(:take_off).with(1).argument }
```
lindseydew: Nice, I can see you are test driving the take off method. 

What would the next test look like to check the logic of the take_off method? What would you need to observe to check that it is doing what you expect?

PR #1753, opened 2020-02-24
Jo Redfern airport challenge

JoRedfern:
please have a look at my code :-)


spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "airport"
+describe Airport do
+  it "allows plane to land" do
+    airport = Airport.new
+    plane = Plane.new
+    airport.stormy? == false
+    expect(airport.land(plane)).to eq(1)
```
lindseydew: This test totally works as written, but what things in the class changes as a result of calling land? Could we assert on that?

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+  it "allows plane to take off" do
+    airport = Airport.new
```
lindseydew: Excellent test design here. It is really clear from your tests what functionality your code has 🥇

spec/airport_spec.rb
lindseydew: Looking at how the test is designed in bank account: https://github.com/makersacademy/skills-workshops/tree/master/week-1/writing_tests in the deposit method, is there a way to take a similar approach for the land and take_off method?

PR #1754, opened 2020-02-24
Olivia's airport submission

othetgyi:


lib/airport.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize
+    @total_planes = 0
+  end
+
+  def land(plane)
+    @total_planes += 1
+    @location = "airport"
+  end
+
+  def number_of_planes
```
tatsiana-makers: This method definitely works to give us visibility on the value of an instance variable. Another way we could do a similar thing would be to define an attribute reader

lib/airport.rb
```diff
@@ -0,0 +1,22 @@
+require_relative 'plane'
+
+class Airport
+  def initialize
+    @total_planes = 0
+  end
+
+  def land(plane)
+    @total_planes += 1
+    @location = "airport"
```
tatsiana-makers: When we set `@location` here, we're setting the instance variable called `@location` in the Airport class. Is this the class where we want this `@location` instance variable to live?

spec/airport_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  describe '#land' do
+    it 'instructs plane to land' do
+      expect(Airport.new).to respond_to(:land).with(1).argument
+    end
+
+    it 'adds one to the total number of planes when planes land' do
+      airport = Airport.new
+      plane = Plane.new
+      expect{airport.land(plane)}.to change{airport.number_of_planes}.by(1)
```
tatsiana-makers: This is so cool! I haven't seen a matcher like this before, nice work!

tatsiana-makers: Also this is a great assertion to convince ourselves that the number of planes at the airport changes when the plane lands !

lib/airport.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def number_of_planes
+    @total_planes
+  end
+
+  def takeoff(plane)
+    @plane
```
tatsiana-makers: What is this line of code doing? Do we need it?

spec/airport_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    it 'sets the planes position to airport' do
+      airport = Airport.new
+      plane = Plane.new
+      allow(airport).to receive(:land).and_return("airport")
+      expect(airport.land).to eq("airport")
```
tatsiana-makers: Does this assertion convince us that the plane's `@position` is the airport? Is there something similar to the test block above that we could build that would be a stronger assertion for this?

PR #1756, opened 2020-02-24
Funmi's Airport challenge in bitesize

FunmiBADev:
Coverage:

Airport.rb
1. Plane_land: covers permit plane to land but stormy? weather check is not covered
2. Plane_take_off: covers permit plane to take off including conditional stormy wether check and also send confirmation that plane has taken off and left the aiport.

Plane.rb
1. Check plane has taken off
2. Confirm plane has taken off and left the airport


lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+class Airport
+
+	attr_reader :planes
+	attr_accessor :capacity
+
+
+	DEFAULT_CAPACITY = 35
+
+	def initialize
+		@planes = []
```
tatsiana-makers: We're not currently using the `@planes` instance variable anywhere in the code. What was your motivation behind having this instance variable? Are there some methods that might use this instance variable in some way?

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+		@capacity = capacity
```
tatsiana-makers: Where is the `capacity` variable on the right hand side of the `=` coming from? It might be useful to do a manual test in `irb` and play around with trying to create an instance of the Airport class with a specific capacity (for example, 5) and testing that it does work.

Similarly, once we've got a manual test in irb, it'd be great to capture that in an rspec test

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+	end
+
+	def plane
+		@plane
+
+	end
+
+
+	def plane_land(plane)
+		@plane = plane
```
tatsiana-makers: I'm curious why we have `@plane` as an instance variable of the `Airport` class - what is the purpose of the airport keeping track of one specific plane?

Could this be a place that we make use of the `@planes` instance variable in some way?

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
+
+require_relative "plane"
+
+class Airport
+
+	attr_reader :planes
+	attr_accessor :capacity
+
+
+	DEFAULT_CAPACITY = 35
```
tatsiana-makers: We're not currently using `DEFAULT_CAPACITY` anywhere in the code. What is the goal of having `DEFAULT_CAPACITY`? Using a manual test in `irb` how would we convince ourselves that the `DEFAULT_CAPACITY` is working? 

Once we've got a manual test in `irb`, it'd be great to capture that in an rspec test

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'airport'
+
+describe Airport do 
+	# context "allows plane to land" do
+it { is_expected.to respond_to(:plane_land).with(1).argument }
+it "plane lands" do
+	plane = double :plane
+	expect(subject.plane_land (plane)).to eq plane
```
tatsiana-makers: If we build off of the examples in the [Translating User Stories into Code workshop](https://github.com/makersacademy/skills-workshops/tree/master/week-1/translating_user_stories_into_code), what manual testing would convince us that the plane landing is working? Can we then translate that into an assertion in `rspec`?

PR #1759, opened 2020-03-21
Airport ready!

hturnbull93:
I enjoyed the challenge! I hope the readme isn't too long! I know the contributions markdown said brief!

In the code are Airport and Plane classes. Airports has methods to land and take off Planes, which adds or removes the Plane from the airport's hangar attribute, and updates the Plane's attribute to reflect if it is in flight or not, using the Planes setter methods. There is also an airport method to return if the weather is currently stormy, in which case no planes may land or take off.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,3 @@
+class Airport
+  
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -17,9 +17,7 @@ class Airport
 
   def take_off(plane)
     preflight_checks(plane)
-    taxi(plane)
-    plane.takes_off
-    plane
+    take_off_sequence(plane)
   end
   
   def stormy?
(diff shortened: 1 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,71 @@
...
+  DEFAULT_CAPACITY = 5
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+  end
+
+  attr_reader :hangar
+
+  def land(plane)
+    landing_checks(plane)
```
alexakearns: I like that you called the method that allows you to  know whether the plane can land or not and it contains all scenarios where it can't

spec/airport_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+        Airport::DEFAULT_CAPACITY.times { subject.land(Plane.new) }
+        expect { subject.land(boeing) }.to raise_error 'Hangar full.'
+      end
+      it 'marks the plane as not in flight' do
+        subject.land(boeing)
+        expect(boeing).to_not be_flying
+      end
+      it 'planes not in flight cannot be landed, raises error' do
+        subject.land(boeing)
+        expect { subject.land(boeing) }.to raise_error 'Cannot land planes that are not in flight.'
```
alexakearns: 👏

PR #1760, opened 2020-03-22
Airport challenge update

FayeCarter:
Recap of week 1 learnings and application to new project

**Included in the change:**
- User stories completed
- Tests passing
- Have not worked on edge cases

**What is incomplete?**
- Edge cases not started
- After implementing stormy? method, I haven't had a chance to refactor the tests. I have a lot of repetition stormy or not stormy airport objects in tests
- Haven't removed magic number in tests

**What I'd like to improve:**
- How to improve use of context in tests
- Refactor tests to make them more readable

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -17,15 +17,16 @@ describe '#land(plane)' do
 
   end
 
-<<<<<<< HEAD
   context 'Airport can #take_off planes' do
 
     it 'Airport responds to #take_off' do
       expect(airport).to respond_to(:take_off)
     end
 
+    it 'Airport responds to #view_planes' do
+      expect(airport).to respond_to(:view_planes)
+    end
+
   end
 
-=======
->>>>>>> parent of 332159d... Airport responds to #take_off
 end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -40,4 +40,14 @@ describe '#land(plane)' do
 
   end
 
+  context 'Given a Capacity' do
+
+    it "Plane  won't land if above capacity" do
+      new_airport = Airport.new(72)
+      72.times { new_airport.land }
+      expect { new_airport.land }.to raise_error 'Airport Full'
+    end
+
+  end
+
 end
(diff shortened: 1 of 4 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'airport'
+
+describe 'Airport' do
```
PhilipVigus: Your general layout of this file is excellent. I like the use of contexts and there is really good consistency with the syntax you've used in the lets

spec/airport_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  subject(:airport) { Airport.new }
```
PhilipVigus: I'm not understanding why you've used a let and a subject for essentially the same thing. My understanding is that a let is generally preferable in this situation, and would suggest using it for both.

spec/airport_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+  let(:plane) { Plane.new }
+
+  context 'Airport can #land planes' do
+
+    it 'Airport responds to #land' do
+      expect(airport).to respond_to(:land)
+    end
+
+    it 'Plane lands in airport' do
+      allow(airport).to receive(:stormy?).and_return false
```
PhilipVigus: The repeated use of this line across multiple tests could be extracted out into a let. This would make your code DRY and much easier to read.

spec/airport_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    it "Plane  won't land if airport is full" do
+      allow(airport).to receive(:stormy?).and_return false
+      airport.land
+      expect { airport.land }.to raise_error 'Airport Full'
+    end
+
+  end
+
+  context 'Airport can #take_off planes' do
```
PhilipVigus: I'm curious about your use of # here as I haven't seen it before. I'd be grateful if you could explain what it means when you use it in this way?

spec/airport_spec.rb
PhilipVigus: As far as I'm aware, you could remove this by tweaking your describe on line 3. If you replace describe 'Airport' with describe Airport, then subject will automatically resolve to an instance of the Airport class.

PhilipVigus:
Generally really impressed with your code. Most of it is really laid out well, although there are some things you can do in the test files to improve readability a bit.

PR #1766, opened 2020-03-23
Submitting Airport Challenge

alexakearns:


README.md
```diff
@@ -88,3 +74,25 @@ Finally, don’t overcomplicate things. This task isn’t as hard as it may seem
...
+
+## Reflections
+---
+
+Whilst being unable to complete this task, my understanding of TDD has increased significantly. After user story 1, and some peaking at [Harry's](https://github.com/hturnbull93/airport_challenge) initial steps to see if I was on the right track, it became clear to me I needed to step back and take some time to read more about TDD and Rspec as my understanding of these was not enough to continue, I was going no where!
+
+My break down of the user story into the steps needed to ensure this was testing correctly and my code was building to meet its requirements, was not enough and the steps in my plan were too big (I was not thinking about it in terms of testing). Going forward my plan for each user story became far more substantial, and this make a big difference in building the tests and ultimately allowed me to complete 4 of the user stories. 
+
+Rspec syntax, which matcher to and how the tests allow us to build and test the code, whilst still a long was from having 'got it' I feel I have a much better understanding on from doing this task. 
+
```
hturnbull93: I love these reflections! 👍

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def collect_capacity
+    @capacity
+  end
+
+  def change_capacity(capacity)
+    @capacity = capacity
+  end
```
hturnbull93: Rather than using manual getters and setters for changing capacity, you could use an attribute accessor
```
attr_accessor :capacity
```

Alternatively, you could allow capacity to be set in the initialise method, as at the moment an airport can have its capacity changed after it is made.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def plane_in_hanger?(plane)
+    @hanger.include?(plane)
+  end
```
hturnbull93: Nice SRP abstraction 🔥

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def land(plane)
+    fail "Airport is full" unless @hanger.count < @capacity
+
+    store(plane)
+  end
+
+  def store(plane)
+    @hanger << plane 
+  end
```
hturnbull93: Also a nice SRP abstraction 💯

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def take_off(plane)
+    "Plane takes off!"
+  end
+
+  def left_airport(plane)
+    @hanger.delete(plane)
+  end
```
hturnbull93: Another great SRP abstraction 🥇

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  def change_capacity(capacity)
+    @capacity = capacity
+  end
+
+  def plane_in_hanger?(plane)
+    @hanger.include?(plane)
+  end
+
+  def land(plane)
+    fail "Airport is full" unless @hanger.count < @capacity
```
hturnbull93: Good guard clause!

To improve you could abstract out `@hanger.count < @capacity` to a hangar_full? method perhaps?

spec/airport_spec.rb
```diff
@@ -0,0 +1,65 @@
+require 'airport'
+
+describe Airport do
+  
+  subject(:city) { Airport.new }
+  let(:jumbo) { Plane.new } 
+ 
+  context '#landing' do
```
hturnbull93: I like the use of contexts to talk about particular methods 🏆

spec/airport_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      expect(subject).to respond_to(:take_off).with(1).argument
+    end
+
+    it "a plane takes off" do
+      expect(subject.take_off(jumbo)).to eq "Plane takes off!"
+    end
+
+    it "plane no longer at airport" do
+      expect([@hanger]).not_to include(jumbo)
+    end
```
hturnbull93: This passes, righ? But was jumbo in the airport to begin with? 

To me it seems like an initially empty hangar will contain no planes 🤔

spec/airport_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    it "stores plane in hanger on landing" do
+      @hanger = subject.store(jumbo)
+      expect(@hanger).to include(jumbo)
+    end
+
+    it "raises error if airport is full" do
+      subject.change_capacity(5)
+      5.times { city.land(jumbo) }
+      expect { city.land(jumbo) }.to raise_error "Airport is full"
+    end
```
hturnbull93: A thought, in this test the same plane is landed 5 times into the same airport. Should that be allowed to happen? Will this create duplicate references to jumbo in the hangar array? ✈️

spec/airport_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    it { is_expected.to respond_to :land }
+
+    it "expects one argument" do
+      expect(subject).to respond_to(:land).with(1).argument
+    end
+
+    it "lands the plane" do
+      @hanger = []
+      expect(subject.land(jumbo)).to be subject.store(jumbo)
+    end
```
hturnbull93: Is this checking if the plane is in the hangar? strictly speaking the store method adds the plane to the hangar array and returns it implicitly, as does the land method. 🤔

PR #1767, opened 2020-03-23
airport challenge 

Gina-Frankel:


lib/airport.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def landing(plane)
+
+    if @planes.include?(plane)
+      @planes
+      error "plane is aleady landed"
+    end
+
+    if stormy? == true || full?
```
Hundredfolds: I like these two checks being reduced in a single line! I'll be refactoring mine to do this haha

README.md
```diff
@@ -88,3 +88,52 @@ Finally, don’t overcomplicate things. This task isn’t as hard as it may seem
...
+|* 2.  prevents 'take_off' if Full?||√but need to change to raise error |
+|* 3. planes that are already flying cannot land|||
+|* 4. Planes that are landed cannot land again and must be in an airport||√|
+
+
+To change:
+
+* In relation to the user stories "As an air traffic controller, To ensure safety
+I want to prevent takeoff/landing when weather is stormy/ full" I have ensured that the @ plane array which is storing the planes does not change. However it would be better to raise an error instead (see the table at the bottom 1 + 2). (I had a lot of difficulty with raising errors - it seemed to cause the rspec file to faill all the tests). 
+
```
Hundredfolds: For this it sounds like an issue of wrapping what the expect is evaluating in RSPEC in curly braces to form a code block, that way your RSPEC should compare the error the code block throws to the one you're testing for - witgout the curly braces it'll just run the code in the brackets and throw and error - not sure if that was the error you were getting.

I don't know the error command in ruby, I used raise instead, I don't imagine they're all that different but that might be somewhere to look into.

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
+require 'airport'
+
+describe Airport do
+  subject(:airport) {described_class.new}
+  let(:plane) {Plane.new}
+  let(:qantas) {Plane.new}
+  let(:singapore) {Plane.new}
```
Hundredfolds: Instead of calling the Plane class here you could instantiate plane doubles using let(:plane) {double :plane } and that way these tests would not fail without access to the Plane class.

lib/airport.rb
Hundredfolds: Also I think you can get away without using the == true component here, as stormy? should be returning a boolean value

PR #1768, opened 2020-03-23
SK

Steven-Klavins:


grahamfalconer:
Spelling error: .gsub("capasity", "capacity")

grahamfalconer:
can_i_land method could potentially renamed to something more fitting in a user story. 
relies on booleans so also maybe useful to include "?" at the end of the method.

PR #1769, opened 2020-03-23
Angelica

Angelica137:


lib/plane.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'airport'
+require_relative 'weather'
+
+
+class Plane
+    
+    def initialize
+      conditions = Weather.new
```
jamesjoshuahill: I like that you created a separate Weather class 💯

PR #1770, opened 2020-04-18
Early pull request

GarethWoodman:


lib/weather.rb
```diff
@@ -0,0 +1,21 @@
+class Weather
+  attr_reader :storm_probability
+  DEFAULT_PROBABILITY = 70
+
+  def initialize(probablity = DEFAULT_PROBABILITY)
```
AndrewHulme, line 5: This method could be refactored (e.g. doesn't need the argument in the method as can just use the DEFAULT_PROBABILITY constant)

spec/airport_spec.rb
```diff
@@ -0,0 +1,24 @@
+require './lib/plane'
+require './lib/airport'
+require './lib/weather'
+
+describe Airport do
+  default_capacity = Airport::DEFAULT_CAPACITY
+  airport = Airport.new
+
+  def clear_weather_at(airport)
```
AndrewHulme, line 9: Good use of method to apply to all consequent terms in rspec

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  def not_docked?(plane)
+    @planes.include?(plane) ? false : plane_not_docked_message
+  end
+
+  private
+  def airport_full_message
+    fail "Airport is full"
+  end
+
+  def plane_not_docked_message
```
AndrewHulme, line 29: Very good use of SRP, all methods are concise and clear and have a specific goal

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def docked?(plane)
+    @planes.include?(plane) ? plane_docked_message : false
+  end
+
+  def not_docked?(plane)
+    @planes.include?(plane) ? false : plane_not_docked_message
+  end
+
+  private
```
AndrewHulme, line 24: Good use of private methods

PR #1771, opened 2020-04-18
Airport Challenge

dpwdec:


spec/weather_reporter_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'weather_reporter'
+
+describe WeatherReporter do
+
+  describe "#check_weather" do
+    it "returns a either 'Sunny' or 'Stormy'" do
+      expect(subject.check_weather).to eq("Sunny").or eq("Stormy")
```
tanilserbes, line 7: that's a good idea to check both return values in one expect ! :) Had no idea to use two eq matchers at same time.

spec/airport_spec.rb
```diff
@@ -0,0 +1,114 @@
+require 'airport'
+
```
collier-jo, line 2: All of your specs are GREAT! Very well organised. Using alot of mocks. The use of shared contexts allow them to be super readable and DRY. Additionally breaking up the RSPEC into #function allows for it to be super clear what you are testing and also compartmentalises so the reader isn't going through one huge Airport.

spec/weather_reporter_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'weather_reporter'
+
+describe WeatherReporter do
```
collier-jo, line 3: I'd be interested to know what coverage you got as while it may seem unnecessary  because obvs the rspec is able to run the method generally as it is returning Stormy for example, I cannot see any respond_to?

spec/airport_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      expect { subject.undock(plane) }.to raise_error("It's too stormy to take off.")
+    end
+  end
+
+  describe "#docked" do
+    it "returns false if plane is NOT docked" do
+      expect(subject.docked?(Plane.new)).to be false
+    end
+
+    context "it is sunny" do
```
collier-jo, line 79: While I like the methods it does seem that maybe your code is starting to be a little repetitive in that you are constantly having to set the context - i.e sunny. Maybe if it was set up like context at the beginning and everything that happens to airport this may remove this??

PR #1772, opened 2020-04-19
Ed Ancerys

EdAncerys:
Airport Challange


lib/airport.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'plane'
+
+class Airport
+  attr_accessor :capacity, :parked_planes
+
+  DEFAULT_CAPACITY = 2
```
JStant95, line 6: Avoided magic numbers well here by using constant (I forgot to do this!)

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'plane'
+
+class Airport
```
JStant95, line 3: Methods within this class all adhere to srp but maybe the class as a whole doesn't. Could maybe move to different class like the plane one and weather one

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+  before { allow(subject).to receive(:stormy?) { false } }
```
JStant95, line 6: Nice use of before

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
```
JStant95, line 4: Spec is very clear re what you are testing

PR #1774, opened 2020-04-19
Andrew Hulme Airport Challenge

AndrewHulme:
Finished version of the airport challenge

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,14 @@
+class Airport
+  attr_reader :name, :list_of_planes, :capacity
+  attr_writer :name, :list_of_planes, :capacity
+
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,14 @@
+class Airport
+  attr_reader :name, :list_of_planes, :capacity
+  attr_writer :name, :list_of_planes, :capacity
+
```
GarethWoodman, line 4: Hey dude, I would re-write the attributes with the attr_accessor keyword as it enables both read & write

Katerina-codes: @GarethWoodman why read and write?

GarethWoodman: @Katerina-codes can you be more specific?

PR #1775, opened 2020-04-19
completed

Patrick-Sawyer:
Hi, think I have pretty much done the requirements although it could do with refactoring and some syntax could be improved (both the main script and the rspec script). It passes all my rspec tests but fails a few of the rubocop ones still.

### DIFF
--- a/lib/airports.rb
+++ b/lib/airports.rb
@@ -1,0 +1,67 @@
+class Airport
+
+  attr_reader :capacity, :hangar, :weather
+  DEFAULT_CAPACITY = 20
+  
+  def initialize(capacity = DEFAULT_CAPACITY)
+   
+    @capacity = capacity
+    @hangar = []
+    @weather = false
+  end
+
+  def stormy?
+    weather
+  end
+
+  def weather_update(weather)
+    @weather = weather
+  end
+end
+
+class Plane
+
+  attr_reader :location
+
+  def initialize(airport = 'SKY')
+    @location = [airport]
+    if airport.is_a?(Airport)
+      if airport.hangar.length < 20
+        airport.hangar << self
+      else
+        raise "Airport full."
+        @location = 'SKY'
+      end
+    end
+  end
+
+  def land(airport)
+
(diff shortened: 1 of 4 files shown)


lib/airports.rb
```diff
@@ -0,0 +1,67 @@
+class Airport
+
+  attr_reader :capacity, :hangar, :weather
+  DEFAULT_CAPACITY = 20
```
will-head, line 4: Excellent work avoiding Magic Numbers. I did not do this :(

lib/airports.rb
```diff
@@ -0,0 +1,67 @@
...
+  def stormy?
+    weather
+  end
+
+  def weather_update(weather)
+    @weather = weather
+  end
+end
+
+class Plane
```
will-head, line 22: Maybe split the Planes calls into its own file, so each file only contains a single class?

Patrick-Sawyer: I shall do that next time.

lib/airports.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  attr_reader :location
+
+  def initialize(airport = 'SKY')
+    @location = [airport]
+    if airport.is_a?(Airport)
+      if airport.hangar.length < 20
```
Patrick-Sawyer, line 29: I need to change '20' to airport.capacity

spec/airports_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+    it "responds true after taking off" do
+      heathrow = double(:hangar => [], :capacity => 20, :stormy? => false, :is_a? => true)
+      plane = Plane.new(heathrow)
+      plane.takeoff
+      expect(plane.takenoff?).to eq true
+    end
+    it "responds false if sat in airport" do
+      heathrow = double(:hangar => [], :capacity => 20, :stormy? => false, :is_a? => true)
+      plane = Plane.new(heathrow)
+      expect(plane.takenoff?).to eq true
```
Patrick-Sawyer, line 133: This bit is wrong too

PR #1779, opened 2020-04-19
My airport challenge

cmb84scd:


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def initialize(capacity = DEFAULT_CAP)
+    @capacity = capacity
+    @planes = []
+    @weather = Weather.new
+  end
+
+  def land(plane)
+    fail "Airport is full" if full?
+
```
LTurns, line 18: Code is really clear and easy to read! :)

spec/airport_spec.rb
```diff
@@ -1,61 +1,52 @@
 require 'airport'
 
 describe Airport do
-  it 'lands a plane' do
-    expect(subject).to respond_to(:land)
+  before(:each) do
+    @airport = Airport.new
+    @plane = Plane.new
```
tatsiana-makers, line 6: Do these variables need the `@` prefix?

spec/airport_spec.rb
```diff
@@ -1,61 +1,52 @@
...
+    expect(@airport.capacity).to eq Airport::DEFAULT_CAP
   end
 
-  describe '#take_off' do
-    it 'takes off a plane' do
-      allow(subject.weather).to receive(:stormy?) { false }
-      plane = Plane.new
-      subject.land(plane)
-      expect(subject.take_off).to eq plane
+  describe '#land' do
```
tatsiana-makers, line 13: Nice use of describe blocks to group your tests!

spec/airport_spec.rb
tatsiana-makers, line 6: Can we use mocking to isolate the `Airport` unit tests from using the actual implementation of the `Plane` class?

tatsiana-makers: Ah, I was thinking of `let` blocks - looks like you need the `@` here. https://dzone.com/articles/rspec-let-vs-before might be an interesting read

PR #1781, opened 2020-04-19
Jed Stanton - Weekend Challenge Submission

JStant95:


spec/weather_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'weather'
+
+describe Weather do
+  subject { Weather.new }
+  it 'generates random weather conditions' do
+    expect(subject.generate).to be_a(String)
```
tanilserbes, line 6: Hey Jed ! it's nice to control the weather generator if it's working or not. Here is another way using stub which enables you to check the specific weather condition. 

```
allow(subject).to receive(:generate).and_return 1"
expect(subject.generate).to eq 'sunny'
```

PR #1784, opened 2020-04-20
Submit airport_challenge for review

will-head:


irb/atc.irb
```diff
@@ -0,0 +1,207 @@
...
+world.add(planes[11])
+world.add(planes[12])
+world.add(planes[13])
+world.add(planes[14])
+world.add(planes[15])
+world.add(planes[16])
+world.add(planes[17])
+world.add(planes[18])
+world.add(planes[19])
+
```
Patrick-Sawyer, line 44: consider using a loop for the above, and all similar instances . e.g 20.times...

lib/air.rb
```diff
@@ -0,0 +1,19 @@
...
+  def initialize
+    @planes = Array.new
+  end
+
+  def add(plane)
+    @planes.push(plane)
+    @planes.last == plane
+  end
+  
+  def del(plane)
```
Patrick-Sawyer, line 13: whats del?

lib/sd.rb
```diff
@@ -0,0 +1,7 @@
+class Sd
```
Patrick-Sawyer, line 1: Your names are very concise and abbreviated making it a bit unclear

lib/world.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def takeoff(plane, airport)
+    return false unless plane.is_a?(Plane) && airport.is_a?(Airport)
+#    return false if where_is(plane) == @air
+    return :stormy if airport.weather == :stormy
+    
+    result = airport.del(plane) if air.add(plane)
+    result == plane
+  end  
+end
```
Patrick-Sawyer, line 66: I am unsure what the world class is for

irb/atc.irb
will-head: Ooo - good point! That would have been way more efficient.

lib/air.rb
will-head: del(ete) but you're right, I should avoid shortening things (even if it means add and del are the same length...)

lib/sd.rb
will-head: Yep. Will defo avoid in future.

lib/world.rb
will-head: World has the Air in it, and you can then put Planes and Airports into it (or take them out). It basically keeps track of all the objects you've created.

PR #1786, opened 2020-04-20
Weekend 1 - Airport Challenge

HResalaty:
Pull request for partial solution.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,5 @@
+require_relative 'plane'
+
+class Airport
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,13 +2,26 @@ require_relative 'plane'
 
 class Airport
 
+  attr_reader :airport_capacity, :current_plane_count
+
+  DEFAULT_CAPACITY = 100
+  CURRENT_PLANE_COUNT = 0
+
+  def initialize (airport_capacity = DEFAULT_CAPACITY)
+    @airport_capacity = airport_capacity
+    @current_plane_count = CURRENT_PLANE_COUNT
+  end
+
   def land
+    fail 'WARNING! AIRPORT FULL!' if current_plane_count >= DEFAULT_CAPACITY
+    @current_plane_count += 1
   end
 
   def take_off
   end
 
   def confirm_location
+    puts "Plane can confirm that no longer at airport"
   end
 
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -18,13 +18,28 @@ class Airport
   end
 
   def take_off
-    fail 'WARNING! AIRPORT EMPTY!' if current_plane_count == 0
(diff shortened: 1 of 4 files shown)


lib/plane.rb
```diff
@@ -0,0 +1,4 @@
+class Plane
+
+
+end
```
Patrick-Sawyer, line 4: This seems unnecessary.

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+class Airport
+
+  attr_reader :airport_capacity, :current_plane_count
+
+  DEFAULT_CAPACITY = 100
+  CURRENT_PLANE_COUNT = 0
+
+  def initialize (airport_capacity = DEFAULT_CAPACITY)
+    @airport_capacity = airport_capacity
+    @current_plane_count = CURRENT_PLANE_COUNT
```
Patrick-Sawyer, line 12: I think you can just declare @current_plane_count as 0, I wouln't have thought there is a need for the constant CURRENT_PLANE_COUNT.

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def land(plane)
+    fail 'WARNING! AIRPORT FULL!' if current_plane_count >= DEFAULT_CAPACITY
```
Patrick-Sawyer, line 16: current plane_count should have an @

spec/airport_spec.rb
```diff
@@ -0,0 +1,45 @@
+require './lib/airport.rb'
+
+describe Airport do
+
+
+  describe "#land arguments" do
+    it "plane should be able to #land" do
+      expect(subject).to respond_to(:land).with(1).argument
```
Patrick-Sawyer, line 8: Doesn't test functionality of 'land', e.g. it should test that the @current_plane_count increases by one each time a plane successfully lands.

spec/airport_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  describe "#takeoff arguments" do
+    it "plane should be able to #land" do
+      expect(subject).to respond_to(:take_off).with(1).argument
+    end
+  end
+  # it { is_expected.to respond_to :take_off }
+
```
Patrick-Sawyer, line 18: Same as land, doesn't test functionality. Only testing the method exists. Also string on line 13 says land, should say takeoff. Will make it difficult to read errors when rspec is run

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    "Plane confirms that no longer at airport"
+  end
+
+  def stormy?
+    weather_condition = rand(0..1)
+    "stormy" if weather_condition == 0
+  end
+  #---------#
+
+end
```
Patrick-Sawyer, line 40: Code and methods are clear, and it is easy to read.

PR #1790, opened 2020-04-20
Airport_challenge

ellistrickett:


lib/airport.rb
```diff
@@ -0,0 +1,58 @@
+require_relative 'plane'
```
HResalaty, line 1: Looks like a comprehensive, well-structured code to me :)

lib/plane.rb
```diff
@@ -0,0 +1,32 @@
+class Plane
```
HResalaty, line 1: Looks like the edge cases have also be covered for in the code, which is good.

my_work/README.md
```diff
@@ -0,0 +1,9 @@
+#  Airport Challemne
```
HResalaty, line 1: Really like the fact that you incorporated a readme file, well-done!

spec/airport_spec.rb
```diff
@@ -0,0 +1,110 @@
+require 'airport'
```
HResalaty, line 1: comprehensive spec files been developed for the plane and airport as well.

PR #1794, opened 2020-05-01
Kate pull request - friday morning

kate-49:


kate-49:
- I did not finish the Boris Bikes challenge, I was looking at Boris bikes this weekend and I would've written this differently if i'd had a chance to finish it. 
- I am not confident with testing the weather class, I did this using stubs and Eddie has told me that because of how i've written it the tests will always pass. If you could advise a different way to write this using doubles/ mocks that would be super helpful. 
- I'm pretty sure my methods are too complicated and I should've simplified them. 
- I also wanted to work on raising errors but didn't manage to do this successfully. 
- I *think* it passes all User stories but I think you have to invent things in a weird way to do that.

Chanpreet24:
- I like the use of the plane id number as a way of scaling the number of planes.
- the number of written tests made it quite difficult to follow. There was quite a high volume which meant it was difficult to review/follow. 
- There was syntax in test file that I had not seen before (kudos but possibly confusing).

PR #1795, opened 2020-05-01
Commit for Pull Request Airport Challenge

JonF1:


JonF1:
I didn't know how to implement the weather part although that was right near the end. Did I have to write an additional RSpec test or method for the override default capacity story as well? Any problems with syntax?

PR #1796, opened 2020-05-01
Introduced an exception for not allowing more planes than capacity bu…

annamvoitika:
…t need to troubleshoot production code

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,26 @@
+class Airport
+  DEFAULT_CAPACITY = 20
+  attr_accessor :capacity
+
+    def initialize(capacity=DEFAULT_CAPACITY)
+      @capacity = capacity
+      @planes = []
+    end
+
+    def land(plane)
+      return "plane landed"
+      fail 'Airport is full' if full?
+      @planes << plane
+    end
+
+    def take_off
+      return "plane no longer in the airport"
+    end
+
+private
+
+def full?
+  @planes.count >= capacity
+end
+
+end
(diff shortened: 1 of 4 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
+# For Eddie: to resolve the last two stormy user stories, I learned from: https://www.youtube.com/watch?v=Vg0cFVLH_EM
```
EdwardAndress, line 1: 👍

annamvoitika:
Hi code reviewer!

I have not removed commented out user stories from my airport_spec, because I didn't realise we would do a review. Otherwise, as a next step to refactoring, I would have removed that.

I have covered the first 6 user stories for airport challenge

spec/airport_spec.rb
```diff
@@ -6,9 +6,10 @@
 # I want to instruct a plane to land at an airport
 
 describe Airport do
+  let(:plane) { double :plane }
```
JPprojects, line 9: Good start to using doubles in your testing

lib/airport.rb
```diff
@@ -9,19 +9,22 @@ def initialize(capacity=DEFAULT_CAPACITY)
...
-  def weather_safe
-    raise 'stormy'
-  end
-
   def full?
   @planes.length >= @capacity
 end
-end 
+
+  def stormy?
```
JPprojects, line 26: Could this be done in a cleaner way so it just return true or false?

PR #1797, opened 2020-05-01
created the class Plane and Airport in seperate files, created a test…

avinhurry:
… for the first user story which currently fails

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,2 @@
+class Airport
+end 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -13,25 +13,3 @@ class Airport
     true
   end
 end
-
-  # class DockingStation
-  #   attr_reader :inventory
-  #
-  #   def initialize
-  #     @inventory = []
-  #   end
-  #
-  #   def release_bike
-  #     Bike.new
-  #   end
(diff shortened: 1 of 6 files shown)


avinhurry:
I'm not sure if I have written the tests for landing when the airport is full in the best way. 

Also not sure if the other two tests could be improved.

spec/plane_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'airport'
+describe Airport do
+
+# As an air traffic controller
+# So I can get passengers to a destination
+# I want to instruct a plane to land at an airport
+
+  it "instructs a plane to land at an airport" do
+
```
JasmineRachel: I like that you have split up the different components in your test, making it easier to understand what is going on and what you're actually testing for

lib/airport.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class Airport
+  attr_accessor :inventory
+
+  def initialize
+    @inventory = []
+  end
+
+  def planes_available?
+    return raise "There are no planes available" if self.inventory.size < 0
```
JasmineRachel, line 11: Doesn't the "raise" command do the same thing(ish) at "return"?

PR #1798, opened 2020-05-01
Airport_challenge, April2020_Chan

Chanpreet24:


Chanpreet24:
I was not sure how to confirm that the plane had taken off and that it was now airborne.

Chanpreet24:
- I struggled with starting (writing the first test).
- I wasn't sure as to the writing of the classes - still need to work on this, hence the reason it is quite short.

george-bit:
Really good clear and simple code addressing the first two user stories. I think it's really clever how you have divided the take off and land methods across the two classes. The only word of caution would be how this would work for following user stories, as the complexity increases it could be an issue dividing the methods across classes. You might be want to look at simplifying a bit further by keeping methods within one class to prevent potential confusion in future. Otherwise, really good stuff Chan!

PR #1799, opened 2020-05-01
first commit

JasmineRachel:


JasmineRachel:
Wasn't sure if I was on the right track

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+  # end
+
+  def land(_plane)
+    fail 'Airport full!' if full?
+  end
+
+  def take_off(_plane)
+    'Plane on route!'
+  end
+
```
avinhurry, line 21: The succinct methods are great!

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+  private
+
+  def full?
+    @planes.count >=10
+  end
+
```
avinhurry, line 27: I'm curious about this private method, what is the reason it's private?

spec/airport_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'airport.rb'
+require 'plane.rb'
+
```
avinhurry, line 3: I used relative require, require didn't work for me. I'm not entirely sure what the difference is. Do you know?

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
+require 'plane.rb'
+
+class Airport
+
+  def initialize
+    @plane = []
+
```
avinhurry, line 7: should there be a space here?

PR #1801, opened 2020-05-01
changes to file/first commit

Adewunmi12:


Adewunmi12:
I completed the first user story and started the second. I had a bit of a block when starting the challenge but once I gained an understanding it went quite well. I would like to know whether my code is ok and if there are any changes I should make on the little code I have.

In terms of help, is finding a better way of knowing what methods/objects to use in different scenarios.

cass-93:
This was a great start, I can see you knew the direction of where to go for the second user story.Glad to see that we had a similar use of the classes. I also like the include syntax which was something I haven't used yet.

I tend to find it easier to outline the methods with comments for each of the stories first before actually getting into the coding. I find that helps me to break the stories into more manageable chunks and figure out what I need the methods todo at each stage. 

One thing which I think you could remove is the 'require 'plane' in your spec file. As you required this already in your airport.rb file it should be able to read this. 

With the knowing the methods, as you can define them you can call them whatever you want. I tend to use ones which explain what I want them to do. Some of the syntaxes are a bit more difficult to remember but I tend to find Googling helps in those situations or reach out to the cohort. I'm always happy to help. 

Overall though great start, you've definitely grasped the concept of TDD, you should be super happy with your progress Ade!

PR #1802, opened 2020-05-01
Joe's Airport

josephhull676:


josephhull676:
I initially struggled to make the classes communicate with each other but solved the problem after some thought. Other than that I can't think of anything in particular that I struggled with.

PR #1803, opened 2020-05-01
Airplane

GlenHani:
1st Pull Request test passed

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,34 @@
+class Airport
+
+  attr_accessor :capacity, :stormy
+
+  def initialize(capacity = 20)
+    @capacity = capacity
+    @planes = []
+    @stormy = false 
+  end
+
+  def land_plane(plane)
+    raise 'Landing: no airport space' if full?
+    raise 'Landing: to stormy to land' if stormy?
+    @planes.push(plane)
+  end
+
+  def take_off
+    raise 'Taking: to stormy to take off' if stormy?
+  end
+
+  def report_storm
+    @stormy = true
+  end 
+
+  private
+
+  def full?
+    @planes.length >= @capacity
+  end
+
+  def stormy?
+    @stomry 
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,53 @@
+require './lib/airport'
+
+RSpec.describe Airport do
+  
+  let(:plane) { double :plane }
+  
+  describe '#land' do
+    it 'Allow plan to land' do
+      expect(subject).to respond_to(:land_plane)
+    end
+
+    it 'Raise error if airport is at fully capacity' do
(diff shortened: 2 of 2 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
+require './lib/airport'
+
+RSpec.describe Airport do
+  
+  let(:plane) { double :plane }
```
EdwardAndress, line 5: Nice! It's definitely the correct decision to use a double here.

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    it 'Raise error if airport is at fully capacity' do
+      ammount = subject.capacity
+      ammount.times do
+        subject.land_plane(plane)
+      end
+      expect { subject.land_plane(plane) }.to raise_error 'Landing: no airport space'
+    end
+    
+    it 'Raise error if airport is at fully capacity' do
+      allow(subject).to receive(:stormy?).and_return true
```
EdwardAndress, line 21: Good use of stubbing.

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @stormy = false 
+  end
+
+  def land_plane(plane)
+    raise 'Landing: no airport space' if full?
+    raise 'Landing: to stormy to land' if stormy?
+    @planes.push(plane)
+  end
+
+  def take_off
```
EdwardAndress, line 17: When the plane lands it's added to the @planes array but when takeoff is called, the @planes array doesn't change.  What could you change here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    it 'raises an error if asked to take off a plane' do
+      allow(subject).to receive(:stormy?).and_return true
+      expect { subject.take_off }.to raise_error 'Taking: to stormy to take off'
+    end
+
+  end
+
+  describe '#weather' do
```
EdwardAndress, line 41: I think I would probably be aiming to test the `report_storm` method by checking that it changes `@stormy` from false to true.  At the moment you're testing the return value of `report_storm`.

EdwardAndress: If you search for "rspec to change by matcher" you'll find something useful!

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
+class Airport
+
+  attr_accessor :capacity, :stormy
+
+  def initialize(capacity = 20)
+    @capacity = capacity
+    @planes = []
+    @stormy = false
```
EdwardAndress, line 8: If you want to take this one step further you could create a weather class. Then you'll need to figure out a way of using a double for the Weather! :)

EdwardAndress:
Hi Glen!  This is a nice elegant solution.  I've left some comments for you.  The most important ones are about the `take_off` method and how you're testing `report_storm`.

PR #1804, opened 2020-05-01
work on airport challenge

george-bit:


george-bit:
Completed passing tests for user stories up to the final to user stories regarding weather in the README. I have begun work on the tests and coding regarding the Weather. I have created files in anticipation. I know that implementation of the Weather class is subject to testing via mocking. Effectively, I know that mocking is required for the input to test the code for the weather class. The weather class would be a random number generator housed in a separate file and equating to true or false with the method of stormy?. This output would be conveyed to the land and takeoff methods respectively.

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+  attr_reader :capacity
+
+  def initialize(capacity=DEFAULT_AIRPORT_CAPACITY)
+    @capacity = capacity
+    @hangar = []
+  end
+
+  def land(plane, weather)
+    fail 'Airport full' if full?
+    fail 'Cannot land: Stormzy' if weather.stormy? == true
```
kate-49, line 14: is Stormzy a spelling error?

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
+class Airport
+  DEFAULT_AIRPORT_CAPACITY = 92
+
+  attr_reader :hangar
+  attr_reader :capacity
```
kate-49, line 5: I think you can assign 2 elements to 1 attr_reader

spec/airport_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it 'can land a plane and add it to the hangar' do
+      plane = 'spitfire'
+      subject.land(plane)
+      expect(subject.hangar).to eq ([plane])
+    end
+
+    it 'prevents landing when weather is stormy' do
+      weather_double = double :true
+      expect{ subject.land 'spitfire' weather_double } .to raise_error 'Cannot land: Stormzy'
+    end
```
kate-49, line 18: I'm so impressed at you using doubles to manage to test this.

spec/airport_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    subject.land('spitfire')
+    subject.takeoff
+    expect(subject.hangar).to eq([])
+  end
+
+    it 'prevents take off when weather is stormy' do
+    end
+  end
+
+  describe 'hangar' do
```
kate-49, line 34: I think it's '#hangar' ?

spec/airport_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it { is_expected.to respond_to(:hangar) }
+
+    it 'confirms that hangar is created as empty array' do
```
kate-49, line 37: This is a really good idea for a test.

spec/airport_spec.rb
george-bit: Unfortunately even though it looks impressive it did not have the intended result :(
I'm gonna come back to this once i understand double syntax better!

spec/airport_spec.rb
george-bit: Yeah that would make sense given the nomenclature

spec/airport_spec.rb
george-bit: Yeah I'm pretty pleased with it in the end. Kind had to speak to Eddie about it and he nudged me in the right direction :)

PR #1806, opened 2020-05-17
Rich's Airport

RichEwin:
It took me a while to get going on this exercise. I struggled to connect what class would take what method, for example does the land method sit in the Airport class or the Plane class? 

What user stories are included:

1, 2, 3, 5, 6 complete.

I had trouble on User Story 4 allowing the capacity to be over-written. 

What is Incomplete: 

User Story 4 allowing the capacity to be over-written. 

What can I improve:

Creating a Weather class. I created a Weather class initially then transferred my code to my Airport Class and included my methods within. 

What can you do:

Create Plane (or 2)
Create Airport (or 2)
Capacity is set to 15
Notify Airport Full 
Alert to stormy weather 

airport.landing(plane)
airport.depart(plane)
airport.weather

plane.land(airport)
plane.take_off(airport)


lib/airport.rb
```diff
@@ -1,4 +1,14 @@
 require_relative 'plane'
 
 class Airport
-end 
+
+  attr_reader :hanger
```
NikitaDouglas, line 5: I really like the name for this instance variable. It really brings the code to life, if that makes sense?

lib/airport.rb
```diff
@@ -11,28 +11,42 @@ def initialize(capacity = DEFAULT_CAPACITY)
...
     fail "Airport Full!" if full?
 
     @hanger << plane
   end
 
-  def take_off(plane)
-    @hanger = nil
-  end
-
   def depart(plane)
```
NikitaDouglas, line 21: Depart definitely makes more sense as a method name here!

lib/airport.rb
```diff
@@ -11,28 +11,42 @@ def initialize(capacity = DEFAULT_CAPACITY)
...
+    check_weather
```
NikitaDouglas, line 22: I like that your depart method does these pre-flight checks.

lib/weather.rb
```diff
@@ -1,16 +1,16 @@
-class Weather
```
NikitaDouglas: I realise you deleted the weather class and incorporated this feature as a method in Airport. I think having Weather as a separate class is a better decision as it increases the cohesion of the Airport class - all the methods in it are working towards the same goal.

spec/airport_spec.rb
```diff
@@ -4,37 +4,52 @@
   subject { Airport.new }
 
   let(:plane) { Plane.new }
-  DEFAULT_CAPACITY = 10
+  DEFAULT_CAPACITY = 15
 
   before do
+    allow(subject).to receive(:rand) { 5 }
```
NikitaDouglas, line 10: You could consider refactoring your tests to reduce the repetition of this line. You can use a `let` clause which looks something like this:

`let(:expect_sunny_weather) { allow(subject).to receive(:rand) { 5 } }`

You can then replace your method stubs that expect sunny weather with `expect_sunny_weather`. I imagine one can even refactor further by using a `before` block, but I'm not sure how to do that yet!

spec/airport_spec.rb
```diff
@@ -4,37 +4,52 @@
...
   it "adds a plane to the airport" do
+    allow(subject).to receive(:rand) { 5 }
     expect(subject.instance_variable_get(:@hanger)).to include(plane)
   end
 
   it "alerts airport is full" do
-    (DEFAULT_CAPACITY - 1).times { subject.land(Plane.new) }
-    expect { subject.land(Plane.new) }.to raise_error("Airport Full!")
+    allow(subject).to receive(:rand) { 5 }
+    (DEFAULT_CAPACITY - 1).times { subject.landing(Plane.new) }
```
NikitaDouglas, line 26: I think I may know what's going on with your default capacity here. I don't think you need to put the default in line 7; you should be able to do as follows in this test:

`Airport::(DEFAULT_CAPACITY - 1).times { subject.land(Plane.new) }` 

I'm not sure this will work, but I know you can access constants in your tests by using that `::` operators I used. It may be worth experimenting with it to see if you can get it to work.

PR #1807, opened 2020-05-17
Rae's Airport Project

RaeRachael:


spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+describe Airport do
+  let(:airport) { Airport.new }
+  let(:plane_a) { double(:plane_a) }
+  let(:plane_b) { double(:plane_b) }
+  let(:weather) { double(:weather) }
+
+  describe '#capacity' do
+    it "should have default capacity" do
+      expect(airport.capacity).to eq(Airport::DEFAULT_CAPACITY)
```
sofyloafy, line 11: Is this argument necessary after you are actively testing capacity? Just a thought

lib/airport.rb
```diff
@@ -0,0 +1,61 @@
...
+    @capacity = cap
+    @planes = []
+    @weather = Weather.new
+  end
+
+  def landing_ok?(plane)
+    unless weather_stormy?
+      if capacity_ok?
+        landing(plane)
+        return true
```
sofyloafy, line 20: I think this boolean is an implicit return of true. Lovely guard clause tho.

lib/airport.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+    false
+  end
+
+  def takeoff_ok?(plane)
+    unless weather_stormy?
+      departing(plane)
+      return true
```
sofyloafy, line 29: I believe this is the same as above - implied return of true.

lib/airport.rb
```diff
@@ -0,0 +1,61 @@
...
+
+class Airport
+
+  DEFAULT_CAPACITY = 100
+
+  attr_accessor :capacity
+
+  def initialize(cap = DEFAULT_CAPACITY)
+    @capacity = cap
+    @planes = []
```
sofyloafy, line 12: Planes might be safer in a private method here.

PR #1810, opened 2020-05-17
Nick Butlers attempt at the Airport Challenge

nwmbutler:


lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+class Airport
+
+  DEFAULT_CAPACITY = 5
+
+  def initialize(capacity = DEFAULT_CAPACITY, weather = rand(5))
```
TimCPB, line 5: Clever idea, making the weather a randomly generated, default attribute of every instance of Airport

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    @planes << plane
+  end
+
+  def take_off(plane)
+    fail "Airport empty" if empty?
+
+    @planes.delete(plane)
+    puts "#{plane} has taken off"
+    self
```
TimCPB, line 23: What does 'self' do here? Asking because I don't know, not because I think it's wrong!

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  def full
+    @planes.length >= DEFAULT_CAPACITY
+  end
+
+  def empty?
+    @planes.length <= 0
+  end
+
+  def rndstorm
+    @weather.odd?
```
TimCPB, line 40: Could storms be less frequent than every odd number?

PR #1811, opened 2020-05-17
As far as I got on the airport challenge this weekend

Sumner1185:


spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
+require_relative '../lib/airport'
+
+describe Airport do
+  let(:subject) { Airport.new }
+  let(:plane) { double :plane }
+
+  it { is_expected.to respond_to :land }
```
RaeRachael, line 7: test covered by later tests?, probably safe to remove?

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    airport = Airport.new(capacity)
+    capacity.times { airport.land(Plane.new) }
+    expect { airport.land(plane) }.to raise_error('Warning: Hangar full')
+  end
+
+  it 'should not allow a plane to land that has already landed' do
+    airport = Airport.new
+    airport.land(plane)
+    expect { airport.land(plane) }.to raise_error('Plane already in hangar')
+  end
```
RaeRachael, line 20: does this test cover situations with multiple airports?

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  describe '#land' do
+    it 'should raise error if @hangar is full' do
+      Airport::DEFAULT_CAPACITY.times { subject.land Plane.new }
+      expect { subject.land Plane.new }.to raise_error('Warning: Hangar full')
+    end
+    it 'should land a plane' do
+      expect(subject.land(plane)).to eq [plane]
+    end
```
RaeRachael, line 29: does the outcome of this method need to be plane to pass the spec?

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  it 'should let a plane take-off and confirm plane no longer in hangar' do
+    allow(plane).to receive(:take_off)
+    allow(subject).to receive(:weather_check) { 'sunny' }
+    subject.land(plane)
+    subject.instruct_take_off(plane)
+    expect(subject.confirm_plane_left(plane)).to eq(false)
+  end
```
RaeRachael, line 46: does this test make sure that the operator gets a confirmation?

spec/plane_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+describe Plane do
+  let(:plane) { Plane.new }
+
+  it { is_expected.to respond_to :take_off }
+
+  it 'should check if it is in flight before taking off' do
+    plane.land_plane
+    expect(plane.flying?).to eq false
+  end
```
RaeRachael, line 11: yes, but should the operator be able to know if the plane is flying or not? 
the test feels like its testing the state of the code and not the outcomes laid out by the user stories

RaeRachael, line 11: should be in the describe #take_off block? or removed? 
these tests feel like they have been commented out in your head?

spec/weather_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+describe Weather do
+  let(:subject) { Weather.new }
+
+  it { is_expected.to respond_to :check_conditions }
+
+  it 'should return sunny more often than stormy' do
+    arr = []
+    50.times { arr.push(subject.check_conditions) }
+    expect(arr.count("sunny") > arr.count("stormy")).to eq true
+  end
```
RaeRachael, line 12: does this test check that "stormy" is a possible outcome?

spec/airport_spec.rb
Sumner1185: Very true - makes sense

spec/airport_spec.rb
Sumner1185: I guess this would only work at this specific airport as the plane will be logged at that hangar so couldn't then land again at that airport

spec/airport_spec.rb
Sumner1185: Not sure - I'll test

spec/airport_spec.rb
Sumner1185: No - could add a string for this

spec/plane_spec.rb
Sumner1185: I guess the operator would know as the plane cannot land if already at the airport. But I understand your comment and will have a think on this

Sumner1185: Fair

spec/weather_spec.rb
Sumner1185: Probably should have test for those outcomes first - thanks

PR #1815, opened 2020-05-17
submitting weekend airport challenge

Katie-McDonagh:


Katie-McDonagh:
For this weekend challenge,

I completed 1,2,3,4 and 6 user stories, 5 is the code #commented out. I had trouble trying to use my weather class in my airport class.

PLAN.md
```diff
@@ -256,4 +256,16 @@ $ irb
 2.6.5 :004 > airport.takeoff(plane)
  => nil
  ```
+ ## Feature tests - .takeoff now shows which plane is to be allowed to take off
```
BeneArinci, line 264: I am leaving this comment here but it takes in consideration all the work you've done in describing your work so far. 
Files PLAN.md and README.md are great. You show everything and it's really easy to follow your reasoning

spec/airport_spec.rb
```diff
@@ -2,8 +2,14 @@
...
   it 'lands the plane' do
     plane = @plane
     expect(subject.land(plane)).to eq plane
   end
-end
+
+  it 'allows planes to take off' do
+    expect(subject).to respond_to :takeoff
+  end
+  end
```
BeneArinci: I recently read that these kind of tests (to respond to) are not meaningful and you can skip them, testing something else regarding the same method

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
+
+  def forecast
```
BeneArinci, line 3: I suggest to use 1 method only and do something like:
def stormy?
  rand(1..10) > 8
end
In this case, the method stormy? will automatically return true or false based in the random result.

spec/airport_spec.rb
```diff
@@ -16,26 +17,32 @@
...
+  it 'allows the SD to set capacity of airport' do
     expect(Airport.new(10).instance_variable_get :@capacity).to eq(10)
-   end
+  end
 
-   it 'does not allow a plane to land if the weather is stormy' do
-     allow(subject).to receive(:forecast) {5}
-     expect { subject.land(@plane) }.to raise_error('too stormy to safely land')
-   end
+  it 'does not allow a plane to land if the weather is stormy' do
```
BeneArinci, line 39: I would suggest to create specific contexts in order to organise the tests in the spec file

PR #1818, opened 2020-05-17
Submission to Airport Challenge

ad13380:


ad13380:
- All User stories implemented
- Not sure if the rspec tests are laid out properly, or if they should have been broken down into more specific tests

PR #1819, opened 2020-05-18
Tim's attempt at the airport_challenge

TimCPB:


spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
+require './lib/airport'
+
+describe Airport do
+
+  subject(:airport) { Airport.new }
+
```
nwmbutler, line 6: Nice clean way to initialize an instance, to be used.

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+    @planes.delete(plane)
+    "#{plane} has departed the airport"
+  end
+
+  def space?
+    @planes.length < @capacity
+  end
+
+  attr_reader :capacity
+
```
nwmbutler, line 29: Good short methods with clear responsibilities.

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    raise "Cannot land, airport is full" unless space?
+
+    @planes.push(plane)
+  end
+
```
nwmbutler, line 18: Good use of edge case.

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+  def take_off(plane)
+    @planes.delete(plane)
+    "#{plane} has departed the airport"
+  end
+
```
nwmbutler, line 23: Suggestion, what if no such plane

PR #1828, opened 2020-06-14
Airport Challenge (user stories 1-4) - Dhara Patel

Dhara-95:


Dhara-95:
User stories 1-4 submitted. 

For user story 3/4 neew to resolve 'NoMethodError: undefined method 'count' for nil::NilClass'
on line:25 of 'airport.rb' and line:16 of "airport_spec.rb"

Would like to finish other user stories to prevent take off & landing in stormy weather and then refactor code to ensure code isn't DRY.

README_NEW.md
```diff
@@ -0,0 +1,19 @@
+Airport Challenge
+=================
+
+User Stories
+---------
+
+* User Story 1: Land method added - Complete
+* User Story 2: Take off method added - Complete
+* User Story 3: Exception raised for full airport - 1 failing test, incomplete
```
JoshuaNg2332, line 9: Awesome! Super clear on where you got up to and clear on what test failed!

lib/plane.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
```
JoshuaNg2332, line 1: Can you clarify why you would need the plane file for?

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+  DEFAULT_CAPACITY = 10
+
+  def initialze
+    @planes = []
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def land(plane)
+    fail 'Airport is at full capacity' if full_capacity?
+
```
JoshuaNg2332, line 14: Why is there a blank line here?
Great use of spacing elsewhere! Makes the code super easy to read!

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  it 'creates an instance of the Airport class' do
+    expect(airport).to be_an_instance_of(Airport)
+  end
+
+  describe '#land' do
+    it "lands a plane" do
+      expect(airport).to respond_to(:land).with(1).argument
+    end
+    it "raises an error when the airport is full" do
+      DEFAULT_CAPACITY.times { airport.land Airport.new }
```
JoshuaNg2332, line 17: Do you feel like you need `Airport.new` in this line and the line below when you've already used `let` at the top to allow the use of just `airport`? These could be changed to use the new naming convention.

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+describe 'Airport' do
+  let(:airport) { Airport.new }
+
+  DEFAULT_CAPACITY = 10
+
+  it 'creates an instance of the Airport class' do
+    expect(airport).to be_an_instance_of(Airport)
+  end
+
+  describe '#land' do
```
JoshuaNg2332, line 12: You don't always need this describe for your main methods from your normal `airport.rb` as you are already referencing the method in the `it` statement with `:land`. This can me a naming convention thing but saves 2 lines per method. This is a double edged sword as sometimes this means your code can be more difficult to read.

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'airport.rb'
+
+describe 'Airport' do
+  let(:airport) { Airport.new }
+
+  DEFAULT_CAPACITY = 10
+
+  it 'creates an instance of the Airport class' do
```
JoshuaNg2332, line 8: Using '  ' on this line but then on other lines, you are using " " for your `it` descriptions.

Dhara-95: Great point, amended this for consistency.

lib/airport.rb
Dhara-95: This was a suggestion from Rubocop to include a line after the fail statement, however i'm not too sure on the reasoning so will look into this.

lib/plane.rb
Dhara-95: Good point, I based this challenge on the boris bikes challenge where we required a separate file for the 'bike' argument/class. However understand now that this is not entirely necessary.

spec/airport_spec.rb
Dhara-95: Great point! Will keep this in mind for the next project.

spec/airport_spec.rb
Dhara-95: Thanks for spotting, will amend this.

PR #1832, opened 2020-06-14
Alessa - Airport Challenge

alessa-lou:


lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  
+  def initialize
+    @hangar = []
+  end  
+    
+  def land(plane)
+    airport = Airport.new
+    @hangar << plane
+    if @hangar.count > airport.capacity
+      raise 'Airport at capacity, no room for more planes'
```
avirene: You could probably do it in a separate method full?
```
  def full?
    @hangar.count >= capacity
  end
```

PR #1833, opened 2020-06-14
Airport Challenge

stupot1:


spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    it 'lands plane when there is space' do
+      expect(airport.land_plane(plane)).to eq "#{plane} landed"
+    end
+    it "raises 'no space' error when airport is full" do
+      (airport.airport_capacity).times {airport.land_plane(plane)}
+      expect {airport.land_plane(plane)}.to raise_error("no space")
+    end
+  end
+
+  describe '#take_off' do
```
indisaurusrex, line 30: is this code reduced so that the other tests can take priority? if so, is it worth commenting them out or deleting them from this change?

lib/airport.rb
```diff
@@ -2,21 +2,26 @@
...
   def land_plane(plane)
+    raise "too stormy to land" if check_weather == "stormy"
     raise "no space" if full?
     @landed_planes << plane
     "#{plane} landed"
   end
 
   def take_off(plane)
+    raise "too stormy to take off" if check_weather == "stormy"
+    raise "plane not at airport" if !plane_at_airport?(plane)
```
indisaurusrex, line 22: line 22, to avoid using a negative if statement, could unless work here? 
raise "plane not at airport" unless plane_at_airport?(plane)

spec/airport_spec.rb
```diff
@@ -2,27 +2,63 @@
...
+      expect {subject.land_plane(plane)}.to raise_error("no space")
     end
+    it "stores the plane in the landed_planes array after landing" do
+      subject.land_plane(plane)
+      expect(subject.landed_planes).to include(plane)
+   end 
+   it 'doesnt allow plane to land if the weather is stormy' do
+    another_airport = Airport.new
+    allow(another_airport).to receive_messages(:check_weather => "stormy")
+    expect {another_airport.land_plane(plane)}.to raise_error("too stormy to land")
```
indisaurusrex, line 26: within curly brackets in these blocks it's best practice to use a space between the bracket and the code block
expect { another_airport.land_plane(plane) }.to

PR #1834, opened 2020-06-14
KJ airport challenge 

katieljones:


mattybwoy:
Good use of one liner syntax, clear concise code well laid out with methods.
Confusing plane variables in lib folder. Could use a stub to write test for random stormy weather generator.
On the whole thank youuuu Katieee

PR #1836, opened 2020-06-14
airport challenge 4/6 user stories completed

Joanne0330:


README.md
```diff
@@ -13,6 +13,16 @@ Airport Challenge
 
 ```
 
+Joanne's Progress:
```
mirimichaelson, line 16: Super small note, but I like the layout of your additions to the README!

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+class Airport
+  DEFAULT_CAPACITY = 20
+  def initialize#(weather: Weather.new)
+    @planes = []
```
mirimichaelson, line 4: Potentially this could be called "hangar" or "landed planes" for more clarity?

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+  def land(plane)
+    fail "No landing allowed. Airport is full." if full?
```
mirimichaelson, line 7: I like the guard clause :)

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    # fail "Landing not allowed due to storm." if weather.stormy?
+    @planes << plane
+  end
+
+  def take_off
+    "Take off allowed, plane no longer in this airport!"
```
mirimichaelson, line 13: Would it be worth removing the plane from the @planes array if it's no longer in the airport?  :)

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+class Airport
+  DEFAULT_CAPACITY = 20
+  def initialize#(weather: Weather.new)
```
mirimichaelson, line 3: Maybe you could set another argument for initialize, so that the capacity of the airport can be changed by inputting an argument? You can also set its default to DEFAULT _CAPACITY so you don't need to enter the capacity :)

You'd need to add another instance variable, maybe called @capacity?

It could look like this (in my mind at least!!):
def initialize(capacity = DEFAULT_CAPACITY)
@capacity = capacity
end

In which case, you can easily change the capacity of the airport! Let me know if that makes sense though :)
If you wanted to do that, line 17 would have to be slightly altered too :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'airport.rb'
+
+# RSpec.configure do |config|
```
mirimichaelson, line 3: I don't think I've seen this before! Could you explain it to me? :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+describe Airport do
+  # describe initialize do
+  #   it 'creates an instance of airport' do
+  #     # airport = Airport.new
+  #     expect(airport).to be_an_instance_of Airport
+  #   end
+  # end
+  describe '#land' do
+    it 'instructs a plane to land' do
+      airport = Airport.new
```
mirimichaelson, line 16: you can set these using the let method, rather than setting them multiple times in the rspec! 
Syntax is:

let (:airport) {Airport.new}

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      plane = Plane.new
+      expect(airport.land(plane)).to eq [plane]
+    end
+  end
+
+  describe '#land not allowed' do
+    it 'raise an error if full of planes' do
+      Airport::DEFAULT_CAPACITY.times do
```
mirimichaelson, line 24: Can you explain the syntax of the double :: to me? I haven't seen it before!

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      subject.land Plane.new
```
mirimichaelson, line 25: I like the use of subject here!

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+#   def stormy?
+#     true
+#   end
+#
+# end
+
+end
+
+class Plane
```
tatsiana-makers, line 30: To answer your question about file structure: it is best practice to put each class in it's own file. This way, someone can get an overview of what classes you have just by looking at the file names. It also means that your files don't get too long 🙂

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+end
```
tatsiana-makers, line 31: There might be some more logic that needs to live within the plane class. To consider why, here are some questions to start out with:
- Can a plane land immediately after it's already landed without taking off?
- Can a plane take off from an airport if it's just taken off without landing?

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'airport.rb'
```
tatsiana-makers, line 1: https://www.tutorialspoint.com/rspec/rspec_test_doubles.htm might be a nice intro to using doubles  🙂

PR #1837, opened 2020-06-14
Miri Airplane Weekend Challenge

mirimichaelson:


README.md
```diff
@@ -13,7 +13,28 @@ Airport Challenge
...
+
+  I want to prevent landing when the airport is full
+
+I would have prevented the same plane landing by calling the include? method on the hangar array.
+
+On User Story 4:
+
+  I would like a default airport capacity that can be overridden as appropriate
+
+I would have used the initialize method, where capacity is set to a default value in the argument.  @capacity is then set to either the overridden value or the default value within the initialize method, which can be called in land_plane method.
```
Joanne0330, line 32: Very clear message to what you've done, and what is already in your thoughts.

lib/airport.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize
+    @landed_planes = []
+  end
+
+  def land(plane)
+    @landed_planes.push(plane)
+  end
+
+  def take_off(plane)
+   "Plane #{@landed_planes[-1]} is taking off}"
```
Joanne0330, line 13: I think there's a syntax error here. An extra } is detected at the end of the string.

lib/airport.rb
```diff
@@ -0,0 +1,16 @@
+require_relative "plane"
```
Joanne0330, line 1: I am just wondering if it's necessary to create another file called plane.rb, with class Plane inside it. Would it be just the same if the class Plane is written in side the airport.rb itself. Let's discuss.

PR #1838, opened 2020-06-15
completed first three user stories

Rabee93:


spec/plane_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'plane'
+require 'airport'
+
+describe Plane do
+    
+  it {is_expected.to respond_to(:land).with(1)}
+  it {is_expected.to respond_to(:takeoff).with(1)}
+   
+  describe'#land' do
```
aravzpatel, line 9: A code linter would be a good tool to use to prevent whitespace or lack of space here.

PR #1839, opened 2020-06-15
My Airport Challenge

04alexklink:
My airport challenge completed up to end of User Story 4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,54 @@
 Airport Challenge
 =================
 
+For this challenge, I began by creating a the Plane and Airport classes. Once done, 
+I created a land method for my Airport Class that could take 1 argument. 
+This solved User Story 1. 
+
+For User Story 2, in order to allow a plane to take-off and know it is no longer in the 
+airport, it seemed like the right time to introduce the ability to store planes in airport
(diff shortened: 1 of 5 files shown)


samlandman:
1) Appreciate at least one thing you like in the pull request. For example: a clear name, something easy to understand, or a good test case.
- Very clear and easy to read code
- Already_Present is a nice method, and well described
- Land method is clear with steps (unlike mine, which was on one line)

Rspec
- I like how you have split the tests by method (#)
- I like how you used the let functionality. 

2) Suggest at least one thing that might improve the code. Try to give a reason why your suggestion would be an improvement.
 - Intermittent lines on coding file (eg 13,15,21 in airport.rb - do these need to be in there?)
- I liked your use of the 'private' mechanism, but if it is not in attr_reader or attr_accessor, is this required as it's already not accessible?
- (Include weather)

Rspec
- Line 4 - airport = Airport.new, but then repeated on 15,19 (15,19 can be deleted as you're creating a new airport)

3) Respond to any questions the author has asked. Illustrate your answers using code snippets and links to relevant resources. If you don't know, then say so.

PR #1841, opened 2020-06-15
Pull

samlandman:
>> I removed the functionality for planes to be removed from the airport, so this is missing currently. 

>> I did not get time to remove the following edge cases:
1) planes can only take off from airports they are in; 
2) planes that are already flying cannot take off and/or be in an airport; 
3) planes that are landed cannot land again and must be in an airport, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 Airport Challenge
-=================
+================
 
 ```
         ______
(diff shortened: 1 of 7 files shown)


jamesAforster:
Haven't seen the underscore in arguments before - didn't know you could do that!

I like that you are using methods that like “full” that just has:

planes.count >= capacity 

I’m guessing that returns a true or false value?

which you can then just put in the ternary operators in “land” I didn’t know you could do that, makes it nice and tidy.

One thing that I spotted is that you have a safe_to_fly variable in airport.rb that returns true or false, it looks like the weather.rb is doing the same. Couldn't see where that was being used? Looks like you could just use to safe_to_fly variable.

PR #1842, opened 2020-07-12
Weekend Challenge - Airport Challenge

phump81:


lib/airport.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'plane'
+
+class Airport
```
harrylb14, line 3: Each method seems to be doing one job, lovely stuff

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  DEFAULT_CAPACITY = 50
+
+  attr_reader :capacity, :hangar, :weather
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+    @weather = rand(10)
```
harrylb14, line 12: potentially would have had Weather as its own class, as its a seperate concept to an Airport

lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+require_relative 'airport'
+
+class Plane
```
harrylb14, line 3: Plane not doing anything, potentially needs some methods for it to have some use? for example, is it landed to make sure it can't land again

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'airport'
```
harrylb14, line 1: tests are nice and concise, but potentially need more of them to cover every line of code (edge cases etc). for example can you check if the plane is in a specific airport, can you make sure it cant take off(/land) if its already taken off(/landed), does taking off remove the plane from the airport etc

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    @hangar << plane
+  end
+
+  def take_off
+    raise "Hangar is empty" if empty?
+
+    raise "May not take off due to inclement weather" if storm?
+
+    @hangar.pop
```
harrylb14, line 28: Might want to review this method to take off a specific plane, rather than the last plane that was landed

README.md
```diff
@@ -1,6 +1,14 @@
 Airport Challenge
```
harrylb14, line 1: excellent use of personalising the readme, definitely something I should have done!

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  it 'changes the capacity' do
+    airport = Airport.new(70)
+    expect(airport.instance_variable_get(:@capacity)).to eq(70)
+  end
+
+  it 'prevents take off in stormy weather' do
+    allow(subject).to receive(:storm?) { false }
+    subject.land(Plane.new)
```
harrylb14, line 29: consider using doubles so there isnt Plane.new in airport spec

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+    "Plane has departed"
+  end
+
+  def full?
+    @hangar.count >= @capacity
+  end
+
+  def empty?
+    @hangar.count.zero?
```
harrylb14, line 37: 1 line methods = lovely to see

PR #1843, opened 2020-07-12
Ralph Mallett - airport challenge

ralphm10:
Work in progress for review.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -4,7 +4,9 @@ class Airport
 
   attr_reader :capacity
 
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -15,12 +17,11 @@ def land(plane)
     @planes.push(plane)
   end
 
-  def full?
-    @planes.length >= @capacity
```
ROCHAAL: What did you understand about the method used in this line? Is it could be better?

PR #1847, opened 2020-07-12
Ryan Viegas - July 2020 Cohort - Week 1 Weekend Airport Challenge

rjkviegas:


lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+require_relative 'plane'
+
+class Airport
+  attr_reader :capacity, :weather, :planes
+
+  DEFAULT_CAPACITY = 100
+
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
+    @capacity = capacity
+    @weather = weather.stormy?
```
melanieanderson1995, line 11: very similar to how we wrote the "broken?" method for boris bikes, good parallel to draw

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    @planes = []
+  end
+
+  def checks_for(plane)
```
melanieanderson1995, line 15: Good check method

lib/plane.rb
```diff
@@ -0,0 +1,19 @@
+require_relative 'weather'
+require_relative 'airport'
+
+class Plane
+
+  def landed
```
melanieanderson1995, line 6: Is this a necessary method? Couldn't you just check if the plane is in the airport in your earlier check method?

spec/airport_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      expect(Airport.new(20).capacity).to eq 20
+    end
+  end
+
+  describe '#checks_for(plane)' do
+
+    it 'tests whether a specific plane is at the airport' do
+      expect(Airport.new.checks_for(Plane.new)).to eq false
+    end
+    it 'tests that true is returned if checked for plane is at the airport' do
```
melanieanderson1995, line 20: Personally when I'm writing code I try to shy away from using object1, object2, etc. Not sure what's best to replace them with but I think it starts to get confusing when instance of new instances you're giving them really specific names

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    planes.include?(plane) ? true : false
+  end
+
+  def full?
+    planes.count >= capacity
+  end
+
+  def land(plane)
+    fail 'Weather conditions do not permit landing' if weather == true
```
melanieanderson1995, line 24: Good fails here

PR #1850, opened 2020-07-12
Weekend Challenge #1

02ship:


spec/airport_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'airport'
+describe Airport do
+  it 'can land a plane' do
+    plane = Plane.new
```
DillonBarker, line 4: Instead of creating a new plane, you can just create a double(:plane)

spec/airport_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    subject.takeoff(plane)
+    expect(subject.planes).not_to include(plane)
+  end
+  it 'has a default capacity of planes' do
+    Airport::DEFAULT_CAPACITY.times { subject.land(Plane.new) }
+    expect { subject.land(Plane.new) }.to raise_error "Airport is at maximum capacity"
+  end
+  it 'has a capacity that can be set when initialized' do
+    airport = Airport.new(30)
+    30.times { airport.land(Plane.new) }
```
DillonBarker, line 24: Remove magic numbers with Airport::DEFAULT CAPACITY

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
...
+    
+    @planes << plane
+  end
+
+  def takeoff(plane)
+    @planes.delete(plane)
+    plane
+  end
+
+  private
```
DillonBarker, line 21: Good use of the private methods

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
```
DillonBarker, line 1: In order to comply with the user stories I think that you can't have a plane that is flying to takeoff, and currently I think this would happen if you made a test. Therefore, you may need (im not sure) a method in planes which sets the plane as flying or grounded. Then you can call this within takeoff and landing so that a plane that is landed doesnt land again.

spec/airport_spec.rb
DillonBarker: like done in above test ^^

PR #1851, opened 2020-07-13
MatteoMicozzi AirPort challenge

MatteoMicozzi:
I would like to have a feedback, Thanks.

### DIFF
--- a/lib/AirPort.rb
+++ b/lib/AirPort.rb
@@ -8,6 +8,7 @@ class AirPort
   end
 
   def land
+    fail 'AirPort reached max capacity!' if @hangar.length >= 200
     @hangar << Plane.new
     'A new plane arrived!'
   end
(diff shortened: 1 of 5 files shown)


lib/AirPort.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'Plane'
+
+class AirPort
```
mark-james-coder, line 3: I believe Airport would be better, as it is 1 word.

lib/AirPort.rb
```diff
@@ -0,0 +1,28 @@
...
+  attr_reader :hangar
+  attr_accessor :capacity
+
+  def initialize
+    @hangar = []
```
mark-james-coder, line 8: I like the naming of the planes array being called 'hangar', rather than just, for example, 'planes'

PR #1853, opened 2020-07-13
Challenge Completion

mark-james-coder:


spec/weather_spec.rb
```diff
@@ -0,0 +1,18 @@
+describe Weather do
+
```
MatteoMicozzi, line 2: For weather I personally prefer to use a method of airport class

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
+require 'plane'
+
```
MatteoMicozzi, line 2: The code is clean and understandable.

PR #1855, opened 2020-07-13
Airport_challenge Patricia 

ROCHAAL:
_  Summarise what user stories are included in the pull request
- Point anything incomplete
_ Draw attention to anything you'd like to improve

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,7 @@
+class Airport
+  def recieve_plane
+    Plane.new
+  end
+  def release(plane)
+  end   
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,5 +1,11 @@
 class Airport
+  def initialize(capacity)
+    @capacity = capacity
+    @planes = []
+  end
   def land(plane)
+    raise 'Cannot land plane: airport full' if @planes.length >= @capacity
+    @planes << plane
   end
 
   def take_off(plane)
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,7 +1,7 @@
 class Airport
-  def recieve_plane
-    Plane.new
+  def land(plane)
   end
-  def release(plane)
-  end   
+
+  def take_off(plane)
+  end
 end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,5 +1,3 @@
 class Plane
-  def landing?
-    true 
-  end
+  true
 end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,5 @@
+class Plane
+  def landing?
+    true 
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,6 +1,10 @@
 require 'airport'
 describe Airport do
-  it {is_expected.to respond_to :recieve_plane}
-
-  end
-  
+ subject(:airport) { described_class.new }
+ it 'instructs a plane to land' do
+   expect(airport).to respond_to(:land).with(1).argument
+ end
+ it 'instructs a plane to take off' do
+   expect(airport).to respond_to(:take_off).with(1).argument
+ end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,10 +1,20 @@
 require 'airport'
 describe Airport do
- subject(:airport) { described_class.new }
+ subject(:airport) { described_class.new(20) }
+
  it 'instructs a plane to land' do
    expect(airport).to respond_to(:land).with(1).argument
  end
+
  it 'instructs a plane to take off' do
    expect(airport).to respond_to(:take_off).with(1).argument
- end
+end
+
+  it 'does not allow landing when at capacity' do
+    plane = double :plane
+   20.times do
+     airport.land(plane)
+   end
+  expect { airport.land(plane) }.to raise_error 'Cannot land plane: airport full'
+end
 end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,6 @@
+require 'airport'
+describe Airport do
+  it {is_expected.to respond_to :recieve_plane}
+
+  end
+  
--- a/spec/plane_spec.rb
+++ b/spec/plane_spec.rb
@@ -1,4 +1,3 @@
 require 'plane'
 describe Plane do
-  it { is_expected.to respond_to :landing?}
 end
--- a/spec/plane_spec.rb
+++ b/spec/plane_spec.rb
@@ -1,0 +1,4 @@
+require 'plane'
+describe Plane do
+  it { is_expected.to respond_to :landing?}
+end


lib/airport.rb
```diff
@@ -0,0 +1,13 @@
...
+    @planes = []
+  end
+  def land(plane)
+    raise 'Cannot land plane: airport full' if @planes.length >= @capacity
+    @planes << plane
+  end
+
+  def take_off(plane)
+  end
+end
```
ralphm10, line 13: Good method names, short and they describe what they are meant to do. The class is well designed

lib/airport.rb
```diff
@@ -0,0 +1,13 @@
+class Airport
+  def initialize(capacity)
+    @capacity = capacity
+    @planes = []
+  end
+  def land(plane)
```
ralphm10, line 6: Leaving a line space here could make code more readable

lib/airport.rb
```diff
@@ -0,0 +1,13 @@
+class Airport
+  def initialize(capacity)
+    @capacity = capacity
+    @planes = []
+  end
```
ralphm10, line 5: Leaving a line space here could make code a bit more readable

lib/airport.rb
```diff
@@ -0,0 +1,13 @@
+class Airport
+  def initialize(capacity)
```
ralphm10, line 2: Here you could add a default capacity

lib/airport.rb
```diff
@@ -0,0 +1,67 @@
+class Airport
+attr_reader :stormy_weather,
+
+def initialize(capacity, stormy_weather)
+  @capacity = capacity
+  @stormy_weather = stormy_weather
+  @plane = []
```
tatsiana-makers, line 7: We don't use this array anywhere in the code yet, but I think we should 🙂 

In which method do we want to push something into this array?
In which method do we want to remove an element from the array?

lib/airport.rb
```diff
@@ -0,0 +1,67 @@
+class Airport
+attr_reader :stormy_weather,
+
+def initialize(capacity, stormy_weather)
```
tatsiana-makers, line 4: One of the user stories says:
"I would like a default airport capacity that can be overridden as appropriate"

How can we implement a default capacity here?

lib/airport.rb
```diff
@@ -0,0 +1,67 @@
...
+   end
+  end
+
+ def land(plane)
+   if @stormy_weather == true
+     false
+   else
+     true
+   end
+ end
```
tatsiana-makers, line 35: I think we should combine the `land` and `instruct_plane_land` methods into a single method, as they are performing the same responsibility.

lib/airport.rb
```diff
@@ -0,0 +1,67 @@
...
+
+def initialize(capacity, stormy_weather)
+  @capacity = capacity
+  @stormy_weather = stormy_weather
+  @plane = []
+end
+
+
+  def instruct_plane_land(plane)
+    if @capacity <=20
```
tatsiana-makers, line 12: In this case, capacity is the maximum number of planes that an airport can hold.

So if the capacity of an airport is 20, and 20 planes land at that airport and none of those planes take off, then no other planes should be allowed to land at the airport until one of the 20 planes takes off and frees up space

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+attr_reader :stormy_weather, :capacity, :good_weather
+
+def initialize(capacity, stormy_weather)
+  @capacity = capacity
+  @stormy_weather = stormy_weather
+  @planes = []
+
+end
+
+  def land(new_plane)
```
siellsiell, line 14: One of the user stories says "I want to prevent landing when the airport is full". I see that you also have a `full` method below but you're not using it. How could you prevent a plane from being added to `@planes` if the airport is full in this method?

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+  def full(plane)
+      fail " Full capacity exceede " if  @capacity >= FULL_CAPACITY
+  end
+
+  def take_off
+    if @stormy_weather == false
+      @planes.pop
+    return @planes.count
+    elsif @stormy_weather == true
+      false
```
siellsiell, line 32: what are you trying to communicate by returning `false`? 

In the other branch of the if you return a numeric value so in this branch it would be good to return a numeric value as well -- we typically want to avoid returning different types of values from the same function.

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
+  true
+end
```
siellsiell, line 3: This class seems a little redundant given that there's nothing in it. Maybe you could add a "name" property to planes so they are easier to distinguish (which also makes it easier for debugging and testing!)? 

That could help you for test driving this user story:

"I want to instruct a plane to take off from an airport **and confirm that it is no longer in the airport**"

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+  end
+
+def confirm_no_plane(plane)
```
siellsiell, line 36: The user story says "I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport"

but you are confirming whether there are 0 planes or not. How would you have to change this function to be able to check whether a **particular** plane is in the airport?

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
+class Airport
+
+FULL_CAPACITY = 21
+
+attr_reader :stormy_weather, :capacity, :good_weather
+
+def initialize(capacity, stormy_weather)
+  @capacity = capacity
+  @stormy_weather = stormy_weather
```
siellsiell, line 9: Can you think of a way to make the weather change randomly to meet the requirement below?

"You will need to use a random number generator to set the weather (it is normally sunny but on rare occasions it may be stormy)."

spec/airport_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    airport = Airport.new(20, false)
+    plane = Plane.new
+    20.times{airport.land(plane)}
+    expect(airport.take_off).to eq(19)
+  end
+
+  it' instructs a plane to not take off when the weather is stormy 'do
+   airport = Airport.new(20, true)
+   airport.take_off
+   expect(airport.take_off).to eq(false)
```
siellsiell, line 26: I think this test isn't guaranteed to test what you want to test -- since your airport is empty at the beginning, no plane can take off even if the weather wasn't stormy. For a stronger test it would be better to land a plane in the airport first before calling `take_off`

siellsiell (commented):
Hey Patricia,

thanks for submitting this for review again! You've got landing and take off working but there are a couple of user stories that you can revisit in more detail:


> I want to instruct a plane to take off from an airport and **confirm that it is no longer in the airport** 


At the moment you're only confirming whether the airport is empty or not.

> I want to prevent landing when the airport is full 


At the moment you have a method that checks whether capacity is reached but it doesn't actually prevent the user from calling `land` anyway and adding another plane.

Once you've revisited those maybe you can look at how to make the weather random :)

lib/plane.rb
siellsiell: The other option is to not have a plane class and just use strings (e.g. `airport.land("Boeing")`)

ROCHAAL: Hi Simo, I haven't fully understood above 'could add a "name" property to planes'.

lib/airport.rb
```diff
@@ -12,33 +12,35 @@ def initialize(capacity, stormy_weather)
 end
 
   def land(new_plane)
-      if @stormy_weather == false
-          @planes.push(new_plane)
-      elsif @stormy_weather == true
+        if @capacity >= FULL_CAPACITY
+          raise " Full capacity "
```
siellsiell, line 16: Nice!

lib/airport.rb
```diff
@@ -12,33 +12,35 @@ def initialize(capacity, stormy_weather)
...
 
   def take_off
-    if @stormy_weather == false
-      @planes.pop
-    return @planes.count
-    elsif @stormy_weather == true
-      false
-    end
+      if  @capacity == 0
+        fail " There is no longer plane in the airport "
```
siellsiell, line 28: Nice catch of the edge case here :)

lib/airport.rb
```diff
@@ -12,33 +12,35 @@ def initialize(capacity, stormy_weather)
...
+        if @capacity >= FULL_CAPACITY
+          raise " Full capacity "
+        elsif @stormy_weather == true
          raise " Not allowed to land stormy weather "
-      else
-      end
+        elsif @stormy_weather == false
+            @planes.push(new_plane)
+        else
+        end
```
siellsiell: You could simplify this a little. Do you really need `elsifs` here or would  `if` statements be enough? Remember that `raise` causes the function to exit early, so if the flow of execution reaches line 16, it can never reach line 17. Similarly, if it reaches line 18, it can never reach line 19. How could you rewrite this functionality and still pass your tests using the following structure?

```
def land(new_plane) 
  if <something>
    raise "Full capacity"
   
  if <something else>
    raise "Not allowed to land in stormy weather"
  
  <do something else>
end
```

lib/airport.rb
```diff
@@ -12,33 +12,35 @@ def initialize(capacity, stormy_weather)
...
-   end
-end
+# def confirm_no_plane(plane)
+#
+#    if @capacity == 0
+#      true
+#    elsif @capacity <= 20
+#      false
+#    end
+# end
```
siellsiell, line 45: it's best to remove commented out code entirely to make things look tidy. You can always find it again using your git commit history.

spec/airport_spec.rb
```diff
@@ -20,11 +9,12 @@
...
   end
 
-  it' instructs a plane to not take off when the weather is stormy 'do
-   airport = Airport.new(20, true)
-   airport.take_off
-   expect(airport.take_off).to eq(false)
+    it' instructs a plane to not take off when the weather is stormy 'do
+    airport = Airport.new(20, true)
+    airport.take_off
+    expect(airport.take_off).to eq(false)
```
siellsiell, line 15: Your indents are a bit off may want to install a tool like rubocop to help with spotting formatting and other linting issues

spec/airport_spec.rb
```diff
@@ -36,42 +26,29 @@
...
-#   airport = Airport.new
-#   plane = Plane.new
-#   plane.instruct_plane_land
-#   expect(airport).to eq('no landing')
-# end
-
+     expect{airport.land(plane)}.to raise_error " Full capacity "
+   end
 
+     it 'confirms that there is no longer plane in the airport'do
```
siellsiell, line 47: Nice test!

The user story asked for a little more though -- it asked whether you can confirm whether **a particular plane is in the airport**. For example, can you write a method that would pass these two tests:

```
airport = Airport.new(2, false)
plane1 = Plane.new
plane2 = Plane.new

airport.land(plane1)
airport.land(plane2)

expect(airport.is_plane_in_airport?(plane1)).to eq(true)
expect(airport.is_plane_in_airport?(plane2)).to eq(true)

airport.take_off(plane1)

expect(airport.is_plane_in_airport?(plane1)).to eq(false)
expect(airport.is_plane_in_airport?(plane2)).to eq(true)

```

siellsiell (commented):
Thanks for iterating on this!

lib/plane.rb
siellsiell: Hmm okay, in that case maybe let's forget about it for now -- it isn't too important, I suggested it as a way to help you with the user story but it's not necessary. I left you some comments that clarify the user story some more in my second review.

PR #1858, opened 2020-08-09
Charlotte - Airport challenge completed 

Charlotte-91:
Completed all user stories.
Total test coverage: 94.23%


lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  attr_reader :capacity
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hanger = []
+    @capacity = capacity
+    @current_weather = [] 
+  end
+
+  def land(plane)
```
kristycer, line 12: really like that your plane.rb only includes the plane class and the rest of the methods used for plane are in the main airport.rb file - makes it very readable and easy to understand what is going on

PR #1860, opened 2020-08-09
Completed Airport Challenge

JR-G:


Djura22:
README has an impressive journal of the process you underwent with each new task. I like how clear this is with screenshots relating to tests etc.

Djura22:
The flying and landed methods perhaps weren't necessary here.

Djura22:
Adherence to the TDD method appears to be strong. It shows in how clean the code is. Simplicity is also there in the refactoring you have done. Great work.

PR #1862, opened 2020-08-10
Partial Solution

yu2491:
This is a partial solution as I have been unable to test for stormy weather affecting take off and landing. I really struggled understanding doubles, stubs and mock behaviour so need to do some further reading before completing this challenge.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,4 +1,5 @@
 require_relative 'plane'
+require_relative 'weather'
 
 class Airport
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,2 @@
+class Airport
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -12,6 +12,8 @@ class Airport
   def land(plane)
     raise "Airport full" if full?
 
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def at_airport(plane)
+    @planes.include?(plane)
+  end
+
+  def stormy?
+    @weather.stormy?
+  end
+
+end
```
Josh-Steadman, line 50: Clear defined and suitably named methods

spec/airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect { subject.take_off(plane) }.to raise_error("Plane not in airport")
+    end
+
+    #it "should raise error if weather is stormy" do
+    #  weather = subject.weather
+    #  plane = Plane.new
+    #  subject.land(plane)
+    #  allow(weather).to receive(:stormy?) { :true }
+    #  expect { subject.take_off(plane) }.to raise_error("Too stormy to take off")
+    #end
```
Josh-Steadman, line 68: same issue as me when I tried this. Don't think the variable weather can be applied to allow

PR #1863, opened 2020-08-10
Airport Challenge

Djura22:


fraserbrookhouse:
Test coverage of the weather class could be improved, right now it tests that it responds to current_conditions but doesn't test any of the other functionality. Perhaps also think about if you need to have a weather class at all or if it could just be a weather method for airport?

fraserbrookhouse:
take_off doesn't take an argument, so currently there is no way to choose a specific plane to take off. This doesn't technically contravene the user stories, but does seem like an odd choice when we take into account our knowledge of airports!

fraserbrookhouse:
Your code is very well laid out and easy to follow, it flows nicely.

fraserbrookhouse:
I like the addition of the Diagram to the Readme showing the interaction of the different parts of code.

fraserbrookhouse:
I see you haven't commented your code, right now that is fine as it is a small project and well laid out, but it probably a good habit to get into for the future.

Djura22:
> I see you haven't commented your code, right now that is fine as it is a small project and well laid out, but it probably a good habit to get into for the future.

This is a valid point and a habit that I wish to cultivate moving forward.

Djura22:
> take_off doesn't take an argument, so currently there is no way to choose a specific plane to take off. This doesn't technically contravene the user stories, but does seem like an odd choice when we take into account our knowledge of airports!

This is something I considered when writing my code. I did however decide that it would not be necessary given the brief, and that requiring an argument would be superfluous to requirements.

Djura22:
> Test coverage of the weather class could be improved, right now it tests that it responds to current_conditions but doesn't test any of the other functionality. Perhaps also think about if you need to have a weather class at all or if it could just be a weather method for airport?

Agreed. The weather class caused me many issues and was something that held up the development of the project.

PR #1864, opened 2020-08-10
added whole project because I forgot to fork at the beginning

fraserbrookhouse:


fraserbrookhouse:
Forgot to fork, had to add whole project in one.

lib/airport.rb
```diff
@@ -0,0 +1,66 @@
...
+  def change_capacity(num)
+    @airport_capacity = num
+  end
+
+  # adds plane to landed_planes array
+  # gives an error if plane tries to land when airport is full, weather is stormy or plane is already at airport
+  def land(plane)
+    raise 'Weather not suitable for landing' if weather? == false
+    raise 'Airport full, no space to land' if full?
+    raise 'Plane has already landed at airport' if landed_planes.include?(plane)
```
JR-G, line 27: Just a small Rubocop formatting comment - when I ran it, it advised that I should add a blank line after the guard conditions

lib/airport.rb
```diff
@@ -0,0 +1,66 @@
...
+  # checks if landed_planes array has hit limit of airport_capacity
+  def full?
+    landed_planes.count >= airport_capacity
+  end
+
+  # sets @weather to one of four possibilites randomly
+  # returns false if weather is stormy, true otherwise
+  def weather?
+    possible_weather = %w[sunny cloudy raining stormy]
+    @weather = possible_weather[rand(1..4)]
```
JR-G, line 63: Would `.sample` work here for randomly selecting a value from the array? Just thinking that if you later on add a value to the array it will also require you to remember updating the rand() values.

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  describe '#land' do
+    it 'adds plane to landed_planes array' do
+      allow(subject).to receive(:weather?) { true }
+      plane = Plane.new
+      subject.land(plane)
+      expect(subject.landed_planes).to include(plane)
+    end
+    it 'throws an error when plane tries to land if airport is full to default capacity' do
+      allow(subject).to receive(:weather?) { true }
+      20.times { subject.land Plane.new }
```
JR-G, line 32: Could you use the `airport_capacity` variable in place of the `20`  to ensure it always the default value that is tested?

lib/airport.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  # removes plane from landed_planes array
+  # gives error if weather is stormy or plane is not at the airport
+  def take_off(plane)
+    raise 'Weather not suitable for flight' unless weather?
+    raise 'Plane is not currently at this airport' unless landed_planes.include? plane
+
+    @landed_planes.delete(plane)
+    puts "#{plane} has departed"
```
JR-G, line 40: I like this approach to remove the planes from the array!

spec/airport_spec.rb
fraserbrookhouse: Great suggestion, I have implemented that. Much simpler and scales well.

lib/airport.rb
fraserbrookhouse: Done, my rubocop was being naughty and didn't fix that for me. He has been suitably punished.

lib/airport.rb
fraserbrookhouse: Great suggestion, I have implemented that. Much simpler and scales well

lib/airport.rb
```diff
@@ -0,0 +1,66 @@
...
+
+class Airport
+  attr_accessor :landed_planes
+  attr_accessor :weather
+  attr_accessor :airport_capacity
+
+  # sets airport capacity to 20 as default
+  # creates @weather set to sunny
+  # creates empty array for landed planes
+  def initialize(default_capacity = 20)
```
JR-G, line 11: Is there any expected behaviour within the initialize method that you think would be worth testing for? i.e. it always initializes with an empty array?

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  it 'responds to attr_accessor for weather' do
+    expect(subject).to respond_to(:weather)
+  end
+  it 'responds to attr_accessor for airport_capacity' do
+    expect(subject).to respond_to(:airport_capacity)
+  end
+
+  describe '#land' do
+    it 'adds plane to landed_planes array' do
+      allow(subject).to receive(:weather?) { true }
```
JR-G, line 25: If you add `before do allow(subject).to receive(:weather?) { true }` before the tests, would this allow you to group together all the tests that require `:weather?` to be true - _inside of `land`_ - without having the write the `allow` statement within each test? And vice versa for the false tests

PR #1865, opened 2020-08-10
improve the code 

kristycer:


docs/airport.rb
```diff
@@ -1,18 +1,29 @@
...
+  def intialize(capacity = DEFAULT_CAPACITY)
     @capacity = capacity
-    @plane = 0
-    end 
+    @planes = []
+  end 
+
+  def land(plane)
+    fail "Cannot land plane: airport is full" if @planes
+    @planes = plane
```
Charlotte-91, line 15: Is this supposed to add plane to the @planes array? If so I'm not sure that @planes = plane would push a plane to the array?
Maybe it would be better to use:
@planes << plane 
or
@planes.push(plane)

kristycer: thank you! tried to do exactly what you are suggesting but my unit test kept failing

PR #1866, opened 2020-08-10
Tam pull request for Airport challenge

just-tam:
**User stories in pull request are:**

As an air traffic controller

As an air traffic controller

As an air traffic controller
To ensure safety

**Incomplete parts of code are:**

For second user story - confirming that plane is no longer in the airport (wasn't sure what test I was supposed to write to demonstrate this

**Things I'd like to improve**

Finishing off my second user story

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -19,6 +19,10 @@ describe Airport do
     expect(airport.land(plane)).to eq([])
   end
 
+  it "Expects Airport to respond_to take_off method" do
+    expect(subject).to respond_to(:take_off).with(1).argument
+  end
+
   it "Expects a plane to take off from the airport" do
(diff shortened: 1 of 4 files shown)


spec/airport_spec.rb
```diff
@@ -1,9 +1,20 @@
...
+# As an air traffic controller
+# So I can get passengers on the way to their destination
+# I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
+
 require 'airport'
 describe Airport do
   it "Expects Airport to respond to landing a plane" do
     expect(subject).to respond_to(:land)
   end
+
```
cabju, line 18: I like the spacing in-between each tests which makes it much easier to read which user story applies to which test

PR #1867, opened 2020-08-10
airport challenge tbc

cabju:
_**Done**_

As an air traffic controller 

**_To do_**

As an air traffic controller 

As an air traffic controller 
To ensure safety 

As the system designer

As an air traffic controller 
To ensure safety 

As an air traffic controller 
To ensure safety 

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,3 @@
+class Airport
+
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,7 @@
+require 'Airport'
+describe Airport do
+  it 'check if there is an airport' do
+    airport = Airport.new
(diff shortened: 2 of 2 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,3 @@
+class Airport
+
```
just-tam, line 2: Suggestion: Think about next steps for instructing a plane to land at the airport. Does the Airport class need some sort of method?

spec/airport_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'Airport'
+describe Airport do
```
just-tam, line 2: Appreciate: This is a well written test I can see that it correctly checks to see that if a new airport is created, it becomes an instance of the Airport class.

PR #1869, opened 2020-09-04
SC Weekend challenge 1 pull request

samcolson4:


obean:
Like:
I really like the encapsulation. Every method is well associated with its own class. 
I also like that there is 100% test coverage and tests are logical. 

Improve:
Airport doesn't actually store instances of planes, the add_plane function doesn't add a plane object to the @planes instance variable. You would need to iterate over every plane object created to get the airport inventory.
Can't set the capacity when creating a new airport.

nouns_n_verbs.md
```diff
@@ -0,0 +1,58 @@
+
```
EdwardAndress, line 1: I really love this! :)

spec/plane_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    subject.land(heathrow, weather)
+    subject.location.should eq(heathrow)
+  end
+
+  it 'Should raise an error if the plane is not currently at an airport' do
+    houston = Airport.new
+    expect { subject.take_off(houston, weather) }.to raise_error('The plane is not at this airport.')
+  end
+
+  it 'Doesn\'t let a plane land if the hangar is full' do
```
EdwardAndress, line 23: "If you use double quotes for the doc string you won't have to escape the apostrophe."

PR #1872, opened 2020-09-04
Changed cockpit window

TheDancingClown:
Is this change trivial enough ;)


olliesmith3:
Good points: 
  1) Easy to understand the code
  2) Methods are short and concise and obey the Single Responsibility Principle.
  3) Tests do not repeat themselves and have a good use of doubles and stubs. They test edge cases such as a plane not being able to take off if it is flying.

Potential Improvements:
  1) Could condense some conditionals and guard clauses into one line statements.
  2) Test coverage could be higher
  3) It looks as if the empty? method is not used. Also I think this method already built into ruby (but I could be wrong :|)

PR #1877, opened 2020-09-06
Airport Challenge

olliesmith3:


olliesmith3:
User stories included in the project are as follows: 

As an air traffic controller 

As an air traffic controller 

As an air traffic controller 
To ensure safety 

As the system designer

As an air traffic controller 
To ensure safety 

As an air traffic controller 
To ensure safety 

Incomplete parts: A plane does not have to be in the air to land at an airport. The weather is an instance variable whereas I could have used a module. The weather is defined when creating a new airport and does not change every time a plane lands/takes off.

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @hangar = []
+    @capacity = capacity
+    rand(6) > 4 ? @weather = :stormy : @weather = :sunny
+  end
+
+  def land_plane(plane)
+    fail "There is no space in the hangar" if room_for_plane
+    fail "It is too stormy to land" if @weather == :stormy
+
+    plane_in_hangar(plane) ? "That plane is already in the hangar" : @hangar << plane
```
TheDancingClown, line 15: How readable are ternary statements? Would depend on prior agreed convention.

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    fail "That plane is not in the hangar" unless plane_in_hangar(plane)
+    fail "You cannot take off whilst the weather is stormy" if @weather == :stormy
+
+    @hangar.delete(plane)
+  end
+
+  private
+
+  def room_for_plane
+    @capacity - @hangar.size < 1
```
TheDancingClown, line 28: Can this be refactored for better readability?

spec/plane_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'plane.rb'
+
+describe Plane do
```
TheDancingClown, line 3: Is this test necessary?

TheDancingClown:
Appreciate - naming conventions are good. Easy readability and I can follow the code easily

spec/airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  describe 'land_plane' do
+
+    let(:gatwick) { Airport.new }
+    let(:boeing747) { double :plane }
+    let(:boeing777) { double :plane }
+    let(:dublin) { Airport.new }
+
+    before 'a group of tests that have a plane in the hangar and the weather is sunny' do
+      gatwick.instance_variable_set(:@weather, :sunny)
```
EdwardAndress, line 17: It's cool that you found this method 👍

EdwardAndress: Nice work!

PR #1880, opened 2020-09-07
airport challenge not completed

c-donald:
need to come back and complete

### DIFF
--- a/lib/airport_station.rb
+++ b/lib/airport_station.rb
@@ -1,0 +1,24 @@
+require_relative 'plane.rb'
+
+class AirportStation
+  attr_reader :plane
+
+  def initialize
+    @plane = []
+  end
+
+  def land_plane(plane) 
+    @plane = plane
+  end 
+  
+  def take_off
+    raise 'airport empty' if empty?
+    @plane.pop
+  end 
+  
+  private
+
+  def empty?
+    @plane == 0
+  end   
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,19 @@
+require 'airport_station.rb'
+require 'plane.rb'
+
+describe AirportStation do
+  it 'responds to let Plane land' do
+    plane = Plane.new
+    expect(plane).to respond_to(:land?)
+  end
+  
+  it 'responds to let plane take off' do
+    plane = Plane.new
+    expect(plane).to respond_to(:take_off?)
+  end  
+
+  it 'raises an error when there are no planes at airport' do
+    subject.take_off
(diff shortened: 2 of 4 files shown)


lib/airport_station.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'plane.rb'
+
+class AirportStation
+  attr_reader :plane
+
+  def initialize
+    @plane = []
```
veronicavlee89, line 7: As this is an array, the name of the variable could be changed to indicate that it may hold more than a single plane - e.g. planes or hangar

spec/airport_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+  
+  it 'responds to let plane take off' do
+    plane = Plane.new
+    expect(plane).to respond_to(:take_off?)
+  end  
+
+  it 'raises an error when there are no planes at airport' do
+    subject.take_off
+    expect { subject.take_off? }.to raise_error(RuntimeError, 'Airport empty')
```
veronicavlee89, line 17: I like that you've specified both the expected error type and the message

spec/plane_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  end 
+  
+  it 'Plane lands' do
+    plane = Plane.new
+    expect(plane).to respond_to :land?
+  end    
+
+  it 'Plane Take off' do
+    subject.take_off?
+    expect(subject).to be_flying
```
veronicavlee89, line 16: what does 'be_flying' do? Is it a custom matcher somewhere?

lib/airport_station.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def land_plane(plane) 
+    @plane = plane
```
veronicavlee89, line 11: @plane is initialised as an array but this is setting it to = a single plane. Instead could use << to add the plane to the @plane array, so you don't override a plane that was already landed in the airport

spec/airport_spec.rb
veronicavlee89: I think the method you want to test here may actually be 'take_off' (on the Airport class) rather than 'take_off?' (on the Plane class).

lib/airport_station.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'plane.rb'
+
+class AirportStation
+  attr_reader :plane
+
+  def initialize
+    @plane = []
+  end
+
+  def land_plane(plane)
```
EdwardAndress, line 10: Here, you could add a line that calls plane.land, which then sets the plane's @flying instace var to false.

PR #1882, opened 2020-09-07
Airport - xavier

XavierDefontaine:


lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+  attr_reader :capacity
+  DEFAULT_CAPACITY = 5
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @grounded_planes = []
+  end
+
+  def land(plane)
+    raise "Cannot land anymore plane, airport is full." if @grounded_planes.count >= @capacity
```
cfujiname, line 14: NIce reading of the code with the if statement

spec/airport_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'airport'
+require 'plane'
+
+plane = Plane.new
```
cfujiname, line 4: Maybe add airport = Airport.new, to prevent repetition in the other tests

PR #1888, opened 2020-10-04

dm-devtech:


spec/airport_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+  let(:new_airport) { Airport.new("Hamburg") }
+  let(:new_flight) { Plane.new("Lufthansa") }
+  let(:new_flight2) { Plane.new("British Airways") }
+
+  it 'airport can be named' do
+    expect(new_airport.name).to eq("Hamburg")
+  end
+
+  it 'plane array empty' do
+    expect(new_airport.planes_in_airport).to eq([])
```
dan-holmes, line 14: I think this would be an example of a 'state' based test rather than 'behaviour' based, which they mentioned you should avoid. I think a good clue to this is if your test makes reference to an array, or any other implementation of the behaviour. Perhaps you could create a method in the airport class called 'empty?' which you could check here was true.

spec/airport_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+  it 'airport reponds to capacity_override method' do
+    expect(new_airport).to respond_to(:capacity_override)
+  end
+
+  it 'airport responds to capacity_override with argument' do
+    expect(new_airport).to respond_to(:capacity_override).with(1).argument
+  end
+
+  it 'capacity can be overriden to new value' do
+    new_airport.capacity_override(8)
```
dan-holmes, line 30: If this test passes then the previous two must also pass. Although they were probably a good middle step, you could possibly remove them now.

spec/airport_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+  it 'airport responds to storm_landing method' do
+    expect(new_airport).to respond_to(:storm_landing)
+  end
+
+  it 'airport responds to storm_landing with one argument' do
+    expect(new_airport).to respond_to(:storm_landing).with(1).argument
+  end
+
+  it 'storm_landing shows option to user when there is a storm' do
+    expect { new_airport.storm_landing(new_flight) }.to output("There is a storm. Do you want the plane to land yes or no:").to_stdout
```
dan-holmes, line 55: I can see why this made the challenge hard to complete. Confirmation is a cool feature, but I don't think in the original spec.

spec/airport_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+  end
+
+  it 'airport reponds to storm_decision' do
+    expect(new_airport).to respond_to(:storm_decision)
+  end
+
+  # it 'check that if user input for storm decision is yes the plane is landed' do
+  # new_airport.stub(:storm_decision) {"yes"}
```
dan-holmes, line 63: I'm not sure if this syntax is also valid, but I looked at the syntax [here](https://relishapp.com/rspec/rspec-mocks/v/2-14/docs/method-stubs/allow-with-a-simple-return-value) and you can use

allow(new_airport).to receive(:storm_decision) {"yes"}.

Also, a small point but I would have used a boolean true/false rather than a string "yes" or "no"

spec/airport_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+  end
+
+  it 'check if plane removed from airport' do
+    new_airport.land_plane(new_flight)
+    new_airport.take_off(new_flight)
+    expect(new_airport.planes_in_airport).to eq([])
+  end
+
+  it 'if attempt is make to take off plane not in airport receive error' do
+    expect(new_airport.take_off(new_flight)).to eq("Plane cannot take off as it is not in the airport")
```
dan-holmes, line 119: Ruby and rspec have a different way of handling errors. Here I would have used the rspec syntax:

expect { code }.to raise_error "Cannot take off".

Which would have required you to use the command "raise" in the Ruby code, rather than returning a string.

https://rollbar.com/guides/ruby-raising-exceptions/

lib/Airport.rb
```diff
@@ -0,0 +1,52 @@
...
+  attr_reader :name, :planes_in_airport, :capacity
+
+  def initialize(name)
+    @name = name
+    @planes_in_airport = []
+    @capacity = 5
+  end
+
+  def capacity_override(new_capacity)
+    if new_capacity.class == Integer
```
dan-holmes, line 11: You could neaten up an if statement like this a little by using a one line if statement to handle the error. i.e.

raise "Capacity must be an integer" if new_capacity.class != Integer
@capacity = new_capacity

This especially gets useful if there are multiple errors that could come up, otherwise you'd end up with a lot of nested if statements

lib/Airport.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def land_plane(plane)
+    if plane_already_in_airport(plane) == "Yes"
+      print "Cannot land plane as it is already in the airport"
+    else
+      if @capacity == @planes_in_airport.count
+        print "Airport full cannot land plane"
+      elsif @capacity > @planes_in_airport.count
+        @planes_in_airport << plane.name
```
dan-holmes, line 35: Arrays can store any kind of object, so storing the name of the plane works, but you could also add the plane itself to the array. Both are valid of course, but some features could be easier to write if you store the plane itself

lib/Airport.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+    end
+  end
+
+  def plane_already_in_airport(plane)
+    if @planes_in_airport.include?(plane.name)
```
dan-holmes, line 41: I mentioned elsewhere about using a boolean, but just to really go on about it. If you had, this method could be refactored to just:

@planes_in_airport.include?(plane.name)

lib/Airport.rb
dm-devtech: Thanks this is a much better approach than mine.  I've now corrected this

lib/Airport.rb
dm-devtech: I agree, my approach is creating superfluous lines of code.  This has now been corrected

spec/airport_spec.rb
dm-devtech: I've now remove the unneeded tests.  Thanks for flagging this

spec/airport_spec.rb
dm-devtech: I agree.  Yes/no as strings aren't required when booleans are available.

PR #1891, opened 2020-10-04
Airport Challenge 

MaggieHibberd:
submission of partial completion, user stories reviewed and key objects and messages defined. TDD approach taking to drive code.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,27 @@
+require_relative 'plane'
+
+CAPACITY_DEFAULT = 20
+class Airport
+
+  attr_reader :plane, :capacity
+
+  def initialize(capacity = CAPACITY_DEFAULT)
+    @planes_grounded = []
+    @capacity = capacity
+  end
+
+  def plane_landing(plane)
+    fail 'Airport full' if @planes_grounded.count >= 20
+    # fail 'Whether conditions do not permit landing' if whether == 20
+    @planes_grounded << plane
+    @plane = plane
+  end
+
+  def airport_full
+  end
+
+  def landed(plane)
+    @plane = plane
+  end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,26 +1,27 @@
 require_relative 'plane'
+require_relative 'weather'
 
 CAPACITY_DEFAULT = 20
 class Airport
+  include Weather
 
-  attr_reader :plane, :capacity
+  attr_reader :plane, :capacity, :planes_grounded
 
   def initialize(capacity = CAPACITY_DEFAULT)
     @planes_grounded = []
     @capacity = capacity
-  end
-
-  def plane_landing(plane)
-    fail 'Airport full' if @planes_grounded.count >= 20
-    # fail 'Whether conditions do not permit landing' if whether == 20
-    @planes_grounded << plane
     @plane = plane
   end
 
-  def airport_full
+  def plane_landing(plane)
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'plane'
+require_relative 'weather'
+
+CAPACITY_DEFAULT = 20
+class Airport
+  include Weather
+
+  attr_reader :plane, :capacity, :planes_grounded
+
+  def initialize(capacity = CAPACITY_DEFAULT)
```
otisvg, line 10: great way of having a default capacity but also being able to change it

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+    @planes_grounded = []
+    @capacity = capacity
+    @plane = plane
+  end
+
+  def plane_landing(plane)
+    raise StandardError.new 'Airport full' if @planes_grounded.length >= 20
```
otisvg, line 17: good use of guard condition
suggestion: make it return a more specific error message (e.g. 'cannot land as this airport is full')

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+module Weather
```
otisvg, line 1: just done some research on modules - great situation to use a module in and good practice of SRP

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
...
+
+attr_reader :forecast
+
+  def weather_rand
+    @random_num = (0..50).to_a.sample
+    weather_forecaster
+  end
+
+  def weather_forecaster
+     @random_num == 20 ? @forecast = "Stormy" : @forecast = "Clear"
```
otisvg, line 11: great way of randomizing the weather
suggestion: could this module be made more DRY, and written using just one method?

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+    it "will respond to plane_landing" do
+      # confirm no landing  - stormy weather
```
otisvg, line 7: your tests all look great, and all seem to serve a good purpose
suggestion: use 'describe #method_name   <your individual test here>    end' to get more specific information in the terminal, and improve test readability

spec/weather_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'weather.rb'
+
+describe Weather do
+
+  it 'produces a random number between 1..100' do
+    allow_any_instance_of(Weather).to receive(:rand).and_return(rand(0..100))
```
otisvg, line 6: you are testing for 0..100 in this test but your weather uses 0..50

spec/plane_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'plane'
+
+describe Plane do
+
+  it "will confirm takeoff" do
+    expect(subject).to respond_to :taking_off
```
otisvg, line 6: even though your methods are empty, you've still written tests to check them, which is great practice

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    @planes_grounded << plane
+    @plane = plane
+    # fail 'Whether conditions do not permit landing' if whether == 20
+  end
+
+  def add_landed_plane(plane)
+    @planes_grounded << plane
```
otisvg, line 25: suggestion: your previous plane_landing method has the same functionality as add_landed_plane, perhaps this method should be able to take off a plane and remove it from your planes_grounded array?

PR #1893, opened 2020-10-04
Airport challenge

jotkafomat:


MYREADME.md
```diff
@@ -0,0 +1,59 @@
...
+## Process
+1. Write a README.md
+2. User story: As an air traffic controller
+So I can get passengers to a destination
+I want to instruct a plane to land at an airport.
+
+Functional representation of that story:
+
+| Objects       | Messages      |
+| ------------- | ------------- |
```
timcastillogill, line 29: This is a great thing to do to help plan and see potential issues arise before they do.

MYREADME.md
```diff
@@ -0,0 +1,59 @@
...
+I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
+Functional representation of that story:
+
+| Objects       | Messages      |
+| ------------- | ------------- |
+|Airport        | gives permission to take off|
+|Airport        |confirm that plane is no longer at the airport|
+
+Airport <---- allows plane to take off <--- after take off plane dissapear from hangar
+
```
timcastillogill, line 50: Again, breaking down a method like this I think is great.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+require_relative './plane.rb'
+
+class Airport
+
+  DEFAULT_CAPACITY = 2
```
timcastillogill, line 5: Is this better in the argument of the class? -> Less lines.. ?

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  attr_reader :hangar, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
```
timcastillogill, line 9: Did you need to create a global variable for the Default Capacity? Could you pass it straight into the class as an argument?

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    @hangar = []
+    @capacity = capacity
+  end
+
+  def landing(plane)
+    fail 'No space available' if full?
```
timcastillogill, line 15: This is a nice clear way of writing this. It reads almost like English. Nice

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    return unless giving_permition_for_landing(plane) == true
+
+    plane.current_position = to_s
+    @hangar << plane
+  end
+
+  def take_off(plane)
+    return unless giving_permition_for_take_off(plane) == true
+
+    @hangar.delete(plane)
```
timcastillogill, line 25: That's good because it's specific about the plane that has left the hangar array.

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def asking_for_permission?
+    true
+  end
+
+  def landing_at(airport)
+    airport.landing(self)
+  end
+
+  def taking_off_from(airport)
```
timcastillogill, line 14: Good naming of method so that it reads easily.

MYREADME.md
```diff
@@ -0,0 +1,59 @@
...
+2. User story: As an air traffic controller
+So I can get passengers to a destination
+I want to instruct a plane to land at an airport.
+
+Functional representation of that story:
+
+| Objects       | Messages      |
+| ------------- | ------------- |
+|a traffic controller|      |
+|Passanger|get  to a destination |
```
timcastillogill, line 31: Do you need this one?

MYREADME.md
```diff
@@ -0,0 +1,59 @@
...
+|Plane|asks for permission|
+|Airport|gives permmision for landing|
+
+Airport <---- allows plane to land <--- after landing plane goes to hangar
```
timcastillogill, line 35: It might have been useful to have the 'push' part of moving the plane to hangar in your message table.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def giving_permition_for_landing(plane)
+    return unless plane.asking_for_permission? == true
+
+    true
+  end
+
+  def giving_permition_for_take_off(plane)
```
timcastillogill, line 34: I like your uses of these 'switch' (true/false) mechanisms for the permissions. Does it function? I'm not totally clear how it works. If the plane that's asking permission is 'true' it will permit the landing. My question is, how does it turn to false?

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
+require 'airport'
+
+describe Airport do
+
+  let(:airbus747) { double :plane }
```
timcastillogill, line 5: I like this, I'm not sure how it works fully yet, but great to see you've used it!

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  before do
+    allow(airbus747).to receive(:asking_for_permission?).and_return(true)
+    allow(airbus747).to receive(:current_position=).and_return(nil)
+  end
+
+  context 'during landing' do
+    it 'checks for landing permission' do
+      subject.landing(airbus747)
+      expect(airbus747).to be_asking_for_permission
```
timcastillogill, line 15: Nice. It reads like English, very easy to read. 

This I'm saying, checks that the plane is 'asking for permission' (that method is functioning as it should)

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    it 'gives permistion to landing' do
+      expect(subject.giving_permition_for_landing(airbus747)).to eq true
+    end
+
+    it { is_expected.to respond_to(:landing).with(1).argument }
+
+    it 'puts plane into a hangar after landing' do
+      subject.landing(airbus747)
+      expect(subject.hangar).to include(airbus747)
```
timcastillogill, line 26: Great. Like reading a novel

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  context 'about an airport'
+  it 'has a hangar' do
+    expect(subject.hangar).to be_an_instance_of(Array)
+  end
+
+  it { is_expected.to respond_to :capacity }
+
+  it 'has a variable capacity' do
+    london_heathrow = Airport.new(50)
+    50.times { london_heathrow.landing airbus747 }
```
timcastillogill, line 73: Nice test.

timcastillogill (commented):
These are my comments - Tim

spec/airport_spec.rb
tatsiana-makers, line 5: Nice use of doubles to mock instances of the Plane class!

spec/plane_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'plane'
+
+describe Plane do
+  let(:airbus747) { Plane.new }
+  let(:luton_airport) { Airport.new }
```
tatsiana-makers, line 5: We're using an actual instance of the Airport class here inside of the unit tests for the Plane class. Is there a way that you can think of using mocking to isolate the unit tests instead?

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
+require 'airport'
+
+describe Airport do
+
+  let(:airbus747) { double :plane }
+
+  before do
+    allow(airbus747).to receive(:asking_for_permission?).and_return(true)
+    allow(airbus747).to receive(:current_position=).and_return(nil)
```
tatsiana-makers, line 9: It might be more readable to have these lines within the individual test blocks where they are needed (even if they're repeated a few times across this file), rather than the reader of the code having to keep remind themselves what the stubbing looks like by scrolling up to this `before` block. What do you think?

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      subject.landing(airbus747)
+      subject.landing(tupolev102)
+      subject.take_off(airbus747)
+      expect(subject.hangar).not_to include(airbus747)
+    end
+  end
+
+  context 'about an airport'
+  it 'has a hangar' do
+    expect(subject.hangar).to be_an_instance_of(Array)
```
tatsiana-makers, line 66: Is this testing behaviour or state? It might be helpful to check out [this practical](https://github.com/makersacademy/skills-workshops/blob/master/practicals/testing/behaviour_not_state.md) about this topic 🙂

tatsiana-makers (commented):
Nice work, it looks like you've had a good go at the challenge! I've left a few comments about the tests.

Let me know if you have any questions or want to discuss my comments further 🙂

spec/airport_spec.rb
jotkafomat: Doesn't make it less DRY?

spec/airport_spec.rb
jotkafomat: It's on my Goal list

PR #1896, opened 2020-10-04
airport_challenge

Joshuamac2:


spec/feature/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'airport'
+require 'plane'
+
+RSpec.describe Airport do
```
benlynch1931, line 4: `RSpec.` shouldn't be needed. Provided you initialised RSpec in the main directory, the file will know what you're talking about.

spec/feature/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  # As an air traffic controller
```
benlynch1931, line 5: Love the use of the user stories here to further show what the test does.

lib/airport.rb
```diff
@@ -0,0 +1,19 @@
+class Airport
+  PLANE_UNITS = 10
+  def initalize(capacity: PLANE_UNITS)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    return nil unless capacity == full
```
benlynch1931, line 9: Not sure if you meant `return nil if capacity == full` or `return plane unless capacity == full`. I feel it is a contradictory line of code?

lib/airport.rb
```diff
@@ -0,0 +1,19 @@
+class Airport
```
benlynch1931, line 1: Great practice of SRP. Each method does exactly one thing at the moment!

spec/feature/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'airport'
```
benlynch1931, line 1: These tests are correct! Don't fear RSpec! It definitely is a confusing concept. There are lots of resources out there to help you understand it better and utilise all of its matchers and options.

PR #1904, opened 2020-10-16
My solution fo the airport_challenge

krzy5zt0f-1:


krzy5zt0f-1:
I have done completed all the user stories and as per request, every bit of code is kept in a separate rb file. 
I do not have any specific needs for the feedback on my mind so you decide :) 
I have tried to name commits in a descriptive manner and added some # comments in code, so it should be clear to review; but should you get confused reviewing it, please let me know :)

README.md
```diff
@@ -60,30 +44,18 @@ To ensure safety
...
+It was ennsured that:
 
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first.
-
-* **Submit a pull request early.**
-
-* Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+* All tests passing
+* Test coverage is 100%
+* The code is elegant: passes all the ```rubocop``` quality tests
```
JoeKershaw, line 60: This is something I have seen in the notes but not looked into too much. It could help improve the tests on my code to include this in my work, so good job for using it :)

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+class Weather
+  def conditions
+    ["stormy", "sunny", "sunny", "sunny"][rand(0..3)]
+  end
+
+  def stormy?
+    conditions == "stormy"
+  end
+end
```
JoeKershaw, line 9: Don't have an issue with this being in a separate module because it is clear and neat, but could this have been reduced if you insert line 3 into the airport class?

Then when you check the "stormy" condition you could have just used == instead of defining a method? see comment in the airport module.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'plane.rb'
+require_relative 'weather.rb'
+class Airport
+  attr_reader :hangar, :capacity, :weather
+  @@default_capacity = 1
+
+  def initialize(capacity = @@default_capacity)
+    @capacity = capacity
+    @hangar = []
+    @weather = Weather.new
```
JoeKershaw, line 10: See comment on the weather module 👍 which means weather doesn't need to be an object :)

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def land(plane)
+    fail "Plane is already at the airport" if in?(plane)
+    fail "Bad weather" if @weather.stormy?
```
JoeKershaw, line 15: replace method on weather object with == ?

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    fail "Airport is full" if full?
+
+    @hangar << plane
+  end
+
+  def take_off(plane)
+    fail "Plane is not at the airport" unless in?(plane)
+    fail "Bad weather" if @weather.stormy?
```
JoeKershaw, line 23: see comment above :)

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  def land(plane)
+    fail "Plane is already at the airport" if in?(plane)
+    fail "Bad weather" if @weather.stormy?
+    fail "Airport is full" if full?
+
+    @hangar << plane
+  end
+
+  def take_off(plane)
+    fail "Plane is not at the airport" unless in?(plane)
```
JoeKershaw, line 22: could include #{self} to tell the user where the airport is setting off from? I'll need repeat this comment but can be applied to wherever we are saying "airport" in a string :)

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  private
+  def full?
+    @hangar.size == @capacity
+  end
+
+  def in?(plane)
+    @hangar.include?(plane)
+  end
```
JoeKershaw, line 36: despite my comment around the weather object, you are probably right to include this methods at the simplest levels because of the single responsibility principle, it certainly makes it really easy to read :) 👍

spec/airport_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+require_relative '../lib/plane.rb'
+require_relative '../lib/weather.rb'
+
+describe Airport do
+  it { is_expected.to respond_to(:land) }
+  it { is_expected.to respond_to(:take_off) }
+
+  describe "creating a new Airport" do
+    it "creates a new airport object" do
+      expect(Airport.new).to be_instance_of Airport
```
JoeKershaw, line 11: because you have used describe Airport you don't need to define Airport.new you can replace this with subject :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+    it "has a corect custom capacity" do
+      expect(Airport.new(5).capacity).to eq 5
+    end
+  end
+  describe "#land" do
+    let(:plane) { double :plane }
+    it "lands a plane at an airport" do
+      allow(subject.weather).to receive(:rand).and_return(1)
```
JoeKershaw, line 20: don't think .weather is needed because you only need to check that the object is receiving the rand method :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,57 @@
+require_relative '../lib/airport.rb'
+require_relative '../lib/plane.rb'
+require_relative '../lib/weather.rb'
+
+describe Airport do
```
JoeKershaw, line 5: If you define let(:plane) {double :plane} here you only need to call it once in your tests :)

spec/weather_spec.rb
```diff
@@ -0,0 +1,32 @@
+require_relative '../lib/weather.rb'
+
+describe Weather do
+  it { is_expected.to respond_to(:conditions) }
+  it { is_expected.to respond_to(:stormy?) }
+
+  describe "creating a new Weather" do
+    it "creates a new weather object" do
+      expect(Weather.new).to be_instance_of Weather
```
JoeKershaw, line 9: don't need weather.new :)

JoeKershaw (commented):
Overall this was great, just a couple of things that can be tidied up. My comments around the weather object is probably a matter of personal preference but since you have followed Single Responsibility Principle brilliantly then I'm not bothered if you leave it :)

PR #1905, opened 2020-10-16
Alexis Baker

etelan:
No modules as no shared methods between plane and airport - because plane doesn't have any methods.

Coded as an air traffic controller, so gives permission to land and take off for the planes.

Feature test added - Plane flies from Paris to Berlin, with nice clear skies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,11 @@ Airport Challenge
                 =  ===(_________)
 
 ```
+My Approach
+---------
+I did strict TTD. As you can see from my commits, I would do Red (Make the test and the empty method), Green (Implement), Blue (Mostly Rubocop). 
+Sometimes I would mess it up, commit it as Red not Green, or miss a step, but I feel like I improved a lot.
+Passes all my tests, passes all rubocop (because I did rubocop after each Green stage)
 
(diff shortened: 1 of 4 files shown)


JamieJoyce01:
- I like that your code is well documented
- I can see from your commits that you have done the TDD process very strictly and well performed.
- Spaces through out your code also make it very easy to read
- I like the fact you had the weather come externally from the airport class as this is how it would be done in the real-world.

- Functions: weather_safe and weather_good? could have done with a few comments as I found it difficult to see weather_safes purpose without looking at the unit tests.
-Tests are well documented and laid out in format with no inconsistencies.

JamieJoyce01 (commented):
- I like that your code is well documented
- I can see from your commits that you have done the TDD process very strictly and well performed.
- Spaces through out your code also make it very easy to read
- I like the fact you had the weather come externally from the airport class as this is how it would be done in the real-world.

- Functions: weather_safe and weather_good? could have done with a few comments as I found it difficult to see weather_safes purpose without looking at the unit tests.

- Tests are well documented and laid out in format with no inconsistencies.

PR #1906, opened 2020-10-16
Duncan Morrison pull request

Duncan-7:
Submitting my solution to the airport challenge.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,25 +2,14 @@ class Airport
   attr_reader :weather, :planes, :capacity
 
   def initialize(capacity = 50)
-    @weather = generate_weather
+    @weather = Weather.new
     @planes = []
     @capacity = capacity
   end
 
-  def generate_weather
-    rng = rand(1..5)
-    return rng == 5 ? "stormy" : "sunny"
-  end
-
   def receive_plane(plane)
-    if bad_weather?
-      puts "Weather too dangerous to land"
-      return
-    end
-    if airport_full?
-      puts "Airport at maximum capacity"
-      return
-    end
+    return if landing_problems?
+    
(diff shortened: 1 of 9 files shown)


lib/weather.rb
```diff
@@ -0,0 +1,12 @@
+class Weather
+  attr_reader :weather
+
+  def initialize
```
danbedford56, line 4: Could add arg to change storm chance

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    if plane.location == :flying
+      @planes << plane
+      plane.land(self)
+    else
+      puts "This plane is not currently flying"
+    end
+  end
+
+  def release_plane(plane)
```
danbedford56, line 21: There is a private method for landing problems but not for taking off problems, indicates that this method has multiple responsibilities

spec/airport_spec.rb
```diff
@@ -0,0 +1,144 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { double(:takeoff => nil, :land => nil, :location => :flying) }
+  
+  describe "creating a new airport" do
```
danbedford56, line 6: Not sure if this test is necessary

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+    if bad_weather?
+      puts "Weather too dangerous to land"
+      return true
+    end
+    if airport_full?
+      puts "Airport at maximum capacity"
+      return true
+    end
+  end
+
```
danbedford56, line 45: Should this method not be in the weather class?

danbedford56 (commented):
Nice job!

PR #1907, opened 2020-10-16
complete but no README update

its-intricate:
methods such as land and take off are called on the plane object with the airport passed as an argument.
airport default capacity set as 10.

### DIFF
--- a/spec/plane_spec.rb
+++ b/spec/plane_spec.rb
@@ -1,0 +1,23 @@
+require 'plane'
+
+RSpec.describe Plane do
+
+  it "lands at an airport" do
+    plane = Plane.new
+    airport = Airport.new
+    plane.land(airport)
+    expect(airport.landed_planes.include?(plane)).to eq true
+  end
+
+  it "takes off from an airport" do
+    plane = Plane.new
+    airport = Airport.new
+    plane.land(airport)
+    plane.take_off(airport)
+    expect(airport.landed_planes.include?(plane)).to eq false
+  end
+
+
+
+
+end
--- a/spec/plane_spec.rb
+++ b/spec/plane_spec.rb
@@ -18,7 +18,6 @@ RSpec.describe Plane do
     plane = Plane.new
     plane.land(airport)
     expect(plane.take_off(airport)).to be_a(String)
-    plane.take_off(airport)
     expect(airport.landed_planes.include?(plane)).to eq false
   end
 
--- a/spec/plane_spec.rb
+++ b/spec/plane_spec.rb
@@ -3,20 +3,33 @@ require 'plane'
 RSpec.describe Plane do
 
   it "lands at an airport" do
-    plane = Plane.new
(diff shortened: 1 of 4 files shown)


deborahadebiyi:
Esther's code was easily readable and incorporated the single responsibility principle really well so I could easily identify what each method was doing. The code was DRY and commits were made at appropriate stages so I could easily identify a progression in the code.

deborahadebiyi (commented):
Classes created in separate files which aids readability as well as a good use of attr reader and instance variables. Example messages are clear and give a good indication of what is being tested

deborahadebiyi (commented):
Good use of edge cases and the code is very elegant. There is sufficient use of unit tests also

spec/plane_spec.rb
```diff
@@ -0,0 +1,101 @@
+require 'plane'
+
+RSpec.describe Plane do
+  airport = Airport.new
+  airport1 = Airport.new
+  airport2 = Airport.new
+
+  it "lands at an airport" do
+    Airport.any_instance.stub(:weather) { false }
```
FabihaHassan123, line 9: Since  _Airport.any_instance.stub(:weather) { false } _ has been called a few times you can put this in a before block on line 7 and remove it from the places you have called it so that you don't have to repeat yourself.

spec/plane_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        Airport.any_instance.stub(:weather) { false }
+        airport = Airport.new(1)
+        plane = Plane.new
+        subject.land(airport)
+        message = 'Airport is full.'
+        expect { plane.land(airport) }.to raise_error(RuntimeError, message)
+      end
+
+      it "does not land if airport is stormy" do
+        Airport.any_instance.stub(:weather) { true }
```
FabihaHassan123, line 34: Keep _Airport.any_instance.stub(:weather) { true }_  in to override the before block :)

spec/plane_spec.rb
```diff
@@ -0,0 +1,101 @@
+require 'plane'
+
+RSpec.describe Plane do
+  airport = Airport.new
+  airport1 = Airport.new
+  airport2 = Airport.new
+
+  it "lands at an airport" do
```
FabihaHassan123, line 8: You could put a _describe "#land" do_  before this to make it more readable.

spec/plane_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+    Airport.any_instance.stub(:weather) { false }
+    subject.land(airport)
+    expect(airport.at_airport?(subject)).to eq true
+  end
+
+  it "takes off from an airport and confirms" do
```
FabihaHassan123, line 14: You can put a _describe "#takeoff" do_ and split the two expect lines into two different it blocks.

FabihaHassan123 (commented):
Your code was really neat and organised! The methods were all concise and all your objects have a single responsibility which is great.

PR #1909, opened 2020-10-16
Airport Challenge - Swizel Cardozo

swizelc:
Final submit

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,57 +2,46 @@ require './lib/plane.rb'
 require './lib/weather.rb'
 class Airport
 
-  attr_reader :planes, :CAPACITY, :weather
+  attr_reader :planes, :capacity, :weather
   def initialize(capacity = 20)
     @planes = []
     @takenoffplanes = []
-    @CAPACITY = capacity
+    @capacity = capacity
     @weather = Weather.new
   end
 
   def land(plane)
-    if full?
-      raise "Hanger full, connot land"
-    elsif possible?(plane, @planes)
-      if !(full?) and !(@weather.stormy)
-        @planes << plane
-        return "#{plane} has landed"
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    raise "Weather is stormy cannot takeoff" if @weather.stormy
+
+    @planes.delete_at(@planes.index(plane))
+    @takenoffplanes << plane
+    return "#{plane}, has taken off"
+
+  end
+
+  private
+  def full?
```
zoecode26, line 36: Really like how all methods carry out only a single function and they are really easy to read and understand.

spec/airport_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    it 'does not allow takeoff if plane already takenoff' do
+      allow(subject.weather).to receive(:stormy).and_return false
+      subject.land(:plane)
+      subject.takeoff(:plane)
+      expect { subject.takeoff(:plane) }.to raise_error "Plane already taken off"
+    end
+  end
+
+  describe '#feature test' do
```
zoecode26, line 59: I also really like the use of a feature test in the testing file - it seems like a really great idea to carry out feature tests here as well as in irb!

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
+require './lib/plane.rb'
+require './lib/weather.rb'
+class Airport
+
+  attr_reader :planes, :capacity, :weather
+  def initialize(capacity = 20)
+    @planes = []
+    @takenoffplanes = []
```
zoecode26, line 8: No improvements as such, just curious about the use of the takenoffplanes array and how this is used as against that plane just being removed from the planes array?

PR #1910, opened 2020-10-17
Submission for review

GDCStoney:
There is a link added to the README file for the Development discussion markdown documents

The early User Story feature tests don't work, due to development code changes, but are left for information.

### DIFF
--- a/docs/initial_review.md
+++ b/docs/initial_review.md
@@ -1,0 +1,74 @@
+# Change Log for Airport Individual Challenge
+
+[README file](../README.md)
+
+[Go to Domain Model](#domain-model-review)
+[Go to Design Discussion](#design-discussion)
+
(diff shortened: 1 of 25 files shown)


docs/initial_review.md
```diff
@@ -0,0 +1,76 @@
+# Initial review for Airport Individual Challenge
```
grace-farren, line 1: Very detailed break down of the task given!

grace-farren: Good self points raised

grace-farren (commented):
Appropriate  and well named variable/methods and objects.
The documents are very well laid out, the lib and spec folders are easy to follow. 
The code meets the requirement set my the user story it also does more. 
The code could be simplified. 
It is clear that the test was wrote before the code.

PR #1913, opened 2020-10-18
Ben Jones - Airport Challenge

BenJonesMakers:
The code has 100% test coverage and was written test first.  

I chose to separate out the weather into its own class to add to the reusability.

I attempted to use all of the areas we had covered during the week including stubs, private methods, the singleton principal, feature tests and full TDD.  I believe I have the edge cases covered too.  I've included the feature test script ft1.rb that I used during development.

### DIFF
--- a/ft1.rb
+++ b/ft1.rb
@@ -1,0 +1,22 @@
+require "./lib/airport"
+require "./lib/weather"
+
+# takeoff a plane and land it back, check number aiport inventory
+weather1 = Weather.new
+p luton = Airport.new(weather1)
+p ba737 = Plane.new("ba737")
+p luton.land(ba737)
+p luton.takeoff("ba737")
+p ba737.status
+p luton.planes
+p luton.land(ba737)
+p ba737.status
+p luton.planes
+
+# takeoff multiple planes and land it back, check number aiport inventory
+
+p ej747 = Plane.new("ej747")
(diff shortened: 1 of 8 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    plane.status = "landed"
+    @planes.push(plane)
+    return plane
+  end
+
+  private
+
+  def flying?(plane)
+    plane.status == "flying"
```
LornaRoberts, line 43: I like how clean this section is- it's really minimalist.

ft1.rb
```diff
@@ -0,0 +1,22 @@
...
+p luton = Airport.new(weather1)
+p ba737 = Plane.new("ba737")
+p luton.land(ba737)
+p luton.takeoff("ba737")
+p ba737.status
+p luton.planes
+p luton.land(ba737)
+p ba737.status
+p luton.planes
+
```
LornaRoberts, line 15: Again, I like how this is coded- particularly with the comment explaining what you are trying to achieve.

LornaRoberts (commented):
Really hard to suggest anything to improve. Excellent test coverage. Neat, tidy, clean code. Reassuring the structure is similar to my own, although that might not be a good thing... Just really well done! 👍

PR #1916, opened 2020-10-18
Paris Higgins Airport Challenge

ParisHiggins:
All uses cases tested. Some difficulty when trying to test edge cases so # commented out .


lib/airport2.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'plane2'
+
+class Airport
```
pgavenyte, line 3: Love the short methods. You followed single responsibility principle very well!

lib/airport2.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  attr_reader :capacity, :landed_planes
+
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
```
pgavenyte, line 10: I think we both didn't do it, but if I remember correctly, it's a good practice to put instances in 'private' so it wouldn't be easy for other to mess with them if it makes sense? It was definitely mentioned in one of the walkthroughs in Boris Bikes.

lib/plane2.rb
```diff
@@ -0,0 +1,27 @@
...
+class Plane
+
+  def initialize
+    @airport
+    @flying = true
+  end
+
+  def land(airport)
+    airport.landed_planes << self
+    fail "Airport is full" if airport.full?
```
pgavenyte, line 12: great use of fail with raise_error!

pgavenyte (commented):
Paula left comments during code review exercise.

PR #1918, opened 2020-10-19
Ollie Airport Challenge

olliekennedy:


spec/airport_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+  end
+
+  describe '#takeoff' do
+    it 'instructs a plane to takeoff from the airport and confirm it is no longer there' do
+      allow_any_instance_of(Airport).to receive(:rand).and_return(4)
+      plane = Plane.new
+      subject.planes << plane
+      expect(subject.takeoff(plane)).to eq "I, #{plane}, have taken off."
```
jckmrrs, line 54: I like how a specific plane takes off, and you confirm to the user which plane it was!

spec/airport_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+        big_airport.land(plane)
+      }
+      extra_plane = Plane.new
+      expect(big_airport.land(extra_plane)).to eq "Landing!"
+    end
+    it 'denies landing if the weather is stormy' do
+      allow_any_instance_of(Airport).to receive(:rand).and_return(1)
+      plane = Plane.new
+      expect(subject.land(plane)).to eq "No landing in stormy weather."
+    end
```
jckmrrs, line 45: I like that you've done a lot of checks here to make sure that all instances of the 'land' method were covered.

spec/feature_test.rb
```diff
@@ -0,0 +1,89 @@
+require 'airport'
+require 'plane'
+
+RSpec.describe do
+
+  describe 'feature_test' do
```
jckmrrs, line 6: I like how you included an in-depth feature test that covered all of the unit tests that were mentioned above. You made it clear what each part of the test was going to include, and why you wanted to do it in that order.

spec/feature_test.rb
```diff
@@ -0,0 +1,89 @@
...
+      one_takes_off
+      then_another_lands
+    end
+
+    def given_4_planes_are_at_the_airport
+      @airport = Airport.new
+      one = Plane.new
+      two = Plane.new
+      three = Plane.new
+      four = Plane.new
```
jckmrrs, line 24: I like how you had a simple, but clear and concise naming convention throughout

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def land(plane)
+    return "Airport full, go away." unless @planes.count < @capacity
+    return "No landing in stormy weather." if weather_bad?
+
+    @planes << plane
+    "Landing!"
+  end
+
+  def takeoff(plane)
```
jckmrrs, line 17: From a consistency point of view, I think this is the only method you've written that was 2 or more words, but wasn't written with an '_' (take_off)?

spec/feature_test.rb
```diff
@@ -0,0 +1,89 @@
...
+      20.times {
+        expect(@airport.weather_bad?).to eq false
+      }
+    end
+
+    def three_land
+      five = Plane.new
+      @six = Plane.new
+      seven = Plane.new
+      [five, @six, seven].each do |x|
```
jckmrrs, line 55: Using 'x' may be a little ambiguous, even though it's clear because there's only one line.

olliekennedy: Thank you, sorted!

spec/feature_test.rb
olliekennedy: Thank you!

spec/feature_test.rb
olliekennedy: Thank you!

spec/airport_spec.rb
olliekennedy: Thank you!

spec/airport_spec.rb
olliekennedy: Cheers!

spec/airport_spec.rb
```diff
@@ -53,37 +53,37 @@
...
-        subject.planes << @plane
+        @wizz_air = Plane.new
+        subject.hangar << @wizz_air
       end
 
       it 'instructs a plane to take_off from the airport and confirm it is no longer there' do
-        expect(subject.take_off(@plane)).to eq "I, #{@plane}, have taken off."
-        expect(subject.planes.include?(@plane)).to eq false
+        expect(subject.take_off(@wizz_air)).to eq "I, #{@wizz_air}, have taken off."
+        expect(subject.hangar.include?(@wizz_air)).to eq false
```
Shamarni96, line 62: I like that you have updated your code to be more specific with variable names as it will help the user to understand it more clearly

olliekennedy: Cheers !

PR #1919, opened 2020-10-19
Paula's Airport Challenge

pgavenyte:


pgavenyte:
- The user stories that are outlined in the README.md are complete
- No test for a different capacity other than a default, No RSpec feature test
- I'd like to improve on using doubles and user story breakdown

lib/weather.rb
```diff
@@ -0,0 +1,6 @@
+class Weather
+  
+  def weather_check
+    ["sunny", "stormy", "sunny"].sample
```
ParisHiggins, line 4: great method for checking the weather that can be nicely implemented throughout the code and when writing tests

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_reader :capacity
+
+  def initialize(capacity = 20)
+    @airport = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    return "Can't land because of weather" if weather_check == "stormy"
+    return "Airport full" if full?
```
ParisHiggins, line 13: could alternatively use fail & raise_error where you have currently used return statements

spec/airport_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(airport.list_planes).to eq []
+    end
+  end
+
+  describe '#full' do
+    it 'checks if airport is full' do
+      expect(airport.full?).to eq false
+    end
+  end
+end
```
ParisHiggins, line 49: Really clear and precise tests that read well for an outsider looking at the code for the first time

PR #1920, opened 2020-10-19
airport - Jamie Joyce

JamieJoyce01:


lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
+  def working?
```
etelan, line 2: Suggest: Is this method used? If not, my suggestion would be to delete it.

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    raise "The weather is too stormy to land." if stormy?
+
+    @hangar << plane
+    return @hangar.last
+  end
+
+  def stormy?
+    rand(1..9) <= 3
+  end
```
etelan, line 30: Appreciate: I like the use of a random variable being generated at the airport. Allows for correct stubbing.

etelan: Suggestion: It looks like the stormy? method gets a number between 1 and 9 inclusive, and then if it is equal to or below 3 returns true. 

What do you think about making this `rand(1..3) == 1` instead?

spec/airport_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  describe "Capacity var" do
+    it { is_expected.to respond_to(:capacity) }
+
+    it 'checks to see if capacity can be changed' do
+      truetest = rand(1..100)
+      airport = Airport.new(truetest)
+      allow(airport).to receive(:stormy?) { false }
+      truetest.times { airport.land(Plane.new) }
+      expect { airport.land(Plane.new) }.to raise_error 'There is not enough space to land another plane!'
+    end
```
etelan, line 14: Suggestion: Since you have the `attr_reader :capacity` in your airport.rb document, you could change this to 

```
 it 'checks to see if capacity can be changed' do
      truetest = 74
      airport = Airport.new(truetest)
      expect(airport.capacity).to eq 74
end
```
Then you could include another test for a capacity error.

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def initialize(capacity = 20)
+    @capacity = capacity
+    @hangar = []
+  end
+
+  def takeoff
+    raise "There are no planes currently in the hangar." if @hangar.empty?
+
+    raise "The weather is too stormy to take off." if stormy?
```
etelan, line 14: Appreciate: Nice use of guard clauses.

spec/airport_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  describe "Takeoff function" do
+    it 'checks to see if the plane has taken off' do
+      allow(subject).to receive(:stormy?) { false }
+      subject.land(Plane.new)
+      plane = subject.takeoff
+      expect(plane).to be_working
+    end
```
etelan, line 37: Suggestion: Perhaps you could change this to check if the plane is in the hanger still. Otherwise, this test will pass if the plane leaves the hanger or not, because the plane is always working.

etelan (commented):
nice work

lib/plane.rb
JamieJoyce01: it is used in the rspec test be_working? however I was using it as a check to see if the plane has successfully taken off after however I believe you might be right about it returning true wether its in the hangar or not.

lib/airport.rb
JamieJoyce01: Yes you are exactly right, quick maths

spec/airport_spec.rb
JamieJoyce01: will add that when improving on my code, wasn't aware of that method.

spec/airport_spec.rb
JamieJoyce01: True it was more to check if the plane still existed after it took off as a kind of confirmation however you are right it will return true in either case unless the instance is destroyed.

lib/airport.rb
JamieJoyce01: thank you!

PR #1921, opened 2020-10-19
Follow user stories on task: Tests and code

KalindiFonda:


KalindiFonda:
By example of the Bikes, I did exactly what the user stories asked for and nothing more. It was a good exercise of trying to think about the simplest thing. 

I then went to look into the reviewing guidelines, and saw there is much more implied. I started working on that.

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+class Airport
+
+  attr_accessor :capacity
+  attr_reader :plane_count
+
```
FabihaHassan123, line 5: Have you thought about setting a default capacity which the user can change when needed?

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+  def land
+    fail "Cannot land, it's stormy" if stormy?
+
+    fail "Error: Airport full" if @plane_count >= @capacity
+
+    @plane_count += 1
+  end
+
+  def take_off
```
FabihaHassan123, line 19: Same as before you can add plane as an argument

PR #1922, opened 2020-10-19
Airport Challenge - Grace Farren 

grace-farren:


grace-farren:
**classes**
 - aeroplane
    - can takeoff
    - can land
    - knows weather status
 - airport
    - can receive planes
    - can release planes
    - can count planes
 - weather
    - can be sunny
    - can be stormy

**content**
 - land
    - true  
        - weather isn't stormy
        - capacity isn't full
    - false
        - weather is stormy
        - capacity is full
 - takeoff
    - true
        - weather isn't stormy
    - false
        - weather is stormy
 - check for weather
    - after landing
    - before takeoff

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  def initialize(options = {})
+    self.capacity = options.fetch(:capacity, capacity)
+  end
+
+  def capacity
+    @capacity == 10
+  end
+
+  def total_planes
+    @total_planes ||= []
```
GDCStoney, line 16: This is using a comparator || to determine if it exists ... I THINK this will assign a blank array if nil ... otherwise return the total array of planes
might be better for initialisation - 
@total_planes = @total_planes if @total_planes.nil?

lib/weather.rb
```diff
@@ -0,0 +1,2 @@
+class Weather
```
GDCStoney, line 1: you haven't done anything with this class ... I assume this was due to time available?

spec/airport_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    it 'a plane can takeoff' do
+      airport.receive(plane)
+      expect { airport.takeoff(plane) }.to change { airport.plane_count }.by(-1)
+    end
+  end
+
+  describe 'air traffic control' do
+    it 'a plane canot land if the airport is full' do
```
GDCStoney, line 23: I like the use of stubbing for the land function - to focus on the error state

notes.md
```diff
@@ -0,0 +1,46 @@
+## Airport challenge ##
+
```
GDCStoney, line 2: I like the breakdown document to capture the requirements in a more systematic / development focused way

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def receive(plane)
+    raise 'Full' if full?
+
+    @total_planes << plane
+  end
+
+  def takeoff(plane)
+    total_planes.delete(plane)
```
GDCStoney, line 27: don't know how the "plane" was selected if there were more than one plane at an airport

GDCStoney (commented):
I like the code that is there and the testing

I think with more time, you would have built out the functionality and extended the testing ... but what is there is great! I will take some of this styling into my own code

PR #1923, opened 2020-10-19
Joe K Airport Challenge

JoeKershaw:
I have included some functionality for all of the user requests but they are all accounted for in the airport class. Something that may be required when we start working is figuring out how to split some methods across classes without adding unnecessary code or making it too complicated.

The feature test hasn't been written so this is something I would need to complete in order to have 100% certainty that it all works as it should do.

I'd potentially like to write my tests for the airport in a tidier more structured way but couldn't see an obvious way to do that!

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,20 @@
+class Airport
+  attr_reader :storage
+  def initialize
+    @storage = []
+  end
+  def land(plane)
+    @storage.length < 1 ? @storage << plane : raise("#{plane} can't land at #{self} as the airport is full!")
+    "#{plane} landed at #{self}"
+  end
+  def takeoff(plane)
+    here = @storage.select {|x| plane}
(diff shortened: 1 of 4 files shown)


krzy5zt0f-1:
Things that went well: 
- Code structure is concise and easy to read, you made your code clear enough that it was readable without any comments :) 
- You have used stubbing and doubles, which is a great practice and makes for better unit testing  (no need to electively change test cases when one class potentially fails) . 
- You mostly kept the methods within each class simple and coherent; each method responds for one single function. 

Even better if: 
- You would keep Weather as a separate class, but I personally understand why you would keep it as part of an airport. It can be thought of a weather at the airport as the airport's attribute. 
- You could have kept testings for each class in a separate rb files, for the clarity of access and documentation. 
- in your TDD process, you could have been starting off with "respond_to" RSpec functionality to enhance your RED to GREEN approach. 
- You already know it ;) , feature test could have been implemented.

PR #1925, opened 2020-10-19
Airport GCE

GiorgioCE:
## What?
For the airport challenge, I've created a plane class and a test to check whether there are too many planes in the airport to stop the planes landing.

## Why?
This tells the planes whether there is space to land. If so, the plane can land without endangering the passengers otherwise, no the plane can't land as the airport is full.

## How?
Using a boolean checks whether there are more than 5 planes in the airport. If true, can't land. If less than 5, plane can land. 

## Testing?
Tests run but have 'hard' answers to get the tests to pass. 

## Anything Else?

### DIFF
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,11 @@
+class Plane
+
+  def initialize(plane_count: 0)
+    @plane_count = plane_count
+  end
+
+  def land?
+    @plane_count < 5
+  end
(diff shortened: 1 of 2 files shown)


Keowashere:
- Rspec test need to be more specific for the plane class
- Have a class for airport and intregrate the feature with plane class

PR #1926, opened 2020-10-19
 Airport Challenge 

Keowashere:
## What?
- Airport Challenge; completed airport class w
- Incomplete plane class, 
## Why?
To create a system for air traffic controller
## How?
using Methods and guard conditions, to match user criteria.
## Testing?
Full coverage on airport, a little for plane


spec/plane_spec.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/plane.rb'
```
GiorgioCE, line 1: Not enough tests - need to complete test class for the plane

lib/plane.rb
```diff
@@ -0,0 +1,21 @@
...
+  def flying?
+    @grounded == nil
+  end
+
+  def landing_process(airport)
+    raise "This plane is on the ground" if flying? == false
+    airport.land(self)
+    @grounded = airport
+  end
+
```
GiorgioCE, line 16: Plane class needs finishing as there's nothing for the plan takeoff, only landing.

PR #1927, opened 2020-10-19
airport-challenge

bhupeshpatel1:
User stories included - to prevent landing when the airport is full, assign a default capacity.

The rest is incomplete.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,9 @@
+class Airport
+
+  attr_reader :capacity
+
+  def initialize
+    @capacity = 1
+  end
+
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,14 @@
+require 'airport'
+
+RSpec.describe Airport do
+
+airport = Airport.new
+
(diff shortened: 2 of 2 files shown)


Eunicekon (commented):
I like the fact that you have used an attribute right at the top

lib/airport.rb
```diff
@@ -0,0 +1,9 @@
+class Airport
+
+  attr_reader :capacity
```
Eunicekon, line 3: This is the attribute my first review was referring to

spec/airport_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'airport'
+
+RSpec.describe Airport do
+
+airport = Airport.new
```
Eunicekon, line 5: Looks like you got the right idea but were you looking at maybe creating a new plane?

lib/airport.rb
```diff
@@ -0,0 +1,9 @@
...
+
+  def initialize
+    @capacity = 1
+  end
+
```
Eunicekon, line 8: Code is clean and easy to read

PR #1928, opened 2020-10-19
Airport - Shamarni 

Shamarni96:
This is the start of my code, still got quite a bit of coding to do.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,13 @@
+class Airport
+
+  def initialize
+    @hangar = []
+
+
+  end
+
+  def land
+    @hangar << plane
+  end
+
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,1 @@
+require 'airport' 


lib/airport.rb
```diff
@@ -0,0 +1,15 @@
+require_relative "plane"
+
+class Airport
+  attr_accessor :hangar, :plane
+
+  def initialize
+    @hangar = []
```
olliekennedy, line 7: Nice naming - hangar, descriptive

lib/airport.rb
```diff
@@ -0,0 +1,15 @@
...
+    @plane = plane
+  end
+
+  def land(plane)
+    @hangar.push("easyjet")
```
olliekennedy, line 12: Could use the argument here 'plane' instead of easyjet - that would push whatever plane you give the method into the hangar rather than pushing easyjet every time

spec/airport_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'airport'
+require 'plane'
+
+RSpec.describe Airport do
```
olliekennedy, line 4: Great indentation and formatting - the whole codebase looks really clean

spec/airport_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  it "creates a new airport" do
+  end
+
+
+  describe "#land" do
+    it "instructs a plane to land at airport" do
+      plane = Plane.new
+      expect(subject.land("easyjet")).to eq subject.hangar
```
olliekennedy, line 12: clear separation of the setup(line 11) and the expect - nice and clear

spec/airport_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'airport'
+require 'plane'
+
+RSpec.describe Airport do
+  it "creates a new airport" do
+  end
+
+
+  describe "#land" do
```
olliekennedy, line 9: Well formatted RSpec with nested describes for each function and #s before your function names. Very clear

lib/plane.rb
```diff
@@ -0,0 +1,4 @@
+require_relative "airport"
+class Plane
```
olliekennedy, line 2: Separating out the plane from the airport as a class is good, clear separation

lib/airport.rb
```diff
@@ -0,0 +1,15 @@
+require_relative "plane"
+
+class Airport
+  attr_accessor :hangar, :plane
```
olliekennedy, line 4: Maybe consider using just attr_reader if accessor access isn't needed. Only give as much access as is needed

olliekennedy (commented):
Nice clear and well formatted. Good separation into classes and it is clear what everything is doing.

PR #1930, opened 2020-11-01
Early pull request

malachispencer:


lib/airport.rb
```diff
@@ -0,0 +1,40 @@
+require_relative './plane.rb'
+require_relative './weather.rb'
+include Weather
+
+class Airport
+  attr_reader :planes, :capacity
+
+  DEFAULT_CAPACITY = 50
+
+  def initialize(capacity = DEFAULT_CAPACITY)
```
CodeZeus-dev, line 10: Good implementation of the initialize method so that capacity can be overridden by the user.

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    raise 'Weather conditions unacceptable for landing' if stormy?
+    @planes.push(plane)
+    plane.landed
+  end
+
+  def take_off
+    raise 'Airport empty' if @planes.empty?
+    raise 'Weather conditions unacceptable for take off' if stormy?
+    plane = @planes.shift
+    plane.flying
```
CodeZeus-dev, line 26: Good implementation of the flying class to label a plane as flying

lib/plane.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def flying
+    raise 'Unauthorized, ATC request' if not_atc_request?
+    @docked = false
+  end
+
+  private
+
+  def not_atc_request?
+    !%w[land take_off].include? caller[1][/`.*'/][1..-2]
```
CodeZeus-dev, line 21: Clever usage of regex to filter out result

lib/weather.rb
```diff
@@ -0,0 +1,19 @@
+module Weather
```
CodeZeus-dev, line 1: Weather module seems quite complex and could be simplified by using the rand method

spec/weather_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  context '#current' do
+    it 'returns sunny or stormy at random' do
+      outputs = []
+
+      until outputs.include?(:sunny) && outputs.include?(:stormy) do
+        outputs.push(subject.current)
+      end
+
+      outputs.each do |output|
+        expect(output).to eq(:sunny).or eq(:stormy)
```
CodeZeus-dev, line 13: Instead of using a loop, could use the following:
`expect(["sunny", "stormy"]).to include()`
Inside the include method, put a random run of the weather generator you've written.

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      expect(airport.capacity).to eq(77)
+    end
+  end
+
+  context '#land' do
+    it 'can be called on an Airport' do
+      expect(subject).to respond_to(:land).with(1).arguments
+    end
+
+    it 'adds a Plane to the Airport array' do
```
CodeZeus-dev, line 16: To make the array of planes at the airport private, you'd need to define it as a private method and then use the following to access it from the tests,
`expect(subject.send(:planes)).to eq(the_planes_array)`

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'airport'
+
+describe Airport do
```
CodeZeus-dev, line 3: Good usage of description labels for their corresponding methods inside the class

PR #1931, opened 2020-11-01
Airport weekend challenge completed

AdamusBG:


lib/airport.rb
```diff
@@ -0,0 +1,22 @@
+class Airport
+  attr_reader :planes_in_hangar, :capacity
+
+  def initialize(capacity = 20)
+    @planes_in_hangar = []
+    @capacity = capacity # default capacity of 20
+  end
+
+  def space?
```
kabligh, line 9: I like how this uses the single responsibility principle, but combines 'empty' and 'full' into one method. Makes the code succinct.

lib/plane.rb
```diff
@@ -0,0 +1,30 @@
+class Plane
+  attr_reader :landed_at
+
+  def initialize(airport = -1)
+    @landed_at = airport # nb that we will consider the value of -1 to mean the airport is in flight
```
kabligh, line 5: Bit confused by 'airport in flight' – assuming you mean plane (and therefore not in the airport)?

lib/plane.rb
```diff
@@ -0,0 +1,30 @@
+class Plane
+  attr_reader :landed_at
+
+  def initialize(airport = -1)
```
kabligh, line 4: In general, a really impressive and thorough response to user story 2 to confirm whether planes are in the airport or not. Though, I almost wonder whether this goes too far in terms of the TDD rule of doing the simplest thing to pass the feature test?

PR #1932, opened 2020-11-01
Attempt at Weekend Challenge

alexleesonmill:
- I haven't used doubles in my tests so they are quite long
- It will take a plane off even if I use a different variable than the one that was called so planes aren't tagged
- I need to test for multiple planes landing
- I haven't utilised a separate class for the weather

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -8,21 +8,22 @@ describe Airport do
 
   it 'lands a plane at the airport' do
     plane = Plane.new
+    allow(subject).to receive(:stormy?).and_return(false)
     expect(subject.land(plane)).to eq [plane]
   end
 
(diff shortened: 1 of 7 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    subject.stormy? == false
+    subject.land(plane)
+    expect(subject.takeoff).to eq plane
+  end
+
+  it 'checks that the plane has taken off' do
+    plane = Plane.new
+    subject.land(plane)
+    length = subject.planes.length
+    subject.stormy? == false
```
rafael-oelmann: Rspec tests are very readable. However there are some duplicated setup in some of the logic. I believe using a before loop in rspec could be useful here.

PR #1933, opened 2020-11-01
Airport Exercise

joshuaretallick:
Got most of the way there but couldn't figure out how to generate random weather events without destroying my code :-(


spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'airport'
+
+describe Airport do
+
+  it { is_expected.to respond_to :land }
+  it { is_expected.to respond_to :take_off }
+  it { is_expected.to respond_to :hangar_count }
```
francescoFH, line 7: Nice one line syntax

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
+class Airport
+
+  attr_reader :hangar
+
+  def initialize(weather = 'sunny')
+    @weather = weather
```
francescoFH, line 6: As you said in your comments, the weather is hardcoded. I would have used a random number generator to set the weather as it was requested, using a private method outside the other codes to set a small probability of stormy weather (e.g. 
def stormy?
rand(1..6) > 5
end
)
Like this, you would have 1 in 5 chances of stormy weather. Happy to help you if it's not clear.

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @hangar = []
+  end
+
+  def land(plane)
+
+    raise "The weather is too stormy!" if @weather == 'stormy'
+
+    raise "Airport at Capacity!" if @hangar.length >= 20
```
francescoFH, line 14: Same as the weather, I would have used a private method to make the code easier to read (I've learnt it thanks to google this weekend :D).
I would have just asked if the hangar was full? and created a private method with @hangar.length >= 20 to return true or false.

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  context "weather is stormy" do
+    it 'plane cannot land' do
+      airport = Airport.new('stormy')
+      expect { airport.land Plane.new }.to raise_error("The weather is too stormy!")
+    end
+  end
```
francescoFH, line 14: Very nice and easy to read! I need to learn how to use context as well! Thanks!

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  end
+
+  describe '#take_off method' do
+
+    it "allows the plane to take_off" do
+      airport = Airport.new
+      plane = Plane.new
+      airport.land(plane)
+      expect { airport.take_off(plane) }.to change { airport.hangar.count }.by(-1)
```
francescoFH, line 37: Nice how you have tested the take_off method removing one plane from array hangar

francescoFH (commented):
I like how easy is to read your airport_spec. Reviewing your code I've learnt how to use context and I will certainly use it as well from now on. 
Also, the methods you have used in the airport.rb are very clear to read.

PR #1934, opened 2020-11-01
Weekend Airport challenge

Walshy92:


lib/airport.rb
```diff
@@ -4,17 +4,25 @@ class Airport
...
     raise "NO SPACE" if hangar.length >= capacity
     @hangar << plane
   end
 
   def take_off(plane)
+    raise "HOLY SH*T ITS WET" if stormy?
     hangar.delete(plane)
   end
+
+  private
```
savannaelbey, line 26: great use of private method

savannaelbey:
well refactored, easy to read and good edge case testing

PR #1935, opened 2020-11-01
The Weekend Airport Challenge 

TamMelPer:
Completed User Story 5, Rspec passes and works in IRB.
Screen Shot 2020-11-01 at 6 27 28 PM

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,4 +1,5 @@
 require_relative 'plane'
+require_relative 'weather'
 
 class Airport
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,12 +2,18 @@ require_relative 'plane'
 
 class Airport
 
-  def land(plane)
-    plane = Plane.new
+attr_reader :plane
+
+  def initialize
+    @plane = []
   end
 
-  def take_off
+  def land(plane)
+    @plane << plane
+  end
 
+  def take_off(plane)
+    @plane.pop
   end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -19,6 +19,7 @@ DEFAULT_CAPACITY = 10
   end
 
   def take_off(plane)
+    fail 'Cant take off in stormy weather' if 1
     @plane.pop
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -3,7 +3,11 @@ require_relative 'plane'
 class Airport
 
   def land(plane)
+    plane = Plane.new
+  end
+
+  def take_off
 
   end
-  
+
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -19,7 +19,9 @@ DEFAULT_CAPACITY = 10
   end
 
   def take_off(plane)
-    fail 'Cant take off in stormy weather' if 1
+    weather = Weather.new
+    current = weather.generate
+    fail 'Cant take off in stormy weather' if current == 1
     @plane.pop
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -9,11 +9,17 @@ attr_reader :plane
   end
 
   def land(plane)
+    fail 'Airport Full' if full?
     @plane << plane
   end
 
   def take_off(plane)
     @plane.pop
   end
+
+  private
+    def full?
+      @plane.count >= 10
+    end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -3,9 +3,13 @@ require_relative 'plane'
 class Airport
 
 attr_reader :plane
+attr_accessor :capacity
 
-  def initialize
+DEFAULT_CAPACITY = 10
+
+  def initialize(capacity=DEFAULT_CAPACITY)
     @plane = []
+    @capacity = capacity
   end
 
   def land(plane)
@@ -19,7 +23,11 @@ attr_reader :plane
 
   private
     def full?
-      @plane.count >= 10
+      @plane.count >= @capacity
+    end
+
+    def empty?
+      @plane.empty?
     end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,9 @@
+require_relative 'plane'
+
+class Airport
+
+  def land(plane)
+
+  end
+  
+end
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,0 +1,5 @@
+class Weather
+  def generate
+    rand(1..6)
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,7 @@
+require 'airport'
+
+describe Airport do
+  it 'responds to #land with one argument' do
+    expect(subject).to respond_to(:land).with(1).argument
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,11 +1,17 @@
 require 'airport'
 
 describe Airport do
-  it 'responds to #land with one argument' do
-    expect(subject).to respond_to(:land).with(1).argument
+
+  describe "#land" do
+    it 'responds to #land with one argument' do
+      expect(subject).to respond_to(:land).with(1).argument
+    end
   end
 
-  it 'responds to #take_off with one argument' do
-    expect(subject).to respond_to :take_off
+  describe "#take_off" do
+    it 'responds to #take_off' do
+      expect(subject).to respond_to(:take_off).with(1).argument
+    end
   end
+
 end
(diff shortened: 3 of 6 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      subject.capacity.times { subject.land Plane.new }
+      expect { subject.land Plane.new }.to raise_error 'Airport Full'
+    end
+  end
+
+  describe "#take_off" do
+    it 'responds to #take_off' do
+      expect(subject).to respond_to(:take_off).with(1).argument
+    end
+    it 'prevents take off if stormy(number 1)' do
```
ehwus, line 47: This test is failing for me, see below

`Airport#take_off prevents take off if stormy(number 1)
     Failure/Error: expect { subject.take_off Plane.new }.to raise_error 'Cant take off in stormy weather'
       expected Exception with "Cant take off in stormy weather" but nothing was raised
     # ./spec/airport_spec.rb:50:in `block (3 levels) in <top (required)>'`

README.md
```diff
@@ -3,8 +3,8 @@ Airport Challenge
 
 ```
         ______
-        _\____\___
-=  = ==(____MA____)
+        _\____\____
+=  = ==(__MEL AIR__)
```
ehwus, line 7: 10/10

README.md
```diff
@@ -13,77 +13,107 @@ Airport Challenge
...
-Note that is a practice 'tech test' of the kinds that employers use to screen developer applicants.  More detailed submission requirements/guidelines are in [CONTRIBUTING.md](CONTRIBUTING.md)
+My goals for the weekend challenge were:
 
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first.
+* To be able to write a meaningful test, based on the user requirements. I'll then be able to write code using TDD, to make the test pass.
+* I'll setup and structure a new Ruby project, and turn user requirements into working code.
+* When my code doesn't work, I'll be able to apply a consistent process to resolve the issue.
 
-* **Submit a pull request early.**
+Process
```
ehwus, line 54: Great to see your process and design model here!

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+attr_reader :plane
+attr_accessor :capacity
+
+DEFAULT_CAPACITY = 10
```
ehwus, line 9: Love the avoidance of magic numbers 😎

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def take_off(plane)
+    weather = Weather.new
+    current = weather.generate
+    fail 'Cant take off in stormy weather' if current == 1
+    @plane.pop
+  end
+
+  private
```
ehwus, line 28: Nice encapsculation of the methods that don't need to be externally called

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    def full?
+      @plane.count >= @capacity
```
ehwus, line 30: good use of helper methods here to follow SRP

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
+  def generate
+    rand(1..6)
```
ehwus, line 3: Would it be useful to be more clear here? Numbers and weather aren't immediately correlated to me, and I'm not immediately sure where I'd look to find out how you've interpreted these numbers. Maybe the weather could be explicitly returned?

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+require 'weather'
+
+describe Airport do
+
+  it "Allows the user to create an instance of Airport" do
+    expect(subject).to be_an_instance_of(Airport)
+  end
+
+  it "has a default capacity" do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
```
ehwus, line 11: Great calling of the constant within the class here, very clear and good to avoid magic number!

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  describe "#initialize" do
+    #it "has a variable capacity" do
```
ehwus, line 15: love how you've changed this, much cleaner - do we need the comments now?

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    let(:plane) { Plane.new }
+    it "allows user to set capacity" do
+      described_class::DEFAULT_CAPACITY.times do
+        subject.land(plane)
+      end
+      expect{ subject.land(plane) }.to raise_error 'Airport Full'
+    end
+  end
+
+  describe "#land" do
```
ehwus, line 29: Can planes land if they are on the ground? Can planes take off in the sky?

spec/weather_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'weather'
+
+describe Weather do
+  it "Allows the user to create an instance of Weather" do
+    expect(subject).to be_an_instance_of(Weather)
+  end
+  it 'responds to #generate' do
+    expect(subject).to respond_to :generate
+  end
+  it "give a result of 1 = stormy or 2 to 6 = sunny" do
```
ehwus, line 10: Could you make sure both expected outcomes are returned?

ehwus (commented):
I think this is a really great effort, good testing and use of SRP and encapsculation!

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      # 10.times { subject.land Plane.new } ## Refactor this as below
+      # Airport::DEFAULT_CAPACITY.times do  ## Refactor again as below
+      #   subject.land Plane.new
+      # end
+      subject.capacity.times { subject.land Plane.new }
+      expect { subject.land Plane.new }.to raise_error 'Airport Full'
+    end
+  end
+
+  describe "#take_off" do
```
tatsiana-makers, line 43: The plane can only takeoff if it isn't the number 1 and in your `take_off` method, you randomly generate the number. In order for the test to consistently pass, you'll need to use stubbing. [This](https://codereporter.de/?p=54) is a resource you might find useful 🙂

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail 'Airport Full' if full?
+    @plane << plane
+  end
+
+  def take_off(plane)
+    weather = Weather.new
```
tatsiana-makers, line 22: It looks like you successfully incorporated weather here, nice!

If you were to do the 6th user story, you'd have to do something similar in the `land` method. But repeating code unnecessary is against the Don't Repeat Yourself (DRY) principle so we probably don't want to repeat the following two lines across both methods:
```
weather = Weather.new
current = weather.generate
```

Instead, can you think of a place in the class where you could create a single instance of the Weather class and then re-use that instance to `generate` the weather in the `land` and `takeoff` methods?

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'plane'
```
tatsiana-makers, line 1: The difference between an instance and class methods is that an instance method is called on and acts on a specific instance of the class, while a class method is not related to a specific instance. https://culttt.com/2015/06/10/understanding-class-methods-verses-instance-methods-in-ruby/ is a resource that I like explaining the difference.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+attr_accessor :capacity
+
+DEFAULT_CAPACITY = 10
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @plane = []
+    @capacity = capacity
+  end
+
+  def land(plane)
```
tatsiana-makers, line 16: Should a plane be able to land at airport2 if it is already landed at airport1 without taking off first?

Scenarios like this might require adding some more code to the `Plane` class 🙂

tatsiana-makers (commented):
This is a nice go at the challenge! I've left a few comments and suggested resources for your questions. 

Let me know if you have any further questions 🙂

spec/airport_spec.rb
TamMelPer: Hi Tatsiana, thank you for sending the resource. I can see that stubbing syntax looks like this: 
expect(plane).to receive(:generate_weather).and_return(:stormy)

Isn't my line 49/50 saying the same thing? Is that stub?

      allow(weather).to receive(:generate) { 1 }
      expect { subject.take_off Plane.new }.to raise_error 'Cant take off in stormy weather'

tatsiana-makers: Is the `weather` on line 48 the same as the one that the airport actually relies on when you call `take_off`?

Note that in `take_off`, the code creates a _new_ instance of weather before calling `.generate` on that instance.

PR #1936, opened 2020-11-01
Airport Challenge submission

ehwus:
My take on the Airport Challenge, full details of my implementation in the README


lib/airport.rb
```diff
@@ -1,7 +1,8 @@
 class Airport
-  attr_reader :capacity
-  
-  def initialize(capacity = 50)
+  attr_reader :capacity, :planes
+
+  def initialize(capacity = 50, planes = [])
```
TamMelPer: What do you think about naming your array "storage" or hanger"?

lib/plane.rb
```diff
@@ -0,0 +1,14 @@
+class Plane
+  attr_reader :call_sign
+  @@plane_id = 0
+
+  def initialize
+    @call_sign = create_call_sign
+  end
+
+  def create_call_sign
```
TamMelPer, line 42: I'm afraid I don't understand the purpose of this code.

PR #1938, opened 2020-11-01
Airport Challenge Submission

margmartelo:
Missing last User Story and edge cases.


lib/airport.rb
```diff
@@ -0,0 +1,34 @@
+
+class Airport
+  DEFAULT_CAPACITY = 15
+
+  def initialize(weather, capacity = DEFAULT_CAPACITY )
+    @weather = weather.weather_status
```
MykeNuLeng, line 6: I really like the idea of @weather!

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    if !stormy?
+      @airplanes.delete(plane)
+      return @airplanes.length
+    else
+      raise StandardError.new("The storm is preventing us to takeoff")
+    end
+  end
+
+  def stormy?
+    @weather == "stormy"
```
MykeNuLeng, line 31: I like this line, it's very clean and efficient code!

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    double(:weather)
+  }
+
+  it "instruct a plane to land at an airport" do
+    allow(weather).to receive(:weather_status).and_return "good"
+    airport = Airport.new(weather, 10)
+    plane = Plane.new
+    expect(airport.land(plane)).to eq 1
+  end
+
```
MykeNuLeng, line 20: I love how you have included user stories here so it's easy to see if you're hitting the objectives.

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
```
MykeNuLeng, line 33: I would perhaps like a method that could check if a particular plane is in the airport, so that you could check if it had left?

margmartelo: I totally agree! Defining a separate method for it would make my code more readable, indeed!

MykeNuLeng (commented):
I really liked the approach and the code is well written!

PR #1939, opened 2020-11-01
Airport Challenge submission

horthbynorthwest:


spec/aiport_spec.rb
```diff
@@ -0,0 +1,81 @@
+require 'airport'
+require 'plane'
+describe Airport do
+
+  it 'responds to land method' do
+    allow(subject).to receive(:bad_weather?) { false }
```
Michaelschats, line 6: What does the { false } represent in this code ?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+
+class Airport
+  def initialize(capacity = 10)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    fail "Airport full" if full?
+    fail "Dangerous weather!" if bad_weather?
```
Michaelschats, line 11: If not a fail, does the if bad_weather? statement go to the bad_weather? method and produce the final result from that method ?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def take_off
+    fail "No planes" if empty?
+    fail "Dangerous weather!" if bad_weather?
+
+    @planes.pop
+    "The plane is in the air"
+  end
+
+  private
```
Michaelschats, line 24: What does private represent ?

spec/aiport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  it 'requires 1 argument' do
+    expect(airport).to respond_to(:land).with(1).argument
+  end
+
+  it 'raises an error when full' do
+    allow(airport).to receive(:bad_weather?) { false }
+    # this is maximum length -1 as we've already landed a plane in line 19!!
+    9.times { airport.land Plane.new }
```
Michaelschats, line 32: May be silly question!, but if capacity is 10 why would you want to raise an error after running the loop 9 times?

spec/aiport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    expect { airport.land Plane.new }.to raise_error 'Airport full'
+  end
+end
+
+describe '#take_off' do
```
Michaelschats, line 37: Does this mean take-off is now the subject of these tests so does not need to be explicitly included in the test ?

spec/aiport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  it 'confirms the plane is in the air' do
+    allow(airport).to receive(:bad_weather?) { false }
+    airport.land(Plane.new)
+    expect(airport.take_off).to eq "The plane is in the air"
+  end
+end
+
+describe '#custom capcity' do
```
Michaelschats, line 57: is this a spelling error, capcity?

PR #1940, opened 2020-11-02
Airport Challenge 

rafael-oelmann:
First Attempt

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,3 @@
+class Airport
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,3 +1,11 @@
 class Airport
 
+def land(plane)
+  return plane
+end  
+
+def take_off
+  return "plane has taken off and longer at airport"
+end
+
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,7 +1,8 @@
 class Airport
+  attr_reader :plane
 
 def land(plane)
-  return plane
+  @plane = plane
 end  
 
 def take_off
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,12 +1,27 @@
 class Airport
-  attr_reader :plane
+  attr_reader :plane, :capacity, :hanger
+  CAPACITY = 10
+
+  def initialize(capacity = CAPACITY) 
+  @capacity = capacity
+  @hanger = []
+  end
 
 def land(plane)
-  @plane = plane
+
+  raise "airport is full" if full?
+
+  @hanger.push(plane)
 end  
 
 def take_off
+  @hanger.pop
   return "plane has taken off and longer at airport"
+end
+
+private
+def full? 
+  @hanger.length >= @capacity
 end
 
 end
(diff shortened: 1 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,27 @@
+class Airport
+  attr_reader :plane, :capacity, :hanger
+  CAPACITY = 10
+
+  def initialize(capacity = CAPACITY)
```
alexleesonmill, line 5: Like the use of a constant here, but wonder whether default capacity can just be set as `def initialize(capacity = 10)` and you could avoid having to use two separate lines? Not sure which way is preferable though!

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+  CAPACITY = 10
+
+  def initialize(capacity = CAPACITY) 
+  @capacity = capacity
+  @hanger = []
+  end
+
+def land(plane)
+
+  raise "airport is full" if full?
```
alexleesonmill, line 12: Good use of a guard!

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+
+def take_off
+  @hanger.pop
+  return "plane has taken off and longer at airport"
+end
+
+private
+def full? 
+  @hanger.length >= @capacity
+end
```
alexleesonmill, line 25: Nice use of a private method to allow you to check if an airport is full

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
+require "airport"
+
+describe Airport do
+  it "allows user to create an airport" do
+    expect(subject).to be_instance_of Airport
+  end
```
alexleesonmill, line 6: I don't think I did it, but I'm not sure if we can't delete tests like this as once the class is created, it's not likely to be something that changes

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  
+  # Landing 
+describe "Landing" do
+  it "land is a function" do
+    expect(subject).to respond_to(:land)
+  end
+end
```
alexleesonmill, line 13: You could be more specific in this test, as you pass the land method an argument. You could test whether the airport responds to the method with arguments

spec/airport_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  
+  # Taking off
+describe "Taking off" do
+  it "can take off" do
+    expect(subject).to respond_to(:take_off)
+  end
+
+  it "can confirm that the plane has left the airport" do
+    expect(subject.take_off).to eq "plane has taken off and longer at airport"
+  end
```
alexleesonmill, line 23: This is good to make sure that the whole of the method is running, as the last line returned is this, however it doesn't test whether the plane 'popped' from the hangar, so think it's testing state rather than behaviour

alexleesonmill (commented):
I like how clear the airport.rb file is to read. Each method makes sense to me what it's trying to do and how it's trying to do it. 

I think similar to my code, the tests require some set up so are testing state rather than the actual behaviour. I'm not certain whether having lengthier tests is best practice, it's something we could go through together!

PR #1941, opened 2020-11-02
Airport_challenge

savannaelbey:


savannaelbey:
I would like to request a code review please :)
thank you

lib/Airport.rb
```diff
@@ -0,0 +1,15 @@
+require_relative 'Weather'
+
+class Airport
+  include Weather
```
Walshy92, line 4: Love the use of weather as a Module, wish I had thought of this.

lib/Plane.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'Airport'
+
+class Plane
+
+  def land(airport)
+    fail "Airport full!" if airport.planes.count >= airport.capacity
```
Walshy92, line 6: Also very handy to learn that the guard conditions don't need a line between them, I know the one after is needed, but this looks much cleaner than mine =)

lib/Airport.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  attr_reader :planes
```
Walshy92, line 6: Just a tip with the attr_reader, you can have muiltiple methods on one line separated by comments, but your way does make them easier to read =)

lib/Plane.rb
```diff
@@ -0,0 +1,18 @@
...
+    fail "Stormy conditions preventing landing" if airport.stormy? == true
+
+    airport.planes.push(self)
+  end
+
+  def take_off(airport)
+    fail "It's stormy, cannot take off!" if airport.stormy? == true
```
Walshy92, line 13: I'm not 100% on this as I didn't use weather as a module, but I think you can leave off the "== true" at the end as I think this should be implicit from the "if" statement =)

spec/Plane_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'Plane'
+
+describe Plane do
+  let(:full_airport_not_stormy) { double :airport, :capacity => 1, :planes => [Plane.new], :stormy? => false }
+  let(:stormy_weather) { double :airport, :capacity => 20, :planes => [], :stormy? => true }
+  it 'responds to #land() method with one argument' do
+    expect(subject).to respond_to(:land).with(1).argument
+  end
```
alicelieutier, line 8: You can remove this test, which is made redundant by the next one.

spec/Weather_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'Weather'
+
+describe Weather do
+
+  describe '#stormy?' do
+    it 'states if weather is stormy or not' do
+      expect(Weather.stormy?).to eq true || false
```
alicelieutier, line 7: This line of code is not really correct.
If you want to say that this method either return true or it returns false, you can write this:
```rb
result = Weather.stormy?
expect(result == true || result == false).to be true
```
But there's a larger thing at play here. Whenever you have code that relies on randomness, the way to isolate your tests is to mock/stub that randomness. There are a few ways to do that.
1 - use rspec stubbing - see https://stackoverflow.com/questions/45761380/how-to-stub-rand-in-rspec
2 - use `srand` (I'll let you do some research on that one.

alicelieutier (commented):
Overall you code reads well and does the job. There are two main points you can improve:
* Testing behaviour
* Decoupling your classes
Both topics are explored more this week.

PR #1943, opened 2020-11-02
Airport Challenge submission

samanthagottlieb:


lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+require 'plane.rb'
+require 'weather.rb'
```
stringiest, line 2: Really clever setting this as a class!

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      subject.capacity.times { subject.land(plane) }
+      expect { subject.land(plane) }.to raise_error(error_message)
+    end
+  end
+
+  describe '#capacity' do
+    it 'sets the default airport capacity' do
+      expect(subject.capacity).to eq(3)
+    end
+  end
```
stringiest, line 59: A possible testing extension may be to test that default capacity can be overwritten.

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    it 'prevents landing if airport full' do
+      allow(subject).to receive(:full?).and_return(true)
+      expect { subject.land(plane) }.to raise_error(error_message)
+    end
+
+    it 'raises an error when full' do
+      subject.capacity.times { subject.land(plane) }
+      expect { subject.land(plane) }.to raise_error(error_message)
+    end
```
stringiest, line 52: Have I understood correctly that the difference between these two tests is that the first allows you to set the state of the airport to 'full' and not accept any further planes, while the second requires you to land each plane up to capacity?

spec/airport_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+describe Airport do
+  plane = Plane.new
+  weather = Weather.new
+
+  it 'allows the user to create an Airport instance' do
+    expect(subject).to be_instance_of(Airport)
+  end
+
+  describe '#land' do
```
stringiest, line 11: I love the use of nested describes, it makes it really easy to see what your tests are doing!

PR #1944, opened 2020-11-02
Symion Edward's Airport Challenge Attempt 02.11.20

sedwards93:


adam2021 (commented):
Code looks clean and logical. It's easy to understand what's going on and all objects have helpfully descriptive names. I particularly like your stormy method which (I think) represents weather on a spectrum and anything over a rating of 8 is considered stormy, rather than using a binary of "stormy" or "sunny" - am I right?

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_reader :planes
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    raise "This plane has already landed" if @planes.include?(plane)
```
tatsiana-makers, line 13: What should happen if the plane is already landed at another airport?

spec/airport_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'airport'
```
tatsiana-makers, line 1: I found this file to be really nicely organised by the different scenarios!

One small improvement might be to use a `let` block to define an instance of a plane, to reduce the number of places that `Plane.new` occurs. https://medium.com/@tomkadwill/all-about-rspec-let-a3b642e08d39 is a resource that shows an example of a let block

tatsiana-makers (commented):
Nice go at the challenge! I found your tests to be nicely organised! Though the test file is lengthy, I think it is necessarily lengthy- there are lots of scenarios after-all 🙂

PR #1946, opened 2020-11-02
Michael Schats - Airport Challenge

Michaelschats:


lib/Airport.rb
```diff
@@ -0,0 +1,17 @@
...
+  def initialize(capacity)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    raise 'Airport at full Capacity' if @planes.count >= @capacity
+    @planes << plane
+  end
+
```
horthbynorthwest, line 13: Does take_off need an argument? You already have @planes which is where you should be taking a plane from or your capacity will not go down

spec/User_Stories_spec.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/Airport'
+require './lib/Plane'
+
```
horthbynorthwest, line 3: These tests are solid but why are these in a separate file?

lib/Airport.rb
```diff
@@ -0,0 +1,17 @@
+require './lib/Plane'
+
+class Airport
```
horthbynorthwest, line 3: initialize method is exactly the same as mine. I think it's really solid

lib/Airport.rb
```diff
@@ -0,0 +1,17 @@
...
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    raise 'Airport at full Capacity' if @planes.count >= @capacity
+    @planes << plane
+  end
+
+  def take_off(plane)
```
horthbynorthwest, line 14: For user story 2 it asks for confirmation that the plane that has taken off is no longer at the airport. At the moment the plane is still at the airport. To remove a plane from the @planes array you could use @planes.pop to make the array go down by 1.

PR #1947, opened 2020-11-02
Airport challenge 

kabligh:


lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+    hangar << plane 
+  end
+
+  def take_off(*)
+    @hangar.pop
+  end
+
+  private
+
+  def full?
```
AdamusBG, line 23: Really like the use of the private method here, keeps the public methods above looking a bit more clean and readable

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+    @hangar = []
+  end
+
+  def land(plane)
+    raise "Airport is full" if full?
+
+    hangar << plane 
+  end
+
+  def take_off(*)
```
AdamusBG, line 17: Like the splat operator here, had to look up what it meant, but after researching I can see where you might have been going. Taking off multiple planes at once?

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+    @hangar.pop
```
AdamusBG, line 18: The .pop method will always remove the last, highest-index, value in an array. So this method will make a plane take off, but you cannot yet instruct a specific plane on what to do.

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'airport'
+
+describe Airport do
+
+  describe "#land" do 
+
+    it "raises an error if airport is full" do 
+      Airport::DEFAULT_CAPACITY.times do
```
AdamusBG, line 8: By using `Airport::DEFAULT_CAPACITY` you are accessing the class variable rather than the instance/object variable. Consider using `subject.DEFAULT_CAPACITY`, this would require the capacity to be accessible by the attr_reader.

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  describe "#take_off" do 
+
+    it "instructs a plane to take off" do
+      expect(subject).to respond_to(:take_off).with(1).argument 
+    end 
+
+    it "confirms a plane that has taken off is not in the airport" do 
+      plane = Plane.new
+      subject.take_off(plane)
+      expect(subject.hangar).not_to include plane
```
AdamusBG, line 29: Be careful for false positives with this test - you are checking to see whether the plane is still in the airport's hangar, but it was never there in the first place.

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'airport'
+
+describe Airport do
+
+  describe "#land" do
```
AdamusBG, line 5: Consider replacing `describe` with `context` when nested inside another `describe` block, it allows you to organise tests in the way you want  and is a little cleaner.

PR #1948, opened 2020-11-02
current version

adam2021:


README.md
```diff
@@ -1,89 +1,17 @@
...
-```
-        ______
-        _\____\___
-=  = ==(____MA____)
-          \_____\___________________,-~~~~~~~`-.._
-          /     o o o o o o o o o o o o o o o o  |\_
-          `~-.__       __..----..__                  )
-                `---~~\___________/------------`````
-                =  ===(_________)
+currently:
```
sedwards93, line 4: Well done for writing a README! I completely forgot to do that...

spec/airport_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'atc'
+require 'plane'
+
```
sedwards93, line 3: Really nice and clear RSpec. Puts mine to shame. The descriptions are spot on and this is very easy to navigate

lib/atc.rb
```diff
@@ -0,0 +1,85 @@
...
+  else
+    puts "Cleared for take-off!"
+    @runways.pop
+    puts "#{plane} is airbourne"
+    puts "There are now #{@free_runways+=1} free runways"
+    puts @runways
+  end
+end
+
+def check_weather
```
sedwards93, line 47: I  _think_ this can be refactored into just @weather == "stormy", since it is a boolean expression anyway and will return true if the statement is true, and false if it isn't.

lib/atc.rb
```diff
@@ -0,0 +1,85 @@
+require_relative 'plane'
+
+class Airport
+
+def initialize
+  @weather = ["stormy","sunny"].sample
```
sedwards93, line 6: I think if you were to have an attr_reader at the top of the Airport class for the instance variables, you would be able to remove the "@"s from the rest of the code. Not the most useful suggestion from me, but I remember they seemed to suggest it was good practice in the Boris Bike exercise!

PR #1949, opened 2020-11-29
airport challenge done

aracho1:
User stories: all 

Request to reviewer: 
is my code easy to follow? are there any methods or variables that you can't follow?
any critiques welcome :) !


spec/weather_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'weather'
```
PiperS52, line 1: You could also use a test double to test the randomness of the weather generated

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
+require_relative '../lib/weather'
```
PiperS52, line 1: Not really such an issue, but I think this can also be written just as:
require_relative 'weather'

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+require_relative '../lib/plane'
```
PiperS52, line 2: I'm not necessarily sure what the correct solution is, and may be wrong here, but from what I can see you have an instance weather for the airport rather than each flight/plane. This would allow for a plane to land in sunny weather, but not take-off if then stormy.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+  def parked?(plane)
+    ramp.include?(plane) # check if plane is already parked in the airport
+  end
+
+  # checks plane position, airport capacity and weather for landing
+  def ready_for_landing(plane)
+    raise 'the plane has already landed in the airport' if parked?(plane)
+    raise 'the plane is already parked in another airport' if plane.location == "ground"
+    raise 'sorry, cannot land as airport has no space' if full?
+    raise 'sorry cannot land or take off due to bad weather conditions' unless sunny?
```
PiperS52, line 44: I tried to land, take-off then land same plane instance after then getting sunny weather and raised error about not being able to land due to bad weather (?)

README.md
```diff
@@ -1,89 +1,21 @@
...
-* Write an RSpec **feature** test that lands and takes off a number of planes
-
-Note that is a practice 'tech test' of the kinds that employers use to screen developer applicants.  More detailed submission requirements/guidelines are in [CONTRIBUTING.md](CONTRIBUTING.md)
-
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first.
-
-* **Submit a pull request early.**
-
-* Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+# Airport Challenge
```
Jestfer, line 1: Great job at adding different sections in your README to explain your approach, criteria and what you learned from the challenge 💯

lib/airport.rb
```diff
@@ -1,20 +1,55 @@
...
+    self # return the airport object
+  end
+
+  def takeoff(plane)
+    ramp.delete(plane) if ready_for_takeoff(plane)
+    plane.location = "air"
+    self
+  end
+
+  def sunny?
```
Jestfer, line 27: Whatever parts of your class that you are not exposing outside, could be declared as private (your functions, for example, are public by default if you do not specify `private`. For example, here the function `sunny?` is not being used outside of the Airport class. Think about other parts of your classes that should not be exposed (used only for internal class logic)

lib/airport.rb
```diff
@@ -1,20 +1,55 @@
...
-    # where planes are parked
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @weather = Weather.new
     @ramp = []
   end
 
+  def land(plane)
+    ramp << plane if ready_for_landing(plane)
+    plane.location = "ground" # update plane location
```
Jestfer, line 17: Using instance variables from a different class on a different one couples the implementation of classes too much (this concept is reviewed further on Week 2, so you will have time to get some more practice). 
For example, here you are modifying `location`, which is an instance variable of a Plane instance, inside the Airport class.

Try to think about defining a public interface for each of your classes (what should be exposed as opposed to private).

lib/weather.rb
```diff
@@ -1,7 +1,9 @@
 class Weather
-  
-  def generate
-    [:sunny, :sunny, :sunny, :stormy].sample
+
+  def sunny?
+    FORECAST.sample == :sunny
   end
 
+  FORECAST = [:sunny, :sunny, :sunny, :stormy]
```
Jestfer, line 7: Ideally, constants should be added at the beginning of the class. One challenge if you're up for it: why can Ruby read what FORECAST is on line 4 if you have declared and assigned a value to it after that (line 7)?

spec/airport_spec.rb
```diff
@@ -1,15 +1,87 @@
-require 'airport'
+require 'weather'
+require 'plane'
 
 describe Airport do
+  let(:plane) { Plane.new }
+  
+  BAD_WEATHER_MSG = 'sorry cannot land or take off due to bad weather conditions'
```
Jestfer, line 7: I really like what you've done here declaring and assigning constants for messages to avoid repetition, well done!

spec/airport_spec.rb
```diff
@@ -1,15 +1,87 @@
-require 'airport'
+require 'weather'
+require 'plane'
 
 describe Airport do
+  let(:plane) { Plane.new }
```
Jestfer, line 5: This is something that can be improved. Your tests are not isolated at the moment as you are using a real Plane instance for your Airport tests. Unit Tests are used to test a particular class / part of your code, without any real dependencies, as adding them would mean there is no single unit being tested anymore, but more than one, and would fail the overall purpose of Unit Testing. Take a look at **Doubles** in RSpec.

spec/airport_spec.rb
```diff
@@ -1,15 +1,87 @@
...
+
+    it 'cannot land again if the plane is already on the ground' do
+      allow(subject).to receive(:parked?).and_return(true)
+      expect { subject.land(plane) }.to raise_error(RuntimeError, 'the plane has already landed in the airport')
+    end
+
+    it 'allows landing if good weather and has capacity, and has not parked before' do
+      allow(subject).to receive(:ready_for_landing).and_return(true)
+      subject.land(plane)
+      expect(subject.ramp.include? plane).to be true
```
Jestfer, line 42: This test is very much testing implementation, not behaviour.

A rule of thumb for testing behaviour, is that you only want to test public methods of the class. You want to test what your implementation is based on the requirements and not how you are doing it, as this could change and make your tests more brittle and having to change them more often, whereas the requirements for an application are less likely to change.

spec/weather_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'weather'
 
 describe Weather do
-  it "says whether it is sunny or stormy" do
-    expect(subject.generate).to eq(:sunny).or eq(:stormy) 
+  it "checks wether it's sunny" do
+    expect(subject.sunny?).to eq(true).or eq(false)
```
Jestfer, line 5: Whenever you have code that relies on randomness, the way to isolate your tests is to mock/stub that randomness. 

There are a few ways to do that.
1 - use rspec stubbing - see https://stackoverflow.com/questions/45761380/how-to-stub-rand-in-rspec
2 - use srand (I'll let you do some research on that one).

Jestfer (commented):
Overall, this is a really good airport challenge. You did many things well:
- Adding feature tests
- Good separation of classes
- Really good organisation of tests and bearing reusability in mind

Things to look at to improve:
- Write Unit Tests that are isolated
- Testing Behaviour and not State
- Defining your public interface for each class and making private the elements that should be private

PR #1950, opened 2020-11-29
Weekend Airport Challenge

Will-Helliwell:
Incomplete (was on the final weather stage - had created and tested stormy weather was generated as expcted). I also haven't had time to look at the code review rubric to refactor the final program.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,2 @@
+class Airport
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -10,12 +10,16 @@ class Airport
 
   def land_plane(plane)
     fail "Airport full!" if self.full?
+    fail "Plane is not airborne!" if plane.airborne == false
     @planes << plane
+    plane.airborne = false
+
   end
 
   def take_off_plane(plane)
     fail "Plane not at airport!" if !@planes.include?(plane)
     @planes.delete(plane)
+    plane.airborne = true
   end
 
   def full?
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -5,4 +5,10 @@ class Airport
     @planes = plane
   end
 
+  def take_off_plane(plane)
+    fail "Plane not at airport!" if @planes != plane
+    @planes = nil
+    plane
+  end
+
 end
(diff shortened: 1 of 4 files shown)


spec/airport_spec.rb
```diff
@@ -75,4 +75,16 @@
     end
   end
 
+  describe "#weather_generate" do
+    it "changes the weather to stormy roughly 1 in 10 times" do
+      a1 = subject
+      weather_frequencies = Hash.new(0)
+      for i in 1..5000
+        a1.weather_generate
+        weather_frequencies[a1.weather] += 1
```
lisabardelli, line 84: This piece of code is not very readable. It might be useful to consider to use a stub (https://relishapp.com/rspec/rspec-mocks/v/2-99/docs/method-stubs/allow-with-a-simple-return-value)

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def take_off_plane(plane)
+    fail "Plane not at airport!" if !@planes.include?(plane)
+    @planes.delete(plane)
+    plane.airborne = true
+  end
+
+  def full?
+    @planes.length == @capacity ? true : false
+  end
```
lisabardelli, line 29: This link might be useful to make .full? method private (https://www.rubyguides.com/2018/10/method-visibility/)

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    expect(Airport).to respond_to(:new).with(1).argument
+  end
+  it "can show planes currently at the airport" do
+    p1 = Plane.new
+    subject.land_plane(p1)
+    expect(subject.planes).to eq([p1])
+  end
+  it "has a default capacity of DEFAULT_CAPACITY" do
+    a1 = subject
+    expect{(Airport::DEFAULT_CAPACITY + 1).times{a1.land_plane(Plane.new)}}.to raise_error("Airport full!")
```
lisabardelli, line 17: Very good test! well done!

spec/airport_spec.rb
Will-Helliwell: I have added comments to explain. The stub sets the outcome of a method and so is not to be used here, this test is testing that the outcome of a method is as expected.

spec/airport_spec.rb
Will-Helliwell: Thanks!

PR #1953, opened 2020-11-29
Airport - WIP(pull request) weekend challenge

lukewickens1989:


lukewickens1989:
Please take a look at my code and any reviews are welcome. There is still one test that is failing but I had an issue with not being able to figure our what the issue was as it looked ok. Any critique on my variable/method names and any issues in understanding the workflow or what a block of code does. I would also appreciate any comments on my tests.

MY_README.md
```diff
@@ -0,0 +1,70 @@
...
+25) I will UT this.
+26) I will write an if/else statement to check hangar before allowing planes.
+27) Watch the test pass.
+28) UT for edge cases and try to pass one thousand planes into the array and see if that causes the error to be raised.
+29) push to GH repo
+30) I want, when the new airport is initialized for a hangar size to be provided by the user and if no hangar size then a default size.
+31) My previous UTs tell me that there is a capacity of five currently set. I now want initialize to take a argument and i will FT this.
+32) refactor tests to use subject and set a paramter for capacity.
+33) UT for default capacity being used to determine capacity and UT to test for specified capacity and check UTs pass.
+34) push to GH repo
```
AJSMonty, line 70: Well written README. nice to see your process

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
+require_relative "weather"
+require_relative "plane"
```
AJSMonty, line 2: Nice use of require_relative

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+class Airport
+
+  attr_reader :hangar
+  DEFAULT_CAPACITY = 10
+
+  include Weather
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @hangar = [].take(@capacity)
```
AJSMonty, line 13: Not too sure of the need for .take(@capacity) here as your full? method negates the airport filling past capacity number.

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    plane_departed(plane)
+  end
+
+  def full?
+    @hangar.length >= @capacity
+  end
+  
+  private
+
+  attr_reader :capacity, :weather
```
AJSMonty, line 38: Should be at the top I think

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def in_hangar(plane)
+    @hangar << plane
+  end
+
+  def plane_departed(plane)
+    @hangar.delete(plane)
+  end
+
+  def correct_airport?(plane)
```
AJSMonty, line 48: I can't see anywhere where this method is used, does it need to be included?

lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+module Weather
+
+  def stormy?
+    Kernel.rand(0..100) > 95
```
AJSMonty, line 4: Wish I lived in this climate!

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
+class Plane
+
+  def initialize
+    @flying = true
```
AJSMonty, line 4: I might be being stupid here, but this implies planes are made in the sky? Would it not make more sense to have planes made when landing, so airports could just 'land' a newly made plane. It might save a load of code..?

AJSMonty (commented):
I've learnt a lot and need to read up on 'before' syntax in rspec.

Good job mate, few comments to keep you busy.

lib/plane.rb
lukewickens1989: Indeed, aren't all planes built mid-air :D. So my thought process behind this was to have a default state of the plane to flying so that once the planes leave the airport they revert to their initial state. But I think I could do with redoing this as your right, it makes zero sense

PR #1954, opened 2020-11-29
Submission For Week 1 Weekend Challenge: Airport

chriswhitehouse:
I've worked through, and created my own README_CW.md file to document process. In addition to the unit tests, I have created a feature test script airport_feature_test.rb.

### DIFF
--- a/README_CW.md
+++ b/README_CW.md
@@ -2,9 +2,9 @@
 
 This document outlines my approach to solving the Airport Challenge.
 
-Firstly I Converted the given user stories into domain models:
 
 ## Domain Models
+Firstly I Converted the given user stories into domain models:
 
 ### User Story 1
 
@@ -20,7 +20,7 @@ I want to instruct a plane to land at an airport
 |Plane      |        |
 |Airport    |land          |
 
-Complete: -[x]
+Complete: - [x]
 
 ### User Story 2
 
@@ -36,7 +36,7 @@ I want to instruct a plane to take off from an airport and confirm it is no long
 |Plane      |    |
 |Airport    |takeoff and confirm departure |
 
-Complete: [x]
+Complete: - [x]
 
 ### User Story 3
 
@@ -52,7 +52,7 @@ I want to prevent landing when the airport is full
 |Plane      |            |
 |Airport    |full?       |
 
-Complete: [x]
+Complete: - [x]
 
 ### User Story 4
 
@@ -67,7 +67,7 @@ I would like a default airport capacity that can be overridden as appropriate
 |-----------|------------|
 |Airport   |capacity            |
 
-Complete: [x]
+Complete: - [x]
 
 ### User Story 5
 
@@ -82,7 +82,7 @@ I want to prevent takeoff when the weather is stormy
 |-----------|------------|
 |Airport   |stormy?     |
 
-Complete: [x]
+Complete: - [x]
 
 ### User Story 6
 
@@ -97,7 +97,7 @@ I want to prevent landing when the weather is stormy
 |-----------|------------|
 |Airport  |stormy?     |
 
-Complete: [x]
+Complete: - [x]
 
 ## Test Driven Development of the User Stories
 Having drafted the domain models, I stepped through each user story in turn. Firstly scripting a feature test, then a unit test, and then scripting to pass the unit test accordingly. Refactoring where necessary.
--- a/README_CW.md
+++ b/README_CW.md
@@ -1,0 +1,114 @@
+# README_CW.md
+
+This document outlines my approach to solving the Airport Challenge.
+
+Firstly I Converted the given user stories into domain models:
+
+## Domain Models
+
+### User Story 1
+
+```
+As an air traffic controller
+So I can get passengers to a destination
+I want to instruct a plane to land at an airport
+```
+
+**Domain Model:**
+|**Objects**|**Messages**|
+|-----------|------------|
+|Plane      |        |
+|Airport    |land          |
+
+Complete: [x]
+
+### User Story 2
+
+```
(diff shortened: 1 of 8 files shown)


README_CW.md
```diff
@@ -0,0 +1,114 @@
+# README_CW.md
+
+This document outlines my approach to solving the Airport Challenge.
+
+
+## Domain Models
```
AmanTank187, line 6: Great use of Domain models

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+require './lib/plane.rb'
+
+class Airport
+  attr_reader :capacity
+
```
AmanTank187, line 5: Code is clear and i can follow along very easily. I like that you created multiple methods that serve a single purpose.

spec/airport_spec.rb
```diff
@@ -0,0 +1,117 @@
+require 'airport'
+
+describe Airport do
+  let(:capacity) { 5 }
```
AmanTank187, line 4: Good use of the let keyword

spec/airport_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+  let(:plane) { Plane.new }
+
+  describe '#new' do
+    it 'should instantiate an Airport object with a given capacity argument' do
+      expect(Airport.new(capacity)).to be_instance_of Airport
+    end
+  end
+
+  describe '#land' do
+    it 'should return an instance of a Plane object' do
```
AmanTank187, line 15: Clear tests which helps me understand what you are doing

lib/airport.rb
AmanTank187: Although the code is clear and i can follow along easily, additional comments would make it clearer.

README_CW.md
chriswhitehouse: Thanks Aman!

AmanTank187 (commented):
Overall good and clear program, can't think of much to change at this moment in time.

lib/airport.rb
chriswhitehouse: Thanks Aman, I'll look to add more comments in the future.

README_CW.md
```diff
@@ -0,0 +1,114 @@
+# README_CW.md
```
tatsiana-makers, line 1: Really awesome to read about your approach to the challenge!

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  def initialize(capacity = 1)
```
tatsiana-makers, line 6: The `1` here is a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)). Instead, you could declare a constant within the class like 

`DEFAULT_CAPACITY = 1`  and then use that constant as the default value for the input parameter.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    @capacity = capacity
+    @airport = []
+  end
+
+  def land(plane)
+    raise "Plane already in airport" unless in_airport?(plane) == false
```
tatsiana-makers, line 12: A cleaner way of writing comparisons like `in_airport?(plane) == false` is `!in_airport?(plane)`

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+
+class Airport
+  attr_reader :capacity
+
+  def initialize(capacity = 1)
+    @capacity = capacity
+    @airport = []
+  end
+
+  def land(plane)
```
tatsiana-makers, line 11: Is there anything which would stop a particular instance of plane from landing in the same airport twice? 

This might require a Plane to keep track of more information than it currently does 🙂

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    @airport.count >= @capacity
+  end
+
+  def chance
+    rand(10)
+  end
+
+  def stormy?
+    chance < 3
+  end
```
tatsiana-makers, line 45: Currently the Airport class is responsible for the motion of planes as well as determining what the weather is. In order to more closely follow the Single Responsibility Principle, we could extract this logic out into a separate class.

spec/airport_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+  end
+
+  describe '#chance' do
+    it 'should return a random number between 0 and 10' do
+      expect(airport.chance).to be_between(0, 10)
+    end
+  end
+
+  describe '#stormy?' do
```
tatsiana-makers, line 98: Nice use of describe blocks to organise tests!

spec/airport_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+
+    it 'should return false when the airport is not stormy' do
+      allow(airport).to receive(:chance) { 8 }
+      expect(airport.stormy?).to eq false
+    end
+  end
+end
+
+describe Plane do
```
tatsiana-makers, line 111: As a convention, just as each class is defined in it's own file, the unit tests for each class are usually in their own file.

tatsiana-makers (commented):
Thank you for asking for feedback! This is a really nice go at the weekend challenge! I've left a few suggestions around things like SRP and magic numbers, as well as a question which might lead to the Plane class to contain more code than it currently does 😄

lib/airport.rb
chriswhitehouse: That's interesting. I did play with both options. But wondered if the `==  false` syntax is more readable?

lib/airport.rb
chriswhitehouse: Cool, thanks @tatsiana-makers. I did feel like the Plane class was a bit empty!

PR #1956, opened 2020-11-29
Airport Challenge - Lisa Bardelli

lisabardelli:


lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+class Air_traffic_controller
+  def take_off(airport, plane)
+    if airport.get_weather == :sunny && plane.is_flying == false
+      if are_same_locations(plane.location, airport.location) == true
+        return true
+        else 
+          raise RuntimeError.new('Location must be wrong')
+        end
+    else 
+      if airport.get_weather == :stormy
```
Will-Helliwell, line 12: I think it might make the code more readable if you put the 'fail if stormy weather' and 'fail if plane already flying' at the start of the method as [guard clauses](https://medium.com/better-programming/refactoring-guard-clauses-2ceeaa1a9da#:~:text=%E2%80%9CIn%20computer%20programming%2C%20a%20guard,avoid%20errors%20during%20execution.%E2%80%9D%20%E2%80%94). It would also stop the rest of the code running if not necessary (e.g. no need to check if the plane is already flying if you have already checked and the weather is stormy).

lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+ 
+  def land(airport, plane)
+    if airport.get_weather == :sunny && airport.is_full == false
+      if are_same_locations(plane.location, airport.location) == true
+        raise RuntimeError.new('Location must be wrong')
+      else 
+        return true
+      end
+    else 
+      if airport.get_weather == :stormy
```
Will-Helliwell, line 36: Same comment for this line and below as for line 12.

lib/plane.rb
```diff
@@ -0,0 +1,25 @@
+class Plane
+  attr_accessor :is_flying, :location
+  def initialize(location)
+    @is_flying = false
+    @location = location
```
Will-Helliwell, line 5: I think it's interesting that you have chosen to use the instance variable @location for each plane to keep track of where they all are, rather than storing landed planes as arrays wrapped up in the airport class (as we did for the boris bike challenge). I think both are equally valid response to the brief but your way is more inventive! 

The advantage of your method is that you can tell the location of a given plane at any time, but the advantage of the airport array method is that you can see all of the planes in a given airport at a given time.

spec/air_traffic_controller_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    expect(atc.land(london_city, easyjet_plane)).to be true
+  end
+  it " raises an error when plane is trying to take_off when the weather is stormy" do
+    atc = Air_traffic_controller.new
+    london_city = Airport.new(50, "LCY")
+    easyjet_plane = Plane.new("LCY")
+    allow(london_city).to receive(:get_weather).and_return(:stormy)
+    expect { atc.take_off(london_city, easyjet_plane) }.to raise_error(RuntimeError, "The plane cannot take_off because of the stormy weather")
+  end
+  it "raises an error when plane wants to take off from the wrong airport " do
```
Will-Helliwell, line 56: I don't think this is strictly to the brief - what happens if an air traffic controller wants to land a plane at the same airport it took off from (e.g. due to a bird strike)?

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  def is_full
+    if @actual_capacity == @max_capacity
+      return true
+    else 
+      return false
+    end
+  end
+
+  def get_weather
+    return [:sunny, :sunny, :sunny, :sunny, :stormy].sample
```
Will-Helliwell, line 18: I like this random weather generator, easy to understand and effective!

spec/airport_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    expect(airport.is_full).to be true
+  end
+
+  it "returns false when the airpot is not full" do
+    airport = Airport.new('20', 'location')
+    airport.actual_capacity = '3'
+    expect(airport.is_full).to be false
+  end
+  
+  it "returns stormy or sunny when calling get_weather" do
```
Will-Helliwell, line 35: You could also add a test to confirm that the weather is being generated as expected (i.e. stormy 20% of the time).

spec/air_traffic_controller_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'airport'
```
Will-Helliwell, line 1: As a general point the testing is great! Very thorough (I imagine you have 100% coverage?) and I can really easily follow all the the descriptions as they are very clear.

Will-Helliwell (changes requested):

PR #1959, opened 2020-11-29
Weekend Challenge: Week 1

chrismabdo:


xavierloos (commented):
Excellent work, your code was very clean and straight forward. I will suggest adding comments everywhere in your code, explaining what it does, explaining methods, etc.

PR #1961, opened 2020-11-29
Airport Challenge

kikidawson:


kikidawson:
The following user stories were used;

As an air traffic controller 

As an air traffic controller 

As an air traffic controller 
To ensure safety 

As the system designer

As an air traffic controller 
To ensure safety 

As an air traffic controller 
To ensure safety 

SarahJoniPower:
I really like that you've included the user stories in the comments! That's really helpful. 
Also personally I actually think your commit messages are very clear and easy to follow.

spec/airport_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+describe Airport do
+  let(:plane) { Plane.new }
+
+  it { is_expected.to respond_to(:land).with(1).argument }
+  it { is_expected.to respond_to(:take_off).with(1).argument }
+
+  it 'removes plane from airport after take off' do
+    allow(subject).to receive(:weather) { "sunny" }
+    subject.land(plane)
+    subject.land(Plane.new)
```
SarahJoniPower, line 13: I think you don't need this line? because you say above plane = Plane.new

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+  def take_off(plane)
+    fail "Weather too stormy for take-off" if poor_weather
+
+    planes.delete(plane)
+    "#{plane} has taken off"
+  end
+
+  def weather
+    possible_weather = ["sunny", "sunny", "sunny", "sunny", "sunny", "stormy"]
+    possible_weather[rand(6)]
```
SarahJoniPower, line 29: I really like how you've done this random weather, it's way neater than the way I did it

SarahJoniPower (commented):
Your code is really easy to follow, it's got great flow!

PR #1962, opened 2020-11-29
Ian Fleming Airport Challenge

iantfleming:


spec/weather_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    weather = Weather.new
+    expect(weather).to be_kind_of(Weather)
+  end
+
+
+  describe "#stormy?" do
+    it 'specifies if the weather is stormy or clear' do
+      weather = Weather.new
+
+      allow(weather).to receive(:sample).and_return(true)
```
natashamcintyre, line 14: I think this is the right way to go about it. From what I understand, we are telling the computer here to apply the .sample method in the Weather class and get a specific result from it. We would use it when we want to test things like "airport.land raises error when weather is stormy", so we write this allow statement then we say `expect { airport.land }.to raise_error "Too stormy" }` or something, in the airport_spec. I don't think we actually test the randomness of the method "sample" as we just have to trust Ruby's inbuilt methods. Would love to chat about this as it's a bit confusing.

natashamcintyre (commented):
Hi Ian, I found your code easy to read. The variables are all named sensibly, all the methods do one thing only an it's very easy to follow the flow. I hope you find my comments useful. 

With regards to testing randomness, I think that's about using stubs/mocks/doubles which I find a bit baffling so I don't really feel confident in explaining them, but would love to talk about it.

README.md
```diff
@@ -13,51 +13,40 @@ Airport Challenge
...
-4. Complete the following task:
+I attempted to solve this challenge using the Boris Bike challenge that we had been working on during the week as a guide. There were many elements of this challenges user stories that were analogous to the user stories found in the Boris Bike challenge. For example, the 'airport' can be likened to a 'docking station'. 'Land' can be likened to 'dock', and 'take off' can be likened to 'release bike'. As there were similarities, I was able to use some of my previous works as a guide for this challenge. I then tried to solve the problem of testing randomness but was unable to complete this task. There is more I could research regarding testing randomness as it is quite a well written about subject. I would hope to use this extra research to complete the challenge.
 
 Task
 -----
 
 We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
 
 ```
```
natashamcintyre, line 26: Also I like how you have included the user stories in your readme so that I can see exactly what the code is supposed to do

PR #1963, opened 2020-11-29
Natasha's submission of Airport Challenge

natashamcintyre:


lib/airport.rb
```diff
@@ -0,0 +1,50 @@
...
+  def land(plane)
+    fail "Airport full" if full?
+
+    fail "Plane already landed" if present?(plane)
+
+    fail "Landing aborted due to stormy weather" if stormy?
+
+    plane.land
+    hangar << plane
+  end
```
iantfleming, line 23: This whole method is very clear both in the way it is set out and the way in which you have named variables and methods.

spec/airport_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    describe '.land' do
+      it 'raises error when trying to land' do
+        message = "Landing aborted due to stormy weather"
+        expect { subject.land(plane) }.to raise_error message
+      end
+    end
+
+  end
+
+end
```
iantfleming, line 132: This could be my lack of expertise talking, but it feels like there are perhaps a few too many tests in your airport spec which can make it difficult to follow exactly what is going on.

lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+class Weather
+
+  def stormy?
+    rand(1..10) == 1
+  end
+
+end
```
iantfleming, line 7: Very clear intentions with your Weather class

spec/airport_spec.rb
natashamcintyre: Yes I did think that too, lots of fail scenarios. I'm not sure how else to do that. Perhaps a #fail? method??

PR #1964, opened 2020-11-29
Week 1 Airport Challenge 

ghost:
User stories completed: 1st, 2nd, 3rd, 5th and 6th. Didn't manage to complete the 4th.


lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+  
+  def takeoff(*)
+    raise StandardError.new "It is too stormy to takeoff plane" if stormy?
+  end
+end
+
+private
+
+def full?(array)
```
calavell, line 27: What do you think about not accepting an argument with this method?

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  attr_reader :capacity
+
+  def initialize
+    @capacity = capacity
+    @stored_planes = []
+  end
+
+  def land(plane)
+    raise StandardError.new "Airport is full" unless full?(@stored_planes) == false
```
calavell, line 13: interesting use of StandardError.new

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+    raise StandardError.new "It is too stormy to land plane" if stormy? 
+
+    @stored_planes << plane
+    @stored_planes[0]
+  end
+  
+  def takeoff(*)
```
calavell, line 20: Do you think there is a name for this method that could better reflect it's functionality?

lib/plane.rb
```diff
@@ -0,0 +1,19 @@
+class Plane
+
+  attr_reader :location
+
+  LOCATION = :air
+
+  def initialize
+    @location = :default_location
```
calavell, line 8: Could you clarify what is the value of default_location?

lib/plane.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  attr_reader :location
+
+  LOCATION = :air
+
+  def initialize
+    @location = :default_location 
+  end
+
+  def plane
```
calavell, line 11: could this method be removed?

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+# As an air traffic controller 
+# so planes can take off safely from my airport
+# I would like to instruct a plane to take off and confirm it is no longer in the airport
+
+  it 'instructs a plane to take off' do
+    allow(subject).to receive(:stormy?).and_return false
+    expect(subject).to respond_to(:takeoff).with(1).argument
+  end
```
calavell, line 22: Could we implement another test to fully satisfy the above user story, specifically;

_I would like to instruct a plane to take off and confirm it is no longer in the airport_

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+describe Airport do
+
+# As an air traffic controller 
+# So I can get passengers to a destination 
+# I want (an airport) to instruct a plane to land at an airport
+
+  it 'instructs a plane to land' do
+    allow(subject).to receive(:stormy?).and_return false
+    expect(subject).to respond_to(:land).with(1).argument
+  end
```
calavell, line 13: Do we think we could implement another test here to test a more specific functionality of :land?

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    expect(subject).to respond_to(:takeoff).with(1).argument
+  end 
+
+# As an air traffic controller 
+# to ensure safety
+# i want to prevent landing when the airport is full
+
+  it 'gives an error when a plane tries to land when the airport is full' do
+    allow(subject).to receive(:stormy?).and_return false
+    expect { 21.times { subject.land(Plane.new) } }.to raise_error "Airport is full"
```
calavell, line 30: maybe we should try to avoid magic numbers. Could we define 21 as something with a bit more context

spec/airport_spec.rb
calavell: I am now realised this was implemented in your plane_spec file

spec/plane_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'airport'
+require 'plane'
+
+describe Plane do 
+
+  let(:airport) { double(:airport) }
```
calavell, line 6: do we need to use a double in this context?

lib/plane.rb
```diff
@@ -0,0 +1,19 @@
+class Plane
+
+  attr_reader :location
+
+  LOCATION = :air
```
calavell, line 5: like use of location and symbols here

lib/plane.rb
WillDixon93: thanks

spec/airport_spec.rb
WillDixon93: thanks

spec/airport_spec.rb
WillDixon93: thanks

spec/airport_spec.rb
WillDixon93: thanks

lib/plane.rb
WillDixon93: thanks

lib/plane.rb
WillDixon93: thanks

spec/plane_spec.rb
WillDixon93: thanks

WillDixon93:
Removed double of airport as wasn't useful, still passed.

PR #1965, opened 2020-11-29
Airport challenge 

IrinaG29:


lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @planes.include?(plane)
+  end
+
+  private
+
+  def full?
+    @planes.size >= @capacity
+  end
+
+  def stormy?
```
ZeenLamDev, line 31: i would personally create a weather class and put this method within the weather class so if you was to create other weather scenarios it'll be all within one place.

ZeenLamDev (commented):
Other than the stormy? method in the airport class i would keep it all as it is.

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
+require_relative "../lib/airport"
+require_relative "../lib/plane"
+
+describe Airport do
+  before do
+    allow(subject).to receive(:stormy?) { false }
```
ZeenLamDev, line 6: i like how you set stormy? as false to allow easier testing

PR #1968, opened 2020-11-30
airport_challenge submission

SarahJoniPower:


lib/airport.rb
```diff
@@ -0,0 +1,39 @@
+require 'plane.rb'
+
+class Airport
+attr_reader :hanger
+attr_reader :weather
+DEFAULT_CAPACITY = 20
+  def initialize
+    @hanger = []
```
kikidawson, line 8: Nice and clearly named variable - hanger.

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+   airport.land(plane)
+   expect(airport.hanger).to include(plane)
+  end
+
+  it 'takes_off plane' do
+   allow(plane).to receive(:take_off)
+   airport.take_off(plane)
+   expect(airport.hanger).not_to include(plane)
+  end
+
```
kikidawson, line 25: Would it be better to allow the maximum number of planes to land first then make another one land, which raises an error? I'm not sure if it really makes a difference but then the error would be being raised on the one plane that can't land and not all of the plane that have already landed.

```
  it 'will not land a plane if hanger is full' do
    allow(airport).to receive(:stormy?).and_return(false)
    Airport::DEFAULT_CAPACITY.times { airport.land(plane) }
    expect { airport.land(plane) }.to raise_error('Hanger is full!')
  end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  let(:airport) { Airport.new }
+  let(:plane) { Plane.new }
+
+  it { is_expected.to respond_to(:land).with(1).argument }
+  it { is_expected.to respond_to(:take_off).with(1).argument }
+
+  it 'expects the hanger array in airport to change in response to calling airport.land(plane)' do
+   expect{ airport.land(plane) }.to change{ airport.hanger }
+  end
+
```
kikidawson, line 13: This all seems very clear and easy to read.

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'airport'
+
+describe Airport do
+  let(:airport) { Airport.new }
```
kikidawson, line 4: Could this line be left out if 'subject' was used instead of 'airport' in the code below. I think using subject uses an instance of the class the unit test is described in. So using subject would use an instance of Airport.

kikidawson (commented):
Good flow and clear to understand. Hope the notes make sense. They're just suggestions and not anything that would change the meaning!

kikidawson:
I also don't understand when you're supposed to use a nested describe.

PR #1969, opened 2020-11-30
Airport Weekend Challenge

AJSMonty:


lib/airport.rb
```diff
@@ -0,0 +1,30 @@
+require './lib/weather.rb'
+require './lib/plane.rb'
+
+class Airport
+  attr_reader :hanger, :capacity, :weather
+  attr_writer :weather
+  DEFAULT_CAPACITY = 20
+  def initialize(capacity = DEFAULT_CAPACITY)
```
AJSMonty, line 8: On initialization of a new Airport instance, a hanger array is created, capacity is set to input if present(otherwise default capacity(set to 20)) and a weather instance is created with a value of true or false(depending on the weather.

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
+require './lib/weather.rb'
+require './lib/plane.rb'
+
+class Airport
+  attr_reader :hanger, :capacity, :weather
+  attr_writer :weather
```
AJSMonty, line 6: Allows for rspec tests to overwrite @weather variable

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+  DEFAULT_CAPACITY = 20
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hanger = []
+    @capacity = capacity
+    @weather = Weather.new.stormy
+  end
+
+  def land(plane)
```
AJSMonty, line 14: The method 'land' checks if airport is full or if the weather is stormy(true) and either lands plane(pushes to @hanger array, or returns an error of 'Too stormy'

lib/plane.rb
```diff
@@ -0,0 +1,6 @@
+class Plane
```
AJSMonty, line 1: Creates Plane class

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
+  attr_accessor :weather, :stormy
+  
+  def initialize
```
AJSMonty, line 4: On initialization, Weather returns a random number between 1 and 10. The stormy method then returns true if the number is equal to or greater than 8 and returns that to the @stormy instance variable

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+    return @hanger.push(plane) unless (full? || @weather == true)
+    return fail 'Too stormy' if @weather == true
+  end
+
+  def take_off(plane)
+    return @hanger.delete(plane) unless @weather == true
+    return fail 'Too stormy' if @weather == true
```
lukewickens1989, line 21: Possibly have set a guard case initially without a return first then you aren't returning too many things in one method. Then attribute the @hangar.delete into a method to reduce how much the take_off function is actually doing.

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  private 
+
+  def full?
+    raise 'Airport Full!' if @hanger.count >= @capacity
```
lukewickens1989, line 27: I like this method, nice and simple and easy to understand.

lib/plane.rb
```diff
@@ -0,0 +1,6 @@
...
+  def initialize
```
lukewickens1989, line 2: I can see that this element is unfinished as explained in the comments.

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
...
+    @weather = rand(1..10)
+    @stormy = stormy?
+  end
+
+  def stormy?
+    false unless @weather >= 8
```
lukewickens1989, line 10: Again nice, simple and easy to understand code.

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'airport'
+require 'weather'
+
+describe Airport do
+
+  let(:plane) { Plane.new }
+  it { is_expected.to respond_to(:land).with(1).argument }
```
lukewickens1989, line 7: Good refactoring to single line

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  it { is_expected.to respond_to(:take_off) }
+
+  it 'removes plane from airport on take off' do
+    allow(subject).to receive(:weather) { false }
+    subject.land(plane)
+    subject.take_off(plane)
+    expect(subject.hanger).not_to include plane
+  end
+
+  it 'Airport takes 1 argument' do
```
lukewickens1989, line 24: I think this test could have been refactored to a single line as the test is pretty self explanatory.

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    20.times { subject.land("plane") }
+    expect { subject.land("plane") }.to raise_error 'Airport Full!'
+  end
+
+  it 'returns stormy if weather >= 8' do
+    allow(subject).to receive(:weather) { true }
+    expect(subject.weather).to eq true
+  end
+
+  it 'raises error if weather is stormy on take_off' do
```
lukewickens1989, line 43: I like this test. I wasn't aware you could add conditions after the raise error condition. I have learnt something from this.

spec/weather_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  it 'can access weather from outside the class' do
+    expect(subject).to respond_to :weather
+  end
+
+  it 'generates weather when initialized' do
+    expect(subject.weather).to be_instance_of Integer
+  end
+
+end
```
lukewickens1989, line 16: All tests look well written and cover all instances

lib/airport.rb
lukewickens1989: ` return @hanger.delete(plane) unless @weather == true ` 
` return fail 'Too stormy' if @weather == true`  
could be
`fail 'Too stormy' if @weather == true` 
`delete(plane)`

lukewickens1989 (commented):
Nice looking code. Very simple and easy to understand and very practical and logical flow.
Nice tests, seem to test for outcomes well and nice use of mocks and stubs.

PR #1972, opened 2020-12-14
Atiyyah's Airport Challenge

notatiyyah:


EdwardAndress:
This is a good solution to the airport challenge, Atiyah.  There's lots of really great stuff going on – the readability in particular is great.  In terms of what you might change / improve, I'll focus on two things: test isolation and testing randomness.

Test isolation
– Ideally tests will only fail when there is a problem with the corresponding class.
– At the moment, you'll probably find that this is not the case
– To isolate your tests, you can use doubles: https://relishapp.com/rspec/rspec-mocks/docs/basics/test-doubles

Testing randomness
– Looks like you created the method to override weather so that you could test that the different weather scenarios are handled correctly. This works, in that sense, but it doesn't ensure that the weather is random.
– You did find a cool way of testing the weather, using some stats but... could you make the test pass in another way, by doing something with the weather that is not random?  Also, you're kind of testing Ruby's built in sample method, which we don't need to do.
– Another approach is to use stubbing.  This is where a method is 'intercepted' and instead of executing the method, we provide a canned response.  The key to this is in figuring out what method to intercept.  Using this approach you can ensure a specific method is being used (sample, for example) and remove randomness by providing the canned response.

https://relishapp.com/rspec/rspec-mocks/v/3-10/docs/basics/allowing-messages
https://relishapp.com/rspec/rspec-mocks/v/3-10/docs/basics/expecting-messages

These two things are advanced topics for the first couple of weeks :)

PR #1977, opened 2020-12-14
Tuyet Le's pull request

toowee-le:
Completed the first four user stories. I created two classes Airport and Plane with test cases to test the first four user stories.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,20 @@
+class Airport
+  DEFAULT_CAPACITY = 20
+
+  attr_accessor :hangar, :capacity
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail "Airport hangar is full" if @hangar.count >= @capacity
+    @hangar << plane
+  end
+    
+  def take_off(plane)
+    @hangar.delete(plane)
+    puts "#{plane} has taken off and is no longer in the airport."
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,20 @@
+class Airport
+  DEFAULT_CAPACITY = 20
+
+  attr_accessor :hangar, :capacity
```
lena0910, line 4: Great use of attributes here!

spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  describe '#land' do
+    it 'instruct a plane to land at airport hangar' do
+      airport = Airport.new
+      plane = Plane.new
+      airport.land(plane)
+      expect(airport.hangar).to include plane
```
lena0910, line 10: include is a really useful method to have here!

spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      expect(airport.hangar).not_to include [plane]
+    end
+  end
+
+  describe '#capacity' do
+    it 'override airport hangar capacity' do
+      airport = Airport.new(30)
+      plane = Plane.new
+      airport.capacity.times { airport.land(plane) }
+      expect { airport.land(plane) }.to raise_error 'Airport hangar is full'
```
lena0910, line 36: all the test descriptions are very easy to read and understand what they are doing

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
+class Airport
+  DEFAULT_CAPACITY = 20
```
lena0910, line 2: Using a constant variable here is a great way to ensure the code is easy to change!

PR #1978, opened 2020-12-14
4/6 user stories complete

rouryjhinds:


lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
+    @plane
```
EdwardAndress, line 2: An example of what you could do here (though it's not specifically asked for in the requirements.

```
def initialize(flight_number)
 @flight_number = flight_number
 @state = 'grounded'
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      plane = Plane.new
+      airport.land(plane)
+      expect(airport.gate).to include(plane)
+    end
+    it 'prevent landing if gate is full' do
+        #assume capacity to be 1
+        airport = Airport.new
+        plane = Plane.new
+        airport.land(plane)
+        expect { airport.land(plane) }.to raise_error('cannot land, airport is full')
```
EdwardAndress, line 17: `expect(airport.land(plane)).to raise_error`
- airport.land(plane) would be executed first
- `expect(airport.gate).to include plane`
- ariport.gate is executed first... and and the return value is the argument for expect
- `expect { airport.land(plane) }.to raise_error('cannot land, airport is full')`
- here we're passing a block to expect
- the code in the block is not executed immediately
```
begin
 <some code that might throw an error, like your block of code for example>
rescue => error
 <do something when the error is thrown>
end
```

PR #1985, opened 2021-01-10
Airport challenge

MegGosnell:
Completed Airport challenge.


acavalla (commented):
- Use subject rather than airport = Airport.new
- Keep code 'dry' - initialise a plane outside the individual examples
- not sure you need land(plane) if inside plane class

PR #1988, opened 2021-01-10
Charlie's Airport Challenge

charliefischer:
missed the describes of the methods on the initial pull!

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,3 +1,4 @@
+$all_grounded_planes = []
 class Airport
   attr_reader :planes_in_airport, :capacity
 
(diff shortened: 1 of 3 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,67 @@
...
+    @planes_in_airport == []
+  end
+
+  def sunny?
+    @weather == 'sunny'
+  end
+
+  def stormy?
+    @weather == "stormy"
+  end
```
twigz826, line 54: Lines 48-54 - I don't think these methods are necessary. The @weather instance variable has already been set by generate_weather method and this can be passed into takeoff/land methods to determine whether the weather is stormy or not.

lib/airport.rb
```diff
@@ -0,0 +1,67 @@
...
+    @planes_in_airport.include?(plane)
+  end
+
+  def plane_elsewhere(plane)
+    $all_grounded_planes.include?(plane)
+  end
+
+end
+
+class Plane
```
twigz826, line 66: Empty class? Could include a status method to determine whether plane is in the air or on the ground.

nouns_verbs.txt
```diff
@@ -0,0 +1,32 @@
...
+Prevent landing -> weather stormy
+
+To create weather
+Random number -> rarely it will be stormy -> needs to be able to be overridden for testing
+
+Defence against edge cases
+Planes can only take off from airports they are in	
+Planes that are already flying cannot take off/be in an airport
+Planes that have landed cannot land again and must be in an airport
+
```
twigz826, line 30: Really useful document here summarising all the user stories, nice!

spec/airport_spec.rb
```diff
@@ -0,0 +1,159 @@
+require 'airport'
+
+describe Airport do
+  describe ".capacity" do
```
twigz826, line 4: I like using two describe blocks, one for the class and one for the method. There's other syntax that you could use in RSpec, for example "context" if multiple tests have the same context.

I think to cut down on repetition, the biggest thing here would be to use something like this at the top so you don't have to define a variable for airport in every test:
let(:airport) { Airport.new }

twigz826 (commented):
Review of airport challenge on 11/01/2021.

PR #1989, opened 2021-01-10
Airport Challenge

KarstenFinlay:


lib/airport.rb
```diff
@@ -0,0 +1,62 @@
+require_relative "plane"
+require_relative "weather"
+
+class Airport
+
+  STANDARD_CAPACITY = 10
+
+  attr_reader :planes, :capacity
+
+  def initialize(weather, capacity = STANDARD_CAPACITY)
```
GlykeriaStr, line 10: Why did you put the weather as an argument and didn't let it be always random?

lib/airport.rb
```diff
@@ -0,0 +1,62 @@
...
+  def stormy?
+    @weather.forecast == "stormy"
+  end
+
+  def depart_plane
+    fail "There are no planes to take off" if empty?
+
+    plane = @planes.pop
+    plane.docked = false
+    return plane
```
GlykeriaStr, line 60: Why did you return plane?

lib/airport.rb
```diff
@@ -0,0 +1,62 @@
...
+      fail "Airport is full" if full?
+
+      fail "Already landed" if plane.docked
+
+      plane.docked = true
+      @planes.push(plane)
+    end
+  end
+
+  def take_off(plane_count = 1)
```
GlykeriaStr, line 29: How come take_off doesn't take a plane as an argument to tell it which plane you want to take off?

GlykeriaStr (commented):
I think it looked great! :)

PR #1990, opened 2021-01-10
Airport Challenge

HolsDuckett:
Airport Challenge

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,7 +2,7 @@ require_relative 'plane'
 
 class Airport
 
-DEF_CAPACITY = 5
+  DEF_CAPACITY = 2
 
   def initialize(capacity = DEF_CAPACITY)
     @capacity = capacity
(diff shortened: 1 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def take_off(plane)
+    fail "Weather looks a bit mental best not to take off." if stormy?
+    @runway.delete(plane)
+    "#{plane} has buggered off."
+  end
+
+  attr_reader :runway
+  attr_reader :capacity
```
ddottyler, line 25: Do these need to be separate? Could it not be:
attr_reader :runway, :capacity

spec/airport_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+subject(:airport) {described_class.new}
```
ddottyler, line 6: I'm not sure what this line means so would be interested in discussing

ddottyler (commented):
I'd say this is really clear and comprehensive with everything easy to read and understand. 
There are a couple of bits of code/syntax that I'm unfamiliar with and would be keen to discuss and learn more about.

PR #1992, opened 2021-01-10
Airport challenge 

bear99a9:
All user stories finished


jshields123 (commented):
Good work just needs a formatting through rubocop.

PR #1995, opened 2021-01-10
Airport Weekend Challenge

CodeRed30:


spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    it 'as instructed' do
+      expect(airport.land(:plane)).to include(:plane)
+    end
+  end
+
+  context 'plane will take off' do
+    it { is_expected.to respond_to(:take_off).with(1).argument }
+
+    it 'will confirm that the plane is no longer in the airport' do
+      airport.land(:plane)
```
alicelieutier, line 20: `plane` is the variable, `:plane` is merely the name of the variable. You should use `:plane` to define it, and then `plane` to use it in your tests.

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      expect { airport.land(:plane) }.to raise_error "Airport is full"
+    end
+  end
+
+  context 'is stormy' do
+    it { is_expected.to respond_to :report_storm }
+
+    it 'will report the weather' do
+      expect(airport.report_storm).to eq true
+    end
```
alicelieutier, line 43: I'm not sure what this tests. When testing, you want to think of the behaviour of a method. If the only behaviour of a method is to set the state (as it is here), you cannot really test it without using other methods that then read the state and do something based on it.
That is the core of the meaning of testing behaviour (vs state).

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+
+      it 'will prevent takeoff' do
+        expect { airport.take_off(:plane) }.to raise_error "Plane cannot take off due to storm"
+      end
+
+      it 'will prevent landing' do
+        expect { airport.land(:plane) }.to raise_error "Plane cannot land due to storm"
+      end
+    end
```
alicelieutier, line 57: This is good behaviour test.

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+  # Tests that can no longer be used as methods are now private
+
+  # it { is_expected.to respond_to :stormy? }
+
+  # it 'will output the weather as stormy' do
+  #   subject.report_storm
+  #   expect(subject).to be_stormy
+  # end
```
alicelieutier, line 66: Remove unused code instead of just commenting it. Git keeps the history of changes anyway, so no need to keep it around.

alicelieutier (commented):
It would be nice to have your feature tests recorded somewhere, as they are the only way to make sure the whole system works together as soon as tests are isolated.
Soon you'll write feature tests in their own rspec files too.

alicelieutier:
Overall, your code is clean and your tests are good. Next step is more classes :)

CodeRed30:
> It would be nice to have your feature tests recorded somewhere, as they are the only way to make sure the whole system works together as soon as tests are isolated.

@alicelieutier could you direct me to any best practice with this please? I can't really envision how I would be doing this at the moment.

PR #1998, opened 2021-01-10
✈️ Airport Challenge - Reeshul

reeshul:


lib/airport.rb
```diff
@@ -0,0 +1,37 @@
+require 'weather_forecast'
+
+class Airport
+
+  DEFAULT_CAPACITY = 20
```
MykeNuLeng, line 5: Love this!

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+  
+  def initialize(weather_forecast, capacity = DEFAULT_CAPACITY)
+    @weather_forecast = weather_forecast
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
```
MykeNuLeng, line 13: Big fan of the refactoring that's taken place here!

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    raise "Cannot land plane - airport full" if full?
```
MykeNuLeng, line 14: You should leave a blank line after a guard clause in Ruby! Love that you're using them though.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    raise "Cannot land plane - stormy weather" if stormy?
+    @planes << plane
+  end
+
+  def take_off(plane)
```
MykeNuLeng, line 19: The logic in this method could use some work! You're not actually removing a plane from the @planes array at any point. So it's not really taking off?

lib/weather_forecast.rb
```diff
@@ -0,0 +1,5 @@
+class WeatherForecast
+  def stormy?
```
MykeNuLeng, line 2: We love Kernel

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'airport'
+
+describe Airport do
+
+  subject(:airport) { Airport.new(weather_forecast, 20) }
```
MykeNuLeng, line 5: Such good doubles! I remember struggling coming to grips with these, so it's great you're using them so fluently

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  let(:plane) { double :plane }
+  let(:weather_forecast) { double :weather_forecast}
+
+  describe '#land' do
+    context 'when not stormy' do
+      before do
```
MykeNuLeng, line 11: These before do's and context blocks are also fantastic for readability, and gives a great structure to your tests.

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+  end
+
+  describe '#take_off' do
+    context 'when not stormy' do
+      before do 
+        allow(weather_forecast).to receive(:stormy?).and_return false
+      end
+      it 'instruct the plane to take off' do
+        expect(airport).to respond_to(:take_off).with(1).argument
```
MykeNuLeng, line 42: Your test here needs to be more specific! Asking it to respond with an argument will almost always pass, however if you asked it to return to you an instance of the plane class, then it'd be more thorough!

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      it 'confirm taken off plane is no longer in the airport' do
+        another_airport = described_class.new(weather_forecast, 20)
+        another_airport.land(plane)
+        expect { airport.take_off(plane) }.to raise_error "Plane taken off - no longer in the airport"
+      end
+    end
+
+    context 'when stormy' do
+      before do
+        allow(weather_forecast).to receive(:stormy?).and_return true
```
MykeNuLeng, line 54: This is also how you can test randomness in other tests! If you don't feel comfortable using srand still, then using doubles works perfectly well as well.

spec/weather_forecast_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'weather_forecast'
+
```
MykeNuLeng, line 2: Amazing!

MykeNuLeng (commented):
Overall a really fantastic attempt at the first weekend challenge!

PR #2000, opened 2021-01-10
Anna's updated solution

acavalla:
Completed except the bonus and the travisbuild.


MegGosnell (commented):
- Organise the different methods a bit more efficiently in your airport file.
- To add methods into your plane class in your plane file.

PR #2001, opened 2021-01-11
Pull request for airport project.

twigz826:


spec/airport_spec.rb
```diff
@@ -0,0 +1,99 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { Plane.new }
+
+  describe 'Tests that need to call #land' do
+    # before do
```
charliefischer, line 7: minor note on this - Josue mentioned we shouldn't have comments left on our code as it should be self explanatory. All the contexts and describes for your tests clearly label everything so feel these could be deleted?

spec/airport_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    # As an air traffic controller
+    # To ensure safety
+    # I want to prevent landing when the airport is full
+    context "raises an error if the airport is full" do
+      it "returns error when airport is at max capacity and another is added" do
+        airport = Airport.new
+        allow(airport.weather).to receive(:weather_report) { "sunny" }
+        20.times { airport.land(Plane.new) }
```
charliefischer, line 56: to avoid calling in classes from outside this class could you make the capacity of the airport 1 and then try to add a second plane that you could save as plane2?

PR #2002, opened 2021-01-11
Salar airport_challenge

SalarGhotaslo:


Reeshul (changes requested):
Inconsistent indentation in the airport spec.

Inconsistent describe / context blocks (e.g. you have one for '#takeoff' but nothing for '#land' )

PR #2003, opened 2021-01-11
Jack's airport challenge attempt - ran out of time..due to stormy weather.

jshields123:


spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      plane = Plane.new
+      subject.land(plane)
+      expect(subject.land(plane)).to include plane
+    end
+  end
+# User story: 2
+  it { is_expected.to respond_to(:take_off) }
+  describe '#take_off' do
+    it 'allows plane to take off and confirms it has taken off' do
+      Plane.new
```
bear99a9, line 19: don't need this Plane.new here

bear99a9 (commented):
Nice work liked your respect divide between methods

PR #2005, opened 2021-01-11
Airport Challenge 

louiswicks:


lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
+  
+  def stormy?
+    if rand_num < 90
+      "Sunny"
+    else
+      "stormy"
```
bateso88, line 7: 'stormy' has no capital letter to begin with but 'Sunny' does - could be amended for consistency?

spec/airport_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'airport'
+
+require 'plane'
```
bateso88, line 3: I don't think you need this line as 'plane' is already required in your 'airport.rb' file

spec/airport_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  it "responds to planes" do
+    expect(subject).to respond_to :planes
+  end
+
+  describe 'land method' do
+    it 'raises an error when airport limit is reached' do
+      airport = Airport.new(20)
+      plane = Plane.new
+      20.times {airport.land(plane) }
```
bateso88, line 36: This is landing the same plane 20 times? You could do "20.times { airport.land(Plane.new) }" to make it 20 different planes

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
```
bateso88, line 1: Great idea to split weather out into its own class

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def land(plane)
+    fail "Airport is full" if full?
+    @planes << plane
+  end
+
+  def take_off
+    #Plane.new
+    @planes
```
bateso88, line 22: I think this will return an array containing instances of the plane class. This might cause issues when there are multiple planes in the airport. If you were to write "@planes.pop" that would return the last plane in the array @planes only and would remove that plane from the array

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
...
+  
+  def stormy?
+    if rand_num < 90
+      "Sunny"
+    else
+      "stormy"
+    end
+  end
+
+  def rand_num
```
bateso88, line 11: Good idea to create separate method for rand_num. Good SRP.

spec/airport_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  it "responds to the take_off method" do
+    expect(subject).to respond_to :take_off
+  end
+
+  it "to take_off a plane and confirm it left airport" do
+    plane = Plane.new
+    airport = Airport.new
+    airport.land(plane)
+    expect(airport.take_off).to eq [plane] #be_instance_of(Plane)
```
bateso88, line 25: I think this test is checking that the plane is still in the airport as [plane] in this case is equivalent to your array @planes

spec/weather_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'weather'
```
bateso88, line 1: Again I think its a smart idea to split out weather into its own class

bateso88 (commented):
Great work! Hard to fault but I think one issue is with your takeoff function, in that the plane remains in the airport. Apart from that all looks good!

PR #2006, opened 2021-01-11
Airport challenge ready for review

Mornevanzyl:
I've completed the airport challenge to a point where I think the stated user requirements are met. Clearly much more development potential on this project, but this is probably as good a point as any to draw an initial line in the sand.


PetarSimonovic (commented):
Really enjoyed reading this. It's very clear, simple and elegant.

It's made me realise how much I can simplify my code.

PR #2007, opened 2021-01-11
Airport -Challenge - Glykeria

GlykeriaStr:


lib/plane.rb
```diff
@@ -0,0 +1,15 @@
...
+    @flying = true
+  end
+
+  def flying?
+    @flying
+  end
+
+  def landed?
+    @flying = false
+  end
```
KarstenFinlay, line 13: You could just have one method called `flying?` Which can set the flying status to false or true.

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  context 'checking if the airport is full' do
+    it 'should return true or false if full' do
+      expect(subject.send(:full?)).to eq false
+      allow(subject).to receive(:stormy?).and_return false
+      subject.capacity.times { subject.land(Plane.new) }
+      expect(subject.send(:full?)).to eq true
+    end
+  end
+
+end
```
KarstenFinlay, line 90: Perhaps easier to read if tests were grouped within methods. 
eg.
```
describe "take off" do
it "should check that plane can take off" do
end

it "raises an error if stormy and trying to take off" do
end
```

PR #2008, opened 2021-01-11
My airport - outdated

alicelieutier:
Demoing opening a pull request.


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 50
```
alicelieutier, line 4: Why is default 50?

spec/features/feature_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'airport'
+require 'plane'
+
+describe 'Land a plane in sunny weather' do
+
+  airport = Airport.new
+  planes = (1..20).map{ |i| Plane.new }
+  planes.each { |plane| plane.land(airport) }
+  planes.each do |plane|
+    if plane.isAirborne?
```
alicelieutier, line 10: `isAirborne?` doesn't seem very rubylike as a name. you may want to use `airbone?`

alicelieutier (commented):
Looks like this could be improved with your current knowledge of ruby.

PR #2009, opened 2021-01-11
Airport challenge - Completed User stories

IliasMariosG:
Would like to chat about:
- the use of doubles
- from the second user story "[...] and confirm that it is no longer in the airport"
- from the gourth user story "[...]default airport capacity that can be overridden as appropriate"


IliasMariosG:
Created a new one

PR #2011, opened 2021-01-29
airport changes

JFindley95:
Airport Challenge


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
+class Airport
+  attr_accessor :capacity, :planes, :weather
+  def initialize
+    @capacity = 0
+    @planes = []
+    @weather = ["sunny", "sunny", "cloudy", "stormy", "sunny"].sample
+  end
+
+  def land(flight_number)
+    if @planes.length >= @capacity
```
EdwardAndress, line 10: what if the airport is full and the weather is stormy?  would you want to give both warnings?

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def take_off(flight_number)
+    if @weather == "stormy"
+      return "you can not take off due to bad weather"
+    end
+    @planes.delete(flight_number)
+  end
+
+  def max_capacity(num)
+    @capacity =+ num
```
EdwardAndress, line 27: this would be more idiomatic
```
def max_capacity=(num)
  @capacity = num
end
```

EdwardAndress: note the use of `=` in the method name and the very small change to the method body

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      airport.weather
+      airport.max_capacity(6)
+      airport.planes << "345"
+      airport.planes << "567"
+      airport.planes << "777"
+      airport.take_off("777")
+       expect(airport.take_off("777")).to eq("you can not take off due to bad weather")
+       expect(airport.planes).to eq ["345", "567", "777"]
+      end
+  end
```
EdwardAndress, line 47: Good tests for how weather impacts on landing / takeoff.  You could use `context` to structure your tests like this

```
    end
  end

    end
  end
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  describe 'max capacity' do
+    it 'when the airport is full, by reaching max capacity, the airport will tell the plane not to land' do
+       srand(2)
```
EdwardAndress, line 50: nice test for capacity!

EdwardAndress:
It's a solid attempt and it's clear that you're picking up the basics of TDD, which is great!  I left some comments in-line and I hope they're helpful.  Some more general comments...

- There are a few extraneous lines in some of your tests.  Make sure that everything you write is actually needed.
- The next step here would be to try and remove some duplication.  Specifically around creating airports (or whatever object is under test / being tested).  How can you DRY up your tests and remove the repeated `airport = Airport.new`?

PR #2012, opened 2021-01-29
Graeme Hedges

GraemeHedges:
My attempted solution for the Airport Challenge.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,30 @@
+require "./lib/plane.rb"
+
+class Airport
+
+    RUNWAY_CAPACITY = 5
+
+    attr_reader :planes, :capacity
+
+    def initialize(capacity=RUNWAY_CAPACITY)
+        @planes = []
+        @capacity = capacity
+    end
+
+    def land(plane)
+        raise "The airport is full" if full?
+        @planes << plane
+    end
+
+    def take_off(plane)
+        @planes.delete(plane)
+    end
+
+    private
+
+    def full?
+        planes.count >= capacity
+    end
+
+end
+
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,4 +1,5 @@
 require "./lib/plane.rb"
+require "./lib/weather.rb"
 
 class Airport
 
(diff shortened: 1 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,39 @@
+require "./lib/plane.rb"
+require "./lib/weather.rb"
+
+class Airport
+
+    RUNWAY_CAPACITY = 5
```
EdwardAndress, line 6: nice use of a constant here

EdwardAndress: but do all the landed planes sit on the runway?

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    private
+
+    def full?
+        planes.count >= capacity
+    end
+
+    def stormy?
+        @weather.current_conditions
+        @weather.conditions == "stormy" ? true : false
```
EdwardAndress, line 35:
```
@weather.conditions == "stormy"
```
This will return a boolean so you don't need the ternary

lib/weather.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class Weather
+
+    attr_reader :conditions
+
+    def initialize
+        @conditions = current_conditions
+    end
+
+    def current_conditions
```
EdwardAndress, line 12: i like the way you separated setting conditions from reading them

spec/airport_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+            subject.land(plane)
+            subject.take_off(plane)
+            expect(subject.planes).not_to include(plane)
+        end
+
+        it "raises an error if the weather is too stormy" do
+            plane = Plane.new
+            subject.land(plane)
+            allow(subject).to receive(:stormy?) {true}
+            expect{raise subject.take_off(plane)}. to raise_error("The weather is too stormy")
```
EdwardAndress, line 33: Good tests for how weather impacts on landing / takeoff.  You could use `context` to structure your tests like this

```
    end
  end

    end
  end
end
```

EdwardAndress:
Good attempt graeme – you've clearly picked up the basics of testing, which is really great! I left some minor comments in-line.  I hope they're helpful!

PR #2013, opened 2021-01-29
Alistair Tait pull request

AlistairTait12:
Please see solution to airport challenge. Any feedback greatly appreciated. Thank you


README.md
```diff
@@ -1,89 +1,86 @@
...
-
-* **Submit a pull request early.**
-
-* Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+# Airport Challenge
+
+## Credits to my solution
+Used advice of peer group 1! (Anna (yy140), Graeme and Anastasia)
+
+## User Stories
```
EdwardAndress, line 6: really nice planning!!

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    raise "Landing denied" if full?
+
+    @planes.push(plane)
+    "#{plane.flight_id} successfully landed at #{@name}"
+  end
+
+  def take_off(plane, weather)
+    raise "Take off denied due to weather" if stormy?(weather)
+    
+    i = @planes.index(plane)
```
EdwardAndress, line 29: try to avoid using single letter variables.  `plane_index` could be an alternative here

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def confirm_status(flight_id)
+    if @planes.map { |plane| plane.flight_id }.include?(flight_id)
+      "#{flight_id} is at #{@name}"
+    else
+      "#{flight_id} is in flight"
+    end
+  end
+
+  private
```
EdwardAndress, line 42: nice that you're using private methods!

lib/weather.rb
```diff
@@ -0,0 +1,12 @@
+class Weather
+  attr_reader :report
```
EdwardAndress, line 2: nice that you're separating weather reporting from randomising the weather

spec/airport_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      expect(airport.take_off(plane, weather)).to eq 'MAK121 successfully taken off from Ruby airport'
+    end
+
+    it 'the list of planes at airport should now reflect that planes have left the airport' do
+      weather = double()
+      weather.stub(:report) { "sunny" }
+      airport = Airport.new('Python airport')
+      plane1 = Plane.new('CPP091')
+      plane2 = Plane.new('RUB091')
+      airport.land(plane1, weather)
```
EdwardAndress, line 54: Passing the weather into the `land` method does make your code more flexible and easier to test.  It's kind of `functional` style and you'll see something a bit like this in other places – watch out for them!

Here it feels a bit odd in some ways though because it allows whoever calls the `.land` method to kind of subvert the system and just say that the weather is good, even if it's not, which is probably bad :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      airport.land(plane2, weather)
+      airport.take_off(plane1, weather)
+      expect(airport.planes).to eq ['RUB091']
+    end
+  end
+
+  describe '#confirm_status' do
+    it 'should, upon request, inform controller when landed at airport' do
+      weather = double()
+      weather.stub(:report) { "sunny" }
```
EdwardAndress, line 64: I'm not familiar with this approach to stubbing, so I have learned something from reading this! :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      airport = Airport.new('Python airport')
```
EdwardAndress, line 65: are you dreaming of using another programming language? ;)

spec/airport_spec.rb
EdwardAndress: there is, however, a more concise way of creating a double with some behaviour associated

```
weather = double :weather, report: "stormy"
```

and if you want to use the same double for more than one test you can declare it outside the it block using like this

```
let(:weather) { double :weather, report: "sunny" }
```

when you do that, the double is scoped to the block in which it is declared and all the nested scopes

spec/airport_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+
+  describe '#stormy?' do
+    it 'denies clearance for landing if weather is stormy' do
+      weather = double()
+      weather.stub(:report) { "stormy" }
+      airport = Airport.new('Makers airport')
+      plane = Plane.new('MAK121')
+      expect { airport.land(plane, weather) }.to raise_error 'Landing denied due to weather'
+    end
+
```
EdwardAndress, line 130: Good tests for how weather impacts on landing / takeoff.  You could use `context` to structure your tests like this

```
    end
  end

    end
  end
end
```

EdwardAndress:
really nice work Alistair!  I left some comments in-line, all pretty minor but I hope they're helpful

PR #2014, opened 2021-01-29
Airport challenge

stacy1999:


lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+  def land(plane)
+    raise 'Cannot land, airport full' if full?
+    raise 'Cannot land, weather is stormy' if stormy?
+    @queue.push(plane) 
+    
+  end
+
+  def take_off(plane)
+    raise 'Cannot take off, weather is stormy' if stormy?
+
```
EdwardAndress, line 18: here, we want the plane to be removed from the `@queue`, right? Which array method can you use for that?

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def hangar
+    @queue
+  end
+
+  private
```
EdwardAndress, line 25: nice that you're using private methods!

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+    @queue
+  end
+
+  private 
+  def full?
+    @queue.length >= @capacity
+  end
+
+  def stormy?
+    Weather.new.stormy?
```
EdwardAndress, line 31: nice that you've created a separate class for weather!

spec/weather_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'weather'
+
+describe Weather do
+  let (:weather) { Weather.new }
+  describe '#stormy?' do
+    it "returns sunny" do
+      allow(Kernel).to receive(:rand).and_return 1
```
EdwardAndress, line 7: nice that you're using Kernel, though it's not the only solution.  If you do `rand` in your class, rather than `Kernel.rand` the implicit recipient is the instance of that class so when you use your stub you would do `allow(weather).to receive(:rand).and return 1`

EdwardAndress: question: what does this mean about the `rand` method?  Why do both approaches work?

spec/airport_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  describe '#take_off' do
+    it 'checks .take_off instructs a plane to take off from an airport' do
+      expect airport.take_off(plane) 
+    end
+
+    it 'checks .take_off prevents take off when stormy' do
+        allow(airport).to receive(:stormy?).and_return true
+        expect { airport.take_off(plane) }.to raise_error 'Cannot take off, weather is stormy'
+      end
+  end
```
EdwardAndress, line 38: Good tests for how weather impacts on landing / takeoff.  You could use `context` to structure your tests like this

```
    end
  end

    end
  end
end
```

EdwardAndress:
nice attempt – you're clearly picking up the fundamentals of TDD :)

there are some comments in-line and I hope they're helpful

PR #2015, opened 2021-01-29
Completed airport challenge

harrythomson1:
I feel like I haven't tested my weather class correctly but I really struggled with how I would do this.

Maybe I could have split it up the "decide_weather" function into two separate functions then I could test each function but I'm not 100% sure.

Also, I know I should've been uploading the GitHub at intervals of my project but it completely slipped my mind! I promise I will get into the habit!

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,57 @@
+require "./lib/plane"
+require "./lib/weather"
+
+class Airport
+
+  attr_reader :planes
+  attr_reader :capacity
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @weather = ""
+    @planes = []
+  end
+
+  def land(plane)
+    checks_safety
+    checks_if_plane_already_in_hangar(plane)
+    checks_current_capacity
+    @planes << plane
+  end
+
+  def takeoff
+    checks_planes_available_to_takeoff
+    checks_safety
+    plane_takeoff_announcement
+    @planes.pop
+  end
+
+  private
+
+  def plane_takeoff_announcement
+    puts "#{@planes[-1]} has left the airport"
+  end
+
+  def checks_current_capacity
+    fail "This airport is currently at maximum capacity" unless @capacity > @planes.length
+  end
+
+  def checks_if_plane_already_in_hangar(plane)
+    return if @planes.empty?
+    fail "#{plane} has already landed" unless @planes.include?(plane) == false
+  end
+
+  def checks_weather
+    @weather = Weather.new.current_weather
+  end
+
+  def checks_safety
+    fail "It is too stormy" unless checks_weather == "sunny"
+  end
+
+  def checks_planes_available_to_takeoff
+    fail "There are no planes in the hangar" unless @planes.empty? == false
+  end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -9,7 +9,7 @@ class Airport
 
   def initialize(capacity = DEFAULT_CAPACITY)
     @capacity = capacity
-    @weather = ""
+    @weather = Weather.new
     @planes = []
   end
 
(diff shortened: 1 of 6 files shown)


harrythomson1:
I've changed how my weather class works and created stubs for the tests.

I've also changed how the weather is checked, before I was creating a new weather class each time now I only create one weather class each time my airport class is initiated.

I still feel like I could come up with more clear and concise variable names though.

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
+require "./lib/plane"
+require "./lib/weather"
+
+class Airport
+
+  attr_reader :planes
+  attr_reader :capacity
+  DEFAULT_CAPACITY = 10
```
EdwardAndress, line 8: nice use of a constant here

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @weather = Weather.new
```
EdwardAndress, line 12: it's great that you've used a separate class for Weather.  how could you make your solution more flexible / less tightly coupled to the Weather class?

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  def checks_current_capacity
+    fail "This airport is currently at maximum capacity" unless @capacity > @planes.length
+  end
+
+  def checks_if_plane_already_in_hangar(plane)
+    return if @planes.empty?
+    fail "#{plane} has already landed" unless @planes.include?(plane) == false
+  end
+
+  def checks_safety
```
EdwardAndress, line 45: nice descriptive method that could be extended to include other safety checks

spec/airport_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    it "stores a landed plane in the airport" do
+      allow(Weather).to receive(:new) {weather}
+      allow(weather).to receive(:stormy?) { "sunny" }
+      expect(subject.land(plane)).to eq([plane])
+    end
+
+    it "prevents landing when the airport is full" do
+      allow(Weather).to receive(:new) {weather}
```
EdwardAndress, line 27: Not sure you need this!

spec/airport_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      allow(weather).to receive(:stormy?) { "sunny" }
+      Airport::DEFAULT_CAPACITY.times { subject.land(Plane.new) }
```
EdwardAndress, line 29: nice use of the constant, which makes your test really flexible (it won't break when the default changes)

spec/airport_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      expect { subject.land(Plane.new) }.to raise_error "This airport is currently at maximum capacity"
+    end
+
+    it "doesn't allow landing during a storm" do
+      allow(Weather).to receive(:new) {weather}
+      allow(weather).to receive(:stormy?) { "stormy" }
+      expect { subject.land(plane) }.to raise_error "It is too stormy"
+    end
+  end
+
```
EdwardAndress, line 39: Good tests for how weather impacts on landing / takeoff.  You could use `context` to structure your tests like this

```
    end
  end

    end
  end
end
```

spec/weather_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "weather"
+
+describe Weather do
+  
+  describe "initialize" do
+    it "creates a storm" do
+      allow(subject).to receive(:rand) { 10 }
```
EdwardAndress, line 7: you mentioned that you're not sure about your tests here...
I think these two tests are good.  You could make them more specific like this

```
allow(subject).to receive(:rand).with(1..10) { 10 }
```

You could also consider doing something like this
```
range = (1..10)
allow(subject).to receive(:rand).with(range) { range.to_a }
```

then call the method 10 times, collect the outputs in an array and ensure that the array is composed correctly.  this would ensure that for each value in range, your program returns the correct result

EdwardAndress:
it's a nice attempt!  Good job.  There are a few comments in-line and I hope they're helpful.  I tried to address your question about testing the weather class but DM me on Slack if you have further questions.

PR #2016, opened 2021-01-29
airport_challenge

girl-interrupted:


lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+
+class Airport
+
+  attr_reader :all_planes
+
+  DEFAULT_CAPACITY = 35
```
EdwardAndress, line 6: nice use of a constant here

spec/airport_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    airport.can_land_a_plane('new_flight_num')
+    airport.can_take_off('new_flight_num')
+    expect(airport.all_planes).not_to include 'new_flight_num'
+  end
+  end
+
+  describe '#full_airport' do
+    it 'does not land planes when the airport is full' do
+      airport.can_land_a_plane('new_flight_num')
+      35.times {airport.can_land_a_plane('new_flight_num')}
```
EdwardAndress, line 26: nice test for capacity!

EdwardAndress:
This is a good attempt and it demonstrates that you've picked up the basics of testing, which is great, so well done.  The next step is to think about how to deal with randomness (in the weather, for example).

lib/weather.rb
```diff
@@ -0,0 +1,8 @@
+require_relative 'airport'
```
EdwardAndress, line 1: Now you have a weather class!  Awesome :)

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def can_land(flight)
+    fail "cannot land at the moment due to stormy weather" if weather.stormy?
+    @all_planes.push(flight)
+  end
+
+  def can_take_off(flight)
+    !weather.stormy?
+    @all_planes.delete(flight)
```
EdwardAndress, line 23: At the moment, the plane will still land even if the weather is stormy.  How can you fix this?

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+describe Airport do
+
+  let(:airport) { Airport.new }
+  let(:plane) { double :plane } # using a double here because I created a separate class
+  let(:weather) { double :weather }
+
+  describe '#can_land_a_plane' do
+    it 'lands a plane at an airport, registers it' do
+      allow(weather).to receive(:stormy?).and_return 'false'
```
EdwardAndress, line 13: Do you need this line?

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      airport.airport_full?
+      airport.can_land(plane)
+      expect(airport.all_planes).to include plane
+    end
+    it 'cannot land a plane if the weather is stormy' do
+      #airport.can_land(plane) # here i am not sure if I need to use the can_land nethod before I check the weather on newxt line
+      allow(weather).to receive(:stormy?).and_return 'true'
```
EdwardAndress, line 20: I'm guessing this is the test that you described on Slack (which doesn't work as intended)... you'll probably fix this by addressing the comment I made about how you're using the weather (above)

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  describe '#airport_full?' do
+    it 'does not land planes when the airport is full' do
+      Airport::DEFAULT_CAPACITY.times { airport.can_land(plane) }
+      expect { airport.can_land(plane) }.to raise_error "airport is full now, cannot land any planes"
+    end
+  end
+
+  describe '#initialize' do
+    it 'has the capacity that can be overridden when needed' do
+      airport.capacity == Airport::DEFAULT_CAPACITY # here I wasn't sure if I need to write 35 or the DEFAULT_CAPACITY const?
```
EdwardAndress, line 43: Here you're comparing the capacity to the default capacity but maybe you're intending to assign a new capacity..?

spec/weather_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'weather'
+
+describe Weather do
+  let(:weather) { Weather.new }
+  describe '#stormy?' do
+    it 'it is not stormy now' do
+      allow(weather).to receive(:rand).and_return 2
```
EdwardAndress, line 7: nice use of a stub!

spec/weather_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+      expect(weather.stormy?).to eq false
+    end
+    it 'it is stormy now' do
+      allow(weather).to receive(:rand).and_return 10
```
EdwardAndress, line 11: and again, here!

PR #2017, opened 2021-01-30
Airport Challenge Complete

yy140:
All tests pass. Weather class needed.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,20 @@
+require 'plane'
+
+class Airport
+  
+  def initialize
+    @plane = []
+  end
+
+  def land(plane)
+    raise 'The airport is full' if @plane.length  == 1 
+    @plane.push(plane)
+  end
+
+  def take_off(plane)
+    @plane.delete(plane)
+    "#{plane} has taken off"
+  end
+
+  
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -8,7 +8,8 @@ class Airport
   end
 
   def land(plane)
-    raise 'The airport is full' if @plane.length  >= @airport_capacity 
+    raise 'The airport is full' if full? 
+    
     @plane.push(plane)
   end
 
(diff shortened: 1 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    @plane.push(plane)
+  end
+
+  def take_off(plane)
+    raise 'Cannot take off due to storm' if stormy? == 3
+     @plane.delete(plane)
+    
+  end
+
+  def stormy?
```
EdwardAndress, line 22: it's convention for methods that end in a `?` to return a boolean value... so this could mislead

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    rand(4)
+  end
+
+  private
```
EdwardAndress, line 26: awesome that you're using private methods!

spec/airport_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'airport'
+
+describe Airport do
+  describe '#land' do
+    it 'raises an error if airport it full' do
+      allow(subject).to receive(:stormy?).and_return(1)
+      subject.airport_capacity.times { subject.land(Plane.new) }
```
EdwardAndress, line 7: ah, really nice way of flexibly reaching capacity!

spec/airport_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      plane = Plane.new
+      expect {raise subject.land(plane)}.to raise_error('The airport is full')
+    end
+
+     it 'raises an error if landing during a storm' do
+      plane2 = Plane.new
+      allow(subject).to receive(:stormy?).and_return(3)
+      expect { subject.land(plane2)}.to raise_error('Too stormy to land')
+     end
```
EdwardAndress, line 16: Good tests for how weather impacts on landing / takeoff.  You could use `context` to structure your tests like this

```
    end
  end

    end
  end
end
```

EdwardAndress:
nice solution!  I've left some comments in-line and hope they're helpful.  One thing to focus on is how to move weather behaviour into it's own class, with test coverage

PR #2018, opened 2021-01-30
Airport challenge

vinc3108:


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'plane'
+
+class Airport
+  attr_accessor :capacity, :planes
+  DEFAULT_CAPACITY = 10
```
EdwardAndress, line 5: nice use of a constant

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+  def plane_count
+    @planes.length
+  end
+
+  def stormy?
+    # rand(1..10) > 7
+    false
+  end
+
+  def check_landing_conditions(tail_number)
```
EdwardAndress, line 29: nice grouping of checks into one method!

EdwardAndress: this makes your solution easier to change and easier to understand

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+    @capacity = capacity
+    @planes = []
+  end
+
+  def approved_for_landing(tail_number)
+    check_landing_conditions(tail_number)
+    @planes << tail_number
+  end
+
+  def approved_for_takeoff(tail_number)
```
EdwardAndress, line 16: do you not need to check weather conditions prior to takeoff?

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+    @planes.delete(tail_number)
+  end
+
+  def plane_count
+    @planes.length
+  end
+
+  def stormy?
```
EdwardAndress, line 24: I'm guessing you commented out the randomness for testing..?

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      expect(airport).to respond_to(:approved_for_landing).with(1).argument
+    end
+  
+    it 'parks a plane once it has landed' do
+      airport.approved_for_landing(plane)
+      expect(airport.planes).to include(plane)
+    end
+
+    context 'when at capacity' do
+      it 'does not allow landing' do
```
EdwardAndress, line 21: nice test!

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  describe '#approved_for_takeoff' do
+    it 'is expected to respond to approved_for_takeoff' do
+      expect(airport).to respond_to(:approved_for_takeoff).with(1).argument
+    end
+
+    it 'takes a plane and removes it from the parking lot' do
+      airport.approved_for_landing(plane)
+      airport.approved_for_takeoff(plane)
+      expect(airport.planes).not_to eq(plane)
```
EdwardAndress, line 55: nice test but would `airport.planes` ever be eq to `plane`?  `plane` is a single plane but `airport.planes` is an array

EdwardAndress:
Nice solution!  I've left some in-line comments and hope they are helpful.  Looks like you were unsure how to deal with the random nature of the weather.  You should look into `stubbing randomness in ruby` ;)

PR #2019, opened 2021-02-01
Airport challenge finished. 

HarryCarton:
Tests not all passing but code is okay. Had trouble with weather tests in airport_spec. Needs lots of refactoring in spec files

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,28 @@
+require_relative './plane'
+require_relative './weather'
+class Airport
+
+  attr_reader :hangar
+ 
+  def initialize(capacity = 10)
+    @capacity = capacity
+    @hangar = []
+    @weather = Weather.new
+  end
+
+  def land_plane(plane)
+    @weather.update_weather
+    fail('Plane is already in hangar') if @hangar.include?(plane)
+    fail('Is it too stormy') unless @weather.stormy == false
+    fail ('Airport is full') unless @hangar.length < @capacity    
+    @hangar.append(plane)
+  end
+
+  def take_off(plane)
+    @weather.update_weather
+    fail('Plane is not in hangar') unless @hangar.include?(plane)
+    fail('Is it too stormy') unless @weather.stormy == false
+    @hangar.delete(plane)
+  end
+
+end
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,0 +1,20 @@
+class Weather 
+  
+  attr_reader :stormy
+  
+  def initialize
+    @stormy = false
+  end
+
+  def storm?
+    rand(10)==0
+  end
+
+  def update_weather
+    if storm?
+      @stormy = true
+    else
+      @stormy = false
+    end
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,71 @@
+require 'airport'
+
+describe Airport do
+  describe '#land_plane' do 
+    it 'when .land_plane is run with string string is in airport hangar' do
+      subject.land_plane('plane_string')
+      expect(subject.hangar).to include('plane_string')
+    end
+    it 'when .land_plane is run with plane the plane is in hangar' do
+      plane = Plane.new 
+      subject.land_plane(plane)
+      expect(subject.hangar).to include(plane)
+    end
+    it 'wont let a plane land if the airport is full' do
+      plane1 = Plane.new
+      plane2 = Plane.new
+      weather = Weather.new
+      airport = Airport.new(1)
+      airport.land_plane(plane1)
+      weather.instance_variable_set(:@stormy, false)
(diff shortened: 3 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+  def initialize(capacity = 10)
+    @capacity = capacity
+    @hangar = []
+    @weather = Weather.new
+  end
+
+  def land_plane(plane)
+    @weather.update_weather
+    fail 'Plane is already in hangar' if @hangar.include?(plane)
+    fail 'It is too stormy' unless @weather.stormy == false
```
EdwardAndress, line 16: If you have a method that returns a boolean you can use it in your if statement like this

`if @weather.stormy?`

and you don't need to do `== true`

it's also convention for ruby methods that return booleans to end with a `?`

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+    fail 'Airport is full' unless @hangar.length < @capacity   
+
+    @hangar.append(plane)
+  end
+
+  def take_off(plane)
+    @weather.update_weather
+    fail 'Plane is not in hangar' unless @hangar.include?(plane)
```
EdwardAndress, line 24: Nice that you're checking whether the plane is actually present before trying to make it fly!  Spotting edge cases like that is really valuable when it comes to working on something important.

lib/weather.rb
```diff
@@ -0,0 +1,20 @@
...
+  
+  def initialize
+    @stormy = false
+  end
+
+  def storm?
+    rand(10)==0
+  end
+
+  def update_weather
```
EdwardAndress, line 13: I like the way that you've separated reading the weather from determining the weather, with this `update_weather` method

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  let(:weather) { Weather.new }
+  
+  before do
+    subject.instance_variable_set(:@weather, weather)
+  end
+ 
+  describe '#land_plane' do 
+    it 'when .land_plane is run with string string is in airport hangar' do
+      expect(weather).to receive(:storm?).and_return(false)
+      subject.land_plane('plane_string')
```
EdwardAndress, line 14: why don't you use the `plane` object here?

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect(subject.hangar).to include('plane_string')
+    end
+    it 'checks update_weather is called in land_plane' do
+      expect(weather).to receive(:update_weather)
```
EdwardAndress, line 18: nice test!

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect(subject.hangar).to include(plane)
+    end
+    it 'wont let a plane land if the airport is full' do
+      airport = Airport.new(0)
+      airport.instance_variable_set(:@weather, weather)
+      expect(weather).to receive(:storm?).and_return(false)
+      expect { airport.land_plane(plane) }.to raise_error('Airport is full')
+    end
+    it 'wont let a plane land if the weather is stormy' do 
+      expect(weather).to receive(:storm?).and_return(true)
```
EdwardAndress, line 33: here, the assertion is that the an error message will be raised so you can / should use `allow` rather than `expect` on line 33

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect { subject.land_plane(plane) }.to raise_error('It is too stormy')
+    end
+    it 'wont let a plane land twice' do 
+      expect(weather).to receive(:storm?).and_return(false)
+      subject.land_plane(plane)
+      expect(weather).to receive(:storm?).and_return(false)
+      expect { subject.land_plane(plane) }.to raise_error('Plane is already in hangar')
```
EdwardAndress, line 40: another nice edge case!

spec/airport_spec.rb
EdwardAndress: same comment for a few other places :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    
+    it 'when .take_off is called with plane as arg it leaves hangar' do
+      expect(weather).to receive(:storm?).and_return(false)
+      subject.land_plane(plane)
+      expect(weather).to receive(:storm?).and_return(false)
+      subject.take_off(plane)
+      expect(subject.hangar).to_not include(plane)
+    end
+
+    it 'doesnt let a plane take off if the weather is stormy' do
```
EdwardAndress, line 54: Good tests for how weather impacts on landing / takeoff.  You could use `context` to structure your tests like this

```
    end
  end

    end
  end
end
```

EdwardAndress:
A nice solution!  I've left a few in-line comments around things that would improve it but you've clearly picked up the basics of TDD.  Good job!

PR #2020, opened 2021-02-01
first commit

mitchell-917:


chill-debug:
**AirPort challenge code review comments**

> Simplistic 

>  Versatile 

> Clean

> Consistent 

> Legible 

> Portable 

**Great** Work***

spec/airport_spec.rb
```diff
@@ -0,0 +1,82 @@
+require 'airport.rb'
+
+describe Airport do
+  let(:subject) { Airport.new("Gatwick", 100) }
+  let(:subject_2) { Airport.new("Heathrow", 100) }
```
EdwardAndress, line 5: What were you going to use this for? :)

EdwardAndress: `subject_2`, that is

spec/airport_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  let(:plane) { Plane.new }
+
+  describe "Testing @weather responds " do
+    it "views weather at given airport" do
+      allow(subject).to receive(:weather).and_return("stormy")
```
EdwardAndress, line 10: You've nailed the syntax here, which is great, but you shouldn't stub the method under test (`weather`, in this case).

EdwardAndress: You might find some other places where the same comment also applies :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  describe "Testing take_off responds " do
+    it "takes off a plane" do
+     allow(subject).to receive(:weather_stormy?).and_return false 
+      expect(subject).to respond_to(:takeoff_plane).with(1).argument
+    end
+  end
+
+  describe "Testing when airport is full" do
+    it "Does not allow landing when the airport is full" do
+      allow(subject).to receive(:weather_stormy?).and_return false
```
EdwardAndress, line 40: nice test for capacity!

spec/airport_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      allow(subject).to receive(:weather_stormy?).and_return true
+      expect { subject.land_plane(:plane) }.to raise_error "The Airport is stormy!"
+    end
+  end
+  describe "Testing that no planes can take off when it's stormy" do
+    it "No planes can takeoff when the airport weather is stormy" do
+      allow(subject).to receive(:weather_stormy?).and_return true
+      expect { subject.takeoff_plane(:plane) }.to raise_error "The Airport is stormy!"
+    end
+  end
```
EdwardAndress, line 58: Good tests for how weather impacts on landing / takeoff.  You could use `context` to structure your tests like this

```
    end
  end

    end
  end
end
```

EdwardAndress:
Hey Mitchell, this is a great attempt – good job!  I've left some comments against the code in specific places.  The summary of those would be that:
- It's really great that you picked up / are practicing the use of stubs
- Be careful to ensure that you're stubbing the 'correct' things

PR #2021, opened 2021-02-01
Add files via upload

Dofassh:


airport.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    def take_off(plane)
+    raise 'Weather is stormy : Plane cannot take off' if stormy?
+
+    end
+
+    private 
+
+    def capacity_full
```
EdwardAndress, line 22: It's nice that you created a method for checking if the capacity it full 👍

airport_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+            plane = Plane.new
+            allow(airport).to receive(:stormy?).and_return false
+            expect(airport) .to respond_to :land
+        end 
+
+        it "does not let plane land when capacity is full" do
+            airport = Airport.new(30)
+            plane = Plane.new
+            allow(airport).to receive(:stormy?).and_return false
+            30.times do
```
EdwardAndress, line 17: Nice test for capacity / exceeding capacity

airport_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+            expect{airport.land(plane)}.to raise_error 'Weather is stormy : Cannot land plane'
+        end
+
+        it "does not allow planes to take off when stormy" do 
+            airport = Airport.new(30)
+            plane = Plane.new
+            allow(airport).to receive(:stormy?).and_return true
+            expect{airport.take_off(plane)}.to raise_error 'Weather is stormy : Plane cannot take off'
+        end
+    end
```
EdwardAndress, line 46: Good tests for how weather impacts on landing / takeoff.  You could use `context` to structure your tests like this

```
    end
  end

    end
  end
end
```

airport.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    @planes << plane
+    end
+
+    def take_off(plane)
+    raise 'Weather is stormy : Plane cannot take off' if stormy?
+
+    end
+
+    private
```
EdwardAndress, line 20: nice use of private methods!

plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
+
```
EdwardAndress, line 2: Looks like you were unsure what the plane should do / what it's behaviours should be.  This is a bit like me being unsure about what Items should do :)

Right now, I think it's OK not to have a plane class.

EdwardAndress:
Hey Parmeet, this is a good attempt and I can see that you've learned the basics of TDD from last week, which is great.  The next step would be to think about how you can move weather related behaviour into a separate weather class.

I particularly like the fact that you've used some private methods.

Also... don't forget to indent things nicely :)

PR #2022, opened 2021-02-01
Completed Airport Challenge 1.31.21

RowHunt:
Rowan's Pull Request

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,28 @@
+class Airport
+
+  attr_reader :hangar
+
+  def initialize
+    @hangar = []
+  end
+
+  def clear_for_landing(flight_number, weather = 1)
+    if weather == 1
+      @hangar << flight_number
+    else
+      puts "Due to inclement weather, landing is disabled."
+    end
+  end
+
+  def clear_for_takeoff(flight_number, weather = 1)
+    if weather == 1
+      @hangar.delete(flight_number)
+    else
+      puts "Due to inclement weather, takeoff is disabled."
+    end
+  end
+
+  def max_capacity(cap_number = 4)
+    @hangar.first(cap_number)
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,118 @@
+# User Story 1:
+# As an air traffic controller 
+# So I can get passengers to a destination 
+# I want to instruct a plane to land at an airport
+
+# Translation:
+# Airport class
+# With a method that allows me to land planes there -> clear_for_landing(plane)
+# It should store the planes somewhere -> array called hangar 
+
+# User Story 2:
+# As an air traffic controller 
+# So I can get passengers on the way to their destination 
+# I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
+
+# Translation:
+# Planes taking off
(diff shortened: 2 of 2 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  describe '#clear_for_landing' do
+    it 'takes a plane and adds it to the hangar' do
+      airport.clear_for_landing("BA123")
+      expect(airport.hangar).to include "BA123"
+    end
+  end
+
+  describe 'clear_for_takeoff' do
+    before(:each) do
```
EdwardAndress, line 70: Nice use of `before`

spec/airport_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+# Translation:
+# Landing disabled when stormy -> Method: disabled
+
+require "airport.rb"
+
+RSpec.describe Airport do
+
+  let(:airport) { Airport.new }
+
+  describe '#clear_for_landing' do
```
EdwardAndress, line 62: good tests for both clear_for_landing and clear_for_takeoff

spec/airport_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      airport.clear_for_landing("NB543")
+      airport.clear_for_landing("GZ765")
+      airport.clear_for_landing("HY890")
+      airport.clear_for_landing("SF267")
+      max_hangar = airport.max_capacity(5)
+      expect(max_hangar).to include("SF267")
+    end
+  end
+
+  describe 'disabled clear_for_landing' do
```
EdwardAndress, line 104: Good tests for how weather impacts on landing / takeoff.  You could use `context` to structure your tests like this

```
    end
  end

    end
  end
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      airport.clear_for_takeoff("BA123")
+      expect(airport.hangar).to_not include("BA123")
+    end
+    it 'returns flight number of the plane that just took off' do
+      actual = airport.clear_for_takeoff("BA123")
+      expect(actual).to eq("BA123")
+    end
+  end
+
+  describe 'max_capacity' do
```
EdwardAndress, line 83: You've implemented a max capacity in a clever way that also allows override but it's a little bit misleading (?) because actually planes can land even if the max capacity is exceeded – they just don't appear in your max_capacity list.

Instead, you could check how many planes are in the hangar as part of your land procedure and only allow landing if the max capacity it not met (and the weather is not stormy).

How would you do this?

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+  def clear_for_landing(flight_number, weather = 1)
+    if weather == 1
+      @hangar << flight_number
+    else
+      puts "Due to inclement weather, landing is disabled."
+    end
+  end
+
+  def clear_for_takeoff(flight_number, weather = 1)
+    if weather == 1
```
EdwardAndress, line 18: Using a number for weather works but is less expressive than using a string that describes the type of weather.  E.g. "stormy".

EdwardAndress:
Hey Rowan!  This is a good attempt and you've demonstrated that you picked up some essential TDD skills from last week.  Good job!  I've left a few comments against the solution in specific places.  Let me know on Slack if you have any questions about that.

PR #2023, opened 2021-02-01
First Attempt

RomanR12:
Took forever to get this far

### DIFF
--- a/airport/lib/airport.rb
+++ b/airport/lib/airport.rb
@@ -1,0 +1,25 @@
+class Airport
+
+  attr_reader :hanger
+
+  def initialize
+    @planes = []
+  end
+
+  def plane_to_land(flight_number)
+    @planes << flight_number
+  end
+
+  def hanger
+    @planes
+  end
+
+  def plane_takeoff(flight_number)
+    @planes.delete(flight_number)
+    p @planes
+  end
+  def no_landing
+    @planes.length
+  end
+
+end
--- a/airport/spec/airport_spec.rb
+++ b/airport/spec/airport_spec.rb
@@ -1,0 +1,36 @@
+# As an air traffic controller
+# So I can get passengers to a destination
+# I want to instruct a plane to land at an airport
+
+# As an air traffic controller
+# So I can get passengers on the way to their destination
+# I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
+
+# As an air traffic controller
+# To ensure safety
+# I want to prevent landing when the airport is full
+
+require "airport"
+
+RSpec.describe Airport do
+  describe "plane_to_land" do
+    it "plane landed so add to hanger" do
+      airport = Airport.new
+      airport.plane_to_land("flight1")
+      expect(airport.hanger).to include "flight1"
+    end
+  end
+  describe "plane_takeoff" do
+    it "plane leaves and confirm" do
+      airport = Airport.new
+      airport.plane_takeoff("flight1")
+      expect(airport.hanger).not_to include "flight1"
+    end
+  end
+  describe "no_landing" do
+    it "checks size of hanger" do
+      airport = Airport.new
+      expect(airport.no_landing).to be <= 4
+    end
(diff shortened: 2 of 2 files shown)


airport/lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+class Airport
+
+  attr_reader :hanger
+
+  def initialize
+    @planes = []
+  end
+
+  def plane_to_land(flight_number)
```
EdwardAndress, line 9: if @planes.length == 5
  raise "Airport is full"
else
  @planes << flight_number
end

EdwardAndress: but... 5 is a magic number here (it's not entirely clear what it means) and we cannot change it for different airports –> they'll all have the same capacity.

airport/lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+class Airport
+
+  attr_reader :hanger
+
+  def initialize
```
EdwardAndress, line 5: Following on from below...

we could do

```
def initialize(capacity)
  @capacity = capacity
  @planes = []
end
```

Then we can use the instance var when checking if the airport is full

airport/lib/airport.rb
EdwardAndress: if we have an instance var for capacity we can do this
```
if @planes.length == @capacity
   raise "Airport is full"
else
  @planes << flight_number
end
```

EdwardAndress: we could also check current number of planes using a method to give more meaning
```
def plane_to_land(flight_number)
  if number_of_planes_at_airport == @capacity
     raise "Airport is full" 
  else
    @planes << flight_number
  end
end

def number_of_planes_at_airport
  @planes.length
end
```

airport/lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+class Airport
+
+  attr_reader :hanger
```
EdwardAndress, line 3: You don't need this.

attr_reader is a method that creates another method

in this case it will create this

```
def hangar
  @hangar
end
```

EdwardAndress: if you didn't have your own hangar method, that reads the @planes variable, airport.hangar would return nil

airport/lib/airport.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def plane_to_land(flight_number)
+    @planes << flight_number
+  end
+
+  def hanger
+    @planes
+  end
+
```
EdwardAndress, line 16:
```
def weather
  rand(100) > 90 ? "stormy" : "sunny"
end
```

```
def weather
  if rand(100) > 90
    "stormy"
  else
    "sunny"
  end
end
```

```
def weather
  case rand(100)
  when 0..10
    "stormy"
  when 11..20
    "blizzard"
  when 21..100
    "sunshine"
  end
end
```

airport/spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      airport.plane_takeoff("flight1")
+      expect(airport.hanger).not_to include "flight1"
+    end
+  end
+  describe "no_landing" do
+    it "checks size of hanger" do
+      airport = Airport.new
+      expect(airport.no_landing).to be <= 4
+    end
+  end
```
EdwardAndress, line 35: describe 'weather' do
  it 'returns stormy when random number is 5' do
     airport = Airport.new(10)
     allow(airport).to receive(:rand).and return 5
     expect(airport.weather).to eq "stormy"
  end
end

EdwardAndress:
Hey Roman!  This is a good attempt and you've demonstrated that you've taken on board some of the key aspects of the TDD week.  It will take (everyone) a while to become really proficient at TDD, so don't be surprised if it's hard but do keep working on it and fluency will come with practice.  We spoke about some specific things yesterday and there are notes above, so I won't repeat all that stuff here!

PR #2024, opened 2021-02-07
completed airport challenge

jshorns:
eek. see readme for notes on my approach and how to use the program.


jshorns:
all user stories have been completed.

- Not sure with the error guard conditions should have their own methods - so that land and take_off just call full? and stormy? by themselves? I think this is the change I would make.
- In contrast to the bikes challenge, I chose to define the land methods within the Plane class rather than the Airport class. This is largely I product of how I read the user stories, and I did contemplate switching them round. However, I felt it made a certain amount of syntactic sense to keep it how it was (and would also make the challenge slightly different, removing the temptation to simply copy and paste code from the challenge in the week.) I did find this created problems with testing the 'full?' method using doubles, as whatever the capacity of the double, its response to the full method would be hard-coded into it. This method is tested in the airport spec though so the coverage is still there.
- I have tried to design error messages so that they are thrown in a logical order (e.g. telling a plane to land when it is already landed throws an error before the question of the weather is even raised.)
- The biggest problem I had was with the weather method as I got caught up on making it reflect real life and making an airport always have a weather attribute. On reflection I felt that the point of the method was to simply get a snapshot of the weather at the exact moment a plane was instructed to land or take off, so I felt this wasn't necessary. I did feel it might make sense in real world application to store the most recent weather status to the Airport object each time it was checked, so that the controller could see this when telling planes where to land. However, building the weather method into the land and take_off methods effectively achieves the same thing, since the error message comes back before the plane is instructed to do anything. There was definitely a lesson in this in that I made it too complicated because of not test-driving it! I reverted a commit and made the method again through TDD and it was much more straightforward.
- Hopefully everything else is self-explanatory! I did initially also have a 'title' attribute for airports so they could be named, which would make the airport status more readable when planes were landed, but I felt this was making the code unnecessarily complicated, and that readability would be factored in when 'translating' this code into an actual user interface.

MJCXII (commented):
Code Review

Layout:
- I would suggest separating the weather into a separate class/module as it's a different object
- applies to above but also having separate class for weather

Optimization
- plane.rb line 5: You have assigned @status as a string here and then later checking the string, this would be a good place to use a boolean (true/false statement) as it's takes up a lot less room in memory and is designed for 'this or that' functions.
- SRP related to fail statements in your land and take_off methods, you can separate your fail statements into their own method and then call that method within land and takeoff.
- DRY related to plane_spec line 56 - 65, this can be replaced by using the .times method, you can also instead of using 10.times if the default capacity was changed use the constant you created by referencing it from the file (Airport::DEFAULT_CAPACITY.times {#code})

Positives
- Code is easy to understand and edit if necessary
- error messages are clear and concise
- stormy method is very optimized
- tests are extensive and cover all of your code adequately

PR #2027, opened 2021-02-07
Jai's solution to the Makers airport_challenge

jai-jk:
## How far did I get? What did I learn? ##

I completed the whole challenge!! Well, sort of.. I didn't code all of the edge cases - I coded the tests and functionality for one edge case, but stopped there. Why? I must have a good reason, right? Absolutely!! The Super Bowl is on in a few minutes and it's Patrick Mahomes vs Tom Brady.. of course I can't miss that! If you have no idea what that means; think of it like the Buddha vs Jesus.. but obviously more aggressive - basically two ridiculously awesome people. But that's besides the point, and I probably prefer Lao Tzu anyway :).

The tests [should] all pass with 100% coverage, including the first edge case (planes trying to land while they're still grounded). It seemed like the best way to write the other edge cases would have probably been to work with the Plane class and spec and I did start them but had to restart a few times - so I chose to conclude it here. This is the one crucial thing that could improve/properly complete the program, and what I would complete if I had another day.

This challenge has really helped me push my synctatical knowledge and skills with RSpec, and I really enjoyed it as I went through it. It was quite frustrating at times, especially because the code itself isn't very difficult to write - I'm still getting confident and effective with the actual syntax and logic of RSpec but I'm coming to really love it. Cycling through Feature tests and Unit tests helped me build a relatively robust program and deepen my understanding. I really enjoy thinking about products and why they're successful, and am super eager to build products that are solid and useful for customers, so the TDD process really appeals to me and I can see why it's useful for building sustainable programs.

I also need to do further research and practice with private and public methods, but did my best with them today.


lib/airport.rb
```diff
@@ -17,15 +17,14 @@ def new_capacity(new_capacity = DEFAULT_CAPACITY)
 
   def land(plane)
     prevent_landing
+    raise "Plane is already at airport, cannot land again." if plane_at_airport?(plane)
     @planes_at_airport << plane
-    return plane
+    plane.to_s + " landing now" if plane_at_airport?(plane)
```
olgachmaj, line 23: I like how you're letting the user know that the landing is being executed.

spec/airport_spec.rb
```diff
@@ -23,12 +23,12 @@
     end
   end
 
-  context 'when a plane has already landed' do
+  context ' - when a plane has already landed' do
     before do
       @airport.land(@plane)
```
olgachmaj, line 28: I like the use of before...do statement, it makes a lot of sense to use it here.

lib/weather.rb
```diff
@@ -1,11 +1,11 @@
 class Weather
 
   def weather_state
-    rand(1..5)
+    rand(1..10)
```
olgachmaj, line 4: I think using .sample method instead of .rand in this case would be easier to test and quicker to read/write. It would also help with some magic numbers in the test.

README.md
```diff
@@ -12,6 +13,46 @@ Airport Challenge
                 =  ===(_________)
 
 ```
```
olgachmaj, line 17: Would be nice to see a 'out of the box' instruction on how to use the application in irb to be able to get going quickly.

lib/airport.rb
```diff
@@ -8,9 +8,9 @@ def initialize
   end
 
   def land(plane)
-      prevent_landing
-      @planes_at_airport << plane
-      return plane
+    prevent_landing
+    @planes_at_airport << plane
+    return plane
```
olgachmaj: You can use implicit return here, but that's me being nitpicky 🤣

lib/airport.rb
```diff
@@ -32,7 +32,7 @@ def full?
   end
 
   def prevent_landing
-    raise "Airport is currently full. You cannot land." if self.full?
+    raise "Airport is currently full. You cannot land." if full?
```
olgachmaj, line 46: Ideally there should be a empty line under this statement.

README.md
```diff
@@ -13,77 +16,57 @@ Airport Challenge
...
+The Weather class contains a method to randomly set the weather_state, and a stormy? method to check if the weather_state is stormy (10% chance). Weather_state selects a random number between 1 and 10, and stormy? == true if weather_state == 10. The stormy? method is called and utilised within the Airport class, to prevent planes from landing and taking off when the weather is unsuitable(stormy?).
 
-Please create separate files for every class, module and test suite.
+All of the tests are currently spread out in proportion to the size of the classes, as this program was written test-first. This ensures that bugs are kept to a minimum, and that we're able to incrementally and effectively build the program to withstand a variety of different possibilities. I found feature tests very helpful in checking or finding any cases that I may have missed in my unit tests.
 
-In code review we'll be hoping to see:
+I used rubocop to check my code for correct formatting (what an awesome tool!), as well as simplecov to check my code's coverage (also a God-send!). Simplecov really helped get me back on track when my code started becoming overly-complicated or unsound.
```
tatsiana-makers, line 56: Really cool to read about an overview of the code and the responsibilities of each of the classes before diving into the Ruby code 🙂

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 20
+  attr_reader :capacity, :weather, :plane, :planes_at_airport
+
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new, planes_at_airport = [])
```
tatsiana-makers, line 8: Nice use of dependency injection!

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  def full?
+    @planes_at_airport.count >= @capacity
+  end
+
+  def prevent_landing
+    raise "Airport is currently full. You cannot land." if full?
+    raise "Weather is currently stormy. You cannot land until it is sunny." if weather.stormy?
+  end
+
+  def prevent_take_off
```
tatsiana-makers, line 50: Can this method be made private? Are there other methods that could be made private?

lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+require_relative 'airport'
+
+class Plane
```
tatsiana-makers, line 3: Here are some questions which could help you fill in this class:
- Can a plane land twice in a row without taking off first?
- Can a plane take off twice in a row without landing first?

spec/airport_spec.rb
```diff
@@ -0,0 +1,105 @@
+require 'airport'
+
+describe Airport do
+
+  before do
+    @airport = Airport.new
+    @plane = Plane.new
```
tatsiana-makers, line 7: In order to isolate unit tests we shouldn't rely on any other actual classes, so in this case, using a plane double would be preferable

spec/airport_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+    allow(@airport.weather).to receive(:weather_state) { 1..9 }
```
tatsiana-makers, line 8: This line does stub the `weather_state` method for the Weather instance, but it'd be preferable to isolate the Airport class unit tests from the Weather class by passing in a weather instance double into `Airport.new`. This way, if we ever add more methods into the Weather class in the future, there's no risk we forget to stub those new methods which would mean that our unit tests wouldn't be isolated anymore.

This pattern is exemplified in the[ dependency injection practical](https://github.com/makersacademy/skills-workshops/blob/master/practicals/object_oriented_design/dependency_injection.md)

spec/airport_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      expect(@airport.land(@plane)).to eq @plane.to_s + " landing now"
+    end
+
+    it 'when :take_off is called' do
+      expect(@airport.take_off(@plane)).to eq @plane
+    end
+
+    it 'when :plane_at_airport is called' do
+      allow(@airport).to receive(:plane_at_airport?) { @plane }
+    end
```
tatsiana-makers, line 23: It's not clear to me what this is testing

spec/weather_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'weather'
+
+describe Weather do
+  it 'is expected to set weather_state to a random number between 1 and 10' do
+    expect(subject.weather_state).to be_between(1, 10)
+  end
+
+  describe 'stormy? is expected to' do
+    it 'return true if weather_state == 10' do
+      allow(subject).to receive(:weather_state) { 10 }
```
tatsiana-makers, line 10: We could also use `srand` here 🙂

tatsiana-makers (commented):
Thanks for asking for a review! Let me know if you have any questions about the feedback I've left 🙂

PR #2029, opened 2021-02-07
My Airport Challenge Solution

AJ8GH:
My attempt at building an Air Traffic Control system using Ruby :)


AJ8GH:
Meets needs of all user stories. challenge completed
Given more time I'd like to:
- Add an optional hash of arguments to Plane's initialize method, allowing it to also be initialized in an airport
- Add verification to the capacity= writer method so that only an integer can be given
- Add a password feature so that only authorised users can change the capacity

lib/exceptions.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+end
+
+# when planes request to land or take off during stormy weather
+class WeatherError < StandardError
+  attr_reader :msg
+
+  def initialize
+    @msg = 'Request Denied: Extreme weather'
+    super(msg)
```
a-nawrocka, line 27: why did you use super ?

lib/air_traffic_control.rb
```diff
@@ -0,0 +1,40 @@
+# prevents landing and take off when guard conditions are not met
+module AirTrafficControl
```
a-nawrocka, line 2: What is a module and why did you used it?

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    @capacity = capacity
+    @weather = weather_station # memoizes weather object
+    @planes = []
+  end
+
+  def full?
+    planes.count >= capacity
+  end
+
+  def weather_station
```
a-nawrocka, line 32: nice SRP :)

lib/weather.rb
```diff
@@ -0,0 +1,12 @@
+# class for simulating random weather conditions
```
a-nawrocka, line 1: I really like atmospheric pressure method

lib/plane.rb
```diff
@@ -0,0 +1,40 @@
+# returns flight status, lands, takes off, sends requests to air traffic control
```
a-nawrocka, line 1: nice inclusion of edge case

lib/exceptions.rb
AJ8GH: super makes a method call to the parent class - in this case the `StandardError` class
I know this is how to define custom errors with messages, but I don't know the exact why and how of what's going on behind the scenes so I'll need to do some research and find out!

lib/air_traffic_control.rb
AJ8GH: module is a collection of methods or constants which can be mixed in  to with `include` 
I knew that I didn't want ATC methods defined within Airport, but I'm still not sure if it's the best approach here, I was also considering an ATC class and have that as an attribute of airport.

lib/airport.rb
AJ8GH: thanks :)

README.md
```diff
@@ -13,77 +15,237 @@ Airport Challenge
...
+- `Feature testing` - Automated feature testing using `rspec`
+- `Edge cases` and `corner cases` tested, ensuring system reliability in extreme and inconsistent states
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+### Design
+- `Domain Modelling` - Translating `user stories` into unit tests, into code
+- `OOP` - Applying `OOD` and `SOLID` principles building a program with maximum flexibility and tolerance for change
```
eoinmakers, line 49: This is great - do you feel you could elaborate on which SOLID principles guided your design?

README.md
```diff
@@ -13,77 +15,237 @@ Airport Challenge
...
+  - Add an optional hash of arguments to Plane's initialize method, allowing it to also be initialized in an airport
+  - Add verification to the `capacity=` writer method so that only an integer can be given
+  - Add a password feature so that only authorised users can change the capacity
+- This exercise has made me more determined to master object oriented design, in particular SOLID principles.
+- My long term goal is to be confident in designing complex programs from scratch, using loosely coupled objects with minimal dependencies and well designed public interfaces.
 
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first.
 
-* **Submit a pull request early.**
+## Mission Accomplished
```
eoinmakers, line 249: Stellar readme in general, well done. Really great context in here about your implementation and how its used.

lib/air_traffic_control.rb
eoinmakers: What made you choose a module for the AirTrafficControl methods, out of interest? It's well written but I'm not sure why you chose it over a class.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  private
+
+  attr_reader :planes, :weather
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @weather = weather_station # memoizes weather object
+    @planes = []
+  end
```
eoinmakers, line 26: Passing the weather in here as a default argument would allow for dependency injection to fully isolate your classes from one another in your tests

lib/exceptions.rb
```diff
@@ -0,0 +1,49 @@
...
+
+# when airborne planes are instructed to take off
+class TakeOffError < StandardError
+  attr_reader :msg
+
+  def initialize
+    @msg = 'Plane already airborne'
+    super(msg)
+  end
+end
```
eoinmakers, line 49: These custom errors are great! Really tidy.

spec/helper_methods.rb
```diff
@@ -0,0 +1,47 @@
...
+end
+
+def make_it_rain
+  make_it(:stormy)
+end
+
+def make_it(weather)
+  allow_any_instance_of(Airport).to receive(:weather_forcast) { weather }
+end
+
```
eoinmakers, line 30: It may be more easily readable to do test set-up in before(:each) blocks in a context for each group of tests instead

spec/feature_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      expect(lhr.contain?(dream_liner)).to be true
+      expect(lhr.contain?(boeing)).to be false
+    end
+  end
+
+  context 'raising errors' do
+    context 'when stormy' do
+      before {
+        let_there_be_sun; boeing.land(lhr);
+        dream_liner.land(lax); make_it_rain
```
eoinmakers, line 64: It's best practice to write one expression per line

lib/air_traffic_control.rb
AJ8GH: I can't remember a strong reason why I chose to use a module, which tells me it's probably not correct! 

I remember thinking when I'd finished that a class may be more appropriate

PR #2030, opened 2021-02-07
Submitting Airport Challenge (Unfinished)

Laura-Keen:


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def change_capacity(number)
+    @capacity = number
+    return "Capacity changed to #{number}"
+  end
+
+  def land(plane)
+    hangar << plane
+  end
```
kjjwhitlock, line 20: love how simple and tight these methods are

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def empty?
+    return "Airport is empty" if hangar.empty? 
+    return "Airport contains plane" if hangar.length >= 1
+  end
+
+  def safety_check
+    return "Space available, please land here" if hangar.length < capacity
+    return "Capacity is full, do not land" if hangar.length >= capacity
```
kjjwhitlock, line 33: this is also really clear.

lib/planes.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
+  
+end
```
kjjwhitlock, line 3: Would land be a message for the plane or the airport?

PR #2033, opened 2021-02-07
Ian_McNicholas Airport_Challenge

ianmcnicholas:
Makers Week 1 end-of-week challenge

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,89 +1,87 @@
-Airport Challenge
-=================
+# IAN MCNICHOLAS AIRPORT CHALLENGE #
 
+This is my submission for the airport challenge that was set at the end of week 1 of the Makers Coding Bootcamp.
+
+This project was approached in the same manner as the Boris Bikes project, using the 'airport' similar to the 'docking station', as the centrepiece of the work.
+
+The project has been built and tested using 'rspec' to meet the following user requirements:
 ```
-        ______
-        _\____\___
-=  = ==(____MA____)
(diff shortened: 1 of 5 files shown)


ianmcnicholas:
All user requests have been fulfiled.
Incomplete - Doubles.  I have used doubles to refer to the Plane class but not for the Airport class.  
This is what I'd like to improve upon so please look to give me feedback on this.

charlierdm:
- The README is clear and concise and gives the user clear instructions of how to use the program. I like the inclusion of 10% chance of storm. 

- Airport = Aiport.new is in every test. To remove these lines of code you could use a let - I’m not 100% sure of these myself but it adds a default value for you to type.

- * describe Factorial do
*   let(:calculator) { Factorial.new }
* 
*   it "finds the factorial of 5" do  
*     expect(calculator.factorial_of(5)).to eq(120)
*   end

That’s an example, I think that calculator us now Factorial.new for every test, saving you creating it each time. This could tidy up the tests.

- I really like the way you have described the weather in your methods and described it as a 10% chance of storm, I think that meets the user requirements very well.

- @planes.slice!(@planes.index(plane)).   I think this syntax could be replaced by @planes.pop.   (again, not 100% sure on this)

- The it "confirms if an plane is in the airport"  test is repeated twice.

- Why did you do the doubling where you did it, out of interest? Would it make more sense to create a double at the beginning of test file and then use it throughout all the tests. Genuine question to myself as well.

PR #2035, opened 2021-02-07
Solution for week 1 weekend challenge.

kerisic:


kerisic:
All user stories included in the code: plane landing, plane taking off, prevent landing when airport is full, can override airport's default capacity, prevent landing and taking off when the weather is stormy.

Haven't been able to make methods/variables private that users shouldn't be able to view, as tests required their visibility.

Would like to improve on the above, and possibly look into more possible edge cases.

AdamP2000:
READEME file very clear
Commit's labelled clearly and code and tests updated efficiently to satisfy new user stories.
Possible refactoring of forecast method to make it clearer what the method is doing.

PR #2036, opened 2021-02-08
Solution

peter-james-allen:


peter-james-allen:
Complete
- Covers all user stories, and some edge cases
- Passes rubocop
- 100% test coverage

Unfinished
- README file has not been updated

To Improve
- Look at applying SRP and DRY
- Write an RSpec feature test that lands and takes off a number of planes

spec/weather_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'weather'
+
+describe Weather do
+
```
jgumoes, line 4: if you use rand to get a random number, you can use srand to test it. srand sets a starting point for rand, so if you call srand(1234) and call rand 5 times, close the script, open it again, call srand(1234) again, when you call rand 5 times you get the same 5 random numbers as before

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
```
jgumoes, line 1: I like that weather has its own class

PR #2038, opened 2021-02-08
first draft of weekend challenge

spaek14:
please refer to the readmesolpaek.md file

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -18,6 +18,10 @@ end
   def land(plane)
     return false if Airport.new.full?
     return true if !Airport.new.full?
+
+    #return false if Weather.new.stormy?
+    #return true if !Weather.new.stormy?
+    
   end
 
   def take_off(plane)
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -4,17 +4,19 @@ attr_reader :capacity, :planes
 
 def initialize
 
-  @capacity = 9
+  @capacity = $def_capacity
   @planes = 10
 
 end
 
   def full?
     return true if @capacity <= @planes
+    return false if @capacity > @planes
   end
 
   def land(plane)
     return false if Airport.new.full?
+    return true if !Airport.new.full?
   end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,11 +1,12 @@
 class Airport
 
-attr_reader :capacity, :planes
+attr_reader :capacity, :planes, :weather
 
 def initialize
 
   @capacity = $def_capacity
   @planes = 10
+  @weather = $def_weather
 
 end
 
@@ -17,6 +18,11 @@ end
   def land(plane)
     return false if Airport.new.full?
     return true if !Airport.new.full?
+  end
+
+  def take_off(plane)
+    return true if !Weather.new.stormy?
+    return false if Weather.new.stormy?
   end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,16 @@
+class Airport
+
+attr_reader :capacity, :planes
+
+def initialize
+
+  @capacity = 9
+  @planes = 10
+
+end
+
+  def full?
+    return true if @capacity <= @planes
+  end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -12,5 +12,9 @@ end
   def full?
     return true if @capacity <= @planes
   end
+  # 
+  # def land(plane)
+  #   @planes += 1
+  # end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -12,9 +12,9 @@ end
   def full?
     return true if @capacity <= @planes
   end
-  # 
-  # def land(plane)
-  #   @planes += 1
-  # end
+
+  def land(plane)
+    return false if Airport.new.full?
+  end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,13 +2,13 @@ class Airport
 
 attr_reader :capacity, :planes, :weather
 
-def initialize
+  def initialize
 
-  @capacity = $def_capacity
-  @planes = 10
-  @weather = $def_weather
+    @capacity = $def_capacity
+    @planes = 10
+    @weather = $def_weather
 
-end
+  end
 
   def full?
     return true if @capacity <= @planes
@@ -21,7 +21,6 @@ end
 
     #return false if Weather.new.stormy?
     #return true if !Weather.new.stormy?
-    
   end
 
   def take_off(plane)
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,0 +1,17 @@
+require 'airport'
+
+class Weather
+
+attr_reader :weather
+
+def initialize
+
+  @weather = $def_weather
+
+end
+
+  def stormy?
+    return true if (@weather == @stormy)
+  end
+
+end
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -4,11 +4,9 @@ class Weather
 
 attr_reader :weather
 
-def initialize
-
-  @weather = $def_weather
-
-end
+  def initialize
+    @weather = $def_weather
+  end
 
   def stormy?
     return true if (@weather == @stormy)
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -4,14 +4,19 @@ require 'plane'
 describe Airport do
 
   it "it is at or past capacity" do
+    $def_capacity = 10
     expect(subject.full?).to eq(true)
+  end
+  it "it is not at capacity" do
+    $def_capacity = 12
+    expect(subject.full?).to eq(false)
   end
 
   it "doesn't land" do
     expect(Airport.new.land(:plane)).to eq(false) if Airport.new.full?
   end
   it "lands" do
-    expect(Airport.new.land(:plane)).to eq(true) if (Airport.new.full? == false)
+    expect(Airport.new.land(:plane)).to eq(true) if (!Airport.new.full?)
   end
 
 end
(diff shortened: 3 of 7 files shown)


spaek14:
user prompts 1, 2, 3, 4 complete
user prompt 5 not quite complete
user prompt 6 complete
user prompt 7 not complete


I am aware that the code is not finished, and that it has not addressed all user prompts. I am also aware that the tests only pass for given values, and do not take a user input - the tests simply pass for what I have offered them. I have struggled a fair amount with this challenge, so I am expecting there to be a lot of errors with my code.

One other thing is that I could perhaps have named my methods/variables a bit more clearly. Hopefully this doesnt make my code too difficult to read or understand. 

I think that the layout of the code should be mostly alright, and that the tests are clear enough in terms of readability - I just hope they work!

I would like to improve on a lot by the end of this. Firstly, I would like to understand more about the things that I did not manage to complete. Secondly, I would like to improve on the functionality of my code, as its only real merit in my eyes is that it passes the tests that I have prepared for it (which makes sense, given that I wrote the tests too). That being said, it does have 100% coverage, but once again that could be an inaccurate representation of the quality of code.

I do not yet feel confident with TDD, and think that this code review could highlight a lot of gaps in my knowledge!

spec/airport_spec.rb
```diff
@@ -21,7 +21,8 @@
   end
 
   it "default capacity can be changed" do
-    expect($def_capacity).to be_a(Integer)
+    $def_capacity = @capacity
+    expect(@capacity).to eq($def_capacity)
```
spaek14, line 25: I don't know about this

lib/weather.rb
```diff
@@ -4,11 +4,9 @@ class Weather
...
 attr_reader :weather
 
-def initialize
-
-  @weather = $def_weather
-
-end
+  def initialize
+    @weather = $def_weather
+  end
```
spaek14, line 9: This commit is more readable than the previous one.

lib/airport.rb
```diff
@@ -18,6 +18,10 @@ def full?
   def land(plane)
     return false if Airport.new.full?
     return true if !Airport.new.full?
+
+    #return false if Weather.new.stormy?
+    #return true if !Weather.new.stormy?
+
```
spaek14: I need to make this part of the land method work, but am unsure how.

spec/airport_spec.rb
```diff
@@ -32,4 +32,12 @@
     expect(Airport.new.take_off(:plane)).to eq(true) if (!Weather.new.stormy?)
   end
 
+  # it "doesn't land if stormy" do
```
spaek14, line 36: This test only passes sometimes, and despite spending an embarrassingly long amount of time to make this work, I could not do so.

spec/airport_spec.rb
```diff
@@ -16,15 +16,15 @@
   it "doesn't land" do
     expect(Airport.new.land(:plane)).to eq(false) if Airport.new.full?
   end
-  it "lands" do
+  it "does land" do
```
spaek14, line 19: This is more readable than the last commit.

spec/airport_spec.rb
```diff
@@ -16,15 +16,15 @@
...
     expect(Airport.new.land(:plane)).to eq(true) if (!Airport.new.full?)
   end
 
   it "default capacity can be changed" do
     expect($def_capacity).to be_a(Integer)
   end
 
-  it "does not take off if stormy" do
+  it "doesn't take off if stormy" do
```
spaek14, line 28: This is also more readable.

lib/airport.rb
```diff
@@ -1,11 +1,12 @@
 class Airport
 
-attr_reader :capacity, :planes
+attr_reader :capacity, :planes, :weather
 
 def initialize
 
   @capacity = $def_capacity
   @planes = 10
+  @weather = $def_weather
```
spaek14: I think this is along the right lines, but I am not entirely sure.

lib/airport.rb
```diff
@@ -19,4 +20,9 @@ def land(plane)
     return true if !Airport.new.full?
   end
 
+  def take_off(plane)
+    return true if !Weather.new.stormy?
```
spaek14, line 27: I think I should be using unless here instead of if.

spec/airport_spec.rb
```diff
@@ -23,4 +24,12 @@
     expect($def_capacity).to be_a(Integer)
   end
 
+  it "does not take off if stormy" do
+    expect(Airport.new.take_off(:plane)).to eq(false) if Weather.new.stormy?
+  end
+
+  it "does take off if not stormy" do
+    expect(Airport.new.take_off(:plane)).to eq(true) if (!Weather.new.stormy?)
```
spaek14, line 33: Once again I think I should be using unless instead of if.

lib/weather.rb
```diff
@@ -0,0 +1,17 @@
+require 'airport'
+
+class Weather
+
+attr_reader :weather
+
+def initialize
+
+  @weather = $def_weather
+
```
spaek14: The layout is inconsistent, but I think that the class has been set up correctly, and although the weather is not random, it interacts with the other files within the directory as intended.

spec/airport_spec.rb
```diff
@@ -19,4 +19,8 @@
     expect(Airport.new.land(:plane)).to eq(true) if (!Airport.new.full?)
   end
 
+  it "default capacity can be changed" do
+    expect($def_capacity).to be_a(Integer)
```
spaek14: This is clearly not the one.

spec/weather_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'weather'
+require 'airport'
+
+describe Weather do
+
+  it "weather is stormy?" do
+    $def_weather = @stormy
+    expect(subject.stormy?).to eq(true)
```
spaek14, line 8: I think this is correct

lib/airport.rb
```diff
@@ -4,17 +4,19 @@ class Airport
 
 def initialize
 
-  @capacity = 9
+  @capacity = $def_capacity
```
spaek14: using a global variable here instead of a set value is good, as it means that the code relies less on already defined constants.

lib/airport.rb
```diff
@@ -4,17 +4,19 @@ class Airport
...
   @planes = 10
 
 end
 
   def full?
     return true if @capacity <= @planes
+    return false if @capacity > @planes
```
spaek14, line 15: I think this makes logical sense, and it seems to be reflected well in the user tests.

spec/airport_spec.rb
```diff
@@ -4,14 +4,19 @@
 describe Airport do
 
   it "it is at or past capacity" do
+    $def_capacity = 10
     expect(subject.full?).to eq(true)
   end
+  it "it is not at capacity" do
+    $def_capacity = 12
```
spaek14, line 12: I think these are acceptable tests, and that defining the global variable within the test is the correct thing to do.

spec/airport_spec.rb
```diff
@@ -4,14 +4,19 @@
...
+    expect(subject.full?).to eq(false)
+  end
 
   it "doesn't land" do
     expect(Airport.new.land(:plane)).to eq(false) if Airport.new.full?
   end
   it "lands" do
-    expect(Airport.new.land(:plane)).to eq(true) if (Airport.new.full? == false)
+    expect(Airport.new.land(:plane)).to eq(true) if (!Airport.new.full?)
```
spaek14, line 20: I think unless is probably better to use here than if.

lib/airport.rb
```diff
@@ -12,9 +12,9 @@ def initialize
   def full?
     return true if @capacity <= @planes
   end
-  # 
-  # def land(plane)
-  #   @planes += 1
-  # end
+
+  def land(plane)
+    return false if Airport.new.full?
```
spaek14, line 19: I don't like that the plane count does not interact with the code. This being said, I think that the definition for landing the plane fits the user prompt.

spec/airport_spec.rb
```diff
@@ -6,9 +6,12 @@
   it "it is at or past capacity" do
     expect(subject.full?).to eq(true)
   end
-  # 
-  # it "lands" do
-  #   expect(@planes += 1).to eq(true)
-  # end
+
+  it "doesn't land" do
+    expect(Airport.new.land(:plane)).to eq(false) if Airport.new.full?
```
spaek14, line 17: This seems correct

spec/airport_spec.rb
```diff
@@ -6,9 +6,12 @@
...
+  end
+  it "lands" do
+    expect(Airport.new.land(:plane)).to eq(true) if (Airport.new.full? == false)
```
spaek14: probably should use unless here.

spec/plane_spec.rb
```diff
@@ -3,17 +3,15 @@
 
 describe Plane do
 
-  let(:plane) {Plane.new}
-  let(:airport) {Airport.new}
+  let(:plane) { Plane.new }
+  let(:airport) { Airport.new }
```
spaek14, line 7: This is more similar to conventions

lib/airport.rb
eds-101: it reads fine

lib/airport.rb
eds-101: I think the logic is good, i get what you're trying to do. Just needs a lil tweak

spec/airport_spec.rb
eds-101: yes the logic is spot on. Here it would be good to use variables. I think airport.new, used twice, will give you two different airports which may be hard to test.

lib/airport.rb
eds-101: I generally get your logic, even though you said it's not working. One example to improve could be describing your plane capacity. You wrote '@planes = 10, this could be more readable with '@plane_capacity = 10.

I think it was good idea to initialise weather for the airport. This is something I didn't think of personally

eds-101 (commented):
Good logic used and some nice ideas, such as initializing airport with weather.
I think it could improve by using some descriptive variables. Feels like there are a lot of concepts on the page.

PR #2039, opened 2021-02-08
AdamP2000 Airport Challenge

AdamP2000:


AdamP2000:
User stories that involve weather are not completed yet.

kerisic:
1. The code is easy to understand and methods are transparent in terms of what it achieves.
2. Perhaps group tests including edge cases according to each method under each describe, e.g., #land, #take_off. Can see that there is one for #land and initialization already so that's great. This may improve readability.
3. Consider the test case "removes planes that take off" - is there another way of testing the removal of a plane?

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
+require_relative 'plane'
+class Airport
+  DEFAULT_CAPACITY = 33
+  attr_accessor :capacity
```
eoinmakers, line 4: I would put a new line under your attribute accessor for neatness.

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail 'Airport full' if full?
+
```
eoinmakers, line 12: I would remove this empty line and keep both fail checks together

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+    fail 'Already in airport' if @planes.include? plane
+
+    @planes << plane
+    @planes
```
eoinmakers, line 16: 15 and 16 should be returning the same thing (the array)

spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'airport'
+describe Airport do
+  it "lands a plane" do
+    airport = Airport.new
```
eoinmakers, line 4: It's a bit neater to refer to subject for Airport.new rather than creating a new instance in each test

spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      Airport::DEFAULT_CAPACITY.times do
+        subject.land Plane.new
+      end
+      expect { subject.land Plane.new }.to raise_error 'Airport full'
+    end
+  end
+  it 'has a default capacity' do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+  describe 'initialization' do
```
eoinmakers, line 19: As each of the following tests create a new instance of plane, you could use a before(:each) block to define plane = Plane.new

eoinmakers (commented):
I think this is great, nice work. :)

eoinmakers (commented):
I think this is great, nice work. :)

PR #2040, opened 2021-02-08
Airport Challenge – TBC!

hannahdesmond:


spec/airport_spec.rb
```diff
@@ -2,6 +2,16 @@
 
 describe Airport do
   it 'allows planes to land' do
-    expect(subject).to respond_to(:land)
+    expect(subject).to respond_to(:land).with(1).argument
   end
+  it 'lands at the airport' do
+    plane = subject.new_plane
+    expect(subject.land(plane)).to eq(plane)
```
sandyMax974: I don't see what this line is testing - can we discussed what you had in mind?

lib/airport.rb
```diff
@@ -1,4 +1,12 @@
...
+
 class Airport
-  def land
+  attr_reader :plane
+  def new_plane
+    Plane.new
+  end
+
+  def land(plane)
+    @plane = plane
```
sandyMax974: I think I can see what you wanted to do, store plane when landed but I'm not sure - maybe the naming here could be different

spec/airport_spec.rb
```diff
@@ -2,6 +2,16 @@
...
+  end
+  it 'checks the plane is in the airport' do
+    plane = subject.new_plane
+    subject.land(plane)
+    expect(subject.plane).to eq(plane)
```
sandyMax974: I think this test is repeating what you did above - what was the intention behind it? Maybe you were trying to test something slightly different?

spec/airport_spec.rb
```diff
@@ -13,5 +13,11 @@
     subject.land(plane)
     expect(subject.plane).to eq(plane)
   end
+  it 'stores planes at the airport' do
+    expect(subject.parking).to be_kind_of(Array)
+  end
+  it 'allows planes to take off' do
+    expect(subject).to respond_to(:take_off).with(1).argument
+  end
```
sandyMax974: Very clear on what these do.

lib/airport.rb
```diff
@@ -1,17 +1,13 @@
 require_relative './plane.rb'
 
 class Airport
-  attr_reader :plane
-  def new_plane
-    Plane.new
+  attr_reader :plane, :hangar
+  def hangar
+    @hangar = []
   end
```
sandyMax974: This is clearer - I think I can see where you changed your approach of the process here

lib/airport.rb
```diff
@@ -1,15 +1,27 @@
...
 
 class Airport
-  attr_reader :plane, :hangar
+  attr_reader :plane
+  def initialize
+    @@hangar = []
+  end
   def hangar
-    @hangar = []
+    @@hangar
```
sandyMax974: If you have @@hangar defined as an attr_reader, do you still need to create a getter method?

lib/airport.rb
```diff
@@ -1,15 +1,27 @@
...
+
+# Test creates a new instance of plane.
+# The plane lands and should be pushed to the hangar array.
+# The hangar array should contain the plane.
+
+# I am having problems storing an array in a class.
+# The plane instance is going into an array in the airport
+# instance. I want a separate array.
+
+# Do I need a class variable...?
```
sandyMax974: We can discuss this - I would have think that what you did worked
```Airport.new.hangar << plane```

File: spec/airport_spec.rb [hunk 8]
```diff
@@ -8,8 +8,12 @@
   it 'stores planes at the airport' do
     expect(subject.hangar).to be_an_instance_of(Array)
   end
+  it 'adds the landed plane to the hangar' do
+    plane = Plane.new
+    subject.land(plane)
+    expect(subject.hangar).to eq([plane])
```
sandyMax974: That a good test easy to read as well

PR #2041, opened 2021-02-08
Incomplete - Rorie_Airport_challenge

rhc07:


lib/airport.rb
```diff
@@ -0,0 +1,61 @@
...
+class Airport
+
+attr_reader :plane
+attr_accessor :capacity
+attr_reader :forecast
+DEFAULT_CAPACITY = 20
+
+	def initialize(capacity=DEFAULT_CAPACITY)
+ 		@hangar_report = []	
+ 		@capacity = DEFAULT_CAPACITY
```
iainaitken, line 13: Possible bug/improvement - @capacity should equal capacity (not DEFAULT_CAPACITY) so that it can be overridden on instantiation

lib/airport.rb
```diff
@@ -0,0 +1,61 @@
...
+ 		@weather = [:sunny, :stormy]
+	 	@forecast = @weather.sample
+	end
+
```
iainaitken, line 17: I like the methods here - really clean, easy to read and understand

lib/airport.rb
```diff
@@ -0,0 +1,61 @@
+require "./lib/Plane"
+
+
+class Airport
+
+attr_reader :plane
+attr_accessor :capacity
+attr_reader :forecast
```
iainaitken, line 8: Possible improvement - can put attributes on a single line:

attr_reader :plane, :forecast

Also apparently it's a bad idea to use attr_accessor as users can then directly access and modify an instance variable. Better to define a method that the user can invoke that will change the value

spec/airport_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+ end
+ it "raises an error when weather is stormy" do
+ 	airport = Airport.new
+ 	plane = Plane.new
+ 	airport.land(plane)
+ 	airport.forecast == :stormy
+ 	expect {airport.take_off(plane)}.to raise_error 'Weather is stormy, cannot take-off.'
+ end 
+end
+end
```
iainaitken, line 76: Your tests are well organised - the groups of tests related to specific methods make them easy to read through

PR #2042, opened 2021-02-08
Up to user story 4

kjjwhitlock:
Have updated README file with information about how I found this challenge - looking forward to learning more so I can finish it!

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,41 @@
+require 'airports'
+require 'planes'
+
+describe Airport do
+  describe '#default_capacity' do
+    it 'has a default capacity' do
+      airport = Airport.new
+      expect(airport.capacity).to eq Airport::DEFAULT_CAPACITY
+    end
+  end
+   describe '#taking_off' do
+     it 'expects airport to respond to taking_off method' do
+      airport = Airport.new
+      expect(airport).to respond_to(:taking_off)
+     end
+    it 'checks that the number of planes in the airport has decreased by one' do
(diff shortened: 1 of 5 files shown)


spec/plane_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'airports'
+require 'planes'
+
+describe Plane do
+  describe '#plane_landing' do
+    it 'responds to the method plane_landing' do
+      airport = Airport.new
+      plane = Plane.new
```
Laura-Keen, line 8: Could use let syntax to save having to do airport = Airport.new and plane = Plane.new each time

spec/airport_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+describe Airport do
+  describe '#default_capacity' do
+    it 'has a default capacity' do
+      airport = Airport.new
+      expect(airport.capacity).to eq Airport::DEFAULT_CAPACITY
+    end
+  end
+   describe '#taking_off' do
+     it 'expects airport to respond to taking_off method' do
```
Laura-Keen, line 12: Could use Rspec one line syntax here to clean up code

spec/plane_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      expect(plane).to respond_to(:land)
+    end
+  end
+  describe '#plane_takeoff' do
+    it 'responds to the method take_off' do
```
Laura-Keen, line 13: Could use one line syntax here to clean up code

PR #2043, opened 2021-02-08
Joanna Airport Challenge

JLBrigham:


lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+  attr_reader :hanger, :capacity
+
+  def initialize(capacity = 2, hanger = [])
+    @capacity = capacity
+    @hanger = []
+  end
+
+  def land(plane)
+    if self.weather_check == "Stormy"
+      raise "Weather too bad to land"
```
eoinmakers, line 13: You can rewrite this in one line to: 

raise "Weather too bad to land" if weather_check == "Stormy". 

Additionally, the 'self' is not needed here as weather_check should be accessible without

eoinmakers: The same can be applied to each raise "..." below this

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    @hanger << plane
+    end
+  end
+
+  def take_off(plane)
+    if self.weather_check == "Stormy"
+      raise "Weather too bad to take off"
+    elsif @hanger.include?(plane)
+    @hanger.delete(plane)
+    @hanger
```
eoinmakers, line 28: 27 and 28 should return the same thing (the hangar array), making 28 unnecessary

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    else
+    @hanger << plane
+    end
+  end
+
+  def take_off(plane)
+    if self.weather_check == "Stormy"
+      raise "Weather too bad to take off"
+    elsif @hanger.include?(plane)
+    @hanger.delete(plane)
```
eoinmakers, line 27: You should indent your code from line 27 onwards to show it falls inside the elsif

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    @hanger
+  else
+    raise "Plane not available to take off"
+  end
+  end
+
+  def full?
+    if @hanger.length >= @capacity
+      true
+    end
```
eoinmakers, line 37: can be rewritten by just writing @hanger.length >= @capacity inside your method and it will return true or false accordingly

lib/plane.rb
```diff
@@ -0,0 +1,12 @@
+require_relative 'airport'
+
+
+class Plane
+
+  def land(airport_name)
+  end
+
+  def take_off(airport_name)
+  end
```
eoinmakers, line 10: Unused code that can be removed

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'airport'
+
+describe Airport do
+
+it 'should store a plane once it lands' do
+  airport = Airport.new(10)
+  expect(airport.land("plane")).to be_instance_of Array
```
eoinmakers, line 7: You could write another test to check that that array now contains a plane, whether by checking the length of the array or that it contains that specific instance

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+ end
+
+ it 'should not let a plane take off that is not in the hanger' do
+  airport = Airport.new(2)
+  plane = Plane.new
+  expect{airport.take_off("plane")}.to raise_error("Plane not available to take off")
+ end
+
+it 'should not let a plane land if it has already landed' do
+  airport = Airport.new
```
eoinmakers, line 40: You can refer to subject instead of writing airport = Airport.new, it will make a new instance for you automatically for each test.

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def initialize(capacity = 2, hanger = [])
+    @capacity = capacity
+    @hanger = []
+  end
+
+  def land(plane)
+    if self.weather_check == "Stormy"
+      raise "Weather too bad to land"
+    elsif self.full?
```
eoinmakers, line 14: no reference to self needed here

eoinmakers (commented):
Nice work - happy to chat some more about the default capacity if you're blocked on getting it working. :)

PR #2045, opened 2021-02-08

MJCXII:


lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    @airport_space.push(plane)
+  end
+
+  def takeoff(plane)
+    takeoff_errors(plane)
+    @airport_space.delete(plane)
+    @airport_space
+  end
+
+  private
```
jshorns, line 23: I really like the organisation and layout of this error methods and it's very good SRP - particularly grouping the error methods themselves into landing errors and take_off errors. I wish I'd done this!

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+module Weather
```
jshorns, line 1: I haven't read the resource on modules so this is probably my own ignorance talking, but I'm not sure why this needs to be a separate thing? Would it not be simpler just to have this as a method in the airport class?

README.md
```diff
@@ -35,29 +35,29 @@ Task
 We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
```
jshorns, line 35: This is not really a code comment, but your commit history is pretty sparse and the commit messages are not very descriptive. I found it very easy to get my head around your code because we basically had the same approach, but for someone coming to it blind, it might be a bit harder to get the hang of your thinking.

PR #2046, opened 2021-02-08
Airport challenge

NicolasBermejo:
I have gone through all 6 user stories, but the last 2 concerning the weather change I got stuck at. The Ruby code works but I did not manage to get the RSPEC to work.

### DIFF
--- a/lib/planes.rb
+++ b/lib/planes.rb
@@ -2,7 +2,6 @@ class Plane
 
 end
 
-
 class Airport
   attr_reader :plane, :capacity, :weather_today
   MAX_AIRPORT_CAPACITY = 20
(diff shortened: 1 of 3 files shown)


lib/planes.rb
```diff
@@ -0,0 +1,35 @@
...
+  WEATHER = [1, 2]
+  def initialize(capacity = MAX_AIRPORT_CAPACITY, weather = WEATHER.sample)
+    @weather_today = weather
+    @capacity = capacity
+    @plane = []
+  end
+
+  def land(plane)
+    if @weather_today == 1
+      fail 'Airport is full' if @plane.length >= @capacity
```
penguat, line 17: Love the use of the ternary operator here! Great job :)

penguat (commented):
Looks good to me! :)

lib/planes.rb
NicolasBermejo: thank you

PR #2049, opened 2021-03-07
completed user story and edge case requirements for Airport challenge

loushark:


lib/airport.rb
```diff
@@ -0,0 +1,54 @@
+require_relative 'planes'
+require_relative 'weather'
+
+class Airport
+
+  attr_reader :apron, :capacity, :weather
```
louballantyne, line 6: In the case that there are 10 planes at the airport and somebody tries to change the capacity to 1, are they able to do this, or does the code stop them?

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def initialize(capacity = 1)
+    @apron = []
+    @capacity = capacity
+    @weather = Weather.new
+  end
+
+  def land(plane)
+    fail "A plane can't land, there's no room!" if apron.length == capacity
```
louballantyne, line 15: Maybe a test of the airport being at capacity could be a method?

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+    fail "This plane is already at another airport" if
+    plane.status == "landed"
+
+    apron << plane.name
+    plane.status = "landed"
+    "#{plane.name} has landed!"
+    # change_status(plane)
+  end
+
+  def take_off(plane)
```
louballantyne, line 31: Can planes take off during bad weather?

spec/weather_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'weather'
+require 'airport'
+
+describe Weather do
+  it "returns a random number" do
+    expect(subject).to receive(:rand)
+    subject.weather_report
+  end
```
louballantyne, line 8: Could you also test whether the weather reports itself as stormy or sunny?

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    colin = Plane.new('colin')
+    expect { heathrow.land(colin) }.to raise_error("A plane can't land, there's no room!")
+  end
+
+  it "allows airports to override the default capacity" do
+    expect(heathrow.capacity).to eq(1)
+    gatwick = Airport.new(5)
+    expect(gatwick.capacity).to eq(5)
+  end
+
```
louballantyne, line 48: Can a user change the capacity of the airport while the program is running?

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    return fail "Plane is not at this airport" unless apron.include?(plane.name)
+
```
louballantyne, line 34: This is a really nice, catch-all way to check if the plane is valid for take off.

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+class Weather
+  def initialize
+  end
+
+  def weather_report
+    # where every number is sunny except for 6 which is a rare occasion of stormy
+    rand(1..9) == 6 ? "stormy" : "sunny"
+  end
```
louballantyne, line 8: Great use of the ternary operator here!

README.md
```diff
@@ -60,30 +43,24 @@ To ensure safety
...
+To feature test in irb:  
+> require './lib/airport'  
+  
+example instances to test:  
+> heathrow = Aiport.new (default capacity is 1)  
+> gatwick = Airport.new(10) (default capacity overriden to be 10)  
+> bertie = Plane.new('Bertie')  
+> colin = Plane.new('Colin')  
+> heathrow.land(bertie) => "Bertie has landed!" (provided weather is sunny)  
+> heathrow.land(colin) => "A plane can't land, there's no room!" (heathrow only has a capacity of 1, bertie is already on the apron)
```
louballantyne, line 65: I love these examples - it's really helpful to understand how the code works.

louballantyne (commented):
You named your methods and variables clearly, making it easy to understand the code. I think it could possibly be refactored slightly so that things are extracted to additional methods. Well done on the full test coverage!

lib/airport.rb
loushark: The capacity is set when the airport instance is created so the capacity shouldn't be able to change. Useful to write an extra test to ensure this cannot be changed though

spec/airport_spec.rb
loushark: Shouldn't be able to. The capacity is only readable and set during Airport.new.

PR #2050, opened 2021-03-07
Airport Challenge

mstrivens:
60% Complete. 100% coverage

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,47 @@
+class Airport
+  attr_reader :capacity
+  attr_accessor :plane
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+  end
+
+  def land(plane)
+    raise "Airport full" unless @plane.nil?
+
+    raise "Plane is at another airport" if plane.grounded
+
+    plane.airborne = false
+    @plane = plane
+  end
+
+  def take_off(plane)
+
+    raise "Plane is already in the air" if plane.flying
+
+    plane.airborne = true
+    @plane = plane
+  end
+end
+
+class Plane
+  attr_accessor :airborne, :airport_id
+
+  def initialize(airborne = true)
+    @airborne = airborne
+    # @airport_id = airport_id
+
+  end
+
+  def flying
+    @airborne == true
+  end
+
+  def grounded
+    @airborne == false
+  end
+end
+
+class Weather
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,10 +1,11 @@
 class Airport
-  attr_reader :capacity
+  attr_reader :capacity, :weather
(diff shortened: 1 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,62 @@
...
+end
+
+class Weather
+  attr_reader :report
+  def initialize(report = self.forecast)
+    @report = report
+  end
+  def forecast
+    number = rand(1..10)
+    if number.odd?
```
inbarvernia, line 56: Since the brief said that stormy conditions are rare, maybe instead of setting all odd numbers to stormy, could just pick one or two numbers that would set to stormy and leave the rest as sunny? But also, I know you haven't finished this part of the challenge (I haven't even started it myself), so possibly this is something you were planning to change later anyway.

stories.txt
```diff
@@ -0,0 +1,87 @@
...
+objects -------------- messages
+
+Controller
+Passengers          get_to_destination
+*Plane               to_land
+*Airport             land_plane
+
+This is running a land method within airport that
+returns a plane which is passed into it as an arguement
+
```
inbarvernia, line 27: I like how you explain the methods and how they relate to the user stories, very clearly laid out.

lib/airport.rb
```diff
@@ -0,0 +1,62 @@
+class Airport
+  attr_reader :capacity, :weather
+
+  DEFAULT_CAPACITY = 20
```
inbarvernia, line 4: Great idea to define it as a constant - I like the clarity.

lib/airport.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
+    @capacity = capacity
+    @weather = weather
+    @planes = []
+  end
+
+  def land(plane)
+    raise "Airport full" unless @planes.length <= DEFAULT_CAPACITY
```
inbarvernia, line 13: Small logic disagreement: I think it should be `@planes.length < @capacity`: `@capacity` rather than `DEFAULT_CAPACITY` because if the user provides a custom capacity it should be based on this rather than the default one; and `<` rather than `<=` because if the airport is already at max capacity (`=`), it should already raise an error, not only when it's over the max.

lib/airport.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    raise "Plane is at another airport" if plane.grounded
+
+    plane.airborne = false
```
inbarvernia, line 17: Good feature here to have the method also affect the plane object - should have done this in mine too!

inbarvernia (commented):
I tried to be a bit nitpicky, so I've probably given a lot more comments than is really necessary. Overall very clear and efficient code! My main suggestions are to group example tests relating to each method together so that they're easier to find (possibly by using nested `describe #method` within `describe Class`), and also using `context` when you are testing the same method under different circumstances.

PR #2053, opened 2021-03-07
Airport challenge completed!

taran314:


lib/airport.rb
```diff
@@ -0,0 +1,17 @@
+require 'plane'
+
+class Airport
+
+  attr_reader :name, :hangar, :capacity
+
+  def initialize(name = 'heathrow', hangar = [], capacity = 1)
```
StuBehan, line 7: Default initialization is good for numerical variables and stuff, but maybe for names not so much? You may end up with multiples of the same name. Depends on how `name` is used elsewhere tbh.

lib/airport.rb
```diff
@@ -0,0 +1,17 @@
...
+    @name = name
+    @hangar = hangar
+    @capacity = capacity
+  end
+
+  def full?
+    @hangar.length == @capacity
+  end
```
StuBehan, line 15: I think methods with question marks that return true or false are neat.

lib/plane.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize(name = 'concord', location = 'air')
+    @name = name
+    @location = location
+  end
+
+  def land(airport)
+    fail 'plane already landed' if @location != 'air'
+    fail 'weather is stormy' if stormy?
+    fail 'hangar at capacity' if airport.full?
+
```
StuBehan, line 16: I didn't know `rubocop` only wanted a tailing newline for `fail`/`raise` statements after the last one

spec/airport_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { Plane.new('concord', 'air') }
+  let(:airport) { Airport.new('heathrow', [], 1) }
```
StuBehan, line 5: You shouldn't need to specify these arguments for plane and airport as they are your defaults in the class files

spec/plane_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'plane'
+
+describe Plane do
+  # default values shown for clairty.
+  let(:plane) { Plane.new('concord', 'air') }
+  let(:airport) { Airport.new('heathrow', [], 1) }
```
StuBehan, line 6: same again - arguments are defaulted in class files

spec/plane_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      airport.capacity.times { airport.hangar << Plane.new }
+      expect { plane.land(airport) }.to raise_error('hangar at capacity')
+    end
+
+    it 'fails if custom set capacity hangar is reached' do
+      large_airport = Airport.new('large', [], 20)
+      large_airport.capacity.times { large_airport.hangar << Plane.new }
+      expect { plane.land(large_airport) }.to raise_error('hangar at capacity')
+    end
+  end
```
StuBehan, line 45: sweet

StuBehan (commented):
I like it, interesting to see how someone else went about the task.

PR #2054, opened 2021-03-07
airport_challenge

dmgreaves:


lib/airport.rb
```diff
@@ -2,20 +2,21 @@
...
+  def initialize(capacity = 2)
     @capacity = capacity
     @landed_planes = []
   end
+
   def take_off(plane)
     @plane = plane
     raise "Plane has already taken off" unless @landed_planes.include?(plane)
     if weather == "Stormy"
       return "Weather is stormy; plane is unable to take off"
```
Emmapr123, line 16: I think you can raise this as an error as well, instead of it returning the message! that way the code is more consistent with the other error above it

lib/airport.rb
```diff
@@ -2,20 +2,21 @@
...
-    @plane
+  attr_reader :landed_planes, :capacity
+
+  def initialize(capacity = 2)
     @capacity = capacity
     @landed_planes = []
   end
+
   def take_off(plane)
     @plane = plane
```
Emmapr123, line 13: You can put @plane = plane in the intializer as well, to prevent any repetition

lib/airport.rb
```diff
@@ -2,20 +2,21 @@
...
     raise "Plane has already taken off" unless @landed_planes.include?(plane)
     if weather == "Stormy"
       return "Weather is stormy; plane is unable to take off"
-    else
-    @landed_planes.delete(plane)
-    "#{plane} has taken off"
+      else
+        @landed_planes.delete(plane)
+        "#{plane} has taken off"
```
Emmapr123, line 19: I love that you are doing #{plane} ! it is great to see which plane is taking of, very clever!

lib/airport.rb
```diff
@@ -25,10 +26,10 @@ def land(plane)
     raise "Airport full; unable to land plane" if @landed_planes.length == @capacity
```
Emmapr123, line 26: In terms of SRP you could have made a function 'full?' that would take care of checking it's capacity. Like we did in Boris Bikes! same goes for 'inlcude' in the take_off function.

spec/airport_spec.rb
```diff
@@ -6,31 +6,30 @@
...
-        subject.land(plane)
-        plane2 = Plane.new
-        allow(subject).to receive(:weather) { "Sunny" }
-        subject.land(plane2)
-        plane3 = Plane.new
-        allow(subject).to receive(:weather) { "Sunny" }
-        expect {subject.land(plane3)}.to raise_error("Airport full; unable to land plane")
-   end
+  it 'raises an error if airport is full' do
+    plane = Plane.new
```
Emmapr123, line 16: you can put plane = Plane.new at the start of your spec file, like this :
let(:plane) { Plane.new } 
that way, any time you mention plane, your code will see Plane.new, unless you change plane into anything else! saves you a lot of lines and time

spec/airport_spec.rb
```diff
@@ -6,31 +6,30 @@
...
+    allow(subject).to receive(:weather) { "Sunny" }
+    subject.land(plane)
+    plane2 = Plane.new
+    allow(subject).to receive(:weather) { "Sunny" }
```
Emmapr123, line 20: You can specify this at the start as well!

spec/airport_spec.rb
```diff
@@ -6,31 +6,30 @@
...
+    subject.land(plane2)
```
Emmapr123, line 21: and you dont have to land all planes seperately, you could say 3.times { subject.land(plane) }, that will land 3 seperate planes for you!

spec/airport_spec.rb
```diff
@@ -6,31 +6,30 @@
...
-   end
+  it 'raises an error if plane has already landed' do
+    plane = Plane.new
+    allow(subject).to receive(:weather) { "Sunny" }
+    subject.land(plane)
+    expect { subject.land(plane) }.to raise_error("Plane has already landed at this airport")
+  end
 
   it 'confirms plane has taken off' do
     plane = Plane.new
```
Emmapr123, line 35: this is not actually testing anything

spec/plane_spec.rb
```diff
@@ -1,14 +1,4 @@
 require 'airport'
 
 describe Plane do
-
```
Emmapr123: now your plane class isnt doing anything, it might be nice for the plane to know if it is flying or if it has landed too? You could maybe show it what the airport knows, so they agree on the location of the plane?

Emmapr123 (commented):
Hope my review makes sence! I did not want to repeat the same comments when they applied on more than one piece of code, so if you are going to use it, do keep in mind that they might apply on more than one thing! see you this afternoon

PR #2055, opened 2021-03-07
initial submission commit

Yates101:


lib/airports.rb
```diff
@@ -0,0 +1,30 @@
+require './lib/planes.rb'
+
+class Airport
+  attr_reader :capacity, :planes, :weather
+
+  def initialize(airport_code = '***', capacity = 10)
```
crotchetycrow, line 6: Perhaps setting a constant rather than relying on the use of magic numbers might make this code easier to maintain.

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    expect { subject.land_plane(Plane.new) }.to raise_error("Airport full!")
+  end
+
+  it 'allows the default capacity to be overridden at Airport initialize' do
+    heathrow = Airport.new("LHR", 15)
+    expect(heathrow.capacity).to eq 15
+  end
+
+  it 'prevents takeoff when weather is bad' do
```
crotchetycrow, line 61: Writing a separate Weather class which a specific method that has your random weather generator in might solve this problem (but it didn't for me so I'm not sure that's good advice)

spec/plane_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'planes.rb'
+
+describe Plane do
```
crotchetycrow, line 3: I like how you've recorded the planes this way. I wish I had thought of that.

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
+require 'airports.rb'
+
+describe Airport do
+
```
crotchetycrow, line 4: Might be my ignorance but you don't seem to have created doubles for testing using let(:plane) { Plane.new } or something. But I don't know about that 100% 🤷

crotchetycrow (commented):
Please accept my humbly written review from someone who doesn't know what they are talking about

spec/airport_spec.rb
crotchetycrow: i meant let (:plane) { double :plane } ***

PR #2056, opened 2021-03-07
Airport challenge (partial completion)

crotchetycrow:


lib/weather.rb
```diff
@@ -0,0 +1,10 @@
+class Weather
```
Yates101, line 1: Nice separation of classes

lib/planes.rb
```diff
@@ -0,0 +1,11 @@
+class Plane
+    attr_accessor :landed
+
+  def initialize(status=true)
+    @landed = status
+  end
+
+  def landed?
```
Yates101, line 8: I think this is superfluous code as attr_accessor, in part is shorthand for an almost identical method

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
+require './lib/airport'
+require './lib/planes'
+require './lib/weather'
+
+describe Airport do
+  let(:plane) { double :plane, :landed= => false, landed?: false }
```
Yates101, line 6: Ahh is this how you do this? Good job **less review than note to self

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  let(:weather) { double :weather, :sunny= => true, sunny?: true }
+  let(:airport) { Airport.new }
+
+  describe "instruct_to_land" do
+    it { is_expected.to respond_to(:instruct_to_land).with(1).argument }
```
Yates101, line 11: Noice shorthand

crotchetycrow:
Struggled with user story 4 and 5. Two failed tests with one test passing intermittently (too stormy to land).

I felt the code could be read and understood, but struggled with the landing portion of the code and the use of attr_reader and attr_accessor (must improve understanding and application better). 

Spec looks to be correct but I threw up errors so who knows.

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    it "prevents landing if airport is full" do
+      allow(plane).to receive(:instruct_to_land).and_return(false)
```
Yates101, line 13: Confused by this line. Tag to discuss later

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+      it 'will not allow a plane to take off when it is stormy' do
+       allow(weather).to receive(:sunny?).and_return false
+       expect { airport.instruct_to_take_off(plane) }.to raise_error("It is too stormy to fly")
+     end
+  end
+
+  describe "instruct_to_take_off" do
+    it { is_expected.to respond_to(:instruct_to_take_off).with(1).argument }
+    it 'will not allow a plane to land when it is stormy' do
```
Yates101, line 25: In the wrong nested describe?

spec/planes_spec.rb
```diff
@@ -0,0 +1,12 @@
+require './lib/planes'
+
+describe Plane do
+  let(:plane) { Plane.new }
+  it {is_expected.to respond_to :landed?}
+
+  it {is_expected.to respond_to :landed}
+
```
Yates101, line 8: It does all this, however, I don't think you've channeled that in to your code under test. Planes are still able to land if they are already landed.

Have you separated :landed and :landed? to avoid one having setter functions?

Yates101 (commented):
Overall, great man. Good separation of classes, shorthand, default values and weighted randomised weather.

Or at least, I think so - for whatever that is worth.

Some code doesn't seem to be implemented in a way that a feature test perhaps might expose? 
- Again, my feedback is essentially guesswork.
But if it was implemented, looks like it would do the job.

PR #2057, opened 2021-03-07
Creating pull request for partial solution to aiport challenge

fg24davies:
updated with README.md with how I found the challenge

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,56 +1,37 @@
-Airport Challenge
-=================
+# Airport Challenge - Weekend 1
+I found this really challenging and have submitted a partial solution. The parts that felt similar to Boris Bike Challenge were doable but I did struggle beyond that.
 
-```
-        ______
-        _\____\___
-=  = ==(____MA____)
-          \_____\___________________,-~~~~~~~`-.._
-          /     o o o o o o o o o o o o o o o o  |\_
-          `~-.__       __..----..__                  )
-                `---~~\___________/------------`````
-                =  ===(_________)
-
-```
-
-Instructions
----------
-
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Steps
--------
-
-1. Fork this repo, and clone to your local machine
-2. Run the command `gem install bundle` (if you don't have bundle already)
-3. When the installation completes, run `bundle`
-4. Complete the following task:
-
-Task
+I have looked at some previous years solutions but not copied from one single one. My code is more of an update of my boris bike
+bike challenge so I didn't manage to integrate a plane landing AT a specific airport.
+I only had airports being passed a plane object but I think you would need to pass a plane an airport object?
 -----
-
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1,56 +1,37 @@
-Airport Challenge
-=================
+# Airport Challenge - Weekend 1
+I found this really challenging and have submitted a partial solution. The parts that felt similar to Boris Bike Challenge were doable but I did struggle beyond that.
```
eatruman, line 2: I felt the same way!

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  it 'can take off plane' do
+    expect(subject).to respond_to(:takeoff)
+  end
+  
+  context 'weather is good' do 
+    before do
+      allow(subject).to receive(:stormy?) { false }
+    end
+    it 'raises an error when full' do 
+      subject.land(Plane.new)
```
eatruman, line 21: If you created a plane underneath your describe Airport do, you could then use that  in all of these tests rather than using (Plane.new) each time.

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      expect { Airport::DEFAULT_CAPACITY.times do
+        subject.land(Plane.new) 
+      end 
+      }.to raise_error 'Airport is full: plane cannot land'
+    end 
+
+    it 'raises an error when empty' do
+      expect { subject.takeoff(Plane.new) }.to raise_error 'Airport is empty: no planes to takeoff'
```
eatruman, line 29: Do you need a line before your expect to make sure your planes_at_airport array is empty before testing this? Otherwise you're just expecting calling takeoff with a new plane to return that the airport is empty when it might not be?

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  it 'has a default capacity' do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+
+  
+
+  context 'weather is stormy' do
+    before do
```
eatruman, line 40: This was a good idea rather than setting Stormy? to true in each test, it makes your code cleaner and easy to read/understand.

lib/plane.rb
```diff
@@ -0,0 +1,19 @@
+class Plane
+  attr_accessor :in_air
+  def initialize
+    @in_air = true 
+  end
+
+  def in_air?
+    @in_air
+  end
```
eatruman, line 9: This was a good way to solve the location of the plane problem.

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+module WeatherConditions 
+  def stormy?
+    rand(10) > 8
+  end
```
eatruman, line 4: This is really clear and concise!

eatruman (commented):
I think overall you did a great job! There were parts of the challenge missing but you completed it in order so you covered all of the user stories and started on the edge cases. I think it was clear what you were testing and your tests were easy to understand. Your code is also clear and easy to read. Going forward, you could improve by creating your own user stories for the edge cases to practice writing tests that are different from ones we have already done (which is what I will be working on too!) Good job Flora!

PR #2060, opened 2021-03-08
Airport Challenge Complete

willspencer16:


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+require 'plane'
```
TataSher, line 1: irb says it can't load 'plane' file. Maybe './lib/plane' needed

PR #2063, opened 2021-03-08
airport challenge

Emmapr123:


spec/weather_spec.rb
```diff
@@ -8,7 +8,7 @@
     end
   end
 
-  describe 'chances_of_rain' do
+  describe 'chance_of_rain' do
```
dmgreaves, line 11: I like how you're testing for a few weather options and expecting your code to return a value in a range, a different approach to using stubs.

lib/airport.rb
```diff
@@ -2,24 +2,26 @@
 require 'weather'
```
dmgreaves, line 2: Good use of separate classes for separate activities: SRP.

dmgreaves (commented):
Really thorough code covering all the project bases and code rubric requirements. Also very creative additions, I can see you had fun creating this!

PR #2065, opened 2021-03-15
airport challenge

renae-stl:


lib/airport.rb
```diff
@@ -0,0 +1,22 @@
...
+      weather.stormy?? (fail 'unsafe conditions') : (@plane.land)
+    
+    end
+
+    def safe_takeoff(plane)
+      weather = Weather.new
+      weather.stormy?? (fail 'unsafe conditions') : (@plane.takeoff)
+    end
+
+end
```
RobertMcCann, line 22: I like the methods you have defined, they are very easy to read and I can already picture the result. I also like how you have used the one line if statement, I always forget that it's a neatened way of writing it :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    expect(boeing747).to respond_to(:takeoff)
+  end
+    
+  it 'checks weather is safe for takeoff' do
+    weather = Weather.new
+    weather.stormy? == true
+    expect(controller.safe_takeoff(boeing747)).to raise_error('unsafe conditions')
+  end
+
+  end
```
RobertMcCann, line 28: I think having separate spec files for each of your classes may help, say if there's an issue with Weather or Plane then you'll be able to just go to that spec file and fix it. :)

RobertMcCann (commented):
Hi Renae, I liked your submission for the airport challenge. I hope my feedback helps :)

lib/airport.rb
```diff
@@ -0,0 +1,22 @@
+require_relative 'plane'
+require_relative 'weather' 
+
+class Controller
```
leoht, line 4: Maybe this class should be named `Airport` (as the file is named `airport.rb` indeed) - also because `Controller` is quite vague and doesn't really express what the program is about

lib/airport.rb
```diff
@@ -0,0 +1,22 @@
...
+    def plane (plane)
+        # This should be edited so that each plane called is 
+        # is one of 20 that are registered at the aiport
+        @plane = Plane.new
```
leoht, line 8: If you want to have more than one planes, maybe you could use an array of planes, like this:
```
@planes = []
```

lib/plane.rb
```diff
@@ -0,0 +1,10 @@
+class Plane
+  def land(plane)
```
leoht, line 2: Since the `Airport` class will take the responsibility of "knowing" when a plane is landed or not, those methods could be removed from `Plane` (the Plane class is not doing anything in the end)

lib/weather.rb
```diff
@@ -0,0 +1,6 @@
+class Weather
+    def stormy?
+        stormy = [true, false].sample
```
leoht, line 3: Nice!

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+describe 'Controller' do 
+  controller = Controller.new
+  boeing747 = controller.plane(boeing747)
+  it 'instructs a plane to land' do
+    expect(boeing747).to respond_to(:land)
+  end
+  
+  it 'checks weather is safe for landing' do
+    weather = Weather.new
+    weather.stormy? == true
```
leoht, line 14: I assume you wanted to tell RSpec: "Hey, given the weather is stormy, my safe landing should raise the "unsafe conditions" error "

The line `weather.stormy? == true` doesn't work in this case, since it's just comparing `weather.stormy?` to `true` and forgetting this result (it doesn't affect our test at all)

In this case, you can "stub" the `Weather` class, which means "forcing" the  `Weather.stormy?` method to return a very specific value, instead of being random. This can be done with the following syntax:

```ruby
# whenever weather.stormy? will be called, true will always be returned (no more random!)

allow(weather).to receive(:stormy?).and_return(true)
```

leoht (commented):
Good separation of concerns, there is still some bits that can be removed from the `Plane` class. Also, there is a way to tell `weather.stormy?` to return a specific result (which I think you tried to do) - you need to use the `allow(...)` method, I've explained more below. Good job! You can revisit this challenge later, understand the motive for some of the feedback, and try making some changes to address it :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  describe '#full_airport' do
+    it 'prevents landing if airport is at capacity' do
+      expect{subject.safe_landing}.to raise_error{'airport is full'}
+    end
+end
+
+
+end
```
Maelvis, line 37: Really easy-to-read tests with very little repetition :)

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+  def initialize(capacity)
+    @capacity = capacity
+    @hangar = [Plane.new]
+
+  end
+
+  def safe_landing
+    @plane = Plane.new
+    weather = Weather.new
+    if weather.stormy == true
```
Maelvis, line 16: You could perhaps collapse this if-statement into fewer lines for better readability?

lib/weather.rb
```diff
@@ -0,0 +1,6 @@
+class Weather
+    def stormy
+        stormy = [true, false].sample
```
Maelvis, line 3: I love this method you used here !

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
+require_relative 'plane'
+require_relative 'weather' 
+
+class Airport
+  attr_reader :capacity, :hangar
+
+  def initialize(capacity)
+    @capacity = capacity
```
Maelvis, line 8: Capacity might have been easier to change if put in a constant instead?

PR #2066, opened 2021-03-15
Airport challenge example commits

leoht:
- There is one test that doesn't work, not sure why, any help appreciated

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,24 @@
+class Airport
+  def initialize
+    @planes = []
+  end
+
+  def land(p)
+    fail 'Airport full' if @planes.length > 10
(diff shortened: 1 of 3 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,24 @@
+class Airport
+  def initialize
+    @planes = []
+  end
+
+  def land(p)
+    fail 'Airport full' if @planes.length > 10
```
leoht, line 7: Test

lib/plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
```
justinhylton, line 1: No plane instance variable initialised.

PR #2067, opened 2021-03-15
Airport challenge

Maelvis:
User stories implemented.
Tests passing.
More refactoring needed.


lib/airport.rb
```diff
@@ -0,0 +1,49 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :plane, :weather
+  attr_accessor :capacity, :planes
```
leoht, line 7: You could simplify this class by assuming that a plane that is "landed" or not depends actually of its presence in the `@planes` array of the Airport class. It gets simpler because then you just say that a plane that's not landed is simply not in the array, and you free yourself for having to manage the state of the Plane in parallel of the state of `Airport.planes` (and also you wouldn't need to check `if plane.flying == false` etc)

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+    "Plane #{plane} has taken off."
+
+  end
+
+  def airport_full
+    @planes.count >= capacity
+  end
+
+  def stormy_weather(weather)
+    return true if weather.stormy?
```
leoht, line 43: You can simplify it by just writing:
```ruby
def stormy_weather(weather)
    weather.stormy?
end
```
since the `stormy_weather` method is just returning the same boolean value of `weather.stormy?`

lib/weather.rb
```diff
@@ -0,0 +1,18 @@
+class Weather
+
+  WEATHER_PREDICTIONS = [:stormy, :sunny, :sunny, :stormy, :sunny, :sunny, :sunny]
```
leoht, line 3: I assume you wanted to have more "range" for randomness here - that could be done by just using `rand` without arguments (which returns a number between 0 and 1 I think) and do something like:
```ruby
def stormy?
  rand > 0.5 # 50% of chance to be stormy
end
```
In this way, the "stormy or sunny?" is abstracted but boils down to just checking some random number

leoht: You also don't need the `:weather_predictions` accessor since you don't want this attribute to be accessible to external classes (other classes should *only* be able to call `stormy?` - this should be the only "interface" with the Weather class)

spec/airport_spec.rb
```diff
@@ -0,0 +1,98 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
```
leoht, line 4: A lot of tests, great! :) 
You might change some of them after modifying your code, but most of them should be OK, and you'll be able to refactor safely your code thanks to them

lib/weather.rb
leoht: Actually I've just noticed you probably wanted to be able to pass more weather predictions value into the class, in this case that code works, but I would probably still remove the accessor to `weather_predictions` (you wouldn't want an external class to access it and just "chose" for itself which weather prediction it wants!)

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+    fail 'The weather is too stormy' if stormy_weather(weather) == true
+    fail 'Plane already flying' if plane.flying == true
+
+    @planes.delete(plane)
+    plane_state_change(plane)
+    "Plane #{plane} has taken off."
+
+  end
+
+  def airport_full
```
leoht, line 38: In Ruby it's idiomatic to end the name of methods that return a boolean with `?`, which would make it `airport_full?`

leoht (commented):
Nice work! I had just a few comments about some details, mainly:
- you could simplify everything by having a simpler model: planes don't know if they are flying or not, but the only matter is their presence in the `planes` of the Airport
- the `Weather` class can be simplified a bit
But the devil is in the details - otherwise it's really good :)

lib/airport.rb
Maelvis: Thank you, Leo!

PR #2068, opened 2021-03-15
Airport challenge commits

Leghnider:
Have not implemented the weather user stories


lib/airport.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  DEFAULT = 40
```
leoht, line 5: Maybe that could be renamed to `DEFAULT_CAPACITY` or something more specific (`DEFAULT` doesn't tell about the default of "what")

lib/plane.rb
```diff
@@ -0,0 +1,19 @@
+require_relative 'airport'
+
+class Plane
```
leoht, line 3: You could simplify your code if you assume that a plane is "flying" or not, just depending on its presence or absence in the `planes` array of Airport. This way, you don't need to have the `Plane` objects managing a "flying" state (and you also remove the cyclic dependency between Airport is Plane, which would be good - usually you want one class to know about the other, but not the other way around, because it makes things more complicated to reason about and to change later)

spec/airport_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      it 'a plane is permitted to take off' do
+        airport = Airport.new
+        airport.receive(plane)
+        expect{ airport.allow_take_off(plane) }.to change {airport.plane_count}.by(-1)
+      end
+      
+    end
+
+    context 'air traffic control based on capacity' do
```
leoht, line 21: Nice! Usually `context` is use with a `before` block to setup the "context" in which the following tests will run. In this example, the "context" could be "airport is at max capacity" and so your `before` block could look like this:
```ruby
context 'air traffic control based on capacity' do
  before do
    40.times{ airport.receive(plane) } # airport is now full
  end

  # ...
```

You can probably look up `context` and `before` in the RSpec docs to see some examples of this

leoht (commented):
Nice! The code could be slightly simplified, mainly to put the responsibility of knowing if a plane is "flying" or not only on the Airport class, not on both classes - otherwise good job!

PR #2069, opened 2021-03-15
code-review

al01010:
Created a weather generator but have not gotten further on the 5th User Story .

### DIFF
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,0 +1,12 @@
+class Weather
+  attr_reader :weather
+
+  def weather
+    num = rand
+    if num > 0.8 
+      return "stormy"
+    else
+      return "sunny"
+    end
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -43,8 +43,13 @@ describe Airport do
     plane2 = Plane.new
     heathrow = Airport.new
     heathrow.land(plane1)
+    expect { heathrow.land(plane2) }.to raise_error('Airport full cannot land') 
+  end
 
-    expect{ heathrow.land(plane2) }.to raise_error('Airport full cannot land') 
+  it 'airport capacity can be overridden' do
+    heathrow = Airport.new
+    heathrow.capacity = 20
+    expect(heathrow.capacity).to eq(20)
   end
 
 end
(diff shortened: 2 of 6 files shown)


lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+class Weather
+  attr_reader :weather
```
leoht, line 2: You don't need the `attr_reader` here, since you already have the method `weather` (using the attr_reader would be useful if you had some instance variable `@weather` you wanted to make accessible, for example)

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  it 'airport capacity can be overridden' do
+    heathrow = Airport.new
+    heathrow.capacity = 20
+    expect(heathrow.capacity).to eq(20)
+  end
+
+  # it 'raises error if stormy and stops take off' do
+  #   # arrange
+  #   today = Weather.new
```
leoht, line 57: This was almost working, but the only thing missing is that your stubbed method is set only on this object `today`, however `Airport` *never uses this specific object* because it creates a new object `Weather` by itself! There are a few options to tackle this:
- use `allow_any_instance_of(Weather)` instead, to force *all* instances of `Weather` to have this method stub
- inject `today` inside `Airport.new`, which would require the Airport's `initialize` method to receive it as an argument, instead of having it created in airport (that's called dependency injection, because we "inject" the dependency into Airport, instead of having Airport create it on its own, so we have more control of this dependency)

leoht (commented):
Nice work! You were almost there with the random weather, I've explained more details below on the bits that were missing

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  it 'raises error if you try and land plane at an airport at max capacity' do
+    plane1 = Plane.new
+    plane2 = Plane.new
+    heathrow = Airport.new
+    heathrow.land(plane1)
+    expect { heathrow.land(plane2) }.to raise_error('Airport full cannot land') 
+  end
+
+  it 'airport capacity can be overridden' do
+    heathrow = Airport.new
```
leoht, line 50: Maybe this should test the `capacity` passed in the constructor instead (which brings the question, do we want to be able to change the airport capacity in the middle of things, once it's created? isn't it dangerous? food for thoughts! :) )

PR #2070, opened 2021-03-15
Airport - Unfinished

alangoodier:
I got an error while completing the 'airport is full' - I was then unsure how to progress and ultimately wanted to start a fresh.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,25 +1,33 @@
-class Airport 
+require_relative 'plane.rb'
 
-  def initialize
-    @planes = (1)
-    @capacity = (10)
+class Airport
+
+MAX_CAPACITY = 10
+
+  attr_reader :plane
+  attr_accessor :planes, :capacity
+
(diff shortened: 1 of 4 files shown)


richLSb (commented):
Great work! Can read it well!

lib/airport.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def full
+    @plane <= @capacity
+    if true
+      return
+    end
+  end
+end
+
+airport = Airport.new
```
leoht, line 25: You probably don't need this line (a class file should just contain the class, and not other code outside of the class - that would be done only in other files, or in tests)

leoht (commented):
Nice! Looks like the project was on good tracks, there are things that could be improved:
- the logic of whether the airport is full or not, and conditional landing (more info below)
- some things could be cleaned/removed
Try to digest this feedback when you got some time, and to think about how you would improve/fix the code following it :)

PR #2071, opened 2021-03-15
Airport challenge 

CallumAhernCode:


KBA99:
Your code is really clear and the tests correspond to your code. 
It's obvious you've taken your time to think about the tests and how they can be used to prevent bugs in the future. 
Really good work and I'll alter my work to be clearer and more precise with the tests like yours.

spec/airport_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'airport'
+
+describe Airport do
+  subject(:airport) { described_class.new(20) }
```
leoht, line 4: It's probably good to put any "magic number" (= a value in the code to which the meaning is not clearly apparent) inside a variable, maybe something like: 
```ruby
let(:capacity) { 20 }
```

lib/airport.rb
```diff
@@ -0,0 +1,12 @@
...
+      @capacity = capacity
+      @planes = []
+    end
+  
+    def land(plane)
+      raise 'Cannot land plane: airport full' if @planes.length >= @capacity
+      @planes << plane
+    end
+  
+    def take_off(plane)
```
leoht, line 12: Looks like the end of the file is missing?

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,19 @@
+    airport = Airport.new(20)
```
leoht, line 1: Weirdly enough looks like this file is truncated at the beginning as well, maybe some bad copy/paste? 🤔

leoht (commented):
Nice job, looks like the code is on good tracks - however be careful of some truncated classes/methods (every `def`, `class`, `describe` and `it` must have a matching `end` indented on the same level) - the next step would be to think about how a "stormy" weather can randomly be generated for the Airport. Feel free to come back to this challenge and try to reflect on the feedback and how you could address it by making changes!

PR #2073, opened 2021-03-15
Pull request for Justin Hylton to code review 15/03/21

Dominic-Taylor-Dev:
Pull request for Justin Hylton to code review 15/03/21.

I believe this covers all the user stories, but please do let me know if you spot issues...!


Phali18 (commented):
Appreciation:

Good READMe file - very detailed and it explains the though process 
Good thorough test suite 

Feedback:

A lot of repetition in the test suite (particularly when instantiating objects) , it would be beneficial to practice DRY where possible.

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+require_relative './weather' #refactor to have multi on one line
+
+# OUTSTANDING - refactor esp tests.
+
+# The main class of the entire programme. The other
+# important interaction is with the Plane class ('require'd)
+class Airport
+  DEFAULT_CAPACITY = 10
+
+  attr_accessor :airport_attributes
```
leoht, line 13: It might be better to either use `attr_reader` or wrap accessing the different attributes inside methods, otherwise the caller of the class can wildly update any attribute without the class having any control on this:
```ruby
a = Airport.new("Gatwick", 100)
a.airport_attributes = {capacity: 10, name: "Gatwick", foo: "bar", some_other_thing: 2}
```
```

File: lib/plane.rb [hunk 2]
```diff
@@ -0,0 +1,18 @@
...
+  attr_accessor :plane_attributes
+
+  def initialize(plane_id)
+    @plane_attributes = {
+      plane_id: plane_id,
+      airport_id: 'IN TRANSIT/AIR'
+    }
+  end
+
+  def check_location
```
leoht, line 15: Looks like this method is returning `@plane_attributes[:airport_id]`, so maybe it could be renamed, or maybe the "check" part of it it's implemented? (When reading `check_****`, it could imply some kind of condition checks inside the method)

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
+  def weather_now
+    %w[sunny sunny sunny sunny stormy].sample
```
leoht, line 3: Maybe this can be simplified (if you want more control on the probabilities, you can use something like `rand` which returns a value between 0 and 1, and compare it with some control value between 0 and 1)

spec/airport_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    # airport know's when it's not full
+    it 'should return false when airport is not full' do
+      new_airport = Airport.new('AIRPORT', 100)
+      expect(new_airport.full?).to eq(false)
+    end
+  end
+
+  # not possible for a plane to land at an airport if full
+  it 'should raise an error "Airport full" if trying to land when airport is already at capacity' do
+    expect { Airport.new('AIRPORT', 0).land(Plane.new('PLANE')) }.to raise_error('Airport full')
```
leoht, line 46: I would suggest adding a test that `land`s enough planes to make the airport full (instead of making the capacity 0 from the start), and then checking that the error is indeed raised - so we can rest assured than this check would work also when the airport fills itself with planes landing (i.e the normal business use case)

spec/airport_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    new_plane = Plane.new('PLANE')
+    expect { new_airport.take_off(new_plane) }.to raise_error('Plane not present at this airport')
+  end
+
+  # not possible for a plane to take off from airport if stormy
+  it "should raise an error 'Weather is stormy' if trying to take off when airport weather is stormy" do
+    allow_any_instance_of(Weather).to receive(:weather_now).and_return('stormy')
+    new_airport = Airport.new('AIRPORT')
+    new_plane = Plane.new('PLANE')
+    new_airport.airport_attributes[:planes] << new_plane
```
leoht, line 86: Maybe it's best to use the `land` method instead of modifying directly the attributes (you could move the `allow_any_instance_of()...` after landing the plane, so the method stubbing would be effective only from that point

spec/airport_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    allow_any_instance_of(Weather).to receive(:weather_now).and_return('clear')
+    new_airport = Airport.new('AIRPORT')
+    new_plane = Plane.new('PLANE')
+    new_plane.plane_attributes[:airport_id] = new_airport.airport_attributes[:airport_id]
+    new_airport.airport_attributes[:planes] << new_plane
+    new_airport.take_off(new_plane)
+    expect(new_plane.plane_attributes[:airport_id]).to eq('IN TRANSIT/AIR')
+  end
+
+  describe '#check_location' do
```
leoht, line 121: If `check_location` is implemented by Plane, its tests should probably be moved to the `Plane` specs

leoht (commented):
Great job! I commented on some details, but it's mostly very good TDD-wise. Modelling wise, and to answer your questions in the README, there is actually a simpler solution for the airport/plane problem: it's simply to have the plane not managing its state at all (the plane doesn't know whether it's flying or not), but simply having the `Airport` class managing it (if the plane is in the @planes array, it's landed, otherwise it's flying)
It might seem too simple, but looking at the user stories, this is probably good enough and we don't need something more complex. If you wanted to have an Airport ID, it would be difficult to enforce uniqueness, unless you had some kind of "Airport registry" class that would centralize management of those Airport IDs. Btw, having the planes airport_id linked to the airport ID is the data model we'll see in Week 6, when we'll work with databases, however when working with objects, we'll usually model this relationship by just having the Airport hold the planes in an array.

```
Objects:

owning side |  relationship | other side
---------------------------------------
Airport     |  array        | Plane

```

PR #2074, opened 2021-03-15
Airport challenge

Phali18:
- Missing plane class methods and plane_spec describe block 

Users stories missing:

As an air traffic controller 

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,30 @@
+class Airport 
+
+attr_reader :capacity, :planes
+DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY) 
+    @planes = []
+    @capacity = capacity
+  end 
+
+  def land(plane)
+    raise 'Cannot land the plane, the airport is full' if @planes.count >= @capacity
+    raise 'The plane cannot land in this weather' if stormy?
+    @planes << plane
+  end 
+
+  def take_off(plane)
+  end 
+
+
+  def instruct_take_off(plane)
+    raise 'The place cannot take off in stormy weather' if stormy?
+    @planes.delete(plane)
+  end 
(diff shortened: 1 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+    @planes << plane
+  end 
+
+  def take_off(plane)
+     raise 'The place cannot take off in stormy weather' if stormy?
+     @planes.delete(plane)
+  end 
+
+  def stormy?
+    rand(1..100) > 75
```
leoht, line 23: It would be nice to move this "responsibility" in a different class, `Weather`, for example. That's because now the `Airport` class is responsible with many things: landing and taking-off planes, but also defining (even randomly) the weather. In this week you'll learn more about why it's better to have classes concerned only with a single responsibility, but it's mainly to ease changes in the code

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    it 'instruct a plane to take off from an airport and confirm this' do
+      expect(airport).to respond_to(:take_off).with(1).argument
+    end 
+
+    it 'has default capacity which can be changed as appopriate' do 
+      capacity = 10
+      expect(airport.capacity).to eq(10)
+    end 
+
+    context 'when the weather is stormy' do
```
leoht, line 38: Nice use of `context` :) you would just need a test checking planes cannot land and take-off (this would be done outside of `before`, in other `it` blocks

leoht (commented):
Nice job! I've commented on some details, also you might want to use rubocop or another code formatting tool (there's likely a plugin for VS code or Atom) to fix some indent levels here and there - otherwise TDD-wise it's really nice, feel free to revisit this challenge with the new things you'll have learned this week, and think about how to address the feedback :)

PR #2075, opened 2021-03-15
Airport challenge

RobertMcCann:
Only managed to complete the first two user stories. I ran out of time :(

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,3 +1,5 @@
 require 'airport'
 
-describe Airport
+describe Airport do
+  it { is_expected.to respond_to :land_plane }
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -2,4 +2,11 @@ require 'airport'
 
 describe Airport do
   it { is_expected.to respond_to :land_plane }
+
+  it { is_expected.to respond_to(:plane_departed?).with(1).argument }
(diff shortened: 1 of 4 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'airport'
+
+describe Airport do
+  it { is_expected.to respond_to :land_plane }
```
renae-stl, line 4: One area you did well in was writing 'clean code', I always find I write things in an overly complicated way but I can see you've made things simple where they need to be.

renae-stl (commented):
So far it looks like all you tests would pass and if you'd like me to have another look at a later stage I'd be happy to do so.

spec/airport_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  it { is_expected.to respond_to(:plane_departed?).with(1).argument }
```
leoht, line 6: Nice!

leoht (commented):
Looks like the project was on good tracks, nice work! You'll be able to revisit it later and think a bit about what the different methods need to do in order for the user story to be complete (what does the `land_plane` method need to do? What does the `plane_departed` need to check? etc) - that will also affect how the code is tested, but you'll probably see this feedback with fresh eyes at the end of this week, with the new things you'll have learned by then :)

PR #2076, opened 2021-03-15
Airport challenge - Rich

RJ21BL:
I done four user stories.
Struggled with capacity of airport, feedback required.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,27 @@
+class Airport
+  def land_plane(plane)
+    @plane = plane
+  end
+
+  def plane_landed?(plane)
+    if @plane == nil
+      return false
+
+    else
+      return true
+    end
+  end
+
+  def take_off(plane)
+    if @plane == nil
+      return true
+    end
+  end
+
+  def receive_plane(plane)
+    raise "plane cannot land when airport is full"
+    if @planes.count >= @capacity
+
+
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+      return true
+    end
+  end
+
+  def take_off(plane)
+    if @plane == nil
+      return true
+    end
+  end
+
```
alangoodier, line 20: My only piece of feedback would be to move this section below 'def initialize' to the top underneath the class heading. Then you can see these attributes first.

alangoodier (commented):
It is very heard to pick out any issues - given that you got further than me in the challenge! But your code is easy to read and it makes sense. There aren't any odd method or variable names that I can see. Overall a great effort and it's easy to read.

leoht (commented):
Nice work, it'll be good to revisit this challenge with the new things you'll have learned, at the end of the week if you want (or at any point later). I've left some comments - a good way of knowing what to write in the methods is to first try to write on a piece of paper what we want to do. For example: 

-> What does it mean for a plane to land at the airport?
-> For our program, it means setting the `@plane` attribute of the airport

So to check the plane is landed, what we're doing is:
-> checking whether the `@plane` attribute is set or not

Now try to think about what it means for a plane to take-off - what will we need to do in respect of the `@plane` attribute?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  attr_accessor :planes, :capacity
+  DEFAULT_CAPACITY = 10
```
leoht, line 22: Nice use of constants! :)

PR #2077, opened 2021-03-15
Second User story

rahmed94:
Thats all I have at the moment.

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,13 @@
+require 'airport.rb'
+describe Airport do
+  let(:plane) {described_class.new}
+  it 'allows plane to land' do
+    expect(Airport.new).to respond_to(:land).with(1).argument
+  end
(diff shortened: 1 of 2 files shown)


Leghnider:
Good job so far, as there isn't much to critique I don't have anything negative to say yet. But it looks like you're on the right track!

spec/airport_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'airport.rb'
+describe Airport do
+  let(:plane) {described_class.new}
```
leoht, line 3: You might want to be careful here, as `described_class.new` will actually create an instance of `Airport` (since it's the currently described class)

leoht (commented):
Looks like you were on the right tracks - now the next step would be to think about what needs to happen for the plane to be "landed", how would we change the state of the Airport instance? And how would this be reflected in the tests?

PR #2079, opened 2021-03-15
The first attempt at the Airport Challenge

justinhylton:
The code doesn't past any tests.

### DIFF
--- a/lib/air_traffic_controller.rb
+++ b/lib/air_traffic_controller.rb
@@ -1,0 +1,19 @@
+
+class Air_traffic_controller
+
+  # attr_accessor
+  attr_reader :first_name, :last_name, :id_number
+  # attr_writer
+
+  def initialize(options={})
+    self.first_name = options[:first_name] || 'Matthew'
+    self.last_name = options[:last_name]
+    self.id_number = (options[:id_number]).to_s
+  end
+
+  def full_name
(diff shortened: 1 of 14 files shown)


lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,19 @@
+
+class Air_traffic_controller
```
leoht, line 2: In Ruby the convention would be to name the class like this: `AirTrafficController` for the class in file `air_traffic_controller.rb` - maybe that could be changed

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
+
+class Airport
+
+  # attr_accessor
+  attr_reader :name, :terminal, :hanger, :runway
+  # attr_writer
+
+  def initialize(options={})
```
leoht, line 8: Nice use of `initialize` and default attributes 👍

leoht (commented):
That's a good start, and the code is readable 👍  however remember that TDD is about *writing the tests and the minimal code to satisfy the user stories* - there's some code here that doesn't link to the user stories of the challenge, so that wouldn't be considered strict TDD! (because we include things we don't need / are not required)

When you'll revisit this challenge, try to think about the tests and the *minimal code* you need for the first user story to be implemented

PR #2094, opened 2021-04-05
Airport challenge

ghost:


lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+  attr_reader :plane
+  def initialize
+    @planes = []
+  end
+
+  def land(plane)
+    @planes.push(plane)
+  end
+
+  def take_off(_)
```
tobyjessup, line 13: (_) is slightly misleading, it would probably be best to have a name for this, just so people can clearly see what you are passing into take_off.

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+    @planes.pop
+  end
+
+  def took_off(_)
```
tobyjessup, line 17: (_) is slightly misleading, it would probably be best to have a name for this, just so people can clearly see what you are passing into took_off.

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+  def initialize
+    @planes = []
+  end
+
+  def land(plane)
+    @planes.push(plane)
+  end
+
+  def take_off(_)
+    @planes.pop
```
tobyjessup, line 14: I believe you may be able to remove .pop here because on initialising, you are creating an empty array from @planes so maybe don't need to do it again?

PR #2102, opened 2021-04-30
Added tests for two user stories

LCzoboriek:
Added tests to the spec files for the airport_spec.rb

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,31 @@
+require './lib/plane.rb'
+
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :capacity
+  attr_reader :plane
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airports = []
+    @capacity = capacity
+  end
+
+  def land_plane(plane)
+    fail 'Airport is full!' if full?
+    
+    @airports << plane   
+  end
+
+  def take_off(plane)
+    puts "Airport currently has #{@airports}"
+    puts "#{plane} has taken off!"
+    @airports.pop
+    puts "Airport now has #{@airports}"
+  end
+
+  private
+
+  def full?
+    @airports.count >= capacity
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,34 @@
+require 'airport.rb'
+describe Airport do
+  it 'will respond to the method ' do
+    expect(subject).to respond_to :land_plane    
+  end
+  
+  describe '#docks_plane' do
+    it 'docks a plane' do
+      plane = Plane.new
+      expect(subject.land_plane(plane)).to eq([plane])
+    end
+  end
+
+  describe '#docks_plane' do
+    it 'raises an error if airport is full' do
+      subject.capacity.times { subject.land_plane Plane.new }
+      expect { subject.land_plane Plane.new }.to raise_error('Airport is full!')
+    end
+  end
+
+  describe '#take_off' do
+    it 'allows a plane to take off' do
+      plane = Plane.new
+      subject.land_plane(plane)
+      subject.take_off(plane)
+      expect(@airports.to_a).to match_array([])
+    end
+  end
+  
+end
+
+# As an air traffic controller 
+# So I can get passengers to a destination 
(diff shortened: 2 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+require './lib/plane.rb'
+
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :capacity
+  attr_reader :plane
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airports = []
```
ChescaRobertson, line 9: Naming: could @airports be changed to @planes? Is it an array of the planes currently in the airport?

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def land_plane(plane)
+    fail 'Airport is full!' if full?
+    
+    @airports << plane   
+  end
+
+  def take_off(plane)
+    puts "Airport currently has #{@airports}"
```
ChescaRobertson, line 20: I like that it puts the array, helps protect against trying to take off if no planes available

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      expect { subject.land_plane Plane.new }.to raise_error('Airport is full!')
+    end
+  end
+
+  describe '#take_off' do
+    it 'allows a plane to take off' do
+      plane = Plane.new
+      subject.land_plane(plane)
+      subject.take_off(plane)
+      expect(@airports.to_a).to match_array([])
```
ChescaRobertson, line 26: This is a good test, it appropriately checks if the plane has left the array,

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'airport.rb'
+describe Airport do
+  it 'will respond to the method ' do
+    expect(subject).to respond_to :land_plane    
+  end
+  
+  describe '#docks_plane' do
```
ChescaRobertson, line 7: Naming: consider making this 'lands_plane' as it is testing the land_plane method of the code

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+require './lib/plane.rb'
+
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :capacity
+  attr_reader :plane
+
+  def initialize(capacity = DEFAULT_CAPACITY)
```
ChescaRobertson, line 8: Good use of default capacity for future proofing

EdwardAndress: Agreed!  Nice use of a constant for the default.

lib/airport.rb
EdwardAndress: What Chesca said!

lib/airport.rb
EdwardAndress: Could you add a guard clause to prevent someone from trying to tell an absent plane to take off?

spec/airport_spec.rb
EdwardAndress: Actually, this would conventionally be `#land_plane`.  I.e. the method name, rather than a description of what it does.

EdwardAndress: For the describe block

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    it 'docks a plane' do
+      plane = Plane.new
+      expect(subject.land_plane(plane)).to eq([plane])
+    end
+  end
+
+  describe '#docks_plane' do
+    it 'raises an error if airport is full' do
+      subject.capacity.times { subject.land_plane Plane.new }
```
EdwardAndress, line 16: Nice use of capacity here!

EdwardAndress: Do you know why the curly braces are needed for this test?

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    @airports = []
+    @capacity = capacity
+  end
+
+  def land_plane(plane)
+    fail 'Airport is full!' if full?
```
EdwardAndress, line 14: Nice that you're using a method to check if the airport is full, rather than doing that logic right here in land_plane

PR #2103, opened 2021-04-30
Airport challenge, partial complete

ChescaRobertson:
I've left comments on each test relating to which bit of the user story I think it is testing

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,35 @@
+class Airport
+
+  attr_reader :plane 
+  attr_accessor :capacity
+
+  DEFAULT_CAPACITY = 25
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail "Airport is full, landing prevented" if full?
+
+    @planes << plane
+  end
+
+  def take_off(_plane)
+    @planes.pop
+  end
+
+  def confirmation
+    "Plane has left the airport"
+  end
+
+  private 
+
+  def full?
+    @planes.count >= DEFAULT_CAPACITY
+  end
+
+ 
+
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,74 @@
+require 'airport'
+
+describe Airport do
+
+  describe "#initialize" do  
+    subject { Airport.new }
+    let(:plane) { Plane.new }
+
+  # I would like a default airport capacity (I have chosen default of 25)
+    it "Airport has a default capacity" do
+      expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+    end
+
+  # I would like the default capacity to be overridden as appropriate
+    it "Airport has a variable capacity" do
+      described_class::DEFAULT_CAPACITY.times do
+        subject.land(plane)
+      end
+      expect { subject.land Plane.new }.to raise_error "Airport is full, landing prevented"
+    end
+
+  end
+  
+  describe "#land" do
+    it "Airport responds to land" do
+      expect(subject).to respond_to(:land).with(1).argument
+    end
+
+    # I want to instruct a plane to land at an airport
+    it "Lands a plane at the airport" do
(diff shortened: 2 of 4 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+  
+  describe "#land" do
+    it "Airport responds to land" do
+      expect(subject).to respond_to(:land).with(1).argument
+    end
+
+    # I want to instruct a plane to land at an airport
+    it "Lands a plane at the airport" do
+      plane = Plane.new
```
lkilk, line 31: could use a double here instead of creating a new class instance, plane = double(:plane)

lkilk, line 31: could use a double here instead of creating a class instance of plane 
`plane = double(:plane)`

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+class Airport
+
```
lkilk, line 2: code is very easy to read and follow 👍

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def land(plane)
+    fail "Airport is full, landing prevented" if full?
+
+    @planes << plane
+  end
+
+  def take_off(_plane)
+    @planes.pop
```
lkilk, line 20: using the delete method you could potentially specify a particular 'plane' to remove from the array? I could be wrong on this but just a thought?
`@planes.delete(plane)`

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      subject.land(plane)
+      expect(subject.take_off(plane)).to eq plane
+    end
+
+    # I want to confirm that it is no longer in the airport
+    it "Confirms that a plane is no longer in the airport" do
+      plane = Plane.new
+      subject.land(plane)
+      subject.take_off(plane)
+      expect(subject.confirmation).to eq("Plane has left the airport")
```
lkilk, line 60: like the fact that there is a message generated here, really clear.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+class Airport
+
+  attr_reader :plane 
+  attr_accessor :capacity
+
+  DEFAULT_CAPACITY = 25
```
EdwardAndress, line 6: Nice use of a constant for the capacity default

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail "Airport is full, landing prevented" if full?
```
EdwardAndress, line 14: nice hat you're using a method for full? rather than including that logic here  in land

lib/airport.rb
EdwardAndress: Nice idea, Liam!

spec/airport_spec.rb
EdwardAndress, line 60: This is a nice feature but the test will always pass, regardless of whether the plane has left the airport or not.  Could you make this confirmation message get generated automatically after a plane has successfully taken off?

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    # I want to instruct a plane to land at an airport
+    it "Lands a plane at the airport" do
+      plane = Plane.new
+      expect(subject.land(plane)).to eq([plane])
+    end
+
+    # I want to prevent landing when the airport is full
+    it "Raises error if airport is full" do
+      subject.capacity.times { subject.land Plane.new }
+      expect { subject.land Plane.new }.to raise_error "Airport is full, landing prevented"
```
EdwardAndress, line 38: Nice!  Do you know why the curly braces are required for this test?

EdwardAndress:
Nice work Chesca! I can see that you've picked up the basics of testing.  The next step for this exercise would be to add a weather class and integrate that to complete the user stories.  I've left some comments against your code – hopefully they make sense and are helpful!

PR #2105, opened 2021-04-30
Airport Challenge

adamroy12:
Completed user stories 1-4. Had trouble with 5 when it came to getting the tests to pass with doubles etc, but the program seems to randomise weather and prevent take offs.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -8,6 +8,8 @@ class Airport
   end
 
   def land_plane(plane)
+    raise "Landing Declined: Airport full" if @planes.any?
+    
     @planes.push(plane)
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,17 @@
+require "plane.rb"
+
+class Airport
+  attr_reader :planes
+    
+  def initialize
+    @planes = []
+  end
+
+  def land_plane(plane)
+    @planes.push(plane)
+  end
+
+  def return_planes
+    @planes
+  end
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -11,7 +11,7 @@ class Airport
   end
 
   def land_plane(plane)
-    raise "Landing Declined: Airport full" if @planes.any?
+    raise "Landing Declined: Airport full" if @planes.length == capacity
     @planes.push(plane)
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,14 +2,16 @@ require_relative "plane"
  
 class Airport
   attr_reader :planes
+  attr_reader :capacity
+  DEFAULT_CAPACITY = 10
     
-  def initialize
+  def initialize(capacity = DEFAULT_CAPACITY)
     @planes = []
+    @capacity = capacity
   end
 
   def land_plane(plane)
     raise "Landing Declined: Airport full" if @planes.any?
-    
     @planes.push(plane)
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -3,6 +3,7 @@ require_relative "plane"
 class Airport
   attr_reader :planes
   attr_reader :capacity
+  attr_reader :weather
   DEFAULT_CAPACITY = 10
     
   def initialize(capacity = DEFAULT_CAPACITY)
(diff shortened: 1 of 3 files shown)


adamroy12:
user stories 1-4 completed. 5 partially completed, it doesn't work.

lib/airport.rb
```diff
@@ -12,15 +13,26 @@ def initialize(capacity = DEFAULT_CAPACITY)
...
     @planes.delete(plane)
     @planes
   end
 
   def return_planes
     @planes
   end
+
+  def weather
+    if rand(10) == 1 || rand(10) == 2
```
ekjones1993, line 31: I like how you've set the weather class up to use rand in an if statement to decide the weather I didn't think of doing it that way but I think its a cleverer approach than how I did it.

EdwardAndress: It's an interesting point of comparison.  This is Emily's approach...

```
def weatherforecast
    ["stormy", "calm"].sample
end
```

Deciding which is 'best' is entirely subjective but we could consider several potential changes / extensions and see how the two approaches would cope.

1.  Adding further types of weather.  For example, snow.
2. Changing the probability that each type of weather will occurr.

I'll leave this for you to think about!

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+require_relative "plane"
+ 
+class Airport
+  attr_reader :planes
+  attr_reader :capacity
+  attr_reader :weather
+  DEFAULT_CAPACITY = 10
```
EdwardAndress, line 7: Nice use of a constant for the default!

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+    
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land_plane(plane)
+    raise "Landing Declined: Airport full" if @planes.length == capacity
```
EdwardAndress, line 15: What if you created a new method called `full?` and then did this instead?

```
raise "Landing Declined: Airport full" if full?
```

Would that make the code more readable?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    @planes.push(plane)
+  end
+
+  def take_off(plane)
+    raise "Weather is stormy" if weather == "stormy"
```
EdwardAndress, line 21: Same here (as for the capacity check)

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+    @planes.delete(plane)
+    @planes
+  end
+
+  def return_planes
```
EdwardAndress, line 26: You don't need this because you have an attr_reader for planes

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  it "air traffic controller instructs planes to take off" do
+    plane = Plane.new
+    subject.land_plane(plane)
+    subject.take_off(plane)
+    expect(subject.return_planes).to eq []
+  end
+
+  it "prevents planes from taking off when weather is stormy" do
+    @weather = double('weather')                                            #this bit doesn't work.
```
EdwardAndress, line 18: Ah, so this is a tricky bit!  You would use a double here if you had a separate Weather class.  As things stand, you'll want to use a stub instead.  A stub is used to intercept a method call and provide a `canned response`.  So, for example, you could stub `weather` and make it return `stormy` rather than actually execute the logic inside the `weather` method.

But then, you wouldn't be testing the weather method.  So what could you do instead?

The weather method uses `rand` and we know that `rand` works because it's part of the ruby core library (and is test there).  So we can stub rand, like this, and make it return a non-random number of our choosing.

```
allow(subject).to receive(:rand) { 1 }
```

What's going on here?  An instance of airport (the subject of your test) is the _implicit receiver_ of `rand`.  So we stub `rand` on `subject`.

Doing this ensures that weather uses rand, because the stub only works for that implementation.

You can go one stage further and ensure that rand is called with a specific argument by modifying the stub.  Give it a try!

EdwardAndress:
Overall a really nice solution and I can see that you've picked up testing nicely.  There are some comments against your code.  I hope they make sense and are helpful!  Drop me a line if you have any questions.

PR #2107, opened 2021-04-30
Airport Challenge - Matt

turboKittyNinja:
I made a simple README that explains where I got up to. I tried getting a method to return a predetermined value, but didn't get too far before realising I needed to go make dinner. I got to the end of the user stories, but the weather is set to manual for now.


lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+class Airport
+  DEFAULT_CAPACITY = 3
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+
+  end
+
+  attr_reader :hangar
```
Jaikido007, line 10: Good terminology/naming

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def take_off(plane)
+    fail 'Cannot takeoff in a storm' if @weather
+    
+    "#{plane} has taken off."
+  end
+
+  def full?
+    @hangar.length >= capacity
```
Jaikido007, line 27: Like the use of .length in this.  Clever method use

lib/airport.rb
EdwardAndress: Agreed!  Good choice of names.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @capacity = capacity
+
+  end
+
+  attr_reader :hangar
+  attr_accessor :capacity
+
+  def land(plane)
+    fail 'Cannot land during a storm' if @weather
+    fail 'Airport is full' if full?()
```
EdwardAndress, line 15: You don't need parentheses on `full?()`... that's some Javascript or Python sneaking in!

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @hangar = []
+    @capacity = capacity
+
+  end
+
+  attr_reader :hangar
+  attr_accessor :capacity
+
+  def land(plane)
+    fail 'Cannot land during a storm' if @weather
```
EdwardAndress, line 14: this is nice but could be refactored for readability...

```
fail 'Cannot land during a storm' if stormy?
```

Then you'd need a new `stormy?` method., of course

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    fail 'Airport is full' if full?()
+
+    @hangar << plane
+  end
+
+  def take_off(plane)
+    fail 'Cannot takeoff in a storm' if @weather
+    
+    "#{plane} has taken off."
```
EdwardAndress, line 23: What else should happen when the plane takes off?

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def full?
+    @hangar.length >= capacity
+  end
+
+  def storm
```
EdwardAndress, line 30: If you commented this out because of failing tests the thing to look for is "how to stub randomness in ruby" :)

EdwardAndress:
This is a nice solution!  I can see that you're picking up TDD well and really appreciated your reflective statement in the readme.  It's always good to see how people think about these challenges.

There are some comments against the code in various places. I hope they make sense and are helful!

lib/airport.rb
turboKittyNinja: This is one of the things I like to do in editor because it changes its colour to yellow for me, so I can see where I'm calling a method amongst all the blue. Helps for my own personal readability, but if it's a big no no. I'll happily take it out! :)

lib/airport.rb
turboKittyNinja: Thank you Eddie, I agree with this. I was heavily debating making a weather class, but at the time I was just trying to get certain functionality to work and never got back to it. Bad naming on my part!

lib/airport.rb
turboKittyNinja: Should remove plane from hangar :0

PR #2110, opened 2021-05-02
All features added and tests passed

charlie-casserly:


spec/weather_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'weather'
+
+describe Weather do
+  it 'randomly selects a number' do
+    expect(subject.random_number).to be_kind_of Integer
+  end
+
+  it 'randomly selects a weather' do
```
BecaLParker, line 8: Consider renaming this test. 

It looks like this is in fact testing 'outputs a weather matching the randomly selected number'.

In the test, you've stubbed out the randomness (great! - I get why). So being pedantic - this one is not actually not a test of randomness.

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
```
BecaLParker, line 4: Your naming throughout is really clear and easy to follow  - keep doing that in future projects
https://github.com/makersacademy/course/blob/master/goals/course_goals.md#i-help-my-teams-succeed

lib/weather.rb
```diff
@@ -0,0 +1,23 @@
...
+      4 => 'heavy rain',
+      5 => 'stormy'
+    }
+
+  def initialize
+    @weather = random_weather
+  end 
+
+  def random_number
+    rand(1..5)
```
BecaLParker, line 17:
```suggestion
    [array, containing, weather, conditions].sample
```
would be another way to achieve this :)
https://www.rubyguides.com/2015/03/ruby-random/

spec/airport_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    it 'raises an error when landing at full airport with default capacity' do
+      allow(subject).to receive(:weather) { 'rain' }
+      Airport::DEFAULT_CAPACITY.times{subject.land(Plane.new)}
+      expect{subject.land(Plane.new)}.to raise_error('Cannot land. The airport is full.')
+    end
+
+    it 'raises an error when landing at full airport with explicit capacity' do
+      airport = Airport.new(500)
+      allow(airport).to receive(:weather) { 'rain' }
+      airport.capacity.times{airport.land(Plane.new)}
```
BecaLParker, line 34: I love that this avoid magic numbering :)

BecaLParker (commented):
From the commit history, it looks like you were working in a TDD process.

Keep on doing this - testing first, then code :)

https://github.com/makersacademy/course/blob/master/goals/course_goals.md#i-can-make-software

PR #2111, opened 2021-05-02
end of unit challenge submission

rdupplaw:


lib/airport.rb
```diff
@@ -15,11 +15,37 @@ def land(plane)
...
-    if @weather.now == 'sunny' && @planes.include?(plane)
+    if plane_can_launch?(plane)
       @planes.delete(plane)
       true
     else
       false
     end
   end
+
+  private
```
fel1xp, line 30: great use of methods, makes the code much shorter/simpler/easy to follow

PR #2118, opened 2021-05-03
airport_challenge mvg

mvgame74:


lib/airport.rb
```diff
@@ -0,0 +1,47 @@
+require "./lib/airplane.rb"
+
+class Airport
+  attr_reader :airplanes, :capacity
```
Dev-ops-true, line 4: great job making them both attr_reader, wish i thought of doing it like that

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def initialize
+    @airplanes = []
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def random_weather
+    weather = ["sunny", "cloudy", "windy","rainy", "stormy"]
+    weather[rand(weather.size)]
+  end
```
Dev-ops-true, line 15: I would suggest a different method for calling the weather (.sample), and maybe to put more "sunny" days in.

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+  
+  def land(airplane)
+    #random_weather
+    #p random_weather
+    #fail 'Too stormy to land' if stormy?
```
Dev-ops-true, line 20: If you update your random weather method this error will work as you intended, you got this.

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def empty?
+    @airplanes.empty?
+  end
+
+  def stormy?
+    random_weather == "stormy"
+  end
+end
```
Dev-ops-true, line 47: Pull your random weather into this method so it can return true or false as required

PR #2126, opened 2021-05-04
inital

ekjones1993:


lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+class Weather
```
adamroy12, line 1: I like that you made a new class for weather, I totally didn't think of that.

spec/airport_spec.rb
```diff
@@ -0,0 +1,27 @@
+require './lib/airport.rb'
+require './lib/weather.rb'
+
+describe Airport do
+
+  describe "#takeoff" do
+    it 'airport responds to takeoff' do
+      expect(subject).to respond_to(:takeoff).with(1).argument
+    end
```
adamroy12, line 9: I'm not sure you need the 'responds to' test anymore. The takeoff method is doing something, so should the takeoff test be testing whether takeoff makes a plane leave the @planes array, instead?
I think the 'responds to' tests are just made to make sure a method actually exists in the code and is in the right class, but can be changed after your method is made to do something. I'm not actually sure, though.

spec/airport_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    it "causes error when no planes at airport to takeoff"  do
+      expect { subject.takeoff Airport.new }.to raise_error "No planes to take off"
+    end
+  end
+
+  describe "#land" do
+    it "airport responds to land" do
+      expect(subject).to respond_to(:land).with(1).argument
+    end
```
adamroy12, line 19: Maybe the same as above here. Should the test for the land method be testing whether the airport responds to land, or should there be a test for whether it successfully adds to the array?
Not sure. I don't know how to code.

adamroy12 (approved):

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def initialize(capacity = 10)
+    @airport = []
+    @capacity = capacity
+    @weather = Weather.new
+  end
+
+  def land(plane)
+    raise "Can't land, stormy weather" if weatherforecast == "stormy"
+    fail "Can't land, airport full" if airportfull?
```
EdwardAndress, line 14: It's nice that you have this `airportfull?` method, rather than doing something like `fail "Can't land, airport full" if @airport.length == @capacity`.

Could you do something equivalent for the line above, where you're checking the weather?

EdwardAndress: If so, what would be the advantage of making that refactor?

lib/weather.rb
EdwardAndress: I agree with Adam! Nice work :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'airport'
+
+describe 'Airport' do
+  airport = Airport.new
+  plane = "easyjet"
```
EdwardAndress, line 5: Ah, nice that you realised you don't necessarily need a plan class (yet).

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'airport'
+
+describe 'Airport' do
+  airport = Airport.new
```
EdwardAndress, line 4: You can omit this line if you do `describe Airport` rather than `describe 'Airport'` and then use `subject` in your tests.  Give it a try!

Also, when you're declaring variables for your tests, it's more idiomatic to use `let` statements like `let(:plane) { "easyjet" }`.

This is because `let` statements are like variable assignments with superpowers: https://medium.com/@tomkadwill/all-about-rspec-let-a3b642e08d39

spec/airport_spec.rb
EdwardAndress: Indeed, these `respond_to` tests are great for getting started with testing but they serve little long term purpose.  One day I might remove them all from the materials!

spec/airport_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  describe '#takeoff' do
+
+    it 'cant take off when its stormy' do
+      allow(airport).to receive(:weatherforecast) { "stormy" }
+      expect { airport.takeoff(plane) }.to raise_error "Can't take off, stormy weather"
+    end
+
+    it 'check if a plane stayed during stormy weather' do
+      allow(airport).to receive(:weatherforecast) { "stormy" }
```
EdwardAndress, line 35: Nice use of stubbing.  We can take this one stage further and create a stub that ensures `Airport` is making correct use of `Weather` but to do this, you also need to use dependency injection, so have a look at that first and then we can chat about the next part, if you like.

spec/weather_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'weather'
+
+describe Weather do
+
+  it 'weather responds to weatherforecast' do
+    expect(subject).to respond_to(:weatherforecast)
+  end
+
+  it "it's calm" do
+    allow(subject).to receive(:weatherforecast) { "calm" }
```
EdwardAndress, line 10: Can you make this test fail by changing the `Weather` class?  Specifically, can you make it fail by changing the `weatherforecast` method?

If not, why not?

EdwardAndress:
Hey Emily, this is a really nice solution with two main highlights
1. You've created a separate class for Weather
2. You've used stubbing!

Nice work.

I hope the suggestions / comments / questions I have added make sense and are helpful.

PR #2128, opened 2021-05-29
airport_challenge

Ashley-Slaney:


TarunTheo13:
Appreciations:

- Clear code
- Clear commit messages
- Good use of mock doubles in RSpec files
- Nice solution for solving edge cases with separate functions in the Plane class

Suggestions:

- Could raise errors if airport is full/weather is stormy instead of returning strings
- Could make certain methods private that don't need to be accessed outside other methods
- Could add personal comments on README file
- Your code will currently still push a plane into an airport even if it is in another airport (I think)

PR #2132, opened 2021-05-31
Merryn's Airport Challenge

merrynhr:


lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    @plane_no = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    raise 'Airport full' if full?
+    raise 'Stormy!' if stormy?
+
+    @plane_no << plane
+
```
craigea92, line 20: I really like the method land(plane), it's simple and easy to understand exactly what it's doing.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+  def take_off(plane)
+    raise 'Stormy!' if stormy?
+    raise 'Wrong airport!' unless @plane_no.include?(plane)
+
+    @plane_no.pop
+  end
+
+  private
+
```
craigea92, line 31: Good use of private and I like the stormy? method

lib/plane.rb
```diff
@@ -0,0 +1,21 @@
+require_relative 'airport'
+class Plane
```
craigea92, line 2: I really like the ability to see the status of the planes, that information would be really useful to a client

PR #2137, opened 2021-05-31
Tarun's Airport Challenge

TarunTheo13:


TarunTheo13:
Addressed the following user stories:

As an air traffic controller 

As an air traffic controller 

As an air traffic controller 
To ensure safety 

As the system designer

As an air traffic controller 
To ensure safety 

As an air traffic controller 
To ensure safety 

TarunTheo13:
Did not address the following edge cases, except for the first one:

Your code should defend against edge cases such as inconsistent states of the system ensuring that planes can only take off from airports they are in; planes that are already flying cannot take off and/or be in an airport; planes that are landed cannot land again and must be in an airport, etc.

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  private
+  def stormy?
+    rand(1..25) == 25
+  end
+
+  def full?
+    @planes.count >= @capacity
+  end
+
```
Ashley-Slaney, line 41: # I think another way to do this would be to have an instance variable in the plane class (called something like 
# @flying) that is set to true or false based on whether or not the plane is in the air.

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+  def in_airport?(plane)
+    @planes.each do |x|
+      if x == plane
+        return true
+      end
+    end
+    return false
+  end
```
Ashley-Slaney, line 49: I think another way to do this would be to have an instance variable in the plane class (called something like
@flying) that is set to true or false based on whether or not the plane is in the air.

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def full?
+    @planes.count >= @capacity
+  end
+
+  def in_airport?(plane)
+    @planes.each do |x|
+      if x == plane
+        return true
+      end
```
Ashley-Slaney, line 46: I think another way to do this would be to have an instance variable in the plane class (called something like
@flying) that is set to true or false based on whether or not the plane is in the air.

PR #2138, opened 2021-05-31
Airport Challenge - Completed User Stories

craigea92:
First submission,  I managed to create tests and code to pass them but I struggled at the further tasks. I used a lot of what I had learned in week 1 when writing this code but I also had to use google a bit for matchers and doubles!


merrynhr:
Really clear code. Excellent use of the weather Class as well! Great use of doubles!

merrynhr:
Perhaps you could have used a plane class to ensure your classes are as specific as possible.

PR #2139, opened 2021-06-01
First pull request

jooomin:


lib/airport.rb
```diff
@@ -0,0 +1,6 @@
+class Airport
```
frodri13, line 3: Remember to write the test for this class first.

PR #2141, opened 2021-06-01
Airport Challange

1ugia:
Up to second /user Story at 96.30% coverage.


lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+# I want to instruct a plane to land at an airport
+
+# As an air traffic controller 
+# So I can get passengers on the way to their destination 
+# I want to instruct a plane to take off from an airport 
+# and confirm that it is no longer in the airport
+
+require_relative 'plane'
+
+class Airport
```
YinnyF, line 12: the naming is very clear and easy to understand.

lib/plane.rb
```diff
@@ -0,0 +1,8 @@
+
+class Plane
+
+  def travelling
```
YinnyF, line 4: It's really smart to have this method defined on Plane class, I did not think about that! It can be multipurpose - useful for future situations.

YinnyF: I think one improvement for this is to define the method with a question mark, as it is a predicate method and you can work with it in RSpec with special matchers.

YinnyF: More on predicate matchers here: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/predicate-matchers

PR #2143, opened 2021-06-04
Emily Fitzsimmons Airport Challenge

EmilyFitzs:
not complete yet but had a go


spec/plane_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'plane'
+
+describe Plane do
+
+
+  it { is_expected.to respond_to(:take_off)}
+
+describe '#take_off' do
+  context 'when plane is flying' do
```
ehelsan, line 9: I like that you have context, within the describe & if methods. I had been looking for a function that achieves/conveys the same thing. I'll be making a note of this one!

lib/plane.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def take_off
+    raise 'Plane already flying!' if @flying
+  end
+
+  def land
+    raise 'Plane is already landed!' unless @flying
+  end
+
```
ehelsan, line 14: This is all very easy to understand, and written out super clearly.

PR #2148, opened 2021-06-07
Ryan Chidwick Pull Request Submission

Ryan734:
Please refer to the README.md files. 

Activity not completed - but as much code as possible submitted.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,4 @@
+class airport
+  def land(plane)
+  end
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,7 +1,51 @@
+require_relative 'weather_reporter'
+
 class Airport
+  DEFAULT_CAPACITY = 20
+  attr_reader :planes
+
+  def initialize(weather_reporter, capacity = DEFAULT_CAPACITY)
+    @weather_reporter = weather_reporter
+    @capacity = capacity
+    @planes = []
+  end
+
   def land(plane)
+    raise 'Cannot land plane: airport full' if full?
+    raise 'Cannot land plane: weather is stormy' if stormy?
+    plane.land(self)
+    add_plane(plane)
   end
 
   def take_off(plane)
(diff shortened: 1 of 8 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,51 @@
+require_relative 'weather_reporter'
+
+class Airport
```
Juel07, line 3: Great use of methods! I like how each method follows the _Single Responsibility Principle_, making the code easy to read and follow.

lib/airport.rb
```diff
@@ -0,0 +1,51 @@
...
+    raise 'Cannot take off plane: weather is stormy' if stormy?
+    raise 'Cannot take off plane: plane not at this airport' unless at_airport?(plane)
+    plane.take_off
+    remove_plane(plane)
+    plane
+  end
+
+  private
+
+  attr_reader :capacity, :weather_reporter
```
Juel07, line 30: Interesting! I assumed that `attr_reader` goes at the beginning of the class. For the purpose of my own understanding, is there a reason you've added a separate `attr_reader` on line 30? :)

lib/weather_reporter.rb
```diff
@@ -0,0 +1,5 @@
+class WeatherReporter
```
Juel07, line 1: Nice! I like that you've created a separate class `WeatherReporter` for the weather instead of mixing the weather behavior within `Airport` or `Plane` classes 👍

spec/airport_spec.rb
```diff
@@ -0,0 +1,98 @@
+require 'airport'
+
+describe Airport do
+  subject(:airport) { described_class.new(weather_reporter, 20) }
+  let(:plane) { double :plane, land: nil, take_off: nil }
+  let(:weather_reporter) { double :weather_reporter }
+
+  describe '#land' do
+    context 'when not stormy' do
```
Juel07, line 9: Good use of `context` and `before` blocks within your specs 💯

PR #2149, opened 2021-06-07

mjconn:


spec/airport_challenge_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    it 'puts plane in airport' do
+      allow(airport).to receive(:stormy?).and_return false
+      expect(airport.land(plane)).to include plane
+    end
+
+    context 'when at capacity' do
+      it 'prevents landing' do
+        allow(airport).to receive(:stormy?).and_return false
```
ciombrik, line 20: could use the before the method to avoid the repetition.

PR #2150, opened 2021-06-07
first commit

emibrand:
Got VERY confused on Friday so I have attempted to redo the challenge using the method used in the airport challenge exemplar which made a lot more sense. Currently on the third user story, not really sure how to incorporate the capacity but will use a @planes array and a raise error for the weather.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,11 @@
+class Airport
+
+  def initialize(capacity)
+    @capacity = capacity
+  end
+  def land(plane)
+    raise 'Cannot land: airport full' if @planes >= @capacity
+  end
+  def take_off(plane)
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,20 @@
+require 'airport'
+
+describe Airport do
+  airport = Airport.new(100)
+
+  it 'tells plane to land' do
+    expect(airport).to respond_to(:land).with(1).argument
+  end
+
+  it 'tells plane to takeoff' do
+    expect(airport).to respond_to(:land).with(1).argument
+  end
+
+  it 'cannot allow landing when airport full' do
+    100.times do
+      airport.land(:plane)
+    end
+    expect { airport.land(:plane) }.to raise_error 'Cannot land: airport full'
+  end
+end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,2 @@
+class Plane
+end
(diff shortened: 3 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,11 @@
+class Airport
+
+  def initialize(capacity)
+    @capacity = capacity
+  end
+  def land(plane)
+    raise 'Cannot land: airport full' if @planes >= @capacity
```
Aroax, line 7: could this line be refactored at all? How might you do so?

lib/airport.rb
```diff
@@ -0,0 +1,11 @@
+class Airport
+
+  def initialize(capacity)
```
Aroax, line 3: could it be beneficial to add a default capacity, so that if a user tries to instantiate an Airport.new without any arguments the program would still start without an error?

lib/airport.rb
```diff
@@ -0,0 +1,11 @@
...
+    @capacity = capacity
+  end
+  def land(plane)
+    raise 'Cannot land: airport full' if @planes >= @capacity
+  end
+  def take_off(plane)
```
Aroax, line 9: might it be worth adding spaces between the method lines, for improved readability? This chain of def and ends is quite a lot to look at! xD

spec/airport_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'airport'
+
+describe Airport do
+  airport = Airport.new(100)
+
+  it 'tells plane to land' do
+    expect(airport).to respond_to(:land).with(1).argument
```
Aroax, line 7: love this use of with().argument here

lib/airport.rb
```diff
@@ -0,0 +1,11 @@
+class Airport
+
+  def initialize(capacity)
+    @capacity = capacity
+  end
+  def land(plane)
+    raise 'Cannot land: airport full' if @planes >= @capacity
+  end
```
Aroax, line 8: how will you finish this method to use @planes?

spec/airport_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  it 'tells plane to takeoff' do
+    expect(airport).to respond_to(:land).with(1).argument
+  end
+
+  it 'cannot allow landing when airport full' do
+    100.times do
```
Aroax, line 15: is there any way of making this more dynamic, rather than hardcoded? May also need to look at line 4, as well as my review suggestions in the production code

Aroax (commented):
Nice going so far!

PR #2151, opened 2021-06-07
GG - airport challenge submission

goncalog:


spec/airport_spec.rb
```diff
@@ -0,0 +1,110 @@
+require "airport"
+
+describe Airport do
+  it "can land an airplane" do
+    airplane = Airplane.new
```
anitaRead, line 5: To avoid creating the same instances of Airport.new throughout the code you could declare it with a subject statement.  Before the first 'it' statement. 

subject(:airport) { described_class.new }

spec/airplane_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "airplane"
+
+describe Airplane do
+  it "can be landed" do
```
anitaRead, line 4: The it message could be "can land" instead of "can be landed"

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  private 
+
+  def full_airport?
+    @airplanes.length == @capacity
+  end
+
+  def in_airport?(airplane)
+    @airplanes.include?(airplane)
+  end
+end
```
anitaRead, line 46: Really well written and easy to understand.  Great use of SRP throughout with clear method/variable names!

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+class Airport
+  def initialize(capacity = 1)
+    @airplanes = []
+    @capacity = capacity
+    @weather = "sunny"
```
anitaRead, line 5: Instead of creating an instance variable for weather and a method for passing in the weather.  You could randomise the weather to make sure it is only stormy some of the time.

spec/airport_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+    heathrow.land(ba)
+    gatwick.land(emirates)
+
+    expect(heathrow.get_airplanes).to eq [tap, ryanair, ba]
+    expect(gatwick.get_airplanes).to eq [emirates]
+    expect(lutton.get_airplanes).to eq []
+    expect(easyjet.landed?).to eq false
+  end
+end
```
anitaRead, line 110: Really well written tests that are easy to understand and do exactly what the user stories have asked!

spec/airport_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  it "an airplane can take-off" do
+    airplane = Airplane.new
+    subject.land(airplane)
+    subject.take_off(airplane)
+    expect(subject.get_airplanes).to eq([])
+    expect(airplane.landed?).to eq false
+  end
+
+  it "not possible to land airplane when full" do
```
anitaRead, line 19: You could use context blocks instead of using 'when' to improve the readability of your code even further.  Right now your code is really easy to read and understand but it may come in handy in longer and more complex projects/code.

PR #2156, opened 2021-06-07
Anita's airport challenge

anitaRead:
Challenge not fully completed.  
User stories 1 through 4.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -10,13 +10,18 @@ class Airport
     end
 
     def land(plane)
-       raise 'Cannot land plane, airport is full!' if @planes.length == @capacity
+       raise 'Cannot land plane, airport is full!' if full?
        @planes << plane
     end
 
     def take_off(plane)
        @planes.delete(plane) 
        @planes.length
+    end
+
+    private
+    def full?
+        @planes.length == @capacity
     end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,8 +2,19 @@ require_relative 'plane'
 
 class Airport
 
+    attr_reader :planes_in_airport
+    
+    def initialize
+        @planes_in_airport = []
+    end
+
     def land(plane)
-       
+       @planes_in_airport << plane
+    end
+
+    def take_off(plane)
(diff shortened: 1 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+       @planes.delete(plane) 
+       @planes.length
+    end
+
+    private
+    def full?
+        @planes.length == @capacity
+    end
+
+    def stormy?
```
goncalog, line 28: love the probabilistic stormy? method. a much more elegant solution vs. using an instance variable. super!

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+describe Airport do
+    subject(:airport) { described_class = new(20) }
+    let(:plane) { double :plane } 
+
+    describe '#land' do
+        
+        it 'instructs a plane to land' do
+            airport = Airport.new(20)
+            allow(airport).to receive(:stormy?).and_return false
```
goncalog, line 11: this ties very well with the stormy? method, and is a great solution when using probabilistic methods

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+            expect(airport).to respond_to(:land).with(1).argument
+        end
+
```
goncalog, line 14: would it help to create a test to confirm that a plane has landed?

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'airport'
+
+describe Airport do
+    # As an air traffic controller 
+    # So I can get passengers to a destination 
+    # I want to instruct a plane to land at an airport
+
```
goncalog, line 7: it's a great idea to add the user stories to the tests. it makes everything much clearer (e.g. why a given test is there and whether it's testing what it's supposed to).

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'plane'
+
+class Airport
+
+    # attr_reader :planes
```
goncalog, line 5: minor thing, but this is probably something that could be removed

PR #2157, opened 2021-06-07
Irina's Airport Challenge Final

IrinaGC:


lib/plane.rb
```diff
@@ -0,0 +1,15 @@
+class Plane
+
+  def initialize
+    @plane_status = status
+  end
+
+  def landed(plane_inqired)
```
usnahakimi, line 7: Argument not clear

spec/airport_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "./lib/airport.rb"
+
+describe Airport do
+
+  it "prevent landing when airport full" do
```
usnahakimi, line 5: Clear and readable test

spec/plane_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "./lib/plane.rb"
+
+describe Plane do
+  plane = Plane.new
+
+  it "has been identified and command sent" do
```
usnahakimi, line 6: Test here could be more to the point/clear.

spec/plane_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect(plane).to respond_to(:land)
+  end
+
+    it "confirms that it has landed" do
+     expect(plane.landed()).to be_an_instance_of Array
```
usnahakimi, line 11: Like the array use here

PR #2158, opened 2021-06-07
Airport Challenge/ Niesoth Pull

NiesothThang:
Attempted the first user story.

### DIFF
--- a/spec/features/user_stories_spec.rb
+++ b/spec/features/user_stories_spec.rb
@@ -1,0 +1,9 @@
+
+describe 'user stories' do
+  it 'so passengers can get to a desitination, instruct the plane to land at the airport' do
+    airport = Airport.new
+    plane = Plane.new
+    expect { airport.land(plane) }.not_to raise_error 
+  end
+end
+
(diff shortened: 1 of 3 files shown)


spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,9 @@
+
+describe 'user stories' do
+  it 'so passengers can get to a desitination, instruct the plane to land at the airport' do
+    airport = Airport.new
+    plane = Plane.new
+    expect { airport.land(plane) }.not_to raise_error
```
kmunro4-ford, line 6: I like this test, i hadnt seen/used it before and it would make testing methods much easier for me :)
I had trouble at first with my tests as i was testing for explict returns when that wsant initally my goal

kmunro4-ford:
good TDD, writing the test before the code, this was one of the largest challenges for me (other than understanding what i was supposed to be doing) i kept starting to write code before the tests then having to comment out my code pretend i hadnt done that, write a test and then see if my code passed when i uncommented 😅

PR #2159, opened 2021-07-03
Finished airport project.

karimabuseer:
All features developed; 100% test coverage; commenting completed and rubocop passed.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -9,4 +9,8 @@ class Airport
   def take_off(plane)
     @plane = nil
   end
+
+  def in_air?(plane)
+    @plane == plane
+  end    
 end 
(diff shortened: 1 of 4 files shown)


isobelpepp:
Lovely code!! The comments before each method make the code very easy to read. You’re tests are great - I don’t really see how they could be cleaner. I haven’t seen the context…do before and that would have definitely benefitted my tests!

isobelpepp:
A couple of suggestions: I would consider putting the weather in a separate class as I think it has separate responsibility to the airport class (but I like how succinctly you have generated the weather)
Checking whether the plane is in or out of the airport could potentially be condensed into one method. But again I like how you can check for .status and not sure if this would definitely make the code cleaner.

isobelpepp:
And one question - I can’t see if you have accounted for landing a plane that is already in the airport and prevents it from landing if so?

PR #2160, opened 2021-07-03
Pav's Airport Challenge

pav0107:


README.md
```diff
@@ -68,22 +57,86 @@ For overriding random weather behaviour, please read the documentation to learn
...
+```
+Plane <-- land  --> Airport
+Plane <-- take-off  --> Airport
+Plane <-- at_airport? --> Airport
+Airport <-- full? --> Plane
+Weather <-- is_stormy? --> Plane
+```
 
-* Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+### Step 5: testing
```
senoodle, line 137: I like that you have explained the steps you took to complete the project :)

spec/plane_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+require "airport"
+
+describe Plane do
+  it { is_expected.to respond_to(:land).with(1).argument }
+
+  it { is_expected.to respond_to(:take_off).with(1).argument }
+
+  it { is_expected.to respond_to(:location) }
+
+  let(:plane1) { Plane.new }
```
senoodle, line 11: Good use of the let method here

lib/airport.rb
```diff
@@ -0,0 +1,16 @@
+class Airport
+
+  attr_accessor :planes, :capacity, :weather
+
+  DEFAULT_CAPACITY = 2
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+    @weather = ["sunny", "sunny", "sunny", "sunny", "sunny", "stormy"][rand(6)]
```
senoodle, line 10: I'm not sure if this is the best suggestion but it might be nice to create a method that randomises the chances of the weather being stormy - so that you can avoid making arrays. This is what I did:
def stormy?
    rand(1..8) > 4  #no particular reason why I chose 8 and 4, it's just so it's a random result
end

I used stormy? in my land method - which is in the same class:
def land(plane)
    raise 'airport is full' if full?
    raise 'cannot land plane, weather stormy' if stormy?
    @planes << plane 
  end

It may be better to keep it your way! I'm not sure if mine is any good 😨

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
+require_relative "./airport"
+
+class Plane
```
senoodle, line 3: This whole class is really well written and super easy to read!

lib/airport.rb
eoinmakers, line 10: There is maybe something a little static about the initialisation of the airport class being what determines the weather - this means that if one airport sees multiple landings and takeoffs over the next few days / weeks, the weather will remain the same.

eoinmakers: A small refactor - you could take away the [rand(6)], and later when you need the weather to be generated you can say @weather.sample. It will pick a random element from this array.

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  attr_accessor :location
+
+  def initialize
+    @location = nil
+  end
+
+  def land(airport)
+    fail "Plane can't land. Airport is full" if airport.full?
+
```
eoinmakers, line 13: I would remove the newlines underneath each of the fails to group them together for readability

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
...
+    fail "Plane can't land in stormy weather" if airport.weather == "stormy"
+
+    fail "Plane has already landed at this airport" if location == airport
+
+    @location = airport
+    airport.planes.push(self)
```
eoinmakers, line 19: We try not to interact directly with variables in other classes, so we can protect our data as much as we can. The way around this is trying to access variables internally in other classes, and allowing other classes to interact with them only through specific methods. For example, here you could call a method in airport that passes this plane into the planes array. This way, you won't need an attribute reader, and instead will only be able to interact with planes in a really specific way, which is safer than having it publicly exposed.

spec/airport_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+require 'plane'
+
+describe Airport do
+
+  let(:plane1) { Plane.new }
+  let(:plane2) { Plane.new }
+  let(:airport) { Airport.new }
+
+  it 'is full' do
+    allow(airport).to receive(:weather) { "sunny" }
```
eoinmakers, line 11: Nice!

eoinmakers (commented):
Well done, you've implemented the user stories well. I made a small suggestion for a change in your weather logic, and this can also be separated out into its own class to better adhere to SRP. Good unit testing too, well done.

README.md
```diff
@@ -68,22 +57,86 @@ For overriding random weather behaviour, please read the documentation to learn
...
+### Step 4: diagramming how our Objects will use Messages to communicate with one another
 
-* **Submit a pull request early.**
+```
+Plane <-- land  --> Airport
+Plane <-- take-off  --> Airport
+Plane <-- at_airport? --> Airport
+Airport <-- full? --> Plane
+Weather <-- is_stormy? --> Plane
+```
```
eoinmakers, line 135: It's not immediately clear to me what information this is relaying, but if it was useful for you in planning then that's the main thing (perhaps you could illustrate it further with a small blurb of text?)

pav0107: To be honest, I tried to replicate the diagram from Boris Bikes but didn't feel confident it was right. From now on I think I'll use Class Diagrams as I find them more intuitive.

PR #2161, opened 2021-07-04
Izzy Airport Challenge. 

isobelpepp:


karimabuseer:
Great airport challenge Izzy!
Airport.rb
* Line 5 - What do you think the benefits of having a constant capacity is?
* Line 15 - Is it common convention to surround raises by blank lines?
* Big fan of your SRP (etc full?)

Weather.rb
* What do you think about alternative random methods?
* Line 25 - Consider moving end up one line

Airport_spec.rb
* Line 6 - What were the motivations for setting up a double for plane?
* Line 14 - Consider adding a newline after line 14

Plane_spec.rb
* Is there a way to test that a class exists?

PR #2162, opened 2021-07-04
Matt Firmston solution for airport challenge

YoFirmy:


lib/plane.rb
```diff
@@ -0,0 +1,24 @@
+class Plane
+  attr_reader :airport
+  DEFAULT_POSITION = 'In flight'
+  
+  def initialize
+    @airport = DEFAULT_POSITION
```
eoinmakers, line 6: I think it's slightly confusing for @airport to be referring to either 'In flight' or an instance of airport. I would have separate variables to store these values, for readability.

lib/plane.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def land(airport)
+    fail 'The weather is stormy' if airport.stormy?
+    fail 'Airport is full' if airport.full?
+
+    @airport = airport
+    airport.planes << self
```
eoinmakers, line 14: If you need to interact with another class in this way, ie. passing some data in, it's safer to expose it via a method rather than with an attribute reader. If you were to create a method in Airport that has the responsibility of passing a plane in to this array, that you could call from this method, it would serve the necessary purpose - it's safer as this is the only means by which something could interact with the array, as opposed to with the attribute reader where it can be accessed directly.

spec/feature_spec.rb
```diff
@@ -0,0 +1,9 @@
+require './lib/plane'
+require './lib/airport'
+
+# airport = Airport.new
+# (Airport::DEFAULT_CAPACITY - 1).times { Plane.new.land(airport) }
+# test_plane = Plane.new
+# test_plane.land(airport)
+# test_plane.take_off
+# Plane.new.land(airport)
```
eoinmakers, line 9: It's good to remove any extraneous or commented out code before opening a PR, for cleanliness.

eoinmakers: (If you have something in progress or mark something 'TO-DO', that's fine.

spec/plane_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    it { is_expected.to respond_to(:land).with(1).argument }
+
+    it "should raise an error when airport is full" do
+      airport = double(:airport, :full? => true, :stormy? => false)
+      expect { subject.land(airport) }.to raise_error 'Airport is full'
+    end
+
+    it "should raise error if weather is stormy" do
+      airport = double(:airport, :full? => false, :planes => [], :stormy? => true)
+      expect { subject.land(airport) }.to raise_error 'The weather is stormy'
```
eoinmakers, line 14: Nice tests for error checking.

eoinmakers (commented):
I thought this was very well written, well done. The classes you have defined are cleanly written and concise, although one small piece of feedback is that it's good to draw out weather into its own class, as its responsibility differs slightly to the others inside the Airport class. 

It's also good to include context in your ReadMe around the user stories you've implemented, and any design or planning stages you worked through too, to help relay your thought process.

PR #2163, opened 2021-07-04
Completed 5/6 User Stories

senoodle:
1,2,3,5,6 completed.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,2 @@
+class Airport
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,2 +1,33 @@
-class Airport
+require_relative 'plane'
+
+class Airport 
+
+  DEFAULT_CAPACITY = 50
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @amount_of_planes_landed =0
+    @planes =[]
+  end
+
+  def land(plane)
+    raise 'airport is full' if full?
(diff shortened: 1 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+    raise 'airport is full' if full?
+    raise 'cannot land plane, weather stormy' if stormy?
+    @planes << plane 
+  end
+
+  def take_off(plane)
+    raise 'cannot take off plane, weather stormy' if stormy?
+  end
+
+  private
```
pav0107, line 23: Good use of private methods

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  
+  subject(:airport) { described_class.new(50)}
```
pav0107, line 6: You've got a much better grasp of the terminology to use than I have! I'll need to look into this and the next line.

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    it 'responds to land' do
+    expect(airport).to respond_to(:land).with(1).argument
+    end
+
+    it 'prevents landing when airport is full' do
+      allow(airport).to receive(:stormy?).and_return false
+      50.times do
+        airport.land(plane)
+    end
+      expect { airport.land(plane)}.to raise_error 'airport is full'
```
pav0107, line 20: Tiny, tiny point to raise is that there are a few inconsistencies with spaces with the opening/closing curly braces.

PR #2164, opened 2021-07-04
Luke Airport Challenge (4/5)

lushcodes:


spec/plane_spec.rb
```diff
@@ -2,35 +2,34 @@
 require 'airport'
 
 describe Plane do
+  let(:airport) { Airport.new }
```
BrettKarlson, line 5: Nice.Maybe subject {} for the first one

spec/plane_spec.rb
```diff
@@ -2,35 +2,34 @@
...
-    subject.land(airport)
+    plane_lands
     subject.take_off(airport)
     expect(airport.hangar).to eq([subject] - [subject])
   end
   describe '#take_off' do
     it 'should confirm that the plane has left the airport' do
-      airport = Airport.new
-      subject.land(airport)
+      plane_lands
```
BrettKarlson, line 20: Good nested describe
Context would not have been as appropriate

lib/airport.rb
```diff
@@ -0,0 +1,14 @@
...
+  attr_accessor :capacity
+  CAPACITY = 2000
+
+  def initialize(capacity = CAPACITY)
+    @hangar = []
+    @capacity = capacity
+  end
+
+  def full?
+    (@hangar.length >= capacity) == true
```
DontKnowZak, line 12: `== true` isn't needed
the comparison operator will return `true` or `false` already

lib/plane.rb
```diff
@@ -0,0 +1,13 @@
+require_relative 'airport'
+
+class Plane
+  def land(airport)
+    (raise "Cannot land - Airport full") if airport.full? == true
```
DontKnowZak, line 5: same as above - `.full?` will return a boolean value so `if airport.full?` is enough

lib/airport.rb
lushcodes: Fair comment.

spec/airport_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+  describe '#initialize' do
```
DontKnowZak, line 6: consider adding a test that checks the capacity is 2000 as expected, or equal to what you pass in
similar to the check that hangar is an empty array at start

PR #2167, opened 2021-07-05
Airport Challenge

JEC1100:


lib/plane.rb
```diff
@@ -1,10 +1,11 @@
 require_relative 'airport'
 
+
 class Plane
-    # def initialize
-    #     @plane = plane
-    # end
 
     def land(airport)
+
```
William-Young-97, line 7: Does land belong here or in airport.rb?

William-Young-97 (commented):
Good stuff James; most of the user stories are fulfilled and you have tests covering them. A small thing that you could do at the end is go through your code and clean up blank lines and see whether you are assigning the same variables multiple times. Also, make sure to add a README.md at the end of each of your projects.

JEC1100:
Second pull request following initial pull request on challenge completion date (5/7/2021)

JEC1100:
Second pull request following initial pull request on challenge completion date (5/7/2021)

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+    @planes << plane
+  end
+        
+  def take_off(airport)
+    raise "Denied permission to take off due to storm" if stormy? 
+  end
+
+  def in_air?(plane)
+    @plane == plane
+  end
```
JohnForster, line 30: Have a think about what this method is for. It seems to be the only usage of the airport's @plane instance variable. It might be worth thinking about what is the purpose of @plane in this class? What real-world plane does it represent? How does it differ from @planes? Do you need it at all?

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+
+describe Airport do
+  let(:plane) { Plane.new }
+
+  describe '#land' do
+    context 'When not stormy' do
+        before do
+            allow(subject).to receive(:stormy?).and_return false
+        end
```
JohnForster, line 12: Good mocking of the `stormy?` method here. 👍

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      
+          it 'raises error if the airport is full' do
+            allow(subject).to receive(:stormy?).and_return false
+            11.times do
+              subject.land(plane)
+            end
+            expect { subject.land(plane) }.to raise_error "Full to capacity: Landing Denied"   
+          end
+  end
```
JohnForster, line 26: Be careful with your indentation. Without keeping your code neatly indented, it can be confusing to match your `do`s with your `end`s. You can use rubocop to help you with this.

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        end
+      end
+
+         
+    context 'when stormy' do
+      it 'raises an error if asked to take_off a plane in a storm' do
+       allow(subject).to receive(:stormy?).and_return true
+        expect { subject.take_off(plane) }.to raise_error "Denied permission to take off due to storm"
+    end
+  end
```
JohnForster, line 41: You could probably combine these 'when stormy' contexts into one, using the `before` method used on line 10

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      expect(subject).to respond_to(:take_off).with(1).argument
+    end
+  end
+
+  it 'Does method to check if plane in airport?' do
+    plane = Plane.new
+    subject.land(plane)
+    subject.take_off(plane)
+    expect(subject.in_air?(plane)).to eq(false)
+  end
```
JohnForster, line 55: I'm not 100% sure what this test is testing.
I'd expect that a plane _would_ be in the air after taking off, but here you're asserting that as being false. Is that a mistake in the test, or a mistake in the code?
Either way, is `in_air?` something that an airport would be responsible for?

JohnForster: Could we also test whether or not the plane is still present in the airport's `@planes` instance variable?

spec/plane_spec.rb
```diff
@@ -2,7 +2,6 @@
 require 'airport'
 
 describe Plane do
-    subject(:plane) { described_class.new }
+  subject(:plane) { described_class.new }
```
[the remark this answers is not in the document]
JohnForster: Seconded ☝️ A good README is invaluable, especially when getting your code reviewed.

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+    DEFAULT_CAPACITY = 11
+    attr_reader :plane, :capacity
+
+    def capacity
+      @capacity
+    end
+
+    # attr_reader :capacity, :planes
+  
+    def initialize(capacity = DEFAULT_CAPACITY)
```
JohnForster, line 13: Nice usage of a default argument here. Could you add a test for different size airports?

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'plane'
+
+class Airport
+    DEFAULT_CAPACITY = 11
+    attr_reader :plane, :capacity
+
+    def capacity
+      @capacity
+    end
```
JohnForster, line 9: This function is _exactly_ what `attr_reader :capacity` is doing. You don't need both.

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def land(plane)
+    raise "Denied permission to land due to storm" if stormy? 
+    raise "Full to capacity: Landing Denied" if full?
+    @planes << plane
+  end
+        
+  def take_off(airport)
+    raise "Denied permission to take off due to storm" if stormy? 
+  end
```
JohnForster, line 26: I don't think you should be taking an airport as an argument here. I think this instead should be the plane that you want to take off. It essentially needs to be the opposite of the land method above. One adds it to this airport's `@planes` array, the other removes it from that array.

lib/airport.rb
```diff
@@ -1,19 +1,29 @@
...
+    @planes << plane
+  end
         
-    def take_off(airport)
-    end
+  def take_off(airport)
+  end
+
+    
+
```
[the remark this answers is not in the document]
JohnForster: Seconded! 💯

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+          end
+        end
+      
+          it 'raises error if the airport is full' do
+            allow(subject).to receive(:stormy?).and_return false
+            11.times do
+              subject.land(plane)
+            end
+            expect { subject.land(plane) }.to raise_error "Full to capacity: Landing Denied"   
+          end
```
JohnForster, line 25: You have good tests for what happens if `land` fails. I think it would be a good idea to add one for if land succeeds. Something along the lines of
```rb
it 'adds a plane to its planes array' do
    allow(subject).to receive(:stormy?).and_return false
    plane = Plane.new
    subject.land(plane)
    expect (subject.planes).to include plane
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
+
+require 'airport'
+
+
+describe Airport do
+  let(:plane) { Plane.new }
+
+  describe '#land' do
+    context 'When not stormy' do
```
JohnForster, line 9: Good use of the context/describe blocks. 👍

JohnForster (commented):
Overall a lot of promising work here.

I see that you mentioned it still giving an error after taking_off a couple of planes. Can you see anything missing in that function? Maybe when compared to the `land` function above?

If you were to TDD a test for the take_off method (ignoring weather for now) what might that test look like?

Feel free to contact me if you have any more questions.

PR #2169, opened 2021-07-23

hjdalton:
Partially complete, a few challenges to overcome.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,2 @@
+class Airport
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,29 +1,34 @@
 class Airport
-  attr_accessor :capacity
+attr_accessor :capacity
 
-  def initialize
-    @capacity = []
+  def initialize(capacity)
+    @capacity = capacity
+    @planes= []
   end
 
   def landing(plane)
     fail 'No space to land' if full?
 
-    fail 'Weather is stormy' if weather == "stormy"
+    fail 'Unsafe to land, bad weather' if weather?
 
-    @capacity << plane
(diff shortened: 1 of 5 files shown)


lib/airport.rb
```diff
@@ -1,15 +1,14 @@
...
   def initialize(capacity)
     @capacity = capacity
-    @planes= []
+    @planes = []
   end
 
   def landing(plane)
     fail 'No space to land' if full?
-
     fail 'Unsafe to land, bad weather' if weather?
```
NazirShaikh9093, line 12: Weather is returning true or false which is for "Stormy" and "Sunny" respectively but if someone else looked at this they'd have no idea why weather is returning a number and what it means if its bigger or less than 8.

lib/airport.rb
```diff
@@ -1,15 +1,14 @@
 class Airport
-attr_accessor :capacity
-attr_reader :planes
+  attr_accessor :capacity
```
NazirShaikh9093, line 2: Good job using attr_accessor for capacity as you'll have to both allow it to be read and re-written if needed.

lib/airport.rb
```diff
@@ -1,15 +1,14 @@
...
-    @planes= []
+    @planes = []
   end
 
   def landing(plane)
     fail 'No space to land' if full?
-
     fail 'Unsafe to land, bad weather' if weather?
 
     @planes << plane
```
NazirShaikh9093, line 14: In your take_off method, line 20 will never be rain and so the plane will never be deleted from the array, this is because there is a return statement above it that will return plane and stop executing the method code. If you swap line 19 and 20 it should be fine :)

PR #2173, opened 2021-07-26
airport challenge

ibnasif:


airport.rb
```diff
@@ -0,0 +1,46 @@
+class Airport
+  attr_accessor :landed_planes, :default_capacity
```
Dylan-P-H, line 2: Nice decision using attr_accessor instead of reader, probably covers a lot of unnecessary code in future.

airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  def change_capacity(new_capacity)
+    @default_capacity = new_capacity
+  end
+
+  def stormy?
+    rand(1..6) > 4 #if rand is greater than 4 it returns true
+
+  end
+
+
```
Dylan-P-H, line 42: I don't know if we are supposed to be reviewing the actual files but it looked like you didn't have a lib folder which could be an easy way to make the project neater from an outside perspective.

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
+require_relative  "../airport.rb"
+require_relative  "../plane.rb"
+
+
+describe Airport do
```
Dylan-P-H, line 5: Under describe you can add something like:

let(:airport) {Airport.new}
let(:plane) {Plane.new}

Just so you don't have to keep repeating declaring the instances inside of it blocks in the spec files.

PR #2174, opened 2021-07-26
Airport Challenge

WesleyHindle:


lib/airport.rb
```diff
@@ -0,0 +1,47 @@
+class Airport
+  attr_accessor :plane_no
```
manumoualek, line 2: I did not find this attribute being called/referenced in any other part of the code? Is it redundant?

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    @planes.select { |airplane| airplane != plane }
+    "#{plane} has departed"
+  end
+  
+  def hangar(plane)  #this is just doing the same as take_off, but with difference text
+    @planes.select { |airplane| airplane != plane }
+    "#{plane} is no longer in the hangar"
+  end
+
+  def add_plane(plane)
```
manumoualek, line 23: It seems like you can push any argument to the add_plane method. This may be a cause for problems as we would only want planes to land. Adding a plane class would help in this case I believe.

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    @planes.push(plane)
+  end
+
```
manumoualek, line 26: It seems like you can push any argument to the add_plane method. This may be a cause for problems as we would only want planes to land. Adding a plane class would help in this case I believe.

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+  def full_hangar?(len)
+    fail 'The hangar is full' unless @planes.length > @capacity
+    # else 'There is space available in the hangar'
+  end
+
+  def airport_capacity(size)
+    @capacity = size
+    "The airport's capacity is #{@capacity}"
+  end
+
```
manumoualek, line 36: I believe you can bypass having a method specifically for changing the airport's capacity by having @capacity as an attr_reader and you can also make it an argument in the initialize method so that you can set the capacity you want the moment you instantiate the class.

PR #2176, opened 2021-07-26
airport-challenge

manumoualek:


spec/airport_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      airport = Airport.new(false)
+      expect(airport.hangar_capacity).to eq(20)
+    end
+  end
+
+  describe "#land" do
+    
+    it "responds to land plane method" do
+      airport = Airport.new(false)
+      expect(airport).to respond_to(:land).with(1).argument
```
WesleyHindle, line 23: This is really nice, super succinct test.

spec/airport_spec.rb
```diff
@@ -0,0 +1,95 @@
+require_relative "../lib/airport.rb"
+require_relative "../lib/plane.rb"
+
+describe Airport do
+
```
WesleyHindle, line 5: Could use:  let(:instance_of_new_var) {ClassName.new}
Then you wouldn't have to do airport.Airport.new() each time.

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+  end 
+
+  def is_stormy?
+    weather_number = rand(1..10)
+    weather_number > 8 ? true : false
+  end  
+
+  private
+
+  def full?
```
WesleyHindle, line 47: Could the 'full?' and 'empty?' methods become one function by using an if statement?

spec/airport_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    it "responds to take_off plane method" do
+      airport = Airport.new(false)
+      expect(airport).to respond_to(:take_off)
+    end
+    
+    it "allows a landed plane to take_off from hangar" do
+      airport = Airport.new(false)
+      airport.land(Plane.new)
+      expect(airport.take_off).to be_instance_of(Plane)
```
WesleyHindle, line 54: Nice use of this matcher.

spec/airport_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    it "raises an error when the hangar is empty(0 planes in hangar)" do
+      airport = Airport.new(false)
+      expect{airport.take_off}.to raise_error("Cannot take off, there are no planes in the hangar")
```
WesleyHindle, line 59: I'm unsure how this would catch there being no planes as the default hangar capacity is 20

WesleyHindle (changes requested):

spec/airport_spec.rb
manumoualek: 🙏

spec/airport_spec.rb
manumoualek: Thanks for this! Made the changes here :)

lib/airport.rb
manumoualek: They could! I copied the boris-bikes work where both of these methods were separated. It makes it easier to know what is being asked of the airport instance I think.

spec/airport_spec.rb
manumoualek: Thanks !

spec/airport_spec.rb
manumoualek: The default capacity is 20(up to 20 planes can be added in the array) but it starts with 0 planes within.

PR #2177, opened 2021-07-26
Pull request for Makers

Dylan-P-H:


ibnasif:
I like the way you used hangar.length instead of creating a variable for it, saves a lot of code.

ibnasif:
I think writing your own READme file would be good to show your processes and approach

ibnasif:
instead of .pop you could use.delete(plane) which deletes the specific plane whereas the former deletes the last element of the array

PR #2182, opened 2021-07-26
Tyler Airport Challenge 

tylercws:


lib/airplane.rb
```diff
@@ -0,0 +1,19 @@
+# Assume planes that are made aren't at the airport and are in flight at first
+
+class Airplane
+  attr_accessor :planes_at_airport, :flying
```
annesamariyam, line 4: I like how you've used an attribute accessor to make your code easier to use

spec/spec_traffic_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  it 'rejects a airplane when it reaches max' do
+    A1 = Airplane.new()
+    subject.land_plane(A1)
+    A2 = Airplane.new()
+    subject.land_plane(A2)
+    A3 = Airplane.new()
+    expect(subject.land_plane(A3)).to eq ("Plane cannot land due to the max capacity being reached")
+  end
+
```
annesamariyam, line 25: ^ there could be an easier way to do this - it would be hard work if your capacity was a lot bigger. I'm not sure how, but would be worth looking into

PR #2187, opened 2021-07-31
Weekend challenge submission - Amelia

sa-mcquanzie:


spec/airport_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  before(:each) do
+    @airport = Airport.new(1, :clear)
+    @stormy_airport = Airport.new(1, :stormy, [Plane.new])
+    @plane_on_ground = Plane.new
+    @plane_in_flight = Plane.new true
```
eoinmakers, line 9: For these variable definitions, I would use 'let' to define them in your tests, instead of before(:each). It's a slightly more efficient implementation, and doesn't require the use of instance variables. 

https://medium.com/@tomkadwill/all-about-rspec-let-a3b642e08d39

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    @planes = planes
+  end
+
+  def receive plane
+    clear_for_landing plane
+    plane.land
+    @planes << plane
+  end
+
+  def release plane
```
eoinmakers, line 18: 'receive' and 'release' method names, although fine, are a little less descriptive than the actions they're carrying out - taking off and landing.

lib/plane.rb
```diff
@@ -0,0 +1,17 @@
+class Plane
+  def initialize in_flight = false
+    @in_flight = in_flight
+  end
+
+  def in_flight?
+    @in_flight
+  end
+
+  def land
```
eoinmakers, line 10: If you change your method names in Airport to land and take_off, you'll want to rename these ones too. As the Airport methods handle a little bit more of the actions that go into landing or allowing a plane to take off, compared to these ones - which update the status of the plane to @in_flight or not, slightly more descriptive method names could be something to that effect. 

eg. update_flight_status (updating the boolean to the opposite of its current value)
or record_airborne vs record_grounded, for example.

eoinmakers (commented):
This is an excellent airport challenge submission, nice work. I think there are clearly defined responsibilities and separation of concerns, and really comprehensive unit testing, including taking edge cases into account. Great work!

PR #2190, opened 2021-08-01
Pull Request for Feedback

Battery0:
Pull Request for feedback on week 1's weekend challenge - Airport


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
+require './lib/plane.rb'
+
+class Airport
+
+  DEFAULT_CAPACITY = 20
+
+  attr_reader :airport, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airport = []
```
eoinmakers, line 10: I would avoid sharing a name between an instance variable and a class. This is representing the hangar that planes will be stored in, so you can name it something descriptive like @planes, @hangar, @plane_storage, etc. instead.

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+      plane.flying = false
+      @airport << plane
+    else
+      raise "The airport is full and can't land anymore planes"
+    end
+  end
+
+  def take_off(plane)
+    plane.grounded = true
+    if !self.airport.empty? && plane.grounded == true
```
eoinmakers, line 27: I don't think you need to refer to self here, as @airport (your hangar instance variable) should be accessible when called directly.

eoinmakers: Equally as in your land method, plane.grounded will always be true, as you set it to true in the line above. This makes it unnecessary, it will never default to true.

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  it 'responds to #land method call' do
+    expect(subject).to respond_to(:land).with(1).arguments
+  end
+
+  # test - check a plane can land at the airport
+  # (plane object added to @airport instance variable (an array))
+  it 'plane lands at airport' do
+    subject.land(airplane)
+    expect(subject.airport).to include(airplane)
+    expect(airplane.grounded).to be true
```
eoinmakers, line 27: The expectation that the airplane has been grounded would rest more with Plane tests, I believe. The test that the airport now includes the plane will be sufficient here.

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  # test - raises error if calling #take_off with no planes at the airport
+  it 'raises error when no planes at airport try to take off' do
+    # subject.take_off(airplane) ## Add this line in to make this test work when using 'rspec' command
```
eoinmakers, line 32: Does this not work with the rspec command otherwise?

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    expect(subject.airport).not_to include(airplane)
+  end
+
+  # test - confirm plane object has left airport
+  it 'confirm plane has left the airport' do
+    subject.land(airplane)
+    expect { subject.take_off(airplane) }.to output("The plane has left the airport\n").to_stdout
+  end
+
+  # test - Error message if more than 20 planes try to land at the airport
```
eoinmakers, line 55: The descriptions of the tests between 'it' and 'do' should suffice, so you don't need comments to supplement usually

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  # test for default airport capacity
+  it 'has a default capacity' do
+    expect(subject.capacity).to eq(Airport::DEFAULT_CAPACITY)
+  end 
+
+  # This is as far as I got - I wasn't able to implement testing for variable capacity
+  #
+  # it 'has a variable capacity' do
+  #   subject(50)
```
eoinmakers, line 71: In order to pass an argument to subject, you need to redefine it with 'let', or else just create the instance manually for this one test: airport = Airport.new(50), otherwise I think it may be going with the default one (possibly why the test wasn't working)

eoinmakers (commented):
Nice work, Ian. :) I think you did well to implement most of the user stories successfully! The main chunk of logic I would feed back on is the implementation of the 'grounded' or 'flying' status for the plane, as I don't think it affects enough of the code to warrant its inclusion in its current form. But well done, good progress here.

PR #2191, opened 2021-08-01
Week 1 weekend challenge

Nicola-Carroll:


spec/plane_spec.rb
```diff
@@ -0,0 +1,5 @@
+require './lib/plane'
```
Xfarya, line 1: slight-syntax: can you not require just 'plane' ?

lib/airport.rb
```diff
@@ -8,12 +8,29 @@ def initialize(capacity = DEFAULT_CAPACITY)
     @planes = []
   end
 
-  def land(plane, weather)
```
Xfarya: incomplete method

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+class Airport
```
Xfarya, line 1: This code is super-clear - good use of doubles and appropriate rspec predicates - all user stories are accounted for - although I didn't understand the use of "parked" exactly?.. I know the Weather Class is a good separation of concerns, and is very clear in what it does, but could there be a more succinct way to present this function, perhaps in the airport class. Also, (just wondering :) ), there are some instances of almost identical lines, is there a way to refactor these? And finally, as you already mentioned during the chat, the program tells you a lot about errors, but there don't seem to be many indications of what happens when things are going right :)

PR #2192, opened 2021-08-01
Airport_Challenge Solution 

gjstirling:
Could use feedback on my work, Thank you !


lib/airport.rb
```diff
@@ -0,0 +1,8 @@
+
+class Airport
```
gjstirling, line 1: Good

PR #2193, opened 2021-08-01
Dan Roche pull request

danroche10:
You can see my approach to the challenge in the README.


Gemfile
```diff
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 group :test do
+  gem 'capybara'
```
eballantine, line 4: Curious what capybara is/does! You'll have to show me

README.md
```diff
@@ -1,3 +1,27 @@
+**Dan's notes on challenge**
```
eballantine, line 1: This is nice! README.md should also have instructions for someone that wants to run your code, and run your tests.

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+
+  DEFAULT_CAPACITY = 5
+
+  attr :capacity
```
eballantine, line 8: What is attr on it's own? Should this be attr_reader attr_writer or attr_accessor?

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def land(plane)
+    fail 'Airport is at full capacity' if full?
+
+    fail "Plane is already at airport" if plane_in_hangar?(plane)
+
+    fail "Plane has already landed somewhere else" if plane.landed
+
+    fail 'Weather is too bad' if stormy?
```
eballantine, line 23: I like these fails, very clear what they're checking for! As Corinne suggested this morning, perhaps can be separated out into something like a clear_to_land? method.

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
+    @planes = []
+    @weather = weather
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail 'Airport is at full capacity' if full?
+
+    fail "Plane is already at airport" if plane_in_hangar?(plane)
```
eballantine, line 19: This guard becomes redundant thanks to the next one. If the plane is already landed somewhere, that includes it's in this hangar. Perhaps helpful in the future to know which airport it's in, but right now is redundant.

lib/plane.rb
```diff
@@ -0,0 +1,16 @@
+class Plane
```
eballantine, line 1: This class is _lovely_ - very clear, very neat. 10/10, gold star, etc.

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
+
+  def stormy?
+    weather_today == :stormy
+  end
+
+  private
+
+  WEATHER_CHART = [:sunny, :stormy, :sunny, :stormy, :sunny, :sunny, :sunny]
```
eballantine, line 9: Minor thing but there's 2 stormy out of 7 here, maybe put them together at start or end of the list for easier reading. Or some other way to quickly convey to reader, 2x stormy & 5x sunny.

spec/airport_spec.rb
```diff
@@ -0,0 +1,140 @@
+require 'airport'
+require 'plane'
```
eballantine, line 2: I think when you're testing Airport class you only need to require airport, and planes should be mocked?

spec/airport_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    it "enables user to override airport capacity" do
+      another_airport = Airport.new(10)
+      expect(another_airport.instance_variable_get(:@capacity)).to eq 10
+    end
+  end
+
+  describe "#lands planes" do
+
+    it "plane that has landed can be found in the airport" do
+      allow(@my_airport.send(:weather)).to receive(:stormy?) { false }
```
eballantine, line 25: This line is repeated in future it blocks. Any way to DRY it up?

spec/airport_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+      @my_airport.send(:planes) << @my_plane
+      @my_airport.take_off(@my_plane)
+      message = "this plane isn't at the airport"
+      expect { @my_airport.take_off(@my_plane) }.to raise_error message
+    end
+  end
+end
+
+describe "#feature test with two airports, three planes and only sunny weather", type: :feature do
```
eballantine, line 93: This is cool, I didn't get as far as feature test, so I'm reading, taking notes, and would love if you could quickly talk me through

spec/weather_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+describe Weather do
+
+  it 'provides the weather from a few different options' do
+    my_airport = Airport.new
+    expect(my_airport.send(:weather).send(:weather_today)).to eq(:stormy).or(eq(:sunny))
+  end
+
+  it "returns false if the weather is sunny" do
+
+    my_airport = Airport.new
```
eballantine, line 12: Throughout, but just commenting on this one as an example, you haven't used doubles. If there was a problem with your Airport class code, your weather tests might start failing because of Airport, rather than because of Weather. To isolate tests, you should make a fictitious 'airport' double that only does what you tell it in the test. i.e. if you need it to give you weather, you can still use something like line 7 to do that.

eballantine (commented):
Nice Dan. Some comments for you to take a look at whilst I try to run code with test suit now...

PR #2194, opened 2021-08-01
Weekend challenge: week one

edpackard:


lib/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+    @plane = plane
+    take_off_checks
+    plane_on_ground?(false)
+  end
+
+  def contains?(plane)
+    hangar.include?(plane)
+  end
+
+  private
```
ashtroboy79, line 34: good use of private methods to keep things secure

spec/airport_spec.rb
```diff
@@ -0,0 +1,103 @@
+require 'airport'
+
+describe Airport do
+  
+  let(:plane) { double :plane }
+
```
ashtroboy79, line 6: liking the use of before to set the test conditions, keeping the tests nicely DRY.  All tests are clear and concise

spec/airport_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+
+  it "allows user to set custom capacity when instantiating new airport object" do
+    expect(Airport.new(100).capacity).to eq 100
+  end
+
+  it "can land planes up to a custom capacity without error" do
+    airport = Airport.new(100)
+    allow(airport.weather).to receive(:stormy?) { false }
```
ashtroboy79, line 59: do you need this allow as you've set it up at the start using before? Its a question as im not 100% sure myself

spec/airport_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    expect { airport.capacity.times { airport.land(plane) } }.not_to raise_error
+  end
+
+  it "raises an error if custom capacity exceeded" do
+    airport = Airport.new(50)
+    allow(airport.weather).to receive(:stormy?) { false }
```
ashtroboy79, line 65: same as line 59.

ashtroboy79 (commented):
Looks like a very nice and clearly written program, with tests that cover everything, good use of SRP and DRY, in both rspec and ruby. Good use of before & let to arrange all the preconditions for your tests.

lib/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    raise "Airport capacity must be a positive integer" if not_positive_integer?(capacity)
+    
+    @capacity = capacity
+    @hangar = Array.new
+    @weather = Weather.new
+  end
+
+  def land(plane)
```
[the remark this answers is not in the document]
edpackard: thanks - not sure I've totally got the hang of SRP yet, but this was an attempt to keep the methods small without losing coherence!

spec/airport_spec.rb
edpackard: good question! the `before` set up at the start only works with `subject`, so when I create a new Airport like this I have to mock the behaviour manually.

PR #2195, opened 2021-08-01
Airport Challenge CB AIR

CorinneBosch:
Submitting pull request for feedback

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,36 +3,44 @@ Airport Challenge
 
 ```
         ______
-        _\____\___
-=  = ==(____MA____)
-          \_____\___________________,-~~~~~~~`-.._
-          /     o o o o o o o o o o o o o o o o  |\_
-          `~-.__       __..----..__                  )
-                `---~~\___________/------------`````
-                =  ===(_________)
+        _\____\_____
+=  = ==(__CB AIR____)
+          \_____\___________________________,-~~~~~~~`-.._
+          /     o o o o o o o o o o o o o o o o o o o o  |\_
+          `~-.__       __..----..__                          )
+                `---~~\___________/---------------------`````
+          =      =  ===(_________)
 
 ```
+## Ladies and gentlemen welcome on board! 
+#### Thank you for flying CB AIR. I hope you have a pleasant stay, please fasten your seatbelts.
 
-Instructions
----------
+## What it does
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+## Setup
 
-Steps
--------
+1. **Fork** this [airport challenge repository](https://github.com/CorinneBosch/airport_challenge) 
+2. Then clone **your** fork to your computer.
+3. Ensure you have `Ruby` and `rspec` installed. 
 
-1. Fork this repo, and clone to your local machine
-2. Run the command `gem install bundler` (if you don't have bundler already)
-3. When the installation completes, run `bundle`
-4. Complete the following task:
+## How to run
 
-Task
------
+You can frature test and run this airport program in any `code editor` or `IRB`.
 
-We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
+## Run test 
+
+To test specific features in your terminal.
+```
+$ cd airport_challenge/lib
+$ rspec spec/file_name.rb
+```
+To test all features at once in your terminal.
+```
+$ cd airport_challenge
+$ rspec
+```
+
+## The airport challenge is built upon the following user stories:
 
 ```
 As an air traffic controller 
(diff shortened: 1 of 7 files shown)


spec/spec_plane.rb
```diff
@@ -0,0 +1,8 @@
+require 'plane.rb'
+
+describe Plane do
+  it { is_expected.to be_instance_of Plane } 
+  it { is_expected.to respond_to(:flying?) } 
+  it { is_expected.to respond_to(:landing) } 
+  it { is_expected.to respond_to(:offtake) } 
+end
```
yousufmohamed92, line 8: You could also have tested for the return values of the methods.

spec/spec_weather.rb
```diff
@@ -0,0 +1,10 @@
+require 'weather.rb'
+
+describe Weather do
+  before(:each) do
+    @subject = Weather.new
+  end
```
yousufmohamed92, line 6: Is the before necessary? Subject is already implicit.

spec/spec_weather.rb
```diff
@@ -0,0 +1,10 @@
...
+  
+  it { is_expected.to be_instance_of Weather } 
+  it { is_expected.to respond_to(:stormy?) } 
+end
```
yousufmohamed92, line 10: You could also test for the return value of #stormy? method.

spec/spec_airport.rb
```diff
@@ -0,0 +1,90 @@
...
+    #   message = 'We allow only planes to takeoff at this airport!'
+    #   expect {@airport.takeoff(chopper)}.to raise_error message
+    # end
+
+    it 'refuse takeoff in empty hangar ' do
+      @airport.hangar = []
+      puts @airport.hangar
+      message = 'There are no planes in our hangar'
+      expect { @airport.takeoff(@plane)}.to raise_error message
+    end
```
yousufmohamed92, line 81: @plane instance is created airborne so you are getting another error message. You need to use a landed plane instead.

spec/spec_airport.rb
```diff
@@ -0,0 +1,90 @@
...
+    #   allow(chopper).to receive(:flying?) {false}
+    #   allow(chopper).to receive(:takeoff)
+      
+    #   message = 'We allow only planes to takeoff at this airport!'
+    #   expect {@airport.takeoff(chopper)}.to raise_error message
+    # end
+
+    it 'refuse takeoff in empty hangar ' do
+      @airport.hangar = []
+      puts @airport.hangar
```
yousufmohamed92, line 78: Delete puts.

README.md
```diff
@@ -3,36 +3,64 @@ Airport Challenge
...
-          `~-.__       __..----..__                  )
-                `---~~\___________/------------`````
-                =  ===(_________)
+        _\____\_____
+=  = ==(__CB AIR____)
+          \_____\___________________________,-~~~~~~~`-.._
+          /     o o o o o o o o o o o o o o o o o o o o  |\_
+          `~-.__       __..----..__                          )
+                `---~~\___________/---------------------`````
+          =      =  ===(_________)
```
yousufmohamed92, line 13: I admire the customisation.

lib/plane.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def landing
+    @airborne = false
+  end
+
+  def offtake
+    @airborne = true
+  end
+end
```
yousufmohamed92, line 20: I like this class. Very simple and efficient.

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  def hangar_full?
+    @hangar.count >= DEFAULT_CAPACITY
+  end
+
+  def hangar_empty?
+    @hangar.count.zero?
+  end
+
+  def weather
+    Weather.new
```
eoinmakers, line 38: In creating an instance of one class in another in this way, it can work and be fully functional. But the reason it's not _optimal_ is because it means the classes are tightly coupled. 

The way around this tight coupling is by using something called dependency injection. To learn more about it, there's a practical I'll recommend for when you have the time, and share an example of how you might change it: 

```
# slight addition to your initialize method in your Airport class
    @weather = weather
end

def weather 
    @weather.get_forecast
end
```

In this example, get_forecast is an example of what the method in your Weather class might be called that actually returns the result. Dependency injection allows you to pass in alternative instances, which means you can fully isolate your tests from one another with mocking, or pass in a different Weather class if needed (lets say you had a different Weather class for each continent). 

And as I included weather = Weather.new in a default argument, it means that if I don't pass in any instance to this argument when creating the Airport instance, it will by default use your Weather class.

spec/airport_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+  let(:plane) { instance_double(Plane) }
+  let(:weather) { instance_double(Weather) }
+  
+
+  before do
+    allow(plane).to receive(:instance_of?).with(Plane).and_return(true)
+    allow(plane).to receive(:flying?).and_return(true)
+    allow(plane).to receive(:landing)
+    allow(plane).to receive(:offtake)
+    allow(weather).to receive(:stormy?) { false }
```
eoinmakers, line 15: Great work with this stubbing!

spec/airport_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      expect(zurich.capacity).to eq 25
+    end
+  end
+
+  describe '#land' do
+    let(:forecast) { instance_double(Weather) }
+
+    it 'stores landed plane in hangar' do
+      subject.land(plane)
+      expect(subject.hangar.empty?).to eq(false)
```
eoinmakers, line 32: It might be slightly more conclusive if you do test that the hangar now contains the Plane instance passed in (as opposed to some other object)

spec/airport_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    it "lands plane objects only" do
+      message = 'We accept only planes at this airport!'
+      expect { subject.land('Turtle') }.to raise_error message
```
eoinmakers, line 37: This is great!

spec/airport_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      12.times { subject.land(plane) }
+      message = 'Our hangar is full, please land somewhere else'
+      expect { subject.land(plane) }.to raise_error message
+    end
+
+    # it 'refuses landing when stormy' do
+    #   allow(forecast).to receive(:stormy?) { true }
+    #   message = 'The landing is delayed due to stormy weather'
+    #   expect { subject.land(plane) }.to raise_error message
+    # end
```
eoinmakers, line 56: The reason why perhaps this test wouldn't work in this instance, is that a fresh instance of Weather is created directly in your Airport class with 'Weather.new'. Passing in a double with stubbed behaviour (emulating the real weather class) would mean that this kind of stubbing would work.

eoinmakers (commented):
I think this is a great airport challenge attempt, well done. I left some notes around how you could continue / understand the implementation of the Weather class, both functionally and in your testing, and do let me know if you have any questions about any of the comments I added. Well done!

PR #2197, opened 2021-08-02
Airport Challenge

yousufmohamed92:


eoinmakers (commented):
Really great work on this, I think it's an excellent airport challenge submission. 

I believe it adheres well to SRP - your methods are concisely written, and each belongs in the class they're defined in. 

I left some feedback around the re-use of method names in two classes - this isn't inherently a bad thing, but I think the methods in Plane can be a little bit more descriptive for the action they're carrying out. 

I think your use of a module is fine! It likely doesn't make too much difference here vs using a class, and you tested well for it and its requirements in the context of the rest of the implementation. 

Nice work!

PR #2202, opened 2021-08-02
Aiport challenge till 4 user story

suhani-zaman:


lib/airport.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+    def land(plane = Plane.new )
+      raise "Airport is full" if full? == true
+      @airport.push(plane)
+    end
+    def take_off(plane = Plane.new)
+      @airport.delete(plane)
+    end
+    def full?
+      if @airport.count >= @capacity
```
Battery0, line 16: Try writing this out as a ternary operator instead of using and if statement
@airport.count >= @capacity

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
+
```
Battery0, line 2: You could add in some instance variables to check if the plane is grounded or flying and with true/false.
You could then use the true/false instance variables as extra conditionals for the Airport.land and Airport.take_off

spec/Airport_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "Airport"
+describe Airport do
+  it "lands the plane at the airport" do
+    port = Airport.new
+    plane1 = Plane.new
+    port.land(plane1)
+    expect(port.airport).to include(plane1)
+  end
+  it "checks if the plan took off" do
```
Battery0, line 9: Maybe raise an error when calling take_off method if the @airport array is empty?

PR #2203, opened 2021-08-02
git pull of airport_challenge

Xfarya:


lib/airport.rb
```diff
@@ -0,0 +1,63 @@
+require_relative 'airplane'
+
+class Airport 
+
+attr_reader :plane, :plane_take_off
+attr_accessor :capacity
```
Nicola-Carroll, line 6: I believe that `attr_accessor` means that you can alter the attribute **after** the instance has been initiated. So this would mean you can create an airport object and then alter its capacity at any point afterwards. Personally I interpreted this user story as stating it would like to be able to create airports of any capacity, as opposed to be able to alter the capacity of existing airports. So I thought that it was best to put `:capacity` under `attr_reader` and use `def initialize(capacity=DEFAULY_CAPACITY)`

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+end
+
+def land(plane)
+    fail 'Airport is full' if full?
+  if stormy?
+    fail 'Cannot land plane. Weather is stormy'
+  end
+
+  if (@plane.include? plane)
+    puts "This plane is already landed"
```
Nicola-Carroll, line 23: could also use a `fail` or `raise` statement here to show it's an error instead of `puts`, that way the program would automatically break if it raised this error, and you wouldn't need to wrap it in an `if..else` statement

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+  else
+    "#{@plane} has landed"
```
Nicola-Carroll, line 25: Good thinking to print a success message to the user, I didn't think to do that to show it was successful 🤔

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+attr_accessor :capacity
+
+DEFAULT_CAPACITY = 20
+
+def initialize 
+  @plane = []
+  @capacity = DEFAULT_CAPACITY
+  @plane_take_off = []
+end
+
```
Nicola-Carroll, line 15: The `land` method below it quite long, I think we can refactor this into smaller ones

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  if (@plane.include? plane)
+    puts "This plane is already landed"
+  else
+    "#{@plane} has landed"
+    @plane << plane
+  end
+
+if plane_take_off.include? plane
```
Nicola-Carroll, line 29: I'm not completely clear on the purpose that the `plane_take_off` attribute serves

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+  if plane.include? plane_take_off
+    plane.delete(plane_take_off)
+  end
+end 
+
+def stormy?
+  luck = rand(0..6)
+    if (luck == 1)
+      return true
+    end
```
Nicola-Carroll, line 55: This whole method could be refactored to

`rand(0..6) == 1`

and it will automatically return either `true` or `false`. I.e. if you want a method to return a boolean value, you just have to put the statement you're evaluating and true/false will be returned (so need to spell out `if .. return true`)

spec/airport_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    plane = Airplane.new
+    landing = subject.land(plane)
+    expect(landing).to be_stormy
+  end
+
+  it "prevents landing if stormy" do
+    plane = Airplane.new
+    landing = subject.take_off(plane)
+    expect(landing).to be_stormy
+  end
```
Nicola-Carroll, line 27: Did these two tests above do what you expected? How did you ensure that the conditions were stormy for these tests?

spec/airport_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      plane = Airplane.new
+      subject.capacity.times { subject.land(plane) }
+      expect { subject.land(plane) }.to raise_error 'Airport is full'
+    end
+  end
+
+  it "has a default capacity" do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+
```
Nicola-Carroll, line 40: I believe these two tests are essentially testing a very similar thing, in that you haven't also tested if the same thing happens when you change the capacity to a non-default value

Nicola-Carroll (commented):
One thing I would suggest is getting into the habit of committing more regularly - that's the whole point of version control :)

PR #2205, opened 2021-08-05
airport challenge

necsanuradu:


necsanuradu:
I did not finished the pull procedure, only now

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+  def initialize (capacity = CAPACITY, runways = RUNWAYS)
+    @grounded = Array.new
+    @capacity = capacity
+    @runways = 1.upto(runways).to_a.map do |n|
+      {number: n-1, status: "", reason: "", status_initiated_at: "unknown",
+        status_check_at: "unknown", flight_number: "none"}
+    end
+  end
+
+  def check_runway
```
eoinmakers, line 15: The concept of a runway isn't mentioned in the user stories for the challenge. It's really important when working to a specification or set of user stories, that you don't include any extra features or requirements. You want to stick to implementing the most accurate representation of the user stories as possible, and the only deviation you would make is taking 'edge cases' into account: perhaps accounting for if a user tries to allow a plane to take off when there are no planes at the airport, for example.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+    runway = @runways.find do |runway| 
+        runway[:status] != :blocked
+    end
+    runway
+  end
+
+  def commit_runway(flight, procedure, time_to_proceed = 6)
```
eoinmakers, line 22: I think this is an overcomplicated implementation of the user stories - there is no requirement for flight procedures or time to proceed, etc.

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+  def register_runway
+    @runways[@number][:status] = :blocked
+    @runways[@number][:flight_number] = @flight
+    @runways[@number][:reason] = @procedure
+    @runways[@number][:status_initiated_at] = Time.now.to_i  
+    @runways[@number][:status_check_at] = Time.now.to_i + @time_to_proceed
+  end
+
+  def plane_landed(plane)
+    @grounded.push(plane)
```
eoinmakers, line 45: Nice!

lib/airport.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def plane_tookoff(plane)
+    @grounded.delete(plane)
+  end
+
+  def is_full?
+    @grounded.count >= @capacity
```
eoinmakers, line 53: Nice and concisely written.

lib/control_tower.rb
```diff
@@ -0,0 +1,44 @@
...
+    if @plane.plane_now[:grounded_request_takeoff] == :now || @plane.plane_now[:grounded_red_takeoff] == :now
+      return (!@airport.check_runway.nil? && @weather != :stormy) ? green_takeoff : red_takeoff
+    end
+    if @plane.plane_now[:grounded_landing] == :now
+      return @airport.plane_landed(@plane)
+    end
+    if @plane.plane_now[:airborne_takingoff] == :now
+      return @airport.plane_tookoff(@plane) 
+    end
+  end
```
eoinmakers, line 25: While there are certain checks that need to be carried out, they should be much simpler than this. 

For example, as the responsibility of your Airport class is allowing the plane to land and take off, and storing the grounded planes, you would also keep your checks there before allowing a plane to land / take off. 

You successfully implemented the capacity check there, which is great, and if you wanted to also check if a plane is already airborne as some learners choose to do, you could keep that in your Airport class too.

lib/planes.rb
```diff
@@ -0,0 +1,43 @@
...
+    @plane_now[next_exe] = :next
+  end
+
+  def clean_state
+    @plane_now.merge!(@plane_now) { |key, value| :none }
+  end
+
+  def get_state
+    @plane_now.find { |key, value| value == :now }
+  end
```
eoinmakers, line 42: If you adhere to the user stories, the plane actually doesn't need to know much at all. Some people include the status of whether or not the plane is already airborne, but in fact that's also unneeded. 

It can be tempting to let our knowledge of real-world objects influence our implementations: we may have experience or knowledge of how these things behave in real life, but it's important to not let that influence the code we write. This leads to unnecessary code and features being written. 

Typically this real-world knowledge would maybe be discussed after tickets have been allocated to developers, to allow for clarification with UX designers and business analysts, etc.

For these projects, you want to work with the **simplest possible interpretation** of the user stories, and to develop only what is specified. If you'd like to clarify user stories, you can also check with a coach afterwards.

lib/weather.rb
```diff
@@ -0,0 +1,10 @@
+class Weather 
+  attr_reader :possible_weather
+  def initialize 
+    @possible_weather = [0..200].map{:normal} + [0..5].map{:stormy}
+  end
+
+  def forecast
+    @possible_weather.sample
```
eoinmakers, line 8: Nice and concisely written.

lib/weather.rb
```diff
@@ -0,0 +1,10 @@
+class Weather 
+  attr_reader :possible_weather
+  def initialize 
+    @possible_weather = [0..200].map{:normal} + [0..5].map{:stormy}
```
eoinmakers, line 4: A slightly simpler version of this would just be an array containing a couple of weather conditions, perhaps 4 strings with the value of 'sunny', and one with 'stormy'.

lib/airport.rb
necsanuradu: I did jumped with no fence, you are absolutely spot on

lib/airport.rb
necsanuradu: I do not read, I mean I read the instruction but after a few words my imagination takes over. I have to stop doing this, it overcomplicates everything.

lib/control_tower.rb
necsanuradu: Yes that would have been resonable

lib/planes.rb
necsanuradu: Best expression in coding: simplest possible interpretation. Yes this is the way. Thank Eóin

PR #2208, opened 2021-08-28
Complete airport challenge build

Legal-Dan:


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+class Airport
+  LOCAL_WEATHER = ["stormy", "sunny", "drizzle", "light cloud", "clement"]
+  CAPACITY = 1
+
+  def initialize
+    @landed_planes = []
+  end
+
+  def weather?
```
Jjake540, line 11: This could probably be a private method, with stormy? being a result of its own method

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  it 'can instruct a plane to depart from an airport' do
+    allow(subject).to receive(:weather?) { "sunny" }
+    plane1 = Plane.new
+    subject.land_plane(plane1)
+    expect subject.takeoff(plane1) == "Plane has now departed"
+  end
+
+  it 'will not allow a plane to land if the airport is full' do
```
Jjake540, line 18: We can sort hand rspec methods like this. to make it easier to read. 
 
it 'a plane can only take off from an airport it is at' do
    message = 'The plane is not currently landed at this airport'
    expect { airport.take_off(plane) }.to raise_error message
  end

Legal-Dan: Not sure how this is different from what I have done for this test starting at line 18 (except that obviously this example is for a different test). The only difference is that I specified the error message I expected as there are a couple of different possibilities for this method.

PR #2213, opened 2021-08-29
Weekend Challenge 1

Maldorana:


README.md
```diff
@@ -1,3 +1,17 @@
+My Approach
+=================
+
+* I identified the classes (Airport, Plane) and methods (land, take_off, weather) required from the user stories.
+* I tied the 'land' and 'take_off' methods to the Airport class because the user stories are from the air traffic controller point of view, and they are tied to airports, not planes. 
+* I tied the weather method to the Airport class because the weather is tied to a place (airport), not an object (plane).
+* The one thing I wasn't sure about was how rarely the stormy weather would occur (I've put 1 out of 4 chances).
```
eoinmakers, line 7: I think this approach is really thoughtfully planned out, and absolutely checks out. Nice! 

As for the stormy weather, any odds are fine here as long as it's the least most frequent weather type, so 1 in 4 sounds good.

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+  DEFAULT_CAPACITY = 5
+  attr_reader :planes, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    weather
```
eoinmakers, line 11: Instead of calling this here, the code that you have on line 13 is actually fine. Calling a method is a little different to calling a variable - you can call other methods in your same class without having to do anything else.

eoinmakers: This calling of 'weather' could actually end up being different to the method call on line 13, because it will just call it again and run the .sample code once more.

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+  def take_off(plane)
+    weather
+    fail "Bad weather" if weather == "Stormy"
+    fail "This plane is not in the airport" unless @planes.include?(plane)
+
+    @planes.delete(plane)
+  end
+
+  def weather
+    ["Sunny", "Sunny", "Sunny", "Stormy"].sample
```
eoinmakers, line 28: This is good! Nice work. 

If you wanted to refactor this to make it even cleaner, you could designate the full responsibility of checking if it's stormy or not to this method, rather than just generating the weather. 

For example: 

```
# inside the land method
fail "Bad weather" if is_stormy?

# new weather method 
def is_stormy?
  ["Sunny", "Sunny", "Sunny", "Stormy"].sample == "Stormy"
end
```

The benefit of this is that it takes some more of the responsibility away from the land method, so if you wanted to make any changes to the weather function, more of the related code would be in the one method, thus easier to change without having alter too much else.

eoinmakers: Another benefit of this is that the condition that would raise the error is easier to read, because you can be super descriptive in your method naming, rather than having to figure out a conditional (which can sometimes be more complicated than in the current example).

spec/airport_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+    it 'should not land if airport is full' do
+      allow(subject).to receive(:weather) { "Sunny" }
+      subject.capacity.times { subject.land(Plane.new) }
+      expect { subject.land(Plane.new) }.to raise_error("Airport is full")
+    end
+
+    it 'should have a default capacity that can be overwritten' do
+      airport = Airport.new(2)
+      allow(airport).to receive(:weather) { "Sunny" }
+      airport.capacity.times { airport.land(Plane.new) }
```
eoinmakers, line 23: For this one, I would hardcode the new capacity that you created this instance of airport with - this current test would still pass even if your Airport instance didn't set the new capacity to 2. Lets say there was a bug which meant it ignored the 2 on line 21 and just went with the default capacity: this test would still pass because airport.capacity will always return whatever value that has - we can't be sure the capacity has been overwritten unless we specify landing planes twice, and then running the test.  

If we have set the capacity to 2, and then land 2 planes and run the expectation, if the instance has for some reason used the default capacity instead, the test will fail (the third plane will land in this case, and we'll know there's a bug here).

spec/airport_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      expect { subject.land(boeing_747) }.to raise_error("Bad weather")
+      expect(subject.planes).not_to include(boeing_747)
+      # TRY TO TAKE OFF AIRBUS A330 (but there is bad weather)
+      expect { subject.take_off(airbus_a330) }.to raise_error("Bad weather")
+      expect(subject.planes).to include(airbus_a330)
+      # THE WEATHER IS SUNNY
+      allow(subject).to receive(:weather) { "Sunny" }
+      # TRY TO LAND AIRBUS A330 (but the plane is already in the airport)
+      expect { subject.land(airbus_a330) }.to raise_error("This plane is already in the airport")
+      expect(subject.planes).to include(airbus_a330)
```
eoinmakers, line 102: Great feature test! It's covering a lot of ground, and I think maybe the one thing you wouldn't need to cover here is the 'edge case' of the plane already being in the airport - just because this isn't mentioned as a requirement in the user stories.

eoinmakers (commented):
Great work! This is a really good implementation of the airport challenge. 

The code can be further improved by thinking about how many responsibilities belong to each class and drawing out a weather class from there, but that will be the focus of the upcoming weekend challenge so you'll definitely get to practice it then.

PR #2215, opened 2021-08-30
Airport challenge attempt

leahcolleen:


spec/airport.rb
```diff
@@ -0,0 +1,12 @@
+class Airport
+  attr_reader :plane
+  
+  def land
+    fail 'Plane has already landed' if @plane?
+    @plane
+  end
+
+  def landed?(plane)
```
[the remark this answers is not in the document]
eoinmakers: In this case, you're passing a plane in as an argument, and assigning it to the @plane variable. If this method's intention was to check if a @plane already exists (thus has landed at the airport already), perhaps you could tie it in with the changes from my previous comment about having a means to store planes in an airport once they land.

spec/airport_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+describe 'Airport' do
+  
+  it { is_expected.to respond_to :land }
+  let(:plane) { Plane.new }
+  
+  describe '#land' do
+    it 'instructs plane to be landed' do
+      expect(plane).to respond_to(:land)
+      subject.land(plane)
```
eoinmakers, line 11: You don't need line 11 here: your test on line 10 is perfectly valid on its own, and will check to make sure the land method exists on the plane object.

eoinmakers: After checking that a method exists, we also want to check if its able to carry out its intended behaviour. 

What is the intended behaviour of the land method that we're testing? It takes a plane as an argument, checks to see if it's already present at the airport, and if it's not, it creates the @plane object. 

In this case, as @plane is the last line inside the land method, it will [implicitly return](https://franzejr.github.io/best-ruby/idiomatic_ruby/implicit_return.html) this object when the method is called. 

So, when shaping your unit test, you want to think about what a method will change or return, and then test that that's happened. 

For an additional unit test for your land method, you could check that the value being returned is a plane object. This will check that the method has done what you intended it to do.

spec/plane.rb
```diff
@@ -0,0 +1,6 @@
+class Plane
+  
+  def landed?
+    true
```
eoinmakers, line 4: This is hardcoded so that the value will never change - and I imagine something you were working on in order to have your classes communicate with one another. 

If you wanted this value to track the status of a plane - whether its landed or not - you could write something like this: 

```
class Plane 
  def initialize
    @grounded = false 
  end

  def is_grounded
    @grounded = true
  end

  def is_airborne
    @grounded = false
  end
end
```

And then, to access this from your Airport class, you could write something like: 

```
class Airport
  def initialize
    # possibly somewhere to create a way to store planes
  end

  def land(plane)
    plane.is_grounded
    # store the plane in the airport
end
```

I just showed a small part of the Airport class to represent the desired behaviour here: updating the plane's airborne status to grounded (true or false), so that later you can check to see if it's already grounded.

eoinmakers: In thinking about what you need to make these classes interact - you'll be passing an instance of the Plane class into an instance of the Airport class, so think about where the user will do this - in IRB, where you'll need to require both files in order to test them. Once you pass the instance of Plane into the instance of Airport, airport will be able to run the methods that are in Plane (eg. plane.is_grounded).

eoinmakers (commented):
Good progress on a single-plane solution, fulfilling the first user story.

It looks like further practice with classes interacting with one another could be useful here, so I'll share a further resource with you on that topic, and we can discuss any of the other points I've mentioned here.

PR #2216, opened 2021-08-30
1st attempt

99ltaylor:


README.md
```diff
@@ -1,3 +1,18 @@
+Reflection:
+
+*As suggested, I time-boxed this exercise. I did not get time to complete the random weather generator.
+
+To improve:
+
+* Struggled with creating the domain model - what should classify as an object, and which messages are being sent where.
```
eoinmakers, line 7: Tomorrow morning's workshop should support with this, but if this is quite a good example made by another coach in how you can use nouns & verbs to break things down into Owners and Properties, which can be another useful step between reading user stories & diagramming: https://docs.google.com/document/d/1QR2C3hw1cjZKGNF84Zfy0J5VjAbjCfM4qprWXTUkmfc/edit

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
+require_relative "../lib/plane"
+
+class Airport
+  attr_accessor :planes, :default_capacity, :weather
+
+  def initialize
+    @planes = []
+    @default_capacity = "50"
```
eoinmakers, line 8: This value would perhaps be better suited to an integer, for ease of user later. 

For example: what's the responsibility of this default_capacity? It's to make sure that the hangar doesn't get full up, so that we can prevent a plane from landing in case there's no more space. The simplest way to do this would be to be able to count how many items are already in '@planes', compare that to @default_capacity, and either allow the landing or raise an error. If @default_capacity is an integer, we won't have to translate it from a String in order to make that comparison.

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
+require_relative "../lib/plane"
+
+class Airport
+  attr_accessor :planes, :default_capacity, :weather
+
+  def initialize
```
eoinmakers, line 6: Have you come across default arguments previously? It can make your program a little more flexible. 

Below, you set the @default_capacity to 50. However, you want to allow users the option of setting their own default capacity (lets say the person that wants to set their own capacity is whoever is running the airport that the instance is representing). 

To do that, you could write: 

```
def initialize(default_capacity = 50)
  @default_capacity = default_capacity
  # etc...
end
```

In this case, if the user passes in an argument, eg. the number 6, it will get assigned to the variable that's called @default_capacity. If the user doesn't pass in an argument, it will automatically receive the value '50', which is the 'default argument'.

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+  def full_capacity(default_capacity)
+    print "WARNING: Airport at full capcacity, plane cannot land"
+    default_capacity = "100"
+    # if current_capacity = "100"
+    #   print "WARNING: Airport at full capcacity, plane cannot land"
+    # end
+  end
+
+  def land(plane)
+    print "Landing ..." + plane.to_s
```
eoinmakers, line 25: You don't actually need to include these print statements, as the user stories didn't specify needing any confirmation of the planes landing or taking off. If used for testing purposes as you developed your implementation, you would typically remove them before pushing your code. :)

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+    @planes = []
+    @default_capacity = "50"
+    @weather = "mild"
+  end
+
+  # def capacity(current_capacity)
+  #   current_capacity = "50"
+  # end
+
+  def full_capacity(default_capacity)
```
eoinmakers, line 16: Your solution for the capacity check is a little more complicated than it needs to be. 

If we think about what it is we need to be able to add this check:
- We want the plane to land if the hangar is not at full capacity. 

To do this, we'll need to know: 
- What is the default capacity? 
- How many planes are currently in the hangar? 
- How do we raise an error if the hangar is full? 

The answers to those questions:
- In my previous example, I defined the capacity as an argument called @default_capacity. 
- To check how many planes are currently in the hangar, we can use an array method called 
- There are a couple of ways to [raise errors in Ruby](https://rollbar.com/guides/ruby/how-to-raise-exceptions-in-ruby/), but in its simplest form, we could write something like the below: 

```
def land(plane)
  raise "This airport is at capacity" if @planes.full? 
  @planes << plane
end

def full?
  @planes >= @default_capacity
end
```

So no additional method is needed for this check, rather it's something that can be done in the same method as the landing / taking off to check if it's possible to run this actions or not.

lib/airport.rb
```diff
@@ -0,0 +1,43 @@
...
+  def take_off(plane)
+    @planes.delete(plane)
+  end
+
+  def no_landing_or_take_off
+    print "WARNING: due to stormy weather no landing or take offs"
+  end
+
+  def stormy_weather
+    @weather = "stormy"
```
eoinmakers, line 38: How could you refactor this to make the value randomly sunny or stormy?

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    boeing_727 = Plane.new
+    heathrow = Airport.new
+    # act/method: airport instructs plane to take off, and prints to screen
+    heathrow.planes << boeing_727
+    heathrow.take_off(boeing_727)
+    # expect that boeing_727, is removed from planes property in the airport object
+    expect(heathrow.planes).not_to include(boeing_727)
+  end
+
+  it "prevents plane landing when airport full" do
```
eoinmakers, line 28: I think this is slightly overcomplicated - a suggestion of an alternative test once my previous suggestions were implemented: 

If you were to set the capacity to 2, pass in two instances of Plane, and then try to pass in a third, you would expect an error to be raised. 

Your test could look a little something like this: 

```
it 'prevents plane from landing when airport is full' do 
  airport = Airport.new(2)
  plane1 = Plane.new
  plane2 = Plane.new
  plane3 = Plane.new

  airport.land(plane1)
  airport.land(plane2)
  expect { airport.land(plane3) }.to raise_error "This airport is at capacity"
end
```

There are a few ways of then refactoring that test to tidy it up (you could use 'times' to automate multiple planes landing instead of writing them separately).

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    # arrange
+    heathrow = Airport.new
+    # act/method: airport tells plane it is at 100/ full capacity so no planes landing
+    default_capacity = "100"
+    # Expect 'plane cannot land' warning if at full capacity
+    expect(heathrow.full_capacity(default_capacity)).to eq("100")
+    # expect(heathrow.full_capacity(current_capacity)).to output.to_stdout "WARNING: Airport at full capcacity, plane cannot land"
+  end
+
+  it "prevents take off when weather is stormy" do
```
eoinmakers, line 38: Running this kind of check to make sure planes don't land or take off uses similar logic as the capacity check, so it's worth exploring that implementation first, and then circling back to this one - resources on mocking shared later this week should also help in trying to stub randomness.

eoinmakers (commented):
Well done on your progress with this, you were able to partially implement the user stories. 

There are some slight over complications in your solution, so try to remember to keep things as simple as possible. Have a read over my suggestions, and while you don't need to go back and make these changes, please do ask me as many questions as you might have about my notes. Happy to share more examples or have a chat about them, so let me know what you think.

I think you'll also find tomorrow's workshop useful in bridging that gap between user stories and implementation.

PR #2217, opened 2021-08-30
Airport weekend challenge, Makers week 1

lildann:
Completed User Stories.
Further research needed: 
- Stubs to override random weather to ensure consistent test behaviour
- Test doubles for overriding random weather behaviour


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+class Airport 
+  DEFAULT_CAPACITY = 5
+  attr_accessor :capacity
+  attr_reader :plane, :airport
```
eoinmakers, line 4: Minor suggestion, I would leave some whitespace underneath this before the method definition, for readability.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airport = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail "Airport is full" if @airport.length >= @capacity
+    fail "Stormy weather: do not land" if weather? == "Stormy"
```
eoinmakers, line 12: Nice work!

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    @plane = plane
```
eoinmakers, line 14: You can actually just pass plane straight into @airport, you don't need to store this one in an instance variable as it won't need to be accessed from outside this method.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    @airport << @plane
+  end
+
+  def take_off(plane)
+    fail "Stormy weather: do not fly" if weather? == "Stormy"
+    fail "Plane not at airport" unless @airport.include?(@plane)
```
eoinmakers, line 20: Careful with this one - @plane won't have the same value as the plane that's just been passed in as an argument, as that '@' ends up creating an entirely different variable. 

If you wanted to include this check, you could just check that @airport.include?(plane) instead of @plane.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def land(plane)
+    fail "Airport is full" if @airport.length >= @capacity
+    fail "Stormy weather: do not land" if weather? == "Stormy"
+
+    @plane = plane
+    @airport << @plane
+  end
+
+  def take_off(plane)
```
eoinmakers, line 18: You actually don't need to specify which plane takes off, any plane will do.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    fail "Stormy weather: do not fly" if weather? == "Stormy"
+    fail "Plane not at airport" unless @airport.include?(@plane)
+
+    @plane = plane
```
eoinmakers, line 22: You also don't need to store this one as an instance variable - you'll only be interacting with this variable 'plane' in this method, so you can just keep it as plane.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+class Airport 
+  DEFAULT_CAPACITY = 5
+  attr_accessor :capacity
+  attr_reader :plane, :airport
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airport = []
```
eoinmakers, line 6: This naming could get confusing - I would avoid giving variables names that are the same as the class name. Alternatives could be @hangar, @planes, @storage, etc.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def take_off(plane)
+    fail "Stormy weather: do not fly" if weather? == "Stormy"
+    fail "Plane not at airport" unless @airport.include?(@plane)
+
+    @plane = plane
+    @airport.each do |grounded_plane| 
+      if grounded_plane == @plane && weather? == "Sunny"
+        @airport.delete(@plane) 
+        return "#{@plane} has departed"
```
eoinmakers, line 26: The printed confirmation that the plane has left is unneeded, as it wasn't specified in the user stories.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+    end
+  end
+
+  def weather?
+    weather = ["Sunny", "Sunny", "Sunny", "Stormy"]
+    weather[rand(4)]
```
eoinmakers, line 33: Nice! 

If you wanted to pull this out into its own class, it would be as simple as something like: 

```
class Weather
  def generate_weather
    weather = ["Sunny", "Sunny", "Sunny", "Stormy"]
    weather[rand(4)]
  end
end
```

The other changes you would need in order to then interact with this method would be: 
- allowing the user to pass in an instance of Weather when they create the Airport class 
- you would call the method that now belongs to your Weather class inside this method, instead of this logic

For example: 

```
# airport.rb

  @weather = weather
  # etc..
end

def weather?
  @weather.generate_weather
end
```

I pass the instance of Weather in with a default value to allow the user to also pass in a double instead while testing.

lib/weather.rb
```diff
@@ -0,0 +1,6 @@
+# class Weather
+#   def random_weather_generator
+#     weather = ["Sunny", "Sunny", "Sunny", "Stormy"]
+#     weather[rand(4)]
+#   end
+# end
```
eoinmakers, line 6: This looks great!

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect_any_instance_of(Object).to receive(:rand).with(4).and_return(3)
+      expect(subject.weather?).to eq("Stormy")
+    end
+  end
+  
+  describe '#landing plane' do
+    it 'error if airport is at capacity' do
+      allow(subject).to receive(:weather?) { "Sunny" }
+      subject.capacity.times { subject.land(Plane.new) }
+      expect { subject.land(Plane.new) }.to raise_error("Airport is full")
```
eoinmakers, line 37: Great test!

eoinmakers (commented):
Really well done on this, you successfully implemented all of the user stories and demonstrated the learning objectives well.

There are really great unit test examples here! 

Drawing out a Weather class does make the implementation a little cleaner as it adheres more closely to Single Responsibility Principle, so let me know what you think about my suggestions and let me know what questions you might have. :)

PR #2218, opened 2021-08-30
Partial solution 

mandyvuong:
Commits are not in sync with the tasks since I've not used Git commands correctly, therefore had to redo the task but this time it's now an incomplete solution since I don't have the capacity to complete this. 

I believe I completed the task on an the Ruby side but not the RSpec

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,3 +133,12 @@ Airport <-- takeoff --> Plane
 Weather(storm) <-- !land ---> Plane
 
 Weather(storm) <-- !takeoff ---> Plane
+
+### Program
+
+I used (some) TDD by repeating the following processes
+
+1. creating imaginary feature unit test
+2. create a failed test
+3. pass a code
+4. test the manual unit test
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
-Airport Challenge
-=================
+# Airport Challenge
 
-```
+``````
         ______
         _\____\___
 =  = ==(____MA____)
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity=DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def lands(plane)
+    weather = Weather.new
+    weather = weather.generates
```
eoinmakers, line 16: Well done on the implementation of the weather feature!

This snippet of code reveals a 'dependency', which means that the Airport class relies on the Weather class in order to do its job. 'Tight coupling' is a term used to describe classes that are tightly dependant on one another - and in some cases, we can mitigate this a little.

The reason we'd like to avoid tight coupling between classes is:

1. For testing purposes - we want our unit tests to be as specific and isolated from one another as possible. For example, if there's a problem in our Weather class, we don't also want our Airport class tests to start failing, because although the Airport class will make use of the Weather class, they have separate responsibilities - so we want to be able to get to the source of the problem as efficiently as possible.

2. To make our code easy to change. This can mean maintaining code, re-using classes or adding features, or also testing. If our classes are tightly coupled, if we want to make changes to our program, we'll have to sift through several classes to make all of the necessary changes, when ideally we'd like to be able to just make changes to the class in question (unless the way in which you use that class is changing in some way).

The solution here is dependency injection, which means that although Airport will still depend on Weather, it will open it up to the use of mocks, or alternative classes being passed in.

eoinmakers: You can learn about dependency injection here: 
https://github.com/makersacademy/skills-workshops/blob/main/practicals/object_oriented_design/dependency_injection.md

How this might look in your code, would require some changes: 

```
def initialize(weather = Weather.new, capacity=DEFAULT_CAPACITY)
  @weather = weather
  @capacity = capacity
  @planes = []
end

def land(plane)
  fail 'Stormy - do not land!' if @weather.generates == : stormy
  # etc . . . 
end
```

The difference in this example to yours is that it gives the opportunity to pass something other than the real Weather class in, and your Airport class will be none the wiser. 

Check out the practical, my example above, and let me know what you think, or what questions you might have. :)

lib/plane.rb
```diff
@@ -0,0 +1,6 @@
+class Plane
+  def active?
```
eoinmakers, line 2: Do you call this method anywhere else in your code? Looks like it may not be needed.

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    it 'plane lands' do
+      plane = Plane.new
+      expect(subject.lands(plane)).to eq [plane]
+    end
+
+    it 'not land when airport is full, take into consideration of capacity - raises an error' do
+      subject.capacity.times { subject.lands(Plane.new) } # Ruby .times loop
+      expect{subject.lands(Plane.new)}.to raise_error 'Airport full - do not land!'
```
eoinmakers, line 20: Very well tested.

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    it 'not land when airport is full, take into consideration of capacity - raises an error' do
+      subject.capacity.times { subject.lands(Plane.new) } # Ruby .times loop
+      expect{subject.lands(Plane.new)}.to raise_error 'Airport full - do not land!' 
+    end
+    
+    it 'not land when stormy - raises an error' do
+      weather = Weather.new
+      weather.generates == :stormy
```
eoinmakers, line 25: So, this is where things can get a bit challenging. 

My earlier comments & example would mean that you could pass in a double instead of your code relying on the real Weather class - right now, whatever you define here, your Airport class will ignore, because in airport.rb on line  15, it will make its own instance of Weather, and ignore anything else you define here. 

Having dependency injection in your Airport class would allow you to do something like this: 

```
weather = double
allow(weather).to receive(:generates).and_return(:stormy)

airport = Airport.new(weather)
# and then the rest of your test to make sure the error is raised 
```

By doing this, your Airport class will use your mocked version of Weather, and the weather will always be stormy.

eoinmakers: I'll be sharing some resources on this tomorrow if you'd like to learn more!

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  describe '#takesoff' do 
+    it 'method def' do
+      expect(subject).to respond_to(:takesoff)
+    end
+
+    it 'takes off and no plane at airport' do
+      plane = Plane.new # new plane
+      subject.lands(plane) # plane is on land
+      expect(subject.takesoff).to eq plane # plane takes off and is not on land
```
eoinmakers, line 38: Naming conventions for Ruby methods are more descriptive of a singular action: for example 'take_off' vs 'takesoff', and 'land' instead of 'lands'.

spec/airport_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    it 'takes off and no plane at airport' do
+      plane = Plane.new # new plane
+      subject.lands(plane) # plane is on land
+      expect(subject.takesoff).to eq plane # plane takes off and is not on land 
+    end
+
+    it 'not takeoff when stormy - raises an error' do
+      weather = Weather.new
+      weather.generates == :stormy
```
eoinmakers, line 43: Similarly for this one, a better suited name for the method generating the weather condition could be generate_weather or generate_forecast

eoinmakers: To add a bit more of a description ^

spec/weather_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+describe Weather do
+  describe '#generates' do
+    it 'method' do
+      expect(subject).to respond_to(:generates)
+    end
+
+    it "returns nil from allowed messages" do
+      weather = double (subject)
+      allow(weather).to receive(:generates).and_return(:stormy)
```
eoinmakers, line 11: This test doesn't test any behaviour of the program - maybe the only test you could do for Weather in this regard is that it returns one of the possible weather conditions (any will do). Where you'll try to mock this behaviour is when you are testing other behaviour that depends on the output of the Weather class (see previous comments).

eoinmakers (commented):
Great work, I think you've done very well here. You successfully implemented most of the user stories! I think the next things to check out would be reviewing my comments about dependency injection, and let me know if you have any questions. Understanding that can open up your understanding of mocking and doubles, so let me know what you think. Great work!

README.md
```diff
@@ -1,7 +1,6 @@
-Airport Challenge
-=================
+# Airport Challenge
```
florence-bain, line 2: Enjoyed reading your code, was very clear to read and understand how your methods worked.

PR #2222, opened 2021-08-30
airport challenge

LauraMaddaluno:
This is part of code for this challenge. 
I still need to clarify concepts

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -8,4 +8,8 @@ class Airport
     @hangar << plane
     plane
   end
+  def take_off(plane)
+    @hangar.pop 
+    return 'Plane no longer in the airport' 
+  end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -12,4 +12,11 @@ class Airport
     @hangar.pop 
     return 'Plane no longer in the airport' 
   end
+  def prevent_landing(plane)
+    fail "Sorry, the airport is full" if full?
+    @hangar << plane
+  end
+  def full?
+    @hangar.size >= 1
+  end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,8 +1,12 @@
 require_relative "plane"
 class Airport 
- 
-  def initialize
+
+  DEFAULT_CAPACITY = 4
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
     @hangar = []
+    @capacity = capacity
   end
   def land(plane)
     @hangar << plane
@@ -17,6 +21,6 @@ class Airport
     @hangar << plane
   end
   def full?
-    @hangar.size >= 1
+    @hangar.size >= DEFAULT_CAPACITY
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,11 @@
+require_relative "plane"
+class Airport 
+ 
+  def initialize
+    @hangar = []
+  end
+  def land(plane)
+    @hangar << plane
+    plane
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -2,10 +2,17 @@ require_relative '../lib/airport'
 
 describe Airport do 
   it { is_expected.to respond_to(:land).with(1).argument}
+  it { is_expected.to respond_to(:take_off).with(1).argument}
   describe "aircraft landing" do 
     it 'plane can land at airport' do
       plane = Plane.new
       expect(subject.land(plane)).to eq plane
+    end
+  end
+  describe "aircraft take off" do
+    it 'plane take off from airport' do 
+      plane = Plane.new
+      expect(subject.take_off(plane)).to eq 'Plane no longer in the airport' 
     end
   end
 end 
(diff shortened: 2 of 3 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+require_relative "plane"
+class Airport 
+
+  DEFAULT_CAPACITY = 4
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+  end
```
eoinmakers, line 10: It's good practice to leave one empty line underneath between methods.

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+  def land(plane)
+    @hangar << plane
+    plane
```
eoinmakers, line 13: You actually don't need to return a plane here - this method will have carried out its purpose once you've run line 12. 

For testing, you can run your code in IRB if you're unsure what line 12 might return. Here's an example of code I would run in IRB to try to find out what kind of test I could write while refactoring: 

```
2.4.4 :001 > class Airport
2.4.4 :002?>   def initialize
2.4.4 :003?>     @hangar = []
2.4.4 :004?>     end
2.4.4 :005?>   def land(plane)
2.4.4 :006?>     @hangar << plane
2.4.4 :007?>     end
2.4.4 :008?>   end
 => :land 
2.4.4 :009 > class Plane
2.4.4 :010?>   end
 => nil 
2.4.4 :011 > plane = Plane.new
 => #<Plane:0x00007fd29aa523e0> 
2.4.4 :012 > airport = Airport.new
 => #<Airport:0x00007fd29aa48ca0 @planes=[]> 
2.4.4 :013 > airport.land(plane)
 => [#<Plane:0x00007fd29aa523e0>] 
2.4.4 :014 > airport.land(plane)
 => [#<Plane:0x00007fd29aa523e0>, #<Plane:0x00007fd29aa523e0>] 
```

It will seemingly return an array of planes, @hangar, so in your test, you could check to make sure the hangar has increased in size, that it is equal to [plane] if it only has one plane that's landed, that it 'includes' a plane, etc. There are a couple of options here.

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+  def take_off(plane)
+    @hangar.pop
```
eoinmakers, line 16: What happens if the @hangar is empty, and the user tries to make a plane take off? Is there an error you could raise here?

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+    @hangar = []
+    @capacity = capacity
+  end
+  def land(plane)
+    @hangar << plane
+    plane
+  end
+  def take_off(plane)
+    @hangar.pop 
+    return 'Plane no longer in the airport'
```
eoinmakers, line 17: It's important to only implement what's asked for in the user stories or specification - so in this case, you actually don't need to return confirmation that the plane has taken off. Rather, you'll test to make sure it's successfully done so (eg. by checking that the hangar has one less plane in it).

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+  def land(plane)
+    @hangar << plane
+    plane
+  end
+  def take_off(plane)
+    @hangar.pop 
+    return 'Plane no longer in the airport' 
+  end
+  def prevent_landing(plane)
+    fail "Sorry, the airport is full" if full?
```
eoinmakers, line 20: This is good! You could actually take this line and add this to your land method in order to prevent a plane from landing in an airport that's already full.

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+  end
+  def land(plane)
+    @hangar << plane
+    plane
+  end
+  def take_off(plane)
```
eoinmakers, line 15: You don't have to specify which plane is taking off, so you can just define this method as def take_off, and then your method of doing so by running @hangar.pop is good.

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    it 'has a default capacity of 4' do
+      expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+    end
+  describe 'raise error' do
+    it 'if default capacity exceded' do 
+      Airport::DEFAULT_CAPACITY.times do
+        plane = Plane.new
+        subject.land(plane)
+      end
+      expect{subject.prevent_landing(Plane.new)}.to raise_error "Sorry, the airport is full"
```
eoinmakers, line 36: Duplicate tests: this one and the one on line 20.

spec/airport_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    it 'plane take off from airport' do 
+      plane = Plane.new
+      expect(subject.take_off(plane)).to eq 'Plane no longer in the airport' 
+    end
+  end
+  describe "prevent landing" do 
+    it "plane can't land if hangar is full" do
+      plane = Plane.new
+      Airport::DEFAULT_CAPACITY.times {subject.land(plane)}
+      expect{subject.prevent_landing(plane)}.to raise_error "Sorry, the airport is full"
```
eoinmakers, line 23: We want the check to prevent planes from landing when the airport is full, but to do this we would want to somehow tie this check in with the usual way that planes land, rather than creating a separate method for this. I left a comment on the actual method for this one, and I think it would be worthwhile reflecting on.

eoinmakers (commented):
Good progress on this. I would say the main thing to reflect on would be my suggestions around where the check should be for the airport capacity, and my comment around thinking about what a test should expect to be returned.

PR #2224, opened 2021-08-31
Airport Challenge

TomisinJ:


lib/airport.rb
```diff
@@ -0,0 +1,14 @@
+require_relative 'plane'
+
+class Airport
+  attr_reader :plane
+
+  def land(plane)
+    @plane = plane
```
eoinmakers, line 7: Valid implementation of a single plane solution, fulfilling the first user story. 

In order to expand on this, how might you accommodate storing multiple planes at this airport? 

Your current solution takes a plane as an argument and stores it in an instance variable named @plane, which will successfully store that one plane. 

Can you think of any other variable types that could store multiple instances and allow a collection of planes instead?

lib/airport.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
+  def take_off_plane
+    Plane.new
```
eoinmakers, line 11: Successfully allows a plane to take off, that's great. 

In this case, your Airport class will always allow a plane to take off - even if no plane has already landed at the airport.

How could you refactor your implementation so that a plane only takes off if one has previously landed at the airport? What other variable or part of your class would know this information, so that this method could use it to determine whether or not there is a plane that can take off?

spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'airport'
+
+describe Airport do
+
+  describe '#land' do
+
+    it 'lands a plane' do
+      plane = Plane.new
+      subject.land(plane)
+      expect(subject.land(plane)).to eq plane
```
eoinmakers, line 10: There are actually two planes landing here: on line 9 and line 10. 

Line 10 is running the code that the user would run, which is just how unit testing works (except the user is RSpec). So if you were to take out line 9, and just keep line 10, it would still land that plane successfully and check that the value that gets returned is in fact 'plane'.

spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+  end
+
+  it { is_expected.to respond_to(:plane)}
+  
+  it 'returns a plane' do
+    plane = Plane.new
+    subject.land(plane)
+    expect(subject.plane).to eq plane
```
eoinmakers, line 19: Ah! Ignore my previous comment - this is a good test, and the only one you need instead of both this and the 'lands a plane' test.

eoinmakers (commented):
Good progress on the first and second user stories. 

I think the next evolution of this implementation would be to think about storing multiple planes - and also having to take the idea of airport capacity into account.

PR #2227, opened 2021-08-31
Jake's airport challenge

Jjake540:
I got stuck with trying to use a random weather method in my tests. Still little unsure with Rspec

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,7 @@
+require 'airport'
+
+describe Airport do
+  
+  it { is_expected.to respond_to :land_plane }
+
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -27,11 +27,23 @@ describe Airport do
     end
   end
 
-  context 'So that the software can be used for many different airports' do
+  # context 'So that the software can be used for many different airports' do
     
-    it 'I would like a default airport capacity that can be overridden as appropriate' do
-      expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  #   it 'I would like a default airport capacity that can be overridden as appropriate' do
+  #     expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  #   end
+  # end
+
+  context 'So that the software can be used for many different airports' do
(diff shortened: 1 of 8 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  it "Confirm if a plane is ready for take off" do
+    allow(subject).to receive(:weather_forecast).and_return('Sunny')
+    expect {subject.take_off}.to raise_error "No flights avialible"
+  end
+
+  context 'To ensure saftey' do 
+    it 'I want to prevent landing when the airport is full' do
+      allow(subject).to receive(:weather_forecast).and_return('Sunny')
+      subject.capacity.times { subject.landing double :plane }
+      expect { subject.landing double :plane }.to raise_error 'Airport is full'
```
eoinmakers, line 21: Great test.

spec/weather_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  
+    it 'Create weather forecast' do
+      expect(subject).to respond_to(:weather_forecast)
+    end
+
+    it 'Weather_forecast' do
+      weather = Weather.new
+      allow(subject).to receive(:weather_forecast).and_return('Stormy')
+      allow(weather.weather_forecast).to receive(:rand).and_return(0)
+      expect(subject.weather_forecast).to eq "Stormy"
```
eoinmakers, line 15: You don't want to stub the behaviour under test - instead, you want to control that randomness so you can test the behaviour that depends on the _result_ of the randomness.

eoinmakers: All that's necessary to test for this class in the weather_spec.rb file is that it returns one of a series of values, as you'll test the behaviours impacted by specific results in your airport_spec.rb (eg. stubbing stormy and sunny weather accordingly so you can test that planes can or can't take off & land).

eoinmakers (commented):
I thought this was a great airport challenge, really well done. :) I left some comments around dependency injection and class responsibilities - have a read and let me know if you have any questions. These are both topics you'll end up exploring further this week, so you don't necessarily have to make the changes to this specific challenge. 

Well done!

PR #2233, opened 2021-09-24
Feature/airport

thomas99baxter:


lib/airport.rb
```diff
@@ -0,0 +1,59 @@
...
+  def attempt_landing(plane)
+    raise "Too stormy to land" if stormy?
+    raise "Airport full!" if airport_full?
+    plane.flying ? land(plane) : raise("Plane is already landed")
+  end
+
+  def attempt_takeoff(plane)
+    raise "Too stormy to takeoff" if stormy?
+    raise "Airport empty!" if airport_empty?
+    is_landed?(plane) ? takeoff(plane) : raise("Your plane isn't in the airport!")
```
troysnowden, line 25: I like this use of the ternary, very niceee

lib/airport.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  private
+
+  def airport_full?
+    @landed_planes.length >= @capacity
```
troysnowden, line 31: can the length of landed planes ever be greater than the capacity? Like it won't push to the array if it is equal to the capacity so shouldn't ever be more than that, assuming you never pre populate it with more than the capacity

lib/airport.rb
```diff
@@ -0,0 +1,59 @@
...
+  def airport_empty?
+    @landed_planes.length == 0
+  end
+
+  def is_landed?(plane)
+    if @landed_planes.include?(plane) or !plane.flying
+      true
+    else
+      false
+    end
```
troysnowden, line 43: Could use your ternary skillz here.
Also, using or instead of ||??? amateur hour

spec/airport_spec.rb
```diff
@@ -0,0 +1,92 @@
+require "airport"
+require "plane"
+CAPACITY = 20
+
+# this needs to be refactored to use stubs rather than the actual classes
+# but i have ran out of time :()
```
troysnowden, line 6: this is redundant now aye?

spec/airport_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  it "should raise an error when trying to land an already landed plane" do
+    @test_airport.attempt_landing(@test_plane)
+    expect(@test_airport.landed_planes).to eq([@test_plane])
+
+    expect { @test_airport.attempt_landing(@test_plane) }.to raise_error "Plane is already landed"
+  end
+
+  it "should raise an error when trying to land a plane at a full airport" do
+    CAPACITY.times do 
+      # TODO: stub this
```
troysnowden, line 39: what you wanting to stub here?

lib/airport.rb
thomas99baxter: yeah, so i did have this as an == but gave it a kick about and thought it'd be better to have >= just incase of future cahoots. should be fine tho

lib/airport.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def takeoff(plane)
+    @landed_planes.delete(plane)
+    plane.flying = true
+  end
+
+  def stormy?
+    @weather.stormy
+  end
```
troysnowden, line 58: Is this worth having, or just calling the method directly on the weather object possibly?

lib/airport.rb
thomas99baxter: happy to change this. think this was introduced in the rubocop commit though so might cause a linting issue

spec/airport_spec.rb
thomas99baxter: i was wanting to use a double rather than the actual plane class

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
+  attr_reader :stormy
+
+  def initialize
+    @stormy = stormy?
+  end
```
troysnowden, line 6: Do you ever use the initial value of 'stormy'? Or do you just call the stormy method each time? As far as I can tell the latter but I haven't fully checked lol

lib/airport.rb
thomas99baxter: yeah, i agree will stick that in.

lib/weather.rb
thomas99baxter: calling the stormy method on the weather object could change the weather each time its called. so its only being set once and then retrieved by the attr reader

lib/airport.rb
thomas99baxter: calling the stormy method on the weather object could change the weather each time its called. so its only being set once and then retrieved by the attr reader

lib/weather.rb
troysnowden: Righttt, I misunderstood, we've implemented it slightly differently. I've just called the rand method each time whereas you set yours and keep the weather the same. That makes sense all good then

lib/airport.rb
troysnowden: Got ya now

PR #2235, opened 2021-09-24
Pull request for Makers airport challenge (September 2021 tiger)

2Rads:
Messed up the commits so copied files into the pull request instead.


lib/Airport.rb
```diff
@@ -0,0 +1,32 @@
+class Airport
+  @maxPlanes
```
sunnydonaldson, line 2: The style guide for Ruby suggests using snake casing for variable and method names.  ie `max_planes`.  To a certain extent I guess it's personal preference so it's really up to you.  It seems like you've been consistent with camel casing throughout the project so I don't see a problem with it.  Just thought it's worth pointing out.

spec/Airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    :landed => false,
+    :landPlane => landed = true,
+    :takeOff => landed = false
+  } 
+
+  it 'create Airport class' do
+    airport = Airport.new
+  end
+
+  it  { is_expected.to respond_to(:landPlane).with(1).argument }
```
sunnydonaldson, line 15: It's considered good practice in RSPEC to specify the method you're testing using a describe block.
For example here, there are quite a few tests on the `.landPlane` method, and it would be much easier to understand what's going on if they were all grouped together inside `describe "#landPlane"`

spec/Airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+describe Airport do
+  let(:plane) { 
+    double :Airplane,
+    :landed => false,
+    :landPlane => landed = true,
+    :takeOff => landed = false
+  } 
+
+  it 'create Airport class' do
+    airport = Airport.new
```
sunnydonaldson, line 12: This is just an alternate approach, but I find it much easier when writing tests, to use `subject`, which is a premade instance of whatever class you're testing.  This way, you don't have to instantiate Airport in every it block.

lib/Airplane.rb
```diff
@@ -0,0 +1,15 @@
+class Airplane
```
sunnydonaldson, line 1: I really like how simple your Airplane class is, it makes it much easier to track where any errors are coming from since all the logic is being done in the Airport class.

PR #2238, opened 2021-09-24
Airport challenge - michelle-watkins

michelle-watkins:
Things I did:
- Created classes
- Created methods within classes
- Wrote unit tests to test the classes and methods

To-do:
- Implement the correct use of instance variables
- Research initialise method and decide on implementation
- Create a class/method to complete the user story on capacity


lib/airport.rb
```diff
@@ -0,0 +1,14 @@
+require_relative 'plane'
+
+class Airport
+  def take_off
+    types = ['sunny', 'stormy']
+    @weather = types.sample
+    if @weather == 'stormy'
+      return "Flight delayed"
```
duckduckboom, line 8: I like the use of sample method. Though, I think the weather may be better as a seperate thing. 

Airport tests pass. 89% due to missing line 8 about the storminess - unsure why.

michelle-watkins:
Second attempt after more practice with TDD

PR #2240, opened 2021-09-25
Asking for review

DannyL90:
Please give feedback on my code

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -9,8 +9,12 @@ class Airport
     end
 
     def take_off(plane)
-        @register.delete(plane)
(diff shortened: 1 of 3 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,25 @@
+class Airport
+
+    def initialize
+        @register = []
```
pinteck, line 4: Since you are using plane ids to keep track of the planes in the register, using a hash would be an appropriate data structure.

pinteck (approved):
I liked that you wrote a very clear README and your code is readable and clean. 
Thank you

PR #2241, opened 2021-09-25
Airport Challenge. Nejla Lowe.

nejlalowe:


lib/airport.rb
```diff
@@ -0,0 +1,23 @@
+class Airport
+  attr_reader :capacity, :planes_hub
+
+  def initialize(capacity = 50)
+    @capacity = capacity
```
LeoBrownPhonic, line 5: Based on the Boris Bikes example I think this should perhaps be a Class Constant with a capital letter?

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+    @planes_hub = []
+  end
+
+  def land(plane)
```
LeoBrownPhonic, line 9: This is where I get confused - user story says the air traffic controller wants to instruct the PLANE to lane.  I put land as a method on the Plane Class but then ran into further considerations (like prevent landing) that caused me to think more about this

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+    fail "this airport is full" if full?
+    @planes_hub << plane
+  end
+
+  def takeoff(plane)
```
LeoBrownPhonic, line 14: This will remove a plane but it assumes the planes land and later takeoff in the same order.  If the planes become unique instances (i.e. each plane is identifiable (Air France45675 = Plane.new) this might remove the wrong plane from the airport list.

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+    @planes_hub.pop
+  end
+
+  private
+
```
LeoBrownPhonic, line 19: This looks right to me - full? as a private method

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "./lib/airport"
+require "./lib/weather"
+
+describe Airport do
+  it "allows a plane to land at airport" do
+    expect(subject).to respond_to(:land).with(1).arguments
+  end
+
+  it "allows a plane to takeoff, and confirms it is no longer in the airport" do
+    airport = Airport.new
```
LeoBrownPhonic, line 10: I interpreted this as sending a message back to the air traffic controller i.e. - I have taken off.  Adjusting the airport data store is fine but doesn't explicitly report this to air traffic control

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    plane2 = Plane.new
+    airport.land(plane1)
+    airport.land(plane2)
+    airport.takeoff(plane2)
+    expect(airport.planes_hub.count).to eq 1
+  end
+
+  it "return an error if airport is full" do
+    airport = Airport.new
+    50.times { airport.land(Plane.new) }
```
LeoBrownPhonic, line 21: Eventually this '50' should be replaced by the default capacity constant (+1) so the test changes automatically to match the altered airport capacity.

spec/plane_spec.rb
```diff
@@ -0,0 +1,6 @@
+require "plane"
+describe Plane do
+  it "confirms that plane is no longer in the airport" do
+    expect(subject.departed).to eq true
```
LeoBrownPhonic, line 4: Should departed be 'flying'? that way it could be also used to prevent already landed planes from landing

PR #2242, opened 2021-09-25
Airport Challenge

FrancineBlanc:
This is my TDD approach to the airport challenge.

### DIFF
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,74 @@
+require 'airport'
+
+describe Airport do
+
+  let(:plane) { Plane.new }
+
+  context "change capacity" do
+    override_capacity = 10
+    subject { Airport.new(override_capacity) }
+    it "can override the default capacity" do
+      expect(subject.capacity).to eq override_capacity
+    end
+  end
+
+  context "normal weather conditions" do
+    before(:example) do
+      Weather.any_instance.stub(:weather_conditions).and_return("sunny")
+    end
+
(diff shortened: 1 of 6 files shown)


spec/plane_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'plane'
+require 'airport'
+
+describe Plane do
+
+  let(:airport) { Airport.new }
+
+  context "is grounded" do
+    it "can take off from an airport" do
+      expect(subject.can_takeoff(airport)).to be
```
ewa-j, line 10: Maybe it would be worth to specify what exactly it should equal to? '.to' be may pass the test, but I think it would be more specific if it said 'to be_truthy' for example

spec/airport_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    it "does not allow take off" do
+      expect { subject.takeoff(plane) }.to raise_error "Weather is stormy. Takeoff not permitted."
+    end
+
+    it "does not allow planes to land" do
+      expect { subject.land(plane) }.to raise_error "Weather is stormy. Landing not permitted."
+    end
+  end
+end
```
ewa-j, line 74: Good use of context, it makes the tests easy to read and understand

ewa-j (commented):
I think the program and the tests are well written and very easy to understand. The methods to their job in the easiest way possible, which makes the code simple and effective. Tests are short and not overcomplicated.

PR #2252, opened 2021-09-26
tom_kellett_airport_challenge

t-kellett:


t-kellett:
Couple of questions I am concerned about:

1. Is it problematic that the airport's landed_planes is an accessor? The plan can technically take things that are not planes if someone acted maliciously.. You could fill up an airport with strings and then not land any planes for example. Though this is unlikely as someone would need to know the underlying code in order to do this...
2. Do I need a Weather class? SRP indicates I might, but also a class with 1 method means that I end up with very fragmented code and this could lead to complex interdependency in testing.

lib/airport.rb
```diff
@@ -0,0 +1,17 @@
...
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @landed_planes = []
+    @capacity = capacity
+    @stormy_weather = stormy?
+  end
+
+  def stormy?
+    rand(9) < 9
+  end
+
```
ConorButler, line 16: Weather can be put in a different class to follow SRP. The airport class is handling the storage of planes **and** generating random weather.

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  it "airport should have a default capacity that can be overridden as appropriate" do
+    expect(subject.capacity).to eq 10 
+    expect(Airport.new(20).capacity).to eq 20
+  end
+
+  it "airport stormy weather should be true when stormy? is true" do
+    allow(subject).to receive(:stormy?) { true }
+    expect(subject.stormy_weather).to eq true
+  end
+
```
ConorButler, line 24: This might be clearer if it was in one method/variable.

spec/plane_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  let(:stormy_airport) { double(:stormy_airport) }
+
+  before do
+    allow(airport).to receive(:landed_planes).and_return([])
+    allow(airport).to receive(:capacity).and_return(1)
+    allow(airport).to receive(:stormy_weather).and_return(false)
+    allow(stormy_airport).to receive(:stormy_weather).and_return(true)
+    allow(stormy_airport).to receive(:landed_planes).and_return([])
+    allow(stormy_airport).to receive(:capacity).and_return(10)
+  end
```
ConorButler, line 14: Comments explaining the process would make it easier to follow.

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
+class Plane
+  attr_reader :plane, :in_flight
+  attr_accessor :current_airport
+  def initialize
+    @in_flight = false
+    @current_airport = nil
+  end
+
```
ConorButler, line 8: If planes start already out of flight and with no airport, there is no way to add them to landed_planes because they won't get past the guard clause on line 11 unless you manually add them.

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  before do 
+    allow(plane).to receive(:land) do |subject|
+      subject.landed_planes << plane
+    end
+    allow(plane).to receive(:take_off) do |subject|
+      subject.landed_planes.delete(plane)
+    end
+  end
+
```
ConorButler, line 14: I'm not clear on what this doing. Is it a stub for the plane double telling it what to do when it receives land and take_off and if so does it test what the Airport class is doing?  Could you comment to explain it?

spec/plane_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  it "prevents take off from an airport you are not at" do
+    message = "You cannot take off from an airport you are not at"
+    second_airport = double("second_airport", :landed_planes => [], :capacity => 10, :stormy_weather => false)
+    subject.current_airport = airport
+
+    expect { subject.take_off(second_airport) }.to raise_error message
+  end
+end
```
ConorButler, line 80: Could do with a test that lands/takes off multiple planes at once.

lib/airport.rb
```diff
@@ -0,0 +1,17 @@
+class Airport
+  attr_reader :capacity, :stormy_weather
+  attr_accessor :landed_planes
```
ConorButler, line 3: Regarding the attr_accessor, you could land objects that aren't planes, but because they are not from the Plane class you would not be able to call the methods only Planes can use. However, you don't call any Plane specific methods on landed_planes (just array methods), so you are right in that other objects could be put in this variable. I'm not sure how to solve this.

PR #2253, opened 2021-09-26
Create airport

Matt-Warnock:


spec/airport_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  describe '#land' do
+    it 'lands a plan at airport' do
+      airport.land(plane)
+
+      expect(airport.hanger.first).to eq plane
+    end
+
+    it 'raises error if airport is already full' do
+      (1..3).each { airport.land(Plane.new) }
```
AJOsmaston, line 30: Would
`3.times { airport.land(Plane.new) }`
be better for these?

PR #2255, opened 2021-09-26
Airport challenge submission

alexboyling:
- Completed all user stories
- One edge case not working: a landed plane can land at a new airport without first taking off

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-Approach
+## Approach
 
 * Tackled each user story one by one
 * For each user story I captured the objects (nouns) and messages (verbs)
@@ -8,13 +8,13 @@ Approach
 * If test coverage was not 100% additional tests were written until this was met
 * This enabled the test coverage to remain high throughout the entire process
 
-Thoughts
+## Thoughts
 
 * Initially I wanted to write the land and takeoff methods for the Plane class
 * This seems 'logical' as a plane lands or takeoffs rather than an airport
 * However I quickly switched to using the Airport class instead as this was a simpler approach that still met the user stories
 
-Outstanding issues
+## Outstanding issues
 
 * Its possible to land a plane at one airport then land it at another without taking off
 * I started adding more functionaility to the plane class to track plane location but then ran out of time and reverted
--- a/README.md
+++ b/README.md
@@ -1,3 +1,25 @@
+Approach
+
+* Tackled each user story one by one
+* For each user story I captured the objects (nouns) and messages (verbs)
+* A minimum test was written first and made to fail
+* Minimum code was then written to make it pass followed by refactoring
+* Rubocop was run after each successful test and refactor to keep the code clean
+* If test coverage was not 100% additional tests were written until this was met
+* This enabled the test coverage to remain high throughout the entire process
+
+Thoughts
+
+* Initially I wanted to write the land and takeoff methods for the Plane class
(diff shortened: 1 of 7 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'airport'
```
blue-luke, line 1: Your tests are very readable, give a good insight into what the code does

spec/weather_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  it { is_expected.to respond_to :status }
+  it { is_expected.to respond_to :random }
+  it "should return either 'sunny' or 'stormy'" do
+    allow(weather).to receive(:random).and_return('sunny','stormy')
+  end
+  it "should return 'sunny' when sunny" do
+    allow(weather).to receive(:status) { 'sunny' }
+    expect(weather.status).to eq 'sunny'
+  end
+  it "should return 'stormy' when stormy" do
```
blue-luke, line 14: These tests seem circular, as it is only testing the double that has just been created, not the underlying code. There should be a way to both test the underlying code and use stubs to ensure that weather takes a specific value

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
+  attr_reader :status
+
+  def initialize
+    @status = random
+  end
+
+  def random
+    # 10% chance of stormy weather
+    rand(10).zero? ? @status = 'stormy' : @status = 'sunny'
```
blue-luke, line 10: I like this way of setting the weather, it's concise

README.md
```diff
@@ -1,3 +1,25 @@
...
+
+## Thoughts
+
+* Initially I wanted to write the land and takeoff methods for the Plane class
+* This seems 'logical' as a plane lands or takeoffs rather than an airport
+* However I quickly switched to using the Airport class instead as this was a simpler approach that still met the user stories
+
+## Outstanding issues
+
+* Its possible to land a plane at one airport then land it at another without taking off
```
blue-luke, line 19: Perhaps you could suggest how you would go about achieving this, even if you don't have time to actually code it. This would be of benefit to future-you, or another person who comes to add to the code

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    raise 'Airport full' if full?
+    raise 'Plane cannot land in stormy weather' if weather.status == 'stormy'
+    raise 'Plane already landed, cannot land again' if @planes.include? plane
+    @planes << plane
+  end
+
+  def takeoff(plane)
+    raise 'Plane cannot takeoff in stormy weather' if weather.status == 'stormy'
+    raise 'Plane cannot takeoff as not in airport' unless @planes.include? plane
+    @planes.reject! { |n| n == plane }
```
blue-luke, line 23: Why did you choose this method of deleting the plane from the array? Are there advantages to it? Just curious.

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather
+  attr_reader :status
+
+  def initialize
```
blue-luke, line 4: Just curious, why did you choose to initialize the weather with a status?

blue-luke (commented):
Overall seems to be high quality and well-supported by tests. A couple of design topics that I'd like to discuss in more detail

PR #2256, opened 2021-09-26
Airport Challenge

dpkerins:


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
+class Airport
+  attr_reader :planes
+  attr_reader :capacity
+  DEFAULT_CAPACITY = 1
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane, weather)
```
emanfolo, line 10: I like the fact you have weather in a different class,

emanfolo (commented):
Really effective and clear code, good test coverage which clearly defines what is expected from the code

PR #2257, opened 2021-09-26
Airport Challenge - Conor Butler

ConorButler:


ConorButler:
- Tried to use doubles to isolate each class when testing
- I'd like specific feedback on how I used them, if it's readable, and how I can refactor it

README.md
```diff
@@ -1,89 +1,11 @@
...
-* Write an RSpec **feature** test that lands and takes off a number of planes
-
-Note that is a practice 'tech test' of the kinds that employers use to screen developer applicants.  More detailed submission requirements/guidelines are in [CONTRIBUTING.md](CONTRIBUTING.md)
-
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first.
-
-* **Submit a pull request early.**
-
-* Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+# Airport challenge
```
t-kellett, line 1: Good that the README has been changed and gives brief explanation of how you tackled the problem. I would perhaps add a section to explain what the whole program does (i.e. 'This is a program which lands and takes off planes from an airport' etc so in future you are able to refer back to this and not need to dig into code to remind yourself what is actually happening)

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+class Airport
+  DEFAULT_CAPACITY = 50
+  attr_reader :planes, :capacity
+
+  def initialize(airport_id, capacity = DEFAULT_CAPACITY)
+    @airport_id = airport_id
```
t-kellett, line 6: Is the airport ID necessary? When a new airport is created, it creates a unique memory specific reference, so you are able to ensure that you are looking at the unique object you think you are. Is this for readability?

t-kellett (commented):
Overall good, creating a new weather instance every time you land or takeoff is the only thing I think *needs* to change. The rest is mainly stylistic and/or reduction in lines of code

spec/airport_spec.rb
```diff
@@ -0,0 +1,143 @@
+require 'airport'
+
+describe Airport do
+  let(:good_weather) { $weather = double("Weather", :stormy => false) }
+  let(:bad_weather) { $weather = double("Weather", :stormy => true) }
+
+  context 'landing:' do
+    airport = Airport.new(1500)
```
leoht, line 8: It's best to use either a `before` block or a `let` binding to declare `airport` here, it can be done in a similar way to how you've setup `plane` values:
```ruby
let(:airport) { Airport.new(1500) }
```

leoht (commented):
Nice work, and quite a lot of tests! I've had mainly comments about how to avoid using global variables in the exercise, replace it with dependency-injection, some minor improvements to the code, etc. Any question on the feedback do let me know :)

PR #2258, opened 2021-09-26
Week 1 - Weekend Airport Challenge

Kntanos:


Kntanos:
I am aware that the specs need to be tided up, but I am unsure what is the best way to do it.

Kntanos:
Split weather into a new class and updated specs so that they pass. I haven't yet written tests for the new class.
My main concern is making the specs a bit more concise, so I don't have to set up every time all the objects etc.

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :airport, :hangar, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airport = airport
```
leoht, line 8: It might be hard to understand what's at play here, but effectively the `attr_reader` defines a method `.airport` that returns the current value of `@airport`. And, by default, all instance variables are initialized to nil. So if we deconstruct this line:
```irb
@airport = airport
# which is
@airport = @airport
# which is
@airport = nil
```

So the value of `@airport` will stay `nil`. But then this attribute is probably not needed, and could be removed, since it's not used anywhere. What's important to keep as the "internal state" of the airport is the planes it contains (and maybe its capacity). The "airport" then is a whole instance of this class

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+  DEFAULT_CAPACITY = 10
+  attr_reader :airport, :hangar, :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airport = airport
+    @hangar = []
+    @capacity = capacity
+  end
+
+  def full
```
leoht, line 13: It might be good to rename it to `full?` to follow Ruby conventions

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+    hangar.size >= capacity
+  end
+
+  def stormy?
+    weather.forecast
```
leoht, line 18: This `weather` name doesn't seem to be defined here (you'll likely get a NoMethodError) — if it's intended to be an instance of the `Weather` class, it could be injected in your `initialize` method:
```ruby
def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
  # ...
  @weather = weather
end
```

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
...
+  def initialize(ground_location = nil)
+    @ground_location = ground_location
+  end
+
+  def land(airport)
+    fail "The weather is bad! You can't land!" if airport.stormy?
+    fail "You can't land here. The airport is full!" if airport.full
+    fail "This plane is already on the ground!" if ground_location == :ground
+    self.ground_location = :ground
+    airport.hangar << self
```
leoht, line 15: Two things here: 
 - It's probably good to keep only one single "source of truth" source of data: either the airport `hangar` array, either the `ground_location` attribute. For example, if we keep only the array, it's enough for us to know whether a plane is "landed" or not (in the array or not). That avoids us potential data inconsistencies and bugs later, since we wouldn't have to maintain two different sources of data when the state is changing
 - If you'd like to keep only `hangar` and modify it here, maybe use a method rather than accessing directly the array. Maybe something like `airport.add_plane(self)`

lib/plane.rb
```diff
@@ -0,0 +1,28 @@
...
+
+class Plane
+  attr_accessor :ground_location
+
+  def initialize(ground_location = nil)
+    @ground_location = ground_location
+  end
+
+  def land(airport)
+    fail "The weather is bad! You can't land!" if airport.stormy?
```
leoht, line 11: Nice use of defensive programming with `fail` clauses :)

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
+attr_reader :weather
+
+def initialize
```
leoht, line 4: A small fix to indentation can be made here

leoht: Also, there seems to be the same "pattern" as in the Airport class:  `@weather` is not needed in the initialize method, since the "weather" will be an instance of this very class (`Weather.new`).

spec/plane_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  describe "#land" do
+    it "should accept one argument" do
+      expect(subject).to respond_to(:land).with(1).argument
+    end
+  
+    it "should allow a plane to land at an airport" do
+      allow(airport).to receive(:stormy?).and_return false
+      expect(subject.land(airport)).to eq(airport.hangar << subject)
```
leoht, line 25: It's probably better to not modify the state of `airport.hangar` within the `eq(...)` matcher, since it could have unexpected results (and is a bit harder to read). Consider maybe this alternative:
```ruby
expected_hangar = airport.hangar + [subject]

expect(subject.land(airport)).to eq(expected_hangar)
```

spec/plane_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    it "should raise an error if the airport is full" do
+      plane = Plane.new
+      airport = Airport.new
+      allow(airport).to receive(:stormy?).and_return false
+      allow(airport).to receive(:full).and_return true
```
leoht, line 32: A better way to test if the airport is full without mocking it (after all, this is part of the behaviour of the class, so we probably want to test it too) is to override the capacity to a low number, maybe 1 plane max, and then try adding one plane, and a second one — then the error should be raised:
```ruby
plane_1 = Plane.new
plane_2 = Plane.new
airport = Airport.new(1)
plane_1.land(airport) # first plane, all good
expect { plane_2.land(airport) }.to raise_error "You can't land here. The airport is full!"
```

spec/plane_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      airport = Airport.new  
+      allow(airport).to receive(:stormy?).and_return true
+      expect { subject.land(airport) }.to raise_error "The weather is bad! You can't land!"
+    end
+
+    it "should raise and error if plane is landed elsewhere" do
+      plane = Plane.new
+      airport = Airport.new
+      allow(airport).to receive(:stormy?).and_return false
+      allow(plane).to receive(:ground_location).and_return :ground
```
leoht, line 46: Same as above here, consider injecting the `ground_location` state when creating the `Plane` instead of mocking:
`plane = Plane.new(:ground)`

spec/plane_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      airport = Airport.new
+      allow(airport).to receive(:stormy?).and_return false
+      plane.land(airport)
+      plane.take_off(airport)
+      expect(plane.departed?(airport)).to be true 
+    end
+  end
+end
+
+describe Airport do
```
leoht, line 86: It's usually better to separate test suites to have one per class, which would result in two files `airport_spec.rb` and `plane_spec.rb`

leoht (commented):
Nice work! The code is clear to read and nicely laid out — I had mainly comments on a few things here and there to simplify the code and improve a bit tests, code design, etc :) Let me know if anything seems unclear on the feedback and we can chat about it

lib/plane.rb
Kntanos: My understanding is that I am dealing with two different scenarios:
a) the airport hangar is full 
b) the airport hangar is not full, but the plane is already on the ground somewhere else, not in the airport it is trying to land into. Hence the status :ground will not allow it to land again somewhere else.  Without this error in irb you could land the same plane in different airports at the same time.
So only checking if the plane is in the hangar is not enough to cover both scenarios. Because it may not be in the hangar, but it is in some other airports' hangar.
Is this what you are saying here or am I misunderstanding?

spec/plane_spec.rb
Kntanos: Good one. I haven't been using or even thinking of testing like this with an 'expected' outcome. Need to keep this in mind.

lib/airport.rb
Kntanos: Yes, you're right. I understand the concept. It must have been left in there from trying out things when building the code. I guess since it's not doing anything, it was also not picked up.

lib/plane.rb
leoht: Ah, that makes sense indeed, because:
 * you need to be able to know in which airport is (to avoid adding it to two airports at the same time)
 * you need to count how many planes are in the airport

In that case it's fine to have two sources of data — I guess that's the disadvantage of this way of modelling it (rather than having the `land` and `takeoff` methods inside the Airport class). You could still simplify it a bit by maybe having a `@current_airport` attribute in the Plane class (rather than `ground_location`), this way you could keep track of which airport the plane is currently "attached" to using this attribute

PR #2259, opened 2021-09-26
Airport Challenge 

Eagle555:


lib/airplane.rb
```diff
@@ -1,5 +1,7 @@
 class AirPlane
+    
   def exist?
```
Pinkish-Warrior, line 3: would this class be essential? once that the airplane is a true statement already...

spec/weather_spec.rb
```diff
@@ -2,10 +2,10 @@
```
Pinkish-Warrior, line 2: would we necessary need a new spec for the weather? Not sure

PR #2260, opened 2021-09-26
(MF) Airport Challenge

maxcfry:


airport/lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+    def random_weather
+        weather = forecast
+    end
+
+    private
+
+    OUTLOOK = [:stormy, :sunny, :sunny, :mild]
+
+    def forecast
+        OUTLOOK.sample
```
TTurvey, line 43: I like your approach to the random weather generator. It hadn't occurred to me that an array could be used for this so well done for thinking outside of the box on this.

airport/lib/airport.rb
```diff
@@ -0,0 +1,49 @@
+require_relative 'plane'
+
+class Airport
+DEFAULT_CAPACITY = 20
+
+    attr_reader :planes
+    attr_accessor :weather
+
+    def initialize(planes = [], full = DEFAULT_CAPACITY, weather = :sunny)
```
leoht, line 9: It'd probably be good to rename this attribute `max_capacity` for example, as "full" indicates more a state rather than a number or capacity

airport/lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def full?
+        @planes.length >= @full
+    end
+
+    def stormy?
+        weather == :stormy
+    end
+
+    def random_weather
```
leoht, line 34: Again here, maybe a better name could be `assign_random_weather` to indicate this method is modifying/updating something (otherwise simply `random_weather` lets us think it's only computing a value and returning it)

Also, to respect the SRP, it might be better to put the "weather" logic in its own class, probably named `Weather`. An instance of this class would be then injected via the `initialize` of the Airport class

airport/spec/airport_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    it "allows plane to take off" do
+        airport = Airport.new([plane])
+        airport.take_off(plane)
+        expect(airport.planes).to be_empty
+    end
+
+    it "does not allow planes to land when airport is full (20 planes)" do
+        Airport::DEFAULT_CAPACITY.times { airport.land Plane.new }
```
leoht, line 20: Rather than relying on the constant, it might be better to pass the maximum capacity to `Airport.new` when creating it (since we're able to override the default capacity!) — and this way we can test that this overriding is working too:
```ruby
# for example
capacity = 10
airport = Airport.new([], capacity)
capacity.times { airport.land(Plane.new) }
```
```

File: airport/spec/airport_spec.rb [hunk 5]
```diff
@@ -0,0 +1,37 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+    plane = Plane.new
```
leoht, line 6: You could use either `let` or a `before` block with instance variables to set those values and reuse them in tests

leoht (commented):
Nice work! I've add a few comments here and there to improve code readability, and other minor things. It'd also be good to extract the weather logic to its own class (the SRP!), and to inject it via the `initialize` method. However I do agree it's simpler to start off the way you did, having everything in the Airport class, and more manageable in the beginning. It's probably a good exercise to then try to extract `Weather` in a further step! Any question on the feedback, let me know :)

PR #2262, opened 2021-09-26
Leo Brown airport_challenge Submission

LeoBrownPhonic:
Submission for the airport_challenge solo project Week3 TDD.  Partially completed.  File includes supporting planning documents in docx format


spec/plane_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    expect(subject.land(airport)).to eq true
+  end
+
+  it "Won't land if weather is stormy" do
+    airport = Airport.new
+    allow(airport).to receive(:weather) { "Stormy" }
+    expect { subject.land }.to raise_error
+  end
+
+  it "Won't land if airport is full" do
```
nejlalowe, line 15: perhaps for when you mentioned you want to introduce airport capacity that can be overridden, you can introduce a default capacity when initializing the airplane class if the user does not enter an argument when creating their airport instance, and have an instance variable below that where the user can enter whatever they wish, thus your full method can be re factored to check if the plane storage is larger than the capacity

PR #2264, opened 2021-09-26
Only and final commit for airport-challenge

charlie83xt:
This challenge nearly killed me. Too much just for the first week!


charlie83xt:
- Thanks for taking the time to review my code.
- What I tried to do the most was getting landed planes on the air and flying planes on ground as long as the weather allows it.
- My test are messy, but all i am looking for is functionality.

airport/lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+
+
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @flying_planes = [] 
+    @landed_planes = []
+  end
+
+  def capacity
```
marazzo, line 16: You can make @capacity instance variable as a attribute reader and no need for this method then

airport/README.md
```diff
@@ -0,0 +1,5 @@
+Software to control the flow of planes at an airport.
```
marazzo, line 1: Very clear and concise README - can understand the scope of the project quite easily from these lines

airport/lib/landing_planes.rb
```diff
@@ -0,0 +1,27 @@
+
+class Landing_planes
+  attr_reader :landed_planes
```
marazzo, line 3: as a little tip, you can have all attr_readers on one line separated by a " , "

airport/lib/landing_planes.rb
```diff
@@ -0,0 +1,27 @@
...
+    @landed_planes = landed_planes
+    flying_plane.each do |plane|
+      while @landed_planes.length < self.maximum do @landed_planes.push(plane)
+      end
+    end
+    flying_plane.each { |item| flying_plane.delete(item) }
+    @landed_planes
+  end
+  
+  def maximum
```
marazzo, line 23: if you have a attr_reader for capacity, this method isn't really adding much I would say.

airport/lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+    @capacity
+  end
+
+  def landing_flow
+    landing_planes = Landing_planes.new(@capacity)
+    airport = Airport.new
```
marazzo, line 22: In this case you are initilializing an airport object in the airport class - perhaps another way of going about this is storing the weather in an attr_reader so that it is accessible from the other classes as well as this one. 
And then you can refer to it as :weather == sunny?

airport/lib/plane.rb
```diff
@@ -0,0 +1,11 @@
+
```
marazzo, line 1: perhaps in the Plane class you could have added a position class variable. Where landing would change position to 'in airport' and leaving would change it to "in air" to give the plane more functionality

airport/spec/landing_planes_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'landing_planes'
```
marazzo, line 1: This has some great tests !

airport/spec/plane_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'plane'
```
marazzo, line 1: Tests the class fully :)

airport/spec/take_off_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    take_off = Take_off.new
+    landed_planes = []
+    flying_plane = []
+    expect(take_off.flying(landed_planes, flying_plane)).to eq("Empty airport. No planes to take_off")
+  end
+
+  it "checks that no plane takes off under stormy weather" do
+    take_off = Take_off.new
+    airport = double("airport")
+    allow(airport).to receive_message_chain(:weather).and_return("sunny")
```
marazzo, line 25: there is no 'expect' function here

marazzo (commented):
Was super interesting reading through your code - I think you met all the user stories and the code is very readable!!

PR #2265, opened 2021-09-26
airport challenge

alexwintle:


airport_docs.md
```diff
@@ -0,0 +1,53 @@
...
+| | take_off | Instruct a plane to take off from an airport and confirm that it is no longer in the airport | So I can get passengers on the way to their destination
+| | airport_full? | So I can get passengers on the way to their destination | To ensure safety
+| | airport_capacity | | 
+| | planes_amount | |
+| | update_planes_amount | |
+| | prevent_flights? | |
+| weather | weather_stormy? | Prevent takeoff AND landing when weather is stormy | To ensure safety
+| | random_weather | set the weather (it is normally sunny but on rare occasions it may be stormy) | 
+| | weather_status | |
+| system_designer | default_capacity(capacity) | Add a default airport capacity that can be overridden as appropriate | So that the software can be used for many different airports
```
kal31, line 51: I think that that diagram is really easy to follow and understand how the user story will be converted to feature stories.  
The code review stated one idea is to use an irb transcript.  Im not sure its useful when the codebase becomes more complex.

spec/air_traffic_control_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  describe "#planes_amount" do
+    it 'should return the amount of planes as an integer' do
+      control = AirTrafficControl.new(20)
+      control.update_planes_amount(18)
+      expect(control.planes_amount).to eq 18
+    end
+  end
+
+  describe "#update_planes_amount" do
```
kal31, line 77: each behaviour is within a new describe test which is great.  I think the code review notes state that each state within the behaviour should have a context when the tests are related by program state (e.g. 'when it is stormy'). its really easy to follow and understand as well.

spec/weather_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'weather'
+
+describe Weather do
+  describe "#random_weather" do
+    it 'should return a random weather state, stormy if true, sunny if false' do
+      weather = Weather.new
+      expect(subject.random_weather).to eq "sunny" if weather.random_weather.eql? true
```
kal31, line 7: is this expect the same thing as using stubs to mock the random method?  i did something different.  it would be good to discuss and compare our code.

PR #2266, opened 2021-09-26
Airport Challenge

SamJT2021:


README.md
```diff
@@ -12,27 +12,37 @@ Airport Challenge
...
 
-Steps
--------
+My approach to this challenge was to work through each individual user story, breaking each component down into the simplest requirement to create each class and method as needed, while applying the **RED, GREEN, REFACTOR** method in parallel. This method requires a test to be created initially to define what each element of the application should be doing, running the test to ensure they are failing in the intended way, then developing the feature using the test requirements. Once the feature is working correctly and the tests are passing, I then revisited my code to improve implementation and readability for better maintainability. As a personal focus, on the day of the Airport challenge, I had set myself the goal of learning more about RSpec's "syntactic sugar" to reduce lines of unnecessary code and improve readability, and in doing so, I have applied some examples of this in my tests.
```
stevej763, line 30: Nice and clear, good to know where you are coming from and what to look out for 👍

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
+    @planes = []
+    @capacity = capacity
+    @weather = weather
+  end
+
+  def land(plane)
+    fail 'Unable to land, the airport is full' if airport_full?
+    fail 'Unable to land due to storm' if weather.stormy?
```
stevej763, line 17: Nice use of guard clauses

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  attr_reader :planes, :capacity, :weather
+
+  DEFAULT_CAPACITY = 30
+
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
```
stevej763, line 9: Passing the weather class in as a parameter instead of initialising in the class itself 👍

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+    @planes = []
+    @capacity = capacity
+    @weather = weather
+  end
+
+  def land(plane)
```
stevej763, line 15: What happens if you try to land the same plane twice?

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+    fail 'Unable to land, the airport is full' if airport_full?
+    fail 'Unable to land due to storm' if weather.stormy?
+    planes << plane
+  end
+
+  def takeoff
```
stevej763, line 21: Could you make it so you can tell the airport which plane you want to take off?

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+    fail 'Unable to take off due to storm' if weather.stormy?
+    planes.pop
```
stevej763, line 23: What would happen if there are no planes?

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+  
+  private
+
+  def airport_full?
+    planes.count >= capacity
```
stevej763, line 29: Tiny thing, preference more than right or wrong, but unnecessary to be >=, can just be ==

lib/plane.rb
```diff
@@ -0,0 +1,10 @@
+class Plane
+
+  def in_airport
+    @landed = true
+  end
+
+  def landed?
+    @landed
+  end 
+end
```
stevej763, line 10: Nice and simple domain obeject, love it

stevej763: Could you make it so the plane knows which airport it is currently at?  What would happen if a plane that was in an airport already was asked to land at another airport?

lib/weather.rb
```diff
@@ -0,0 +1,7 @@
+class Weather
+
+  def stormy?
+    rand(10).zero?
+  end
+
+end
```
stevej763, line 7: Nice use of SRP, a separate class to deal with the weather 👍

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { Plane.new }
```
stevej763, line 4: Could have used a double here

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { Plane.new } 
+
+  before do
+    allow(subject.weather).to receive(:stormy?) { false }
```
stevej763, line 7: Good use of a stub to make sure the weather is always good 👍

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end 
+
+    it 'should allow the default capacity to be overridden' do
+      large_airport = Airport.new(100)
+      expect(large_airport.capacity).to eq 100
+    end 
+  end
+
+  describe '#land' do
+    it { is_expected.to respond_to(:land).with(1).argument }
```
stevej763, line 26: Good use of Rspec single line syntax 👍

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    it 'should allow a plane to land' do
+      expect(subject.land(plane)).to eq [plane]
+    end
+
+    it 'should prevent landing when airport is full' do
+      subject.capacity.times { subject.land double :plane }
+      expect { subject.land double :plane }.to raise_error 'Unable to land, the airport is full'
```
stevej763, line 34: Looks like you are using a double here, but not in the 'should allow a plane to land' test above?

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    it 'should prevent landing when stormy' do
+      allow(subject.weather).to receive(:stormy?) { true }
```
stevej763, line 38: Stubbing weather to be stormy for this specific test 👍

spec/airport_spec.rb
stevej763: I can see you reference this variable in some places in this unit test, but in others do create a double. If you moved that double declaration up here you could have used this throughout, would be something like: 

`let(:plane) { double :plane }`

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    it 'should prevent take off when stormy' do
+      subject.land(plane)
+      allow(subject.weather).to receive(:stormy?) { true }
+      expect { subject.takeoff }.to raise_error 'Unable to take off due to storm'
+    end
+  end
+
+  describe 'feature' do
```
stevej763, line 64: This feature test should probably have it's own file

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+    fail 'Unable to land due to storm' if weather.stormy?
+    planes << plane
+  end
+
+  def takeoff
+    fail 'Unable to take off due to storm' if weather.stormy?
+    planes.pop
+  end
+  
+  private
```
stevej763, line 26: Nice one making private methods 👍

stevej763 (commented):
Looks good Sam! Good explanations of everything you have done. Well formatted, following OOP with separate classes for each object. Follows the SRP well and uses private methods. Testing is thorough and everything is easy to read and understand. Most of my constructive feedback is for things that were probably out of scope for this exercise - such as the same plane landing/taking off multiple times etc so by no means do I think that detracts from the work. Clearly showing your understanding of TDD and OOP. Nice job!

lib/airport.rb
stevej763: actually forget that, you are right to do it the way you did

stevej763 (commented):
Looks good Sam! Good explanations of everything you have done. Well formatted, following OOP with separate classes for each object. Follows the SRP well and uses private methods. Testing is thorough and everything is easy to read and understand. Most of my constructive feedback is for things that were probably out of scope for this exercise - such as the same plane landing/taking off multiple times etc so by no means do I think that detracts from the work. Clearly showing your understanding of TDD and OOP. Nice job!

PR #2267, opened 2021-09-26
Airport Challenge User Stories 1 to 4

Anamaria-Mocanu:


lib/airport.rb
```diff
@@ -0,0 +1,23 @@
+class Airport
```
AishaDorsett, line 1: I like that you have created 2 separate files & classes for airport, planes and weather

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
+require "Airport"
+
+describe Airport do
+  
+  describe "#initialize" do
+    it "checks the set capacity of the hangar" do
+      air = Airport.new("Heathrow")
+      # expect(Airport.instance_variable_get(:@hangar)).to equal(3)
```
AishaDorsett, line 8: *improvement* As this is dead code, it should be taken out of your test file

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      expect(air.status.length).to eq(0)
+    end
+
+    # before { Airport.instance_variable_set(:@hangar, 3) }
+    it "checks if you can change the capacity of the hangar" do
```
AishaDorsett, line 13: I like your 'it' statements. very clear about what the test does

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    it "diplays content of hangar" do
+      airport = Airport.new("Heathrow")
+      airport.add_plane("plane 1")
+      expect(airport.status).to eq ["plane 1"]
+    end
+  end
+
+  describe "#remove_plane" do
+    it 'removes plane from hangar' do
+      airport = Airport.new("Heathrow")
```
AishaDorsett, line 46: *Improvement* as this is tested within the airport spec, I believe you can use "subject" instead of writing 'airport = airport.new' for each test, which will make your tests more concise. (Or, I believe there is a way to specify this at the beginning of the file so you don't have to repeat it in each test)

spec/airport_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  describe "#add_plane" do 
+    it "adds plane to hangar" do
+      airport = Airport.new("Heathrow")
+      expect(airport.add_plane("plane 1")).to eq ["plane 1"]
+
+    end
+  end
+
+  describe "#status" do
```
AishaDorsett, line 36: I like that you used a describe block for each new method. - very clear

spec/weather_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "Weather"
+RSpec.describe Weather do
+  describe "#weather_conditions" do
+    it "checks that weather_conditions works" do
+      w = Weather.new
```
AishaDorsett, line 5: can use "subject" here as well instead of setting a new weather instance to a variable. it will Dry your code out. "subject" is basically equal to writing "w = Weather.new" :)

spec/weather_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+    it "chooses the stormy weather" do
+      w = Weather.new
+      allow(w).to receive(:weather_conditions) { "stormy" }
+      expect(w.weather_conditions).to eq "stormy"
+    end
+    it "chooses the sunny weather" do
+      w = Weather.new
+      allow(w).to receive(:weather_conditions) { "sunny" }
+      expect(w.weather_conditions).to eq "sunny"
```
AishaDorsett, line 16: I like how you only used 2 weather conditions so It was easier to make tests for when it is stormy - instead of writing multiple different weather conditions in an array(would have made my life easier if I did this too!). This however, could have also be a chance to explore how you would test the randomness of an array with multiple weather conditions inside it, to ensure the test passes each time.

README-Challenge.md
```diff
@@ -0,0 +1,35 @@
+Description of the challenge:
+We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.
+
+User stories:
```
AishaDorsett: love that you included the user stories in your README - it tells me what the goal was so I understand what you're doing in this project. Also like that you specified what steps you took to set up the project. You could maybe add a brief paragraph summing up the user stories and add in how to run the tests/programme.

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+  def add_plane(plane)
+    @hangar.push(plane)
+  end
+
+  def status
+    @hangar
+  end
+
+  def remove_plane(plane)
+    @hangar - [plane]
```
AishaDorsett, line 20: if this method removes the specific (plane) from the hanger list rather than popping the last one off the end (when it has multiple planes inside it) then I love how concise this code is!

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def space?
+    return @hangar if @hangar.length.zero?
+  end
+
+  def add_plane(plane)
+    @hangar.push(plane)
+  end
+
+  def status
```
AishaDorsett, line 15: maybe say what this is a status for to make it more clear e.g naming it "hanger_status"

lib/plane.rb
```diff
@@ -0,0 +1,20 @@
+require_relative "Airport"
+class Plane
+  attr_reader :space, :airport, :add_plane, :hangar
```
AishaDorsett, line 3: love the use of attribute reader

lib/plane.rb
```diff
@@ -0,0 +1,20 @@
...
+  def initialize(plane)
+    @plane = plane
+    @airport = Airport.new("Heathrow")
```
AishaDorsett, line 6: like that you've give the airport a name argument - therefore you can add other airports with other names whilst using the same code

lib/plane.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def land(plane)
+    @airport.space? ? @airport.add_plane(plane) : (raise "No space!")
```
AishaDorsett, line 10: really like your use of these 1 line if statements instead of the longer format - I need to learn how to write these!

spec/plane_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "plane"
+
+describe Plane do
```
AishaDorsett, line 3: I like that you included all of these methods in the plane file instead of adding them in the airport file. it makes it clear to see what the plane does and separates each object so well!

PR #2268, opened 2021-09-26
Airport Challenge Khalid 

kal31:


Airport_README.md
```diff
@@ -0,0 +1,21 @@
...
+
+| Objects | Messages |
+|---------|----------|
+| plane | | land at an airport 
+| plane | take off from aiport  | confirm, no longer in the airport 
+| plane | | prevent takeoff when weather is stormy |prevent landing when weather is stormy 
+
+| Airport | prevent landing when airport is full|
+| Airport | would like a default aiport capacity that can be over written at any time|
+
```
BethLambertMatthews, line 14: I like this objects / messages table and find it straightforward to read through.

lib/airport_class.rb
```diff
@@ -0,0 +1,35 @@
+class Airport 
+  def initialize(capacity)
+    @capacity = capacity
+    @stored_plane = []
+  end       
+  def add_capacity
+    @capacity += 1
+  end
```
BethLambertMatthews, line 8: Does line 7 mean "add 1 to the capacity"? 
What would the user need to do if they wanted to add 50? 
Could the "add" value maybe be passed in as a parameter, like your initialize method?

lib/airport_class.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize(capacity)
+    @capacity = capacity
+    @stored_plane = []
+  end       
+  def add_capacity
+    @capacity += 1
+  end   
+  def land(plane)
+    @plane = plane
+    weather
```
BethLambertMatthews, line 11: I haven't seen this Ruby syntax before - does it mean that you are declaring weather, just by writing it?

lib/airport_class.rb
```diff
@@ -0,0 +1,35 @@
...
+    if @weather == 1 
+      return "Weather is stormy: do not land"
```
BethLambertMatthews, line 13: I really like that you have considered the weather aspect (I didn't get to this). And I like that the user would see a message with an explanation of WHY they can't land.

lib/airport_class.rb
```diff
@@ -0,0 +1,35 @@
...
+    elsif @weather == 2 || @weather == 3
+      if @stored_plane.length < @capacity
+        return @stored_plane.push(@plane)
+      else
+        return "Cannot land Airport is full"
```
BethLambertMatthews, line 18: Helpful message for the user, like this feature.

lib/airport_class.rb
```diff
@@ -0,0 +1,35 @@
...
+        return "Cannot land Airport is full" 
+      end   
+    end         
+  end          
+  def take_off(plane)
+    weather 
+    if @weather == 1 
+      return "Weather is stormy: do not take off"
+    elsif @weather == 2 || @weather == 3
+      @stored_plane.include?(plane)
```
BethLambertMatthews, line 27: Does this check inside @stored_plane for the plane?
What happens if the plane was not found?

lib/plane_class.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
```
BethLambertMatthews, line 1: From the user stories in your ReadME - would some of the methods (like land, take off) also fit inside the Plane class? I found it really difficult to decide where to put them!

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    
+    expect(airport.land(plane)).to eq [plane]
+  end   
+  it 'should instruct plane to take off
+  and confirm if not in airport' do
+    airport = Airport.new(2)
+    plane = Plane.new
+    allow(airport).to receive(:rand).and_return(2)
+    airport.land(plane)
+    expect(airport.take_off(plane)).to eq "Plane has taken off"
```
BethLambertMatthews, line 18: I like this approach to the test - that you can look for the message to see if the method has worked.

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  it 'prevent landing when the airport is full' do
+    airport = Airport.new(2)
+    plane = Plane.new
+    plane2 = Plane.new
+    plane3 = Plane.new
+    allow(airport).to receive(:rand).and_return(2)
+    first_plane = airport.land(plane)
+    second_plane = airport.land(plane2)
+    third_plane = airport.land(plane3)
+    expect(third_plane).to eq "Cannot land Airport is full"
```
BethLambertMatthews, line 29: Again, I really like using a message as the return for a test.

spec/airport_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    allow(airport).to receive(:rand).and_return(1)
+
+    expect(airport.land(plane)).to eq("Weather is stormy: do not land")
+  end   
+
+  it 'land when weather is not stormy' do
+    airport = Airport.new(2)
+    plane = Plane.new
+    
+    allow(airport).to receive(:rand).and_return(2)
```
BethLambertMatthews, line 44: Reading this line - so does allow mean that you can specify what return the airport gives, regardless of which number was generated by rand? This looks really helpful.

BethLambertMatthews (commented):
Hi Khalid, I really enjoyed reading your code. 

I wonder how we decide which methods go with the Airport class, and which would fit better in the Plane class?

PR #2269, opened 2021-09-26
Airport Challenge

pinteck:


DannyL90:
Hi,
I like how you worked on your project. The commits are drove me true your developing process step by step. Your code is clean and easy to read, well organised.
I have only one suggestion about your code. I just recommend adding one extra function to your take_off method to prevent that situation when that plane would like to take off who not at the airport.

PR #2270, opened 2021-09-26
End of week 1 challenge, airport challenge

blue-luke:


README2.txt
```diff
@@ -0,0 +1,16 @@
+I copied the user stories into a text document then made a note of the nouns and verbs, therefore the objects and methods,
```
starf0g, line 1: Nice and thorough readme

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+  def land(plane)
+    (hangar.push(plane) && plane.plane_has_landed) if (landing_permitted && !full? && plane.landed == false)
+  end
+
+  def check_plane_in_hangar(plane)
+    @hangar.include?(plane)
+  end
+
+  def weather
+    ["stormy", "fine"].sample
```
starf0g, line 26: This is a neat way to randomise the weather apart from the 50/50 chance. You could have added multiple "fine"s to the array but that would be a bit messy. Another option would be to use rand.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def land(plane)
+    (hangar.push(plane) && plane.plane_has_landed) if (landing_permitted && !full? && plane.landed == false)
+  end
+
+  def check_plane_in_hangar(plane)
+    @hangar.include?(plane)
+  end
+
+  def weather
```
starf0g, line 25: I think for SRP this should be its own class

supporting_docs/user_stories.txt
```diff
@@ -0,0 +1,56 @@
+plane - instruct to land
```
starf0g, line 1: I like how you've broken the user stories down into their objects/messages

spec/plane_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'plane'
+
+plane = Plane.new
+
+describe Plane do
+
+  let(:airport) { double :airport }
```
starf0g, line 7: I don't think this is necessary as its not used in any of the tests

spec/plane_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'plane'
+
+plane = Plane.new
```
starf0g, line 3: As you mentioned in the readme you could use subject here :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,109 @@
+require 'airport'
+
+checker = Airport.new
```
starf0g, line 3: I feel like it would have been clearer in the tests to use airport as the variable rather than checker

spec/airport_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    expect(checker.hangar.length).to eq 1
+  end
+
+  it "lets a plane land" do
+    checker = Airport.new
+    allow(checker).to receive(:landing_permitted).and_return(true)
+    allow(plane).to receive(:landed).and_return(false)
+    allow(plane).to receive(:plane_has_landed)
+    checker.land(plane)
+    expect(checker.hangar.empty?).to eq false
```
starf0g, line 43: Another option would be to check if the plane instance exists in the hangar array

spec/airport_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  it "prevents landing when stormy" do
+    checker = Airport.new
+    allow(plane).to receive(:plane_has_landed)
+    allow(checker).to receive(:landing_permitted).and_return(false)
+    expect(checker.land(plane)).to eq nil
```
starf0g, line 50: I think it would be clearer if the land method raised an error here rather than equalling nil

starf0g (commented):
Overall this was really well laid out and clear to read.

leoht (commented):
Great work! I've +1 most of Alex's comments already, and added a few more ones on how to refactor the code and do some minor improvements, also extracting the weather logic in its own class :) Let me know if you've got any question on that!

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    @hangar = []
+  end
+
+  def take_off(plane)
+    raise StandardError.new "That plane is not in the hangar" unless @hangar.include?(plane)
+    (hangar.delete(plane) && plane.plane_is_flying) if landing_permitted
+  end
+
+  def land(plane)
+    (hangar.push(plane) && plane.plane_has_landed) if (landing_permitted && !full? && plane.landed == false)
```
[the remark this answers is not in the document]
blue-luke: Thanks for the comments on guard clauses, I've made this change. I'm going to think more about the best place and way to store the 'plane has landed' attribute, as all of the avenues seem sub-optimal at the moment

lib/airport.rb
blue-luke: I will extract this class and re-do the tests as my next step

PR #2273, opened 2021-09-26
Airport Challege - Jacob

jtc27:


README.md
```diff
@@ -1,89 +1,65 @@
...
-
-As an air traffic controller 
-To ensure safety 
-I want to prevent takeoff when weather is stormy 
-
-As an air traffic controller 
-To ensure safety 
-I want to prevent landing when weather is stormy 
+require 'airport'
+
```
bernardleanse, line 16: nice! like how that gets you thinking about how to structure tests straight from the stories. might try this myself text time

spec/weather_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'weather'
+
+describe Weather do
+
+    it 'If weather is bad' do
+        weather = Weather.new
```
bernardleanse, line 6: could use subject?

PR #2277, opened 2021-09-27
Airport Challenge Alex Nakov

alexnakov:


alexnakov:
Complete everything except overriding weather randomness in RSpec

spec/airport_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    it "raises and error if the plane about to be landed is has already landed" do
+      airport = Airport.new
+      plane = Plane.new
+      airport.land(plane)
+      expect { airport.land(plane) }.to raise_error "You can't land a plane that's already at the airport"
+    end
+  end
+
+  describe ".take_off" do
+    it "raise an error when a plane instructed to take off is not at the airport" do
```
tommyoswinwilliams, line 41: This is a good check!

spec/airport_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      capacity.times do
+        airport.land(Plane.new)
+      end
+      expect { airport.land(Plane.new) }.to raise_error "Cannot land, airport is full"
+    end
+  end
+
+  describe ".land" do
+    it "lands the plane at an empty space in the airport" do 
+      plane = Plane.new()
```
tommyoswinwilliams, line 17: You could use [doubles](https://github.com/makersacademy/course/blob/main/pills/doubles.md) to separate the testing of your Airport and Plane code.

lib/plane.rb
```diff
@@ -0,0 +1,13 @@
+class Plane
```
tommyoswinwilliams, line 1: I like how your plane does something (mine is an empty class, lol).

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
+require_relative './plane'
+
+class Airport
+  @@weather_possibilities = ['sunny', 'sunny', 'sunny', 'stormy']
```
tommyoswinwilliams, line 4: I separated our Weather as a separate class - not sure if that's a good or bad way to do it but worth considering.

PR #2278, opened 2021-09-27
Partial airport_challenge solution

TTurvey:
This is a partial solution. I got carried away with trying to implement unique planes and airports into an array of hashes. This was not technically part of the challenge but I thought it would be good to learn how to do.


lib/docs/airport.rb
```diff
@@ -0,0 +1,83 @@
+require_relative "plane"
+
+class Airport
+  # Default capacity is made a constant and set to 5.
+  DEFAULT_CAPACITY = 5
```
maxcfry, line 5: I'm think DEFAULT_CAPACITY needs to be moved back one space (lined up with class Airport)

lib/docs/airport.rb
```diff
@@ -0,0 +1,83 @@
...
+  # Created instances of the Airport class.
+  # These start empty with the default capacity of 5 planes.
+  def initialize(capacity = DEFAULT_CAPACITY)
+    # Initially there are no airports as no instances of the class have been created.
+    @@airport_log = airport_log
+    @@existing_airport_count = existing_airport_count
+    @@airport_capacity = capacity
+  end 
+
+  # Creates an airports with increasing numerical IDs
```
maxcfry, line 18: I'm not too sure what the purpose of this method is - and it probably makes testing your code very difficult! 
If you're creating a hash to store the planes? I think this is an overcomplicated way of doing it (I'd just use an empty array which you can add planes to in the spec tests). Additionally, it seems like the method is doing several things at once (incrementing existing_airport_count, adding an airport with an id, capacity, status and empty string in a hash, and adding each hash to an array!) I think this resembles something we did in the Villains Academy precourse exercise - which I didn't understand fully myself! instead - you want to stick to the Principle of Single Responsibility (Each method should have one responsibility - this helps to simplify testing and makes the code more readable (and easier to write and keep track of variables being passed to each methods if any).

lib/docs/airport.rb
```diff
@@ -0,0 +1,83 @@
...
+  # Prints the number of airports that have been created/exist.
+  def self.count
+    puts "There are #{@@existing_airport_count} airports existing."
+  end
+
+  # Lands a specified plane in a specified airport.
+  def self.land(plane, airport)
+    plane_id = plane
+    airport_id = airport
+    if @@airport_log.include?(plane_id)
```
maxcfry, line 42: I'm not sure if this is another factor impeding your ability to test - but you're using class variables (@@) - and while I'm sure there's probably a way to use them - I used instance variables (@). I honestly could not explain the differences between them but single @'s seemed to work for me!

lib/docs/airport.rb
```diff
@@ -0,0 +1,83 @@
...
+      }
+    end
+  end
+
+  # Prints the number of airports that have been created/exist.
+  def self.count
+    puts "There are #{@@existing_airport_count} airports existing."
+  end
+
+  # Lands a specified plane in a specified airport.
```
maxcfry, line 38: I think you'd only want the plane landing (not the airport!) - perhaps would be easier to just take one argument for this method(plane) and do something such as - store it in an array(called planes) - and this will also save a lot of time and effort because if you want to know how many planes are in the airport - you can put planes.length (and possibly set it to a capacity variable if necessary.

lib/docs/airport.rb
```diff
@@ -0,0 +1,83 @@
...
+    elsif full?(airport_id) == true
+      puts "Cannot land plane: Capacity is full"
+    elsif stormy? == true
+      "Cannot land plane: weather is stormy"  
+    else
+      @@airport_log[:airport_id, :landed_planes] << plane_id
+    end
+  end
+  
+  def self.takeoff(plane, airport)
```
maxcfry, line 53: Again, you're in the airport class - so I would avoid calling an airport in the method (that would be like creating Heathrow within Gatwick!)

lib/docs/airport.rb
```diff
@@ -0,0 +1,83 @@
...
+  #  fail 'Cannot takeoff plane: airport is empty' if empty?
+  #  fail "Cannot takeoff plane: weather is stormy" if stormy?
```
maxcfry, line 55: Instead - call the instances of airports in the spec method. 
Then think airport = Airport.new 
plane = Plane.new 
airport.takeoff(plane).to eq planes(the array).empty (if there was only one plane in there for example)

lib/docs/plane.rb
```diff
@@ -0,0 +1,35 @@
+class Plane
```
maxcfry, line 1: I found it much easier to keep the Plane class empty - and keep all the methods in the Airport. After all, the planes don't really need to do anything but take off and land (provided the airport capacity isn't full and the weather conditions permit it)

lib/docs/plane.rb
```diff
@@ -0,0 +1,35 @@
...
+  # Plane instances are created and logged.
+  def initialize
+    @@plane_log << "Plane#{@@existing_plane_count + 1}"
+    puts "New plane created. ID: Plane#{@@existing_plane_count + 1}"
+  end 
+
+  # Creates planes with increasing numerical IDs
+  def self.create(number)
+    number.times do
+      @@existing_plane_count += 1
```
maxcfry, line 16: Again, I'm not completely sure of the difference between class variables (@@) and instance variables (@) - but I think class variables mean you can only access them within a given class (whereas perhaps instance is what you want because they are instances invoked every time you create a new instance of a class.

lib/spec/airport_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'airport'
```
maxcfry, line 1: You don't need to require the airport (that's supplied by the class) 
Instead you require to bring other classes (kept in other files) into use when you call the Plane class.
For example, require "plane" and/or require "weather" (to bring the info in these classes into your plane class)

lib/spec/airport_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
```
maxcfry, line 2: Another trick is to write require_relative (which just makes your life easier - mean you don't have to specify the path of the plane or weather files).
So, require_relative "plane" 
As opposed to require "plane"

lib/spec/airport_spec.rb
maxcfry: Apologies - ignore this - didn't realise this was a spec file haha

lib/spec/airport_spec.rb
maxcfry: Ditto (though the require_relative is a useful trick nonetheless)

PR #2280, opened 2021-09-27
My Week 3 Individual Challenge: Airport Challenge submission 

feedXia:


lib/airport.rb
```diff
@@ -21,8 +21,8 @@ def takeoff(plane, weather = plane.weather)
     @planes.delete(plane)
   end
 
-  def at_airport?(plane)
+  def at_airport?(plane, airport)
```
Joseph112358, line 24: I'm not entirely sure, but I believe if you are checking a given airport e.g London Heathrow, you can use "self" instead of passing in the airport you are working on.

lib/plane.rb
```diff
@@ -1,3 +1,22 @@
...
+    @flying << plane
+  end
+
+  def flying?(plane) 
+    @flying.include? plane
+  end
+
+  def edge_flying(plane, airport)
+    fail 'Cannot takeoff plane, plane already in flight!' if plane.flying?(plane)
+    fail 'Plane not at airport. Plane already in flight!' unless airport.at_airport?(plane, airport)
```
Joseph112358, line 20: If the plane is at another airport (not in flight), is this error message  incorrect / misleading?

lib/weather.rb
```diff
@@ -1,7 +1,7 @@
 class Weather
   def stormy?
     # Sunny on average 80% of the time
-    n  = rand(9)
-    n < 1 ? true : false
+    n = rand(9)
+    n < 1
```
Joseph112358, line 4: Will n always be less than 1? is this just to pass a test case?

spec/airport_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+describe Airport do
+
+  let(:plane) { Plane.new }
+  let(:weather) { double(:Weather, :stormy? => false) }
+
+  describe '.land' do
+    context 'given a plane' do
+      it 'lands the plane at an airport' do
+        expect(subject.land(plane, weather)).to contain_exactly plane
```
EdwardAndress, line 12: nice use of a double!

spec/airport_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+    end
+
+    context 'when landing a plane' do
+      it 'returns error if airport is full' do
+        10.times { subject.land(plane, weather) } 
+        expect { subject.land(plane, weather) }.to raise_error 'Airport is full.'
```
EdwardAndress, line 19: Do you know why the curly brackets are needed on line 19?

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+    fail 'Safety: unable to land due to storm.' if weather.stormy?
+    @planes << plane
+  end
+
+  def takeoff(plane, weather = plane.weather)
+    fail 'Safety: unable to takeoff due to storm.' if weather.stormy?
+    @planes.delete(plane)
+  end
+
+  def at_airport?(plane, _airport)
```
EdwardAndress, line 24: Looks like you've been doing some python? :). Maybe?

You don't need this second argument (`_airport`)

lib/plane.rb
```diff
@@ -0,0 +1,31 @@
+require 'plane'
+require 'airport'
+require 'weather'
+
```
EdwardAndress, line 4: I like that your plane "knows" if it's flying or not and, also, that you have a method for checking if it's flying.

What do you think about this alternative implementation (I'm only recreating the relevant part)?

```ruby
class Plane
  def initialize
    @flying = false
  end

  def flying?
    @flying?
  end
end
```

EdwardAndress: Which version do you prefer and why?

lib/plane.rb
```diff
@@ -0,0 +1,31 @@
...
+class Plane
+  def initialize
+    @flying = []
+  end
+
+  def planes_flying(plane)
```
EdwardAndress, line 10: What happens if you change this method to ...

```ruby
def planes_flying
  @flying << self
end
```

What is `self`?

EdwardAndress:
Hello!  This is a nice solution for the Airport Challenge!  Well done.  I've left some comments in line against the code.  There are some questions for you as well.  If you want me to see you answers, please send them in Slack or send me a message to come back and take a look as I won't be notified by the answers on GitHub :)

PR #2281, opened 2021-09-27
Submission for Airport Challenge

Anmolbhati30:


docs/domainmodelling.md
```diff
@@ -0,0 +1,41 @@
+# Domain models for user stories
```
pzrsa, line 1: Here you did a good job separating the different user stories and creating separate domain models for them.

lib/airport.rb
```diff
@@ -0,0 +1,58 @@
...
+  include Weather
+  
+  DEFAULT_CAPACITY = 10
+  
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    raise ArgumentError.new("Please enter a positive integer") unless valid_capacity?(capacity)
+    
+    @planes = []
```
pzrsa, line 12: Could possibly change the naming of where you store the planes to something real world "Hangar" or "Cellar"

lib/airport.rb
```diff
@@ -0,0 +1,58 @@
+require_relative 'weather'
```
pzrsa, line 1: Importing the weather class might go against the DRY principle. Could be best to import it in IRB

lib/plane.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
+  attr_reader :id
+  
+  def initialize(id)
+    @id = id
```
pzrsa, line 5: Good that you assigned each plane an id, definitely helps with removing a specific plane when it takes off

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+module Weather
+  def check_weather
+    generate_random_number > 7 ? :stormy : :sunny
+  end
+
+  def generate_random_number
+    rand(1..10)
```
pzrsa, line 7: I think it would be much simpler to store the types of weather in an array, and use the ".sample" method to pick one randomly without having to do it manually

pzrsa:
Overall, very impressed with everything you managed to complete. You went beyond what it asked and assigning the ID was definitely a good feature to implement! Keep it up :)

PR #2285, opened 2021-09-27
Airport Challenge

troysnowden:


lib/plane.rb
```diff
@@ -0,0 +1,53 @@
...
+    clear_to_takeoff?
+    # if no errors raised, plane takes off
+    @at_airport.remove_plane(self)
+    @at_airport = nil # sets it back to nil signalling the plane is not at an airport
+    "Plane has taken off"
+  end
+
+  # if variable not assigned then the plane is not at an airport
+  def at_airport?
+    !@at_airport.nil?
```
thomas99baxter, line 29: I like how compact this is - but not sure i wouldve understood it without the comment - potential refactor?

lib/plane.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  private
+
+  # method raises error if the plane is not clear to land at the given airport
+  def clear_to_land?(airport)
+    raise_error("This plane is already in an airport, " +
```
thomas99baxter, line 36: why are you concatenating a string here?

lib/plane.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  # method raises error if the plane is not clear to takeoff
+  def clear_to_takeoff?
+    raise_error("This plane is not at an airport, " + 
+      "land it at an airport before instructing it to takeoff") unless at_airport?
+    raise_error("The weather is stormy, this plane cannot takeoff") if @weather.forecast_stormy?
+  end
+  
+  def raise_error(error)
```
thomas99baxter, line 49: could you leave this out and just `raise` an error rather than using this function? seems redundant but happy to discuss

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
```
thomas99baxter, line 4: love how clean these tests are :D

spec/weather_gen_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'weather_gen'
+
+describe WeatherGenerator do
+  subject(:weather) { described_class.new }
+
+  it "it should have a boolean stormy property" do
+    expect([TrueClass, FalseClass].include?(weather.forecast_stormy?.class)).to eq(true)
```
thomas99baxter, line 7: very familiar?

lib/plane.rb
troysnowden: Yep completely agree, was probably just adding complexity for the sake of it.

lib/plane.rb
troysnowden: To satisfy Rubocop the cheeky minx. The line was too long cause I made the world's longest error msg

lib/plane.rb
troysnowden: Yeah fair point, I just liked the idea of checking whether the plane was at an airport and which airport it was at in the same variable. Maybe changing the method name to 'in_air?' and taking out the '!' before the nil check?

lib/plane.rb
```diff
@@ -0,0 +1,53 @@
...
+  def instruct_plane_to_land(airport)
+    clear_to_land?(airport)
+    # if no errors raised, land plane
+    airport.add_plane(self)
+    @at_airport = airport # overrides the variable with the airport instance object
+    "Plane landed at airport"
+  end
+
+  def instruct_plane_to_takeoff
+    clear_to_takeoff?
```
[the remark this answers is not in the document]
troysnowden: Yes possibly, I just didn't want a redundant if/else because the method call raises an error and none of the lines below run if an error is raised. Definitely don't disagree though in terms of readability

lib/plane.rb
thomas99baxter: yeah that makes more sense, happy with that one

lib/plane.rb
thomas99baxter: @troysnowden  ee the little tinker, i'll let you off.

lib/plane.rb
thomas99baxter: nice one, resolved

lib/plane.rb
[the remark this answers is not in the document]
thomas99baxter: nice one. think you could move 22 , 23, 24 into a seperate 'takeoff' method so youre keeping things nice and granular?

PR #2287, opened 2021-09-27
Airport Challenge

amin-newcastle:
All user stories were completed and 11 successful tests passed


amin-newcastle:
I used the following video to look at stubs

https://www.youtube.com/watch?v=oyMPzA-ZWkE

docs/planning.txt
```diff
@@ -0,0 +1,40 @@
+Story 1
+As an air traffic controller 
+So I can get passengers to a destination 
+I want to instruct a plane to land at an airport
+
+--Objects: Plane, airport
```
oceansize, line 6: The existence of this document is great. I would expect 'Objects' to contain all the nouns from the user story

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
+require_relative 'plane'
+
+class Airport
+  def initialize
+    @my_airport = []
+    @max_cap = 3
```
oceansize, line 6: What is `max_cap`

oceansize: Also, what is 3? In our case here, 3 is what's known as a 'magic number'

Better would be something like this:
```ruby
class Airport
    DEFAULT_CAPACITY = 3
    
    def initialize
       @max_capacity = DEFAULT_CAPACITY
       ...
```

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def check_ap_capacity
```
oceansize, line 9: Abbreviations!

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+    @max_cap
+    current_cap = @my_airport.length
+    if current_cap + 1 <= @max_cap
+      return true
```
oceansize, line 13: Here we have a method that returns a boolean or nil.

It's worth noting that when we have methods that return booleans, we need to think about communicating that kind of return value in the name of the method:
```ruby
airport_has_space? # indicates that there will be a boolean return value
```

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+  end
+
+  def land_plane
+    # Should add plain to my_airport[]
```
oceansize, line 18: Here is an example of a comment that should not exist. Your code should be clear enough to explain what is intended (and I think you have achieved that with the name `land_plane`

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+    if check_ap_capacity == true
```
oceansize, line 19: Theoretically you could just have this line read `if check_ap_capacity` because if statements work on boolean values

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+      if check_weather == "Sunny"
+        @plane = Plane.new
```
oceansize, line 21: Here we have an airport that **creates a plane** when checking the weather.

Unless we have a method in our airport that says `build_plane`, I would avoid this. (Actually I would avoid building planes inside an airport at all costs).

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+        @my_airport.push(@plane)
+        puts "The plane has landed" + @my_airport.to_s
+        return @my_airport            
+      else
+        puts "The weather is STORMY!"
+      end
+    else
+		puts "Your airport is full, wait until another plane has taken off"
```
oceansize, line 29: Minor: formatting

lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
+  def initialize
+    plane_num = "Boeing 737 Max"
```
oceansize, line 3: Minor: abbreviation and potentially suggesting "Boeing 737 Max" is a number, instead of say make_and_model

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'airport'
+
+describe Airport do
+  it { is_expected.to respond_to :land_plane }
```
oceansize, line 4: This is nice

spec/plane_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'plane'
+
+describe Plane do
+
```
oceansize, line 4: I would expect that the test for plane exists before any code for a plane — therefore I deduce there was no TDD here

PR #2288, opened 2021-09-27
add Contribution - Alex Annani

Alex969:
Added my 4 first features to the airport challenge, I didn't get the time to fully complete it but added a section in my README.md to explain I what would do to implement a weather class


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'plane'
+
+class Airport
+  attr_reader :landed, :capacity  
+  DEFAULT_CAPACITY = 10 
+
+  def initialize (capacity=DEFAULT_CAPACITY)
```
arthurfincham, line 7: I thought of doing this but I ended up defining capacity through an argument with a default value.

``` ruby
def initialize(capacity = 1)
...
end
```

Then when you create an airport you can change it like so:
``` ruby
heathrow = Airport.new(120)
```
So Heathrow has a capacity of 120.

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+    landed << plane
+  end
+
+  def take_off
+    raise "No planes ready to take off" if empty?
+    landed.each_with_index {|plane,index| landed.delete_at(index) ; return plane}
+  end
+
+  def full?
+    @landed.length >= capacity
```
arthurfincham, line 23: Small thing but I think it can just be `=` instead of `>=`. The error will be thrown before the length can outsize capacity!

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+  attr_reader :landed, :capacity  
+  DEFAULT_CAPACITY = 10 
+
+  def initialize (capacity=DEFAULT_CAPACITY)
+    @landed = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    raise 'Not accepting more planes to land' if full?
```
arthurfincham, line 13: Nice single line guard clause 💪

lib/plane.rb
```diff
@@ -0,0 +1,8 @@
+class Plane
+  
+private
+
+  def planes
```
arthurfincham, line 5: Nice and simple. Only thing I would add is a way of tracking the plane's **location**...
``` ruby
def initialize
@location = 'Sky'
end
```
... for example.

lib/airport.rb
arthurfincham: Sorry ignore this - your code does exactly this anyway!

spec/airport_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'airport'
+require 'plane'
+
+describe Airport do 
+  it { is_expected.to respond_to(:land).with(1).argument }
```
arthurfincham, line 5: Just below `describe Airport do` I would create a double for plane:
``` ruby
let(:plane) { Plane.new }
```
This also means you don't have to repeat creating an instance with `plane = Plane.new` in each test!

README.md
```diff
@@ -13,77 +13,4 @@ Airport Challenge
...
-* Write an RSpec **feature** test that lands and takes off a number of planes
-
-Note that is a practice 'tech test' of the kinds that employers use to screen developer applicants.  More detailed submission requirements/guidelines are in [CONTRIBUTING.md](CONTRIBUTING.md)
-
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first.
-
-* **Submit a pull request early.**
-
-* Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+The only feature I haven't had the time to implement yet is the feature where planes aren't landing/taking off if the weather is stormy. The way I would approach doing that, I believe, would be to create a seperate weather.rb file with a weather classe that would take an instance variable @weather = true or false in reference to stormy. I would use an attribute reader to access that value outside of the weather class/file and would call it on airports. If weather = true then raise an error if a Plane instance tries to take off or land
```
arthurfincham, line 16: That's pretty much what I did for the Weather class! 👍

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'plane'
+
+class Airport
+  attr_reader :landed, :capacity
```
arthurfincham, line 4: Nice 👍

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def land(plane)
+    raise 'Not accepting more planes to land' if full?
+    landed << plane
+  end
+
+  def take_off
+    raise "No planes ready to take off" if empty?
+    landed.each_with_index {|plane,index| landed.delete_at(index) ; return plane}
```
arthurfincham, line 19: This looks cool!

lib/airport.rb
Alex969: Truuue ! I'm always confused with those operators tbh

lib/plane.rb
Alex969: Yes and also I was supposed to use that instance variable to access @planes in the Airport class and make it less crowded..

lib/airport.rb
Alex969: it's a bit unnecessary lol but thanks !

PR #2290, opened 2021-09-27
Submit airport challenge

pzrsa:
This PR is for submitting my partially completed airport challenge. I have also created a solution README which goes into more detail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,14 @@
-Airport Challenge
-=================
+# My solution
 
-```
+I wasn't able to fully finish it, but I'm aware of what I had to in order to complete the challenge. I missed out on adding a capacity and making sure that a plane doesn't land when its full, or taking off when there is no plane left.
+
+1. I made sure that the plane is able to land safely. This is done by having the plane stored inside of an array (hangar), but only when the weather is sunny. If the weather is Stormy, then I won't allow any landing, and no plane will be add to the array.
+2. I made sure that the plane can take off. This is done by simply having the plane taken out the array that was stored previously, however once again I cannot allow taking off if the weather is stormy. Also, in order to test weather I can take off, I allowed landing by making sure the weather is sunny, but when taking off it turns stormy.
+3. I made sure that the weather is completely randomised. I have "Sunny" and "Stormy" stored in a constant array, so when I need a random weather that is easily possible.
+
+## Airport Challenge
+
+``````
         ______
         _\____\___
 =  = ==(____MA____)
@@ -11,53 +18,50 @@ Airport Challenge
                 `---~~\___________/------------`````
                 =  ===(_________)
 
-```
+``````
 
-Instructions
----------
+## Instructions
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+- Feel free to use google, your notes, books, etc. but work on your own
+- If you refer to the solution of another coach or student, please put a link to that in your README
+- If you have a partial solution, **still check in a partial solution**
+- You must submit a pull request to this repo with your code by 9am Monday morning
 
-Steps
--------
+## Steps
 
 1. Fork this repo, and clone to your local machine
 2. Run the command `gem install bundler` (if you don't have bundler already)
 3. When the installation completes, run `bundle`
 4. Complete the following task:
 
-Task
------
+## Task
 
-We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
+We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off. Here are the user stories that we worked out in collaboration with the client:
 
 ```
-As an air traffic controller 
-So I can get passengers to a destination 
+As an air traffic controller
+So I can get passengers to a destination
 I want to instruct a plane to land at an airport
 
-As an air traffic controller 
-So I can get passengers on the way to their destination 
+As an air traffic controller
+So I can get passengers on the way to their destination
 I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent landing when the airport is full 
+As an air traffic controller
+To ensure safety
+I want to prevent landing when the airport is full
 
 As the system designer
 So that the software can be used for many different airports
 I would like a default airport capacity that can be overridden as appropriate
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent takeoff when weather is stormy 
+As an air traffic controller
+To ensure safety
+I want to prevent takeoff when weather is stormy
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent landing when weather is stormy 
+As an air traffic controller
+To ensure safety
+I want to prevent landing when weather is stormy
 ```
 
 Your task is to test drive the creation of a set of classes/modules to satisfy all the above user stories. You will need to use a random number generator to set the weather (it is normally sunny but on rare occasions it may be stormy). In your tests, you'll need to use a stub to override random weather to ensure consistent test behaviour.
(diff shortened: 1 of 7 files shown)


user_stories.md
```diff
@@ -0,0 +1,33 @@
...
+
+As an air traffic controller
+To ensure safety
+I want to prevent takeoff when weather is stormy
+
+As an air traffic controller
+To ensure safety
+I want to prevent landing when weather is stormy
+
+| Objects         | Messages                                |
```
Anmolbhati30: Good domain modelling, listed the important objects and methods

lib/airport.rb
```diff
@@ -1,12 +1,14 @@
 require "plane"
 
 class Airport
-  def hangar
+  attr_reader :hangar
```
Anmolbhati30, line 4: Hangar is publicly readable, might be good to consider privatising later on.

lib/airport.rb
```diff
@@ -12,5 +12,6 @@ def land(plane)
   end
 
   def takeoff
+    @hangar.pop
```
Anmolbhati30, line 17: Currently taking off last plane always, room to refactor to take off specific plane

lib/weather.rb
```diff
@@ -1,2 +1,5 @@
 class Weather
+  attr_reader :CONDITIONS
+
+  CONDITIONS = ["Sunny", "Stormy"].freeze
```
Anmolbhati30, line 4: freeze only freezing the array, variables inside can still be mutated. 'CONDITIONS' is already a constant so should be safe from mutation anyways?

lib/weather.rb
```diff
@@ -1,2 +1,5 @@
 class Weather
+  attr_reader :CONDITIONS
```
Anmolbhati30: CONDITIONS is publicly readable, could be privatised later on

spec/weather_spec.rb
```diff
@@ -2,4 +2,5 @@
 
 describe Weather do
   it { is_expected.to respond_to(:CONDITIONS) }
+  it { is_expected.to respond_to(:get_condition) }
```
Anmolbhati30: Might become redundant later on through implicitly tested alongside other tests, consider removing then

lib/weather.rb
```diff
@@ -2,4 +2,8 @@ class Weather
   attr_reader :CONDITIONS
 
   CONDITIONS = ["Sunny", "Stormy"].freeze
```
Anmolbhati30, line 4: "Sunny" and "Stormy" could be stored as symbols instead of strings

spec/weather_spec.rb
```diff
@@ -2,5 +2,12 @@
 
 describe Weather do
   it { is_expected.to respond_to(:CONDITIONS) }
-  it { is_expected.to respond_to(:get_condition) }
```
Anmolbhati30: Good job removing redundant test

spec/weather_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "weather"
 
 describe Weather do
-  it { is_expected.to respond_to(:CONDITIONS) }
+  it { is_expected.to respond_to(:conditions) }
```
Anmolbhati30, line 4: Reasoning behind changing to variable? Are there going to be more weather conditions added?

spec/airport_spec.rb
```diff
@@ -9,11 +9,12 @@
...
   describe "#land" do
     it { is_expected.to respond_to(:land) }
-    it { is_expected.to respond_to(:land).with(1).argument }
+    it { is_expected.to respond_to(:land).with(2).arguments }
 
     it "lands a new plane and adds to hangar" do
       plane = Plane.new
-      subject.land(plane)
+      weather = Weather.new
+      subject.land(plane, weather)
```
Anmolbhati30: Passing a weather object every time, could be refactored to include weather in the Airport class and call that instance instead

lib/airport.rb
```diff
@@ -7,7 +7,7 @@ def initialize
     @hangar = []
   end
 
-  def land(plane)
+  def land(plane, weather = Weather.new)
```
Anmolbhati30, line 10: Right track, moving this argument inside Airport's initialize and using one instance every time could be the next refactor

spec/airport_spec.rb
```diff
@@ -12,10 +12,11 @@
     it { is_expected.to respond_to(:land) }
     it { is_expected.to respond_to(:land).with(2).arguments }
 
-    it "lands a new plane and adds to hangar" do
+    it "lands plane and adds to hangar if weather is sunny" do
       plane = Plane.new
-      weather = Weather.new
-      subject.land(plane, weather)
+      sunny_weather = Weather.new
+      allow(sunny_weather).to receive(:get_condition) { "Sunny" }
```
Anmolbhati30: Testing state rather than behaviour I think, but that might be the only way since using sample in get_condition

Anmolbhati30: Good use of allow instead of expect though

lib/airport.rb
```diff
@@ -8,6 +8,7 @@ def initialize
   end
 
   def land(plane, weather = Weather.new)
+    fail "Weather is dangerous to land in" if weather.get_condition == "Stormy"
```
Anmolbhati30: Good guard condition

spec/airport_spec.rb
```diff
@@ -29,10 +29,12 @@
   describe "#takeoff" do
     it { is_expected.to respond_to(:takeoff) }
 
-    it "removes plane from hangar after taking off" do
+    it "removes plane from hangar after taking off if weather is sunny" do
       plane = Plane.new
-      weather = Weather.new
```
Anmolbhati30: Currently this test could fail if there is a problem with weather class. Should there be spec for weather class too? Decouple tests from each other

spec/airport_spec.rb
```diff
@@ -34,9 +34,17 @@
       sunny_weather = Weather.new
       allow(sunny_weather).to receive(:get_condition) { "Sunny" }
       subject.land(plane, sunny_weather)
-      # the takeoff method is failing because when its stormy nothing is being added to the array after trying to land, hence the error
-      subject.takeoff
+      subject.takeoff(sunny_weather)
```
Anmolbhati30, line 38: Take off simply takes off the last plane added. Could be refactored to take off specific planes

lib/airport.rb
```diff
@@ -8,12 +8,12 @@ def initialize
   end
 
   def land(plane, weather = Weather.new)
-    fail "Weather is dangerous to land in" if weather.get_condition == "Stormy"
+    fail "Weather is dangerous to land in" if weather.condition == "Stormy"
```
Anmolbhati30, line 11: .condition is a bit vague, seems to refer to an attribute rather than a check. .condition? might be more readable

README.md
```diff
@@ -1,7 +1,14 @@
-Airport Challenge
-=================
+# My solution
 
-```
+I wasn't able to fully finish it, but I'm aware of what I had to in order to complete the challenge. I missed out on adding a capacity and making sure that a plane doesn't land when its full, or taking off when there is no plane left.
+
+1. I made sure that the plane is able to land safely. This is done by having the plane stored inside of an array (hangar), but only when the weather is sunny. If the weather is Stormy, then I won't allow any landing, and no plane will be add to the array.
+2. I made sure that the plane can take off. This is done by simply having the plane taken out the array that was stored previously, however once again I cannot allow taking off if the weather is stormy. Also, in order to test weather I can take off, I allowed landing by making sure the weather is sunny, but when taking off it turns stormy.
+3. I made sure that the weather is completely randomised. I have "Sunny" and "Stormy" stored in a constant array, so when I need a random weather that is easily possible.
```
Anmolbhati30, line 7: Array was changed from constant to normal variable though?

README.md
```diff
@@ -1,7 +1,14 @@
-Airport Challenge
-=================
+# My solution
 
-```
+I wasn't able to fully finish it, but I'm aware of what I had to in order to complete the challenge. I missed out on adding a capacity and making sure that a plane doesn't land when its full, or taking off when there is no plane left.
+
+1. I made sure that the plane is able to land safely. This is done by having the plane stored inside of an array (hangar), but only when the weather is sunny. If the weather is Stormy, then I won't allow any landing, and no plane will be add to the array.
```
Anmolbhati30, line 5: Room for Improvement
- Give planes an identifier and pop specific ones passed
- Store a weather object as an instance variable and call to that instead of passing a new on every time
- Handle for edge cases: hangar full, plane not in hangar, plane not given

README.md
Anmolbhati30, line 7: Room for Improvement

Give planes an identifier and pop specific ones passed
Store a weather object as an instance variable and call to that instead of passing a new on every time
Handle for edge cases: hangar full, plane not in hangar, plane not given

- Weather is randomised with 50/50 right now. Could be refactored to a different percentage, maybe a definable percentage?
- Weather class doesn't seem to have tests right now, should create spec for Weather class too.

spec/airport_spec.rb
```diff
@@ -26,6 +26,7 @@
       expect { subject.land(plane, stormy_weather) }.to raise_error "Weather is dangerous to land in"
     end
   end
+
```
Anmolbhati30, line 29: This whitespace single handedly passed every test in your code and made rubocop retire forever :D

Anmolbhati30: Seriously though good job with the challenge, these commits are awesome, I am actually ashamed I didn't commit early and often like you, would make reviewing my code so much easier, really impressed with the workflow!

pzrsa: Thanks Anmol, don't be ashamed. Over time you will find your own workflow that works best for you!

PR #2291, opened 2021-09-27
Completed-Airport-Challenge

stellamartin222:


lib/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+        @capacity = capacity
+    end
+    
+    def plane_land(plane)
+        if !plane_can_fly?
+            raise "Weather warning, cannot land plane"
+        elsif full?
+            raise "The airport is full"
+        else
+            @landed_planes.push(plane)
```
lewiscj97, line 24: You could take this out of the if/else statement using guard statements:
`raise "Weather warning..." if !plane_can_fly?` and then have line 24 on it's own?

lib/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+        end
+    end
+
+    def plane_takeoff
+        if !plane_can_fly?
+            raise "Weather warning, plane cannot take off"
+        elsif !empty? 
+            rand_index = rand(0..@landed_planes.length)
```
lewiscj97, line 32: Did you choose to use a random index for any reason?

lib/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+            @landed_planes.delete_at(rand_index)
+            puts "The plane has taken off"
+        else
+            raise "No planes in the airport"
+        end
+    end
+
+    def get_landed_planes
+        return @landed_planes
+    end
```
lewiscj97, line 42: Since you have an attribute reader for `landed_planes`, you could probably delete this method (or the attribute reader!)

lib/weather.rb
```diff
@@ -0,0 +1,15 @@
+class Weather
```
lewiscj97, line 1: Minor: could maybe turn this into a module rather than a class, as I'm not sure whether you'd want to instantiate an instance of the Weather class? I guess this is just preference, but maybe something to consider?

lib/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+
+class Airport
+    attr_reader :plane
+    attr_reader :weather
+    attr_reader :capacity
+    attr_reader :landed_planes
+    DEFAULT_CAPACITY = 20
+
+    def initialize(plane, weather, capacity = DEFAULT_CAPACITY)
+        @plane = plane
```
lewiscj97, line 12: Maybe the `plane` could be an optional argument or something? I think it creates an interdependence on the two classes if you have to pass a plane argument into the airport on instantiation of the airport class. What do you think?

lib/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def get_landed_planes
+        return @landed_planes
+    end
+
+
+    private
+
+    def plane_can_fly?
+        if @weather.instance_variable_get(:@weather) == "calm"
```
lewiscj97, line 48: I've not seen this method before - could you have just used `@weather.weather` here (because of the attr_reader in the Weather class for the weather instance variable), or would that not work?

lib/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def get_landed_planes
+        return @landed_planes
+    end
+
+
+    private
+
+    def plane_can_fly?
```
lewiscj97, line 47: This method could probably be a one-liner, if you just had `@weather.instance_variable_get(:@weather) == "calm"`, the method would return true if it was true, or false if not - letting you take this out of an if/else

spec/airport_spec.rb
```diff
@@ -0,0 +1,57 @@
+require "Airport"
+require "weather"
+require "plane"
+test_airport = nil
+
+
+describe "Airport" do
+    let (:new_plane) {double :plane}
```
lewiscj97, line 8: Good use of doubles to not have any reliance on the plane class!

spec/airport_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    before(:each) do
```
lewiscj97, line 9: This is good too - nice DRY code

lib/airport.rb
```diff
@@ -0,0 +1,70 @@
...
+    def initialize(plane, weather, capacity = DEFAULT_CAPACITY)
+        @plane = plane
+        @weather = weather
+        @landed_planes = []
+        @capacity = capacity
+    end
+    
+    def plane_land(plane)
+        if !plane_can_fly?
+            raise "Weather warning, cannot land plane"
```
lewiscj97, line 20: Good, clear exception messages that are straightforward and easy to understand. I really like that you extracted the conditional `plane_can_fly?` into a private method too, so much easier to read and great protection of methods.

PR #2292, opened 2021-09-27
Airport Challenge

tommyoswinwilliams:


spec/airport_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'airport'
+
+describe Airport do
+
+  let(:plane) { double :plane }
```
alexnakov, line 6: Nice set up using doubles for the planes, I didn't think of that.

lib/airport.rb
```diff
@@ -1,3 +1,15 @@
...
 
+  def initialize
+    @planes = []
+  end
+
+  def land(plane)
+    @planes.push(plane)
+  end
+
+  def take_off(plane)
```
alexnakov, line 21: You are not using the parameter plane.

lib/airport.rb
```diff
@@ -1,3 +1,15 @@
...
+    @planes.pop
```
alexnakov: Imagine you landed plane1, plane2 and plane3 in that order.
`@planes = [plane1, plane2, plane3]`
Now you want to instruct plane1 to take off.
The `array.pop` method removes the last element of the array.
This means you would instruct plane3 to take off.
Use `@planes.delete(plane)` which deletes plane from `@planes`
i.e. `[plane1,plane2,plane3].delete(plane1)`

PR #2293, opened 2021-09-27
Airport-Challenge: Hassan Saleem

hassansaleem194:
Airport-Challenge: Hassan Saleem

### DIFF
--- a/spec/airport_land_spec.rb
+++ b/spec/airport_land_spec.rb
@@ -1,12 +1,6 @@
 require 'airport_land'
 
 RSpec.describe AirportLand do
-  
-  # expect(my_airport).to receive(:rand).and_return(3)
-
-  # random number generator, each number assigned to weather
-  # put the expect line in the it do block
-  #   put line at top of test
 
   it 'creates an hangar' do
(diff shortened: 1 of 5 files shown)


spec/airport_land_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    my_airport = AirportLand.new
+    plane = Plane.new
+    hangar = my_airport.hangar
+    expect(my_airport).to receive(:rand).and_return(2)
+    my_airport.land_plane(plane)
+    expect(my_airport).to receive(:rand).and_return(2)
+    my_airport.land_plane(plane)
+    expect(my_airport).to receive(:rand).and_return(2)
+    my_airport.land_plane(plane)
+    expect(hangar).to eq([plane, plane])
```
amin-newcastle, line 34: I tried my best not to hard code a result I am expecting. If we keep it generic then we can futureproof our tests. However, when I got stuck, I hardcoded the results i.e ([plane, plane]).

amin-newcastle (commented):
Well done Hassan. I've actually learned from reviewing your code. Thank you.

PR #2294, opened 2021-09-27
Airport challenge submission (Joseph)

Joseph112358:
Finished Airport challenge.

### DIFF
--- a/lib/Airport.rb
+++ b/lib/Airport.rb
@@ -8,7 +8,11 @@ class Airport
   end
   
   def land(plane)
+    if airport_full == false
     terminal.push(plane)
+    else
+      return "Airport is full, cannot land"
+    end
   end
 
   def take_off(plane,airport)
--- a/lib/Airport.rb
+++ b/lib/Airport.rb
@@ -1,5 +1,4 @@
 
-
 class Airport
   attr_accessor :terminal
   attr_accessor :capacity
@@ -60,7 +59,6 @@ class Plane
 
 end
 
-
 lhr = Airport.new
 plane = Plane.new
 plane2 = Plane.new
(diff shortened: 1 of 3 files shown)


lib/Airport.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+  end
+
+  def search_plane(airport,plane)
+    array = airport.terminal
+    array.include?(plane)
+  end
+
+  def airport_full
+    if terminal.size >= capacity
```
feedXia, line 37: Maybe can use 'fail/raise' here to be more concise & clearly highlight cases that would not work

lib/Airport.rb
```diff
@@ -0,0 +1,73 @@
...
+      return false
+    end
+  end
+
+  def overide_capacity(new_capacity)
+   @capacity = capacity
+  end
+
+  def weather1(num)
+  if num > 5
```
feedXia, line 49: Can use a random number generator for the weather:
e..g.
```
num = rand(5)
num > 4
```

spec/Airport_spec.rb
```diff
@@ -0,0 +1,81 @@
+require_relative '../lib/Airport.rb'
+
+describe Airport do
+  it 'directs a plane to an airport' do
```
feedXia, line 4: Can use:
```
let(:lhr) { Airport.new }
let(:flight0001) {Plane.new}
```
at the top to avoid repeating 
```
lhr = Airport.new 
flight0001 = Plane.new
```

spec/Airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    lhr = Airport.new 
+    flight0001 = Plane.new
+    lhr.land(flight0001)
+    expect(lhr.terminal[0]).to eq (flight0001)
+  end
+  it 'checks to see if plane is in airport (plane is in airport)' do
+  lhr = Airport.new
```
feedXia, line 11: Maybe can use `subject` instead of creating a new airport instance every time as it's in the Airport class by default

spec/Airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    expect(jfk.terminal.empty?).to eq true
+  end
+  it 'returns airport is full' do
+    den = Airport.new
+    10.times {den.land(Plane.new)}
+    expect(den.airport_full).to eq true
+  end
+  it 'checks if airport is full, (it is), prevents landing' do
+    den = Airport.new
+    10.times {den.land(Plane.new)}
```
feedXia, line 42: Great use of the times method!

PR #2295, opened 2021-09-27
Sergei Palmer Airport Challenge Updated

s-palmer:


lib/airport.rb
```diff
@@ -0,0 +1,44 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 5
+  attr_accessor :capacity
```
Kntanos, line 6: Is an attr-accessor required or attr_reader would be enough?

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def weather_station
+    # @weather = Weather.new
+    @weather.clear?
+  end
+
+  def takeoff(plane)
+    check_hangers
+    if weather_station
```
Kntanos, line 31: Can be refactored with a ternary expression

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+      puts "Skies are clear. Plane taking off."
+      airport.delete(plane)
+    else
+      puts "It's too stormy to fly"
+    end
+  end
+
+  def land(plane)
```
Kntanos, line 39: Is not checking if the weather is not stormy.

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+    fail "No space" if full?
+    @airport << plane
+    puts "Plane has landed"
```
Kntanos, line 42: I like the return comments confirming the action.

lib/weather.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def clear?
+    case rand(1..4)
+    when 1..3
+      @weather = true
+    when 4
+      @weather = false
+    end
+    # 1..3 = good weather
```
Kntanos, line 15: Good comment here.

lib/weather.rb
```diff
@@ -0,0 +1,19 @@
...
+  attr_reader :weather
+
+  def initialize
+    @weather = weather
+  end
+
+  def clear?
+    case rand(1..4)
+    when 1..3
+      @weather = true
```
Kntanos, line 11: Since you have an attr_reader for @weather, you don't need to keep using instance variables

lib/weather.rb
```diff
@@ -0,0 +1,19 @@
...
+    when 4
+      @weather = false
```
Kntanos, line 13: Since you have an attr_reader for @weather, you don't need to keep using instance variables

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 5
+  attr_accessor :capacity
+  attr_reader :airport
```
Kntanos, line 7: Add :weather to attr_reader, so you don't need to use instance variables in the rest of the code

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+    @airport.size >= capacity
+  end
+
+  def check_hangers 
+    fail "No Planes." if @airport.empty?
+    puts "Plane ready for take-off"
+  end
+
+  def weather_station
+    # @weather = Weather.new
```
Kntanos, line 25: Good one explaining what this is doing

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 5
+  attr_accessor :capacity
+  attr_reader :airport
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @airport = []
```
Kntanos, line 11: Why not call this variable hangers rather than airport? (I thing it's hangar)

lib/weather.rb
s-palmer: Done thanks  :)

lib/airport.rb
s-palmer: Done thanks!

lib/airport.rb
s-palmer: Not yet!

README.md
```diff
@@ -60,30 +47,49 @@ To ensure safety
...
 
-Your task is to test drive the creation of a set of classes/modules to satisfy all the above user stories. You will need to use a random number generator to set the weather (it is normally sunny but on rare occasions it may be stormy). In your tests, you'll need to use a stub to override random weather to ensure consistent test behaviour.
-
-Your code should defend against [edge cases](http://programmers.stackexchange.com/questions/125587/what-are-the-difference-between-an-edge-case-a-corner-case-a-base-case-and-a-b) such as inconsistent states of the system ensuring that planes can only take off from airports they are in; planes that are already flying cannot take off and/or be in an airport; planes that are landed cannot land again and must be in an airport, etc.
-
-For overriding random weather behaviour, please read the documentation to learn how to use test doubles: https://www.relishapp.com/rspec/rspec-mocks/docs . There’s an example of using a test double to test a die that’s relevant to testing random weather in the test.
+## How to use
 
-Please create separate files for every class, module and test suite.
+```
```
leoht, line 52: Nice detailed README with the instructions!

leoht (commented):
Nice work! The code is quite readable and clean, I mainly commented on things like:
 * Instance variables and their use or purpose — they're sometimes not needed when you don't need to store a state on the object
 * Plain strings in the code printed by `puts` — this is OK to try out things in the terminal, but be mindful of having them with your "main logic" code, since printing out things deals with presenting or displaying info to the user, so it probably deserves to be in its own class or component, to respect the SRP
 * Some minor things about refactoring, renaming, simplifying, etc :)

Let me know if some of the feedback is unclear and we can chat!

PR #2296, opened 2021-09-27
airport_challenge

ewa-j:
airport_challenge Ewa Jablonska


spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  let(:plane_two) { Plane.new }
+  
+  describe "#land" do
+    it "instructs a plane to land at an airport" do
+      expect(subject.land(plane, fine_weather)).to eq(plane)
+    end
+
+    context "airport is full" do
+      it "raises and error when airport is full" do
+        Airport::DEFAULT_CAPACITY.times { subject.land(plane, fine_weather) }
```
FrancineBlanc, line 20: Instead of initialising 10 new planes, you could try overriding the default capacity to say 1 or 2 planes just for this test so that if the default capacity changes in the code (e.g. it becomes 50) you won't need 50 instances of a new plane to be created to test the method, which could slow down the running of the program.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    plane.flying
+    @planes.delete(plane)
+    puts report_taking_off
+    plane
+  end
+
+  def report_taking_off
+    return "The plane successfully took off."
+  end
+
```
FrancineBlanc, line 32: I like that you have include a confirmation statement to confirm a successful take off.

lib/airport.rb
```diff
@@ -0,0 +1,37 @@
...
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane, weather)
+    raise 'Airport is full' if full?
+    raise 'Too stormy to land.' if weather.stormy?
+    plane.landed
+    @planes.push(plane)
+    plane
```
FrancineBlanc, line 16: Instead of returning the plane here, could you check that this plane is included in the planes array in the tests? This would then take away a line of code from this method :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      expect(subject.report_taking_off).to eq("The plane successfully took off.")
+    end
+    
+  end
+
+  it "sets capacity to default value" do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+
+end
```
FrancineBlanc, line 76: I find all your tests very clear and easy to read

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    context "stormy weather - no landing possible" do
+      it "raises an error when plane tries to land during a storm" do
+        expect { subject.land(plane, stormy_weather) }.to raise_error 'Too stormy to land.'
+      end
+    end
+
+  end
+  
+  describe "#takeoff" do
+
```
FrancineBlanc, line 34: Could the `fine_weather` scenarios be grouped together, and a variable set for `subject.land(plane, fine_weather)` so avoid repeating this statement in some of the tests?

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+    end
+
+    context "plane is not in the airport, so it cannot take off from here" do
+      it "raises an error when plane is not in the airport" do
+        subject.land(plane, fine_weather)
+        expect { subject.takeoff(plane_two, fine_weather) }.to raise_error "Plane cannot take off if it's not in the airport."
+      end
+    end
+
```
FrancineBlanc, line 59: I like that you have used the `context` method here to set the scene

spec/airport_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    context "airport empty - there's no plane that could take off" do
+      it "raises an error when there are no planes in the airport" do
+        expect { subject.takeoff(plane, fine_weather) }.to raise_error "There are no planes in the airport."
+      end
+    end
+
+    context "plane is not in the airport, so it cannot take off from here" do
+      it "raises an error when plane is not in the airport" do
+        subject.land(plane, fine_weather)
```
FrancineBlanc, line 55: Does this instance of `plane` need to land for the test?

FrancineBlanc (commented):
The code was well written and easy to read and covers all of the user stories. Each object is within its own class and I can understand the reasons for each test. I overall enjoyed reading and reviewing this approach to the airport challenge.

PR #2297, opened 2021-09-27
Week 3 Airport Challenge :) 

AishaDorsett:
My attempt at the airport challenge

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,29 +1,40 @@
 class Airport
-    attr_accessor :capacity
-    attr_reader :planes_at_airport
+	attr_accessor :capacity
+	attr_reader :planes_at_airport
+	attr_accessor :weather_today
 
-    DEFAULT_CAPACITY=100
+	DEFAULT_CAPACITY=100
 
-    def initialize(capacity = DEFAULT_CAPACITY)
-        @capacity = capacity
-        @planes_at_airport = []
-    end
+	def initialize(capacity = DEFAULT_CAPACITY)
+		@capacity = capacity
+		@planes_at_airport = []
+		@weather_today = Weather.new
+	end
 
-    def full?
-        @planes_at_airport.length == @capacity
-    end
+	def full?
+		@planes_at_airport.length == @capacity
+	end
 
-    #planes land at airport and get added to the list of planes at the airport
-    def airport_dock(plane)
-        fail 'airport is full' if full?
-        @planes_at_airport << plane
-        @plane = plane
+	#planes land at airport and get added to the list of planes at the airport
+	def land_plane(plane)
+		fail 'airport is full' if full?
+		@planes_at_airport << plane
+		@plane = plane
 
-    end
+	end
 
-    def plane_takeoff_confirmation
-        @planes_at_airport.pop 
-        p "WE OUTTA HERE!!"
-    end
+	def plane_takeoff_confirmation
+		@planes_at_airport.pop 
+		p "WE OUTTA HERE!!"
+	end
+
+	def prevent_takeoff(plane)
+		fail 'Unable to take off due to stormy weather' if @weather_today.stormy?
+		@plane = plane
+	end
+
+	def prevent_landing
+
+	end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,40 +1,42 @@
 class Airport
-	attr_accessor :capacity
-	attr_reader :planes_at_airport
-	attr_accessor :weather_today
+	 attr_accessor :capacity
+	 attr_accessor :weather
+	 attr_reader :planes_at_airport
 
-	DEFAULT_CAPACITY=100
+	 DEFAULT_CAPACITY = 5
 
-	def initialize(capacity = DEFAULT_CAPACITY)
-		@capacity = capacity
-		@planes_at_airport = []
-		@weather_today = Weather.new
-	end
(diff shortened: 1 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,42 @@
+class Airport
+	 attr_accessor :capacity
+	 attr_accessor :weather
+	 attr_reader :planes_at_airport
+
+	 DEFAULT_CAPACITY = 5
```
Anamaria-Mocanu, line 6: Like the use of the constant, I might add it in mine as well :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'airport'
+require 'plane'
```
Anamaria-Mocanu, line 2: RSpec might not need the require for the other files(plane and weather), you might be able to just leave require "airport" and it should hopefully still work ( I think Rspec knows to look in the folder of the file that you required and read the other files for the methods you are calling

spec/airport_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+	 describe '#successfully_land' do
+			it 'lands plane at airport by adding it to list' do
+				airport = Airport.new
+				plane = Plane.new
+				airport.successfully_land(airport)
+				expect(airport.planes_at_airport).to contain_exactly(airport)
+			end
+		
+			it 'rasies an error when plane tries landing in a full airport' do
```
Anamaria-Mocanu, line 30: Typo in ' ', raises

spec/airport_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+RSpec.describe Airport do
+	 describe '#initialize' do
+		  it 'allows user to set airport capacity and be overridden' do
+			   port = Airport.new(200)
+			   expect(port.capacity).to eq 200
+		  end
+	 end
+
+	 describe '#full?' do
+			it 'returns true or false if airport is full' do
```
Anamaria-Mocanu, line 14: maybe separate this into two, one that returns true and one that returns false, so that you can cover the whole basis of the  full? method

spec/plane_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'plane'
+require 'airport'
+
+RSpec.describe Plane do 
+  describe '#working?' do
```
Anamaria-Mocanu, line 5: might need to add a section that expects working? to return true

spec/plane_spec.rb
```diff
@@ -0,0 +1,10 @@
...
+    it ' confirms Plane can rspond to working? methods' do
```
Anamaria-Mocanu, line 6: typo 'respond'

spec/weather_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+				allow_any_instance_of(Weather).to receive(:stormy?).and_return(true)
+				expect(subject.stormy?).to eq true
+			end
+
+			it 'returns false if random weather condition is not eq to stormy' do
+				expect(subject.stormy?).to eq false unless subject.stormy?
+			end
+	 end
+
+	 describe '#random_conditon' do
```
Anamaria-Mocanu, line 16: love the  testing of  the randomiser

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+class Weather 
+	attr_accessor :condition
+
+	 WEATHER_CONDITIONS = ["Stormy", "Sunny", "Raining", "Mild", "Windy"]
```
Anamaria-Mocanu, line 4: this constant is very realistic, it covers more possible outcomes

spec/airport_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'airport'
+require 'plane'
+require 'weather'
+
+RSpec.describe Airport do
```
Anamaria-Mocanu, line 5: I think if we were to use let, or subject to instantiate our classes, rather than writing plane = ... or airport = ... for each of our "it" statements will make our code cleaner and shorter. We should definitely look into how those work. However, as we are at the beginning, I enjoy the repetition as it helps understand the way the tests work.

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+	
+	 def successfully_land(plane)
+		  fail 'airport is full' if full?
+		  @planes_at_airport << plane
+		  return plane
+	 end
+	
+	 def successful_takeoff(plane)
+		  while @planes_at_airport.include?(plane)
+			  @planes_at_airport.delete_at(@planes_at_airport.index(plane))
```
Anamaria-Mocanu, line 26: Really like the way this was implemented, it ensures you remove the exact plane out of the array.

spec/airport_spec.rb
AishaDorsett: thank you!

spec/airport_spec.rb
AishaDorsett: Thank you! will do

spec/plane_spec.rb
AishaDorsett: think I just added this to put something in the plane class haha. would love to separate some of my airport methods and put them in the planes file how you did it.

spec/airport_spec.rb
AishaDorsett: very true! I started by writing plane = Plane.new then refactored it to the 'subject' version once it worked :)

PR #2298, opened 2021-09-27
My airport challenge

sunnydonaldson:
I didn't have time to refactor and of the tests, and most of the code, or separate the weather from the airport class.
I completed all user stories.


2Rads:
This is great, well thought out and neatly presented. Tests are thorough and cover most aspects.

I made comments on a two things and the first commit had too many changes at once, but was otherwise very well done.

PR #2299, opened 2021-09-27
Airport challenge

ann-ya:
Airport challenge Anna Jablonska

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,24 @@
+class Airport
+
+  attr_reader :plane
+  attr_reader :capacity
+
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @plane = plane
+    @planes_in = []
+  end
+  
+  def land(plane)
+    fail 'Airport full' if full?
+    @planes_in.push(plane)
+    @plane = plane
+  end
+
+  def full?
+    @planes_in.length == DEFAULT_CAPACITY
+  end  
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -19,6 +19,16 @@ class Airport
 
   def full?
     @planes_in.length == DEFAULT_CAPACITY
-  end  
+  end
+  
+  def empty?
+    @planes_in.length == 0
+  end
+
+  def take_off(plane)
+    fail 'Airport empty' if empty?
+    @planes_in.delete(plane)
+    "Plane took off"
+  end
 
 end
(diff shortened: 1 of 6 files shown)


lib/weather.rb
```diff
@@ -0,0 +1,21 @@
...
+  def set_status
+    if (1..7).include?(@random_number)
+      status = 'Sunny'
+    else
+      status = 'Stormy'
+    end
+  end
+
+  def gen_rand_numb
+    rand(1..10)
```
dannehdan, line 19: You can just use `rand(11)` here to generate a random number between 0 and 11 (so a random number that's either 1, 2, 3, 4, 5, 6, 7, 8, 9 or 10)

https://blog.appsignal.com/2018/07/31/generating-random-numbers-in-ruby.html

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+  
+  def land(plane)
+    fail 'Airport full' if full?
+    @planes_in.push(plane)
+    @plane = plane
+  end
+
+  def full?
+    @planes_in.length == DEFAULT_CAPACITY
```
dannehdan, line 21: This check here is testing if the `@planes_in` is equal to the default capacity, but if the capacity has been changed in the initialize, this `full?` method won't use that `@capacity` value

spec/weather_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'weather'
+
+describe Weather do
+
+  it "set status to sunny" do
+    # allow(subject).to receive(:set_status) { 'Sunny' }
+    stub(:rand) {7}
```
dannehdan, line 7: Great use of the `stub` here - much more concise than mine!

lib/weather.rb
```diff
@@ -0,0 +1,21 @@
+class Weather
+
+  attr_reader :status, :random_number
+
+  def initialize
+    @status = status
+    @random_number = gen_rand_numb
+  end
+
+  def set_status
```
dannehdan, line 10: Could you use something here that allows you to accept more than 2 weather conditions? It might be that we need to add in 'snowy' as an option further down the line

spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'airport'
```
dannehdan, line 1: Really like your airport_spec file - the tests are nice and short and easy to read

dannehdan:
Really good job overall! A couple of general points

1. I can't see an updated README.md in your repo?
2. Maybe commit more often? I think your first commit was almost all your completed code, which would make it hard to go back if you found an error further along in your code

PR #2301, opened 2021-09-27
Airport Challenge

arthurfincham:
- Completed all the user stories.
- Think that the doubles are good but not sure. Please take a look for me!
- Test coverage is approx 97%.
- All tests passed.


lib/airport.rb
```diff
@@ -0,0 +1,50 @@
+require 'plane'
+require 'weather'
+
+class Airport
+
+  attr_reader :tarmac, :capacity
+
+  def initialize(capacity = 1)
+    @capacity = capacity
+    @tarmac = []
```
Alex969, line 10: I was looking for that word, good on you that you're using it, it just makes more sense

Alex969, line 10: @tarmac : I was looking for that word, good on you that you're using it, it just makes more sense

spec/airport_spec.rb
```diff
@@ -0,0 +1,77 @@
+require 'airport' 
+
+describe Airport do
+  let(:plane) { Plane.new }
+  subject(:airport) { described_class.new }
```
Alex969, line 5: I'm really not sure, but I think you don't necessarily need line 5 here since you're using a double for plane and are only describing on Airport class

spec/airport_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      expect(airport.capacity).to eq 10
+    end
+  end
+
+  describe '#landing' do
+    before do
+      allow(plane).to receive(:land)
+    end
+
+    context 'weather is sunny' do
```
Alex969, line 19: using context makes things so much easier to read/write ! I'll definitely refactor my Boris bikes rspec file with that

Alex969 (commented):
Your code is great and so well written  ! There's not much to review lol I was mostly learning from you so thank you !

spec/airport_spec.rb
arthurfincham: Nice! Yeah I think it is the same as _describe_ but it makes code more readable.

spec/airport_spec.rb
arthurfincham: Yes you're right - still a bit confused about doubles

lib/airport.rb
arthurfincham: 👍

PR #2302, opened 2021-09-27
Airport challenge

Langan84:
I did not finish a README.md
I did not test for edge cases. 
I would like feedback on my Tests.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -12,13 +12,14 @@ attr_reader :plane
     raise 'The airport is full!' if full?
     raise 'You cannot land the weather is stormy' if stormy?
     @planes << plane
+    plane
     
   end
 
   def take_off(plane)
-  fail "You cannot take off the weather is stormy" if stormy?
-  @plane = plane
-  plane
+    fail "You cannot take off the weather is stormy" if stormy?
+    @plane = plane
+    plane
   end
 
   private
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -9,20 +9,26 @@ attr_reader :plane
   end
 
   def land(plane)
-    fail "The airport is full!" if full?
+    raise 'The airport is full!' if full?
+    raise 'You cannot land the weather is stormy' if stormy?
     @planes << plane
-    plane
+    
   end
 
   def take_off(plane)
+  fail "You cannot take off the weather is stormy" if stormy?
   @plane = plane
-  @plane
+  plane
   end
 
   private
 
   def full?
     @planes.length >= @capacity
+  end
+
+  def stormy?
+  rand(1..6) > 5
   end
 end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,28 @@
+
+require_relative 'plane'
+class Airport
+attr_reader :plane
+
+  def initialize(capacity = 10)
+    @capacity = 10
+    @planes = []
+  end
+
+  def land(plane)
+    fail "The airport is full!" if full?
+    @planes << plane
+    plane
+  end
+
+  def take_off(plane)
+  @plane = plane
+  @plane
+  end
+
+  private
+
+  def full?
+    @planes.length >= @capacity
+  end
+end
+
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -4,11 +4,12 @@ describe Airport do
   let (:plane) { double :plane }
  
     it 'responds to land' do
-      expect(airport).to respond_to(:land).with(1).argument
       allow(airport).to receive(:stormy?).and_return false
+      expect(airport).to respond_to(:land).with(1).argument
     end
 
     it 'instructs a plane to land' do
+      allow(airport).to receive(:stormy?).and_return false
       expect(airport.land(plane)).to(eq(plane))
     end
 
@@ -20,7 +21,7 @@ describe Airport do
 
     it 'throws an error if a planes tries to land when stormy.' do
       allow(airport).to receive(:stormy?).and_return true
-      airport.land(plane).to raise_error 'You cannot take off the weather is stormy'
+      expect{airport.land(plane)}.to raise_error 'You cannot land the weather is stormy'
     end
   
     it 'responds to take off' do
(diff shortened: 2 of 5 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  let (:airport) { described_class.new(10)}
+  let (:plane) { double :plane }
+ 
+    it 'responds to land' do
+      allow(airport).to receive(:stormy?).and_return false
+      expect(airport).to respond_to(:land).with(1).argument
+    end
+
+    it 'instructs a plane to land' do
+      allow(airport).to receive(:stormy?).and_return false
```
serenabertozzi, line 12: To avoid this repetition, I've found double and stubs very useful! (Although, I couldn't properly understand at first because they can be confusing!)
What I did was to, change "subject" to include an instance double and stub, and divided the tests into "stormy weather" and "not stormy'. It's a bit long but I'm happy to share it with you in person! 👍🏼

serenabertozzi (commented):
Hi Kirstie! 
I've tried to check your code carefully, hope my comments are clear (and my English isn't too bad!). Good job on completing all the user stories! 👍🏼

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    raise 'The airport is full!' if full?
+    raise 'You cannot land the weather is stormy' if stormy?
+    @planes << plane
+    plane
+    
+  end
+
+  def take_off(plane)
+    fail "You cannot take off the weather is stormy" if stormy?
+    @plane = plane
```
alicelieutier, line 21: Why do you need that instance variable? Your method returns the plane already, so there shouldn't be a problem accessing the plane that has taken off.. Is there something I'm missing? As a rule, it's good to have fewer rather than more instance variables.

lib/airport.rb
```diff
@@ -9,20 +9,26 @@ def initialize(capacity = 10)
...
+  plane
   end
 
   private
 
   def full?
     @planes.length >= @capacity
   end
+
+  def stormy?
```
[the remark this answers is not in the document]
alicelieutier: Having two classes also presents the advantage of making it easier to test. It would be a good next step :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'airport'
+describe Airport do
+  let (:airport) { described_class.new(10)}
+  let (:plane) { double :plane }
+ 
+    it 'responds to land' do
+      allow(airport).to receive(:stormy?).and_return false
```
alicelieutier, line 7: You are stubbing a method of the class under test here, which is not good practice. Look up `srand` if you'd like to fix randomness for a test.

spec/airport_spec.rb
alicelieutier: Same here.

spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      expect(airport).to respond_to(:land).with(1).argument
```
alicelieutier, line 8: The respond_to matcher is ok as a first test, but soon you should try to create tests around behaviour instead. What's the behaviour of the "land" method? As a user, how would you know a plane has landed?
In your code, one behaviour I can see is that if you land 10 planes, then the airport should be full. That would be a good way to test that land actually works.

spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it 'instructs a plane to land' do
+      allow(airport).to receive(:stormy?).and_return false
+      expect(airport.land(plane)).to(eq(plane))
+    end
+
+    it 'throws an error if a plane tries to land when full' do
+      allow(airport).to receive(:stormy?).and_return false
+      10. times {airport.land(plane)}
+      expect{airport.land(plane)}.to raise_error "The airport is full!"
+      end
```
alicelieutier, line 20: Yes! Great behaviour based test (but do change to srand)

spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    it 'throws an error if a planes tries to land when stormy.' do
+      allow(airport).to receive(:stormy?).and_return true
+      expect{airport.land(plane)}.to raise_error 'You cannot land the weather is stormy'
+    end
```
alicelieutier, line 25: again, good behaviour test.

spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  
+    it 'responds to take off' do
+      expect(airport).to respond_to(:take_off).with(1).argument
+    allow(airport).to receive(:stormy?).and_return false
+    end
```
alicelieutier, line 30: I would remove this test, it doesn't test behaviour. It's only useful while you build the class, and even then could be skipped once you are confident with your coding.

spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    it 'instructs a plane to take off' do
+    allow(airport).to receive(:stormy?).and_return false
+    expect(airport.take_off(plane)).to(eq(plane))
+    end
+  
+    it 'throws an error if a planes tries to take_off when stormy.' do
+        allow(airport).to receive(:stormy?).and_return true
+        expect{airport.take_off(plane)}.to raise_error 'You cannot take off the weather is stormy'
+    end
```
alicelieutier, line 40: Good.

spec/plane_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'plane'
+describe Plane do
```
alicelieutier, line 2: It makes sense that you don't need tests for an empty class :)

alicelieutier (commented):
Overall your code is clean and well organised.
You have an extra instance variable that serves no purpose and should be removed.
Your tests are typically good, except that they rely on `allow(airport).` stubbing, which is a stubbing of the class under test, something that is not ideal. I'd encourage you to use `srand` to make these tests better, or make the Weather a separate class, and then use a double with a stub for it when testing `Airport`.
Some tests are unnecessary (`.respond_to` matcher).

More comments below.

PR #2307, opened 2021-10-17
first 4 completed weather condition left

NenehB:


spec/weather_spec.rb
```diff
@@ -0,0 +1,6 @@
+describe Weather do 
+  it 'should check for weather condition' do 
+    weather = Weather.new
+    expect(weather.Weather_condition).to be_truthy
```
jjegarcia, line 4: not sure what "be_truthy" means. I hope it covers also that I amy be also fault at time. Otherwise please write another test to improve coverage

PR #2308, opened 2021-10-17
Jamie Bowers - Airport Challenge Pull Request

JRMBowers:

### DIFF
--- a/README_CHALLENGE.md
+++ b/README_CHALLENGE.md
@@ -8,4 +8,6 @@ I tested the rspec and code using IRB in order to be sure that everything was fu
 
 After the completion of the code I used rubocop, rspec and simplecov to ensure that I cover as much of the code with tests as I can and that the code has been produced in an acceptable way. The code will not be submitted until both of these have been satisfied.
 
-Tried setting up travis but could not get it to work for some reason. I will look into why this is the case.
+Tried setting up travis but could not get it to work for some reason. I will look into why this is the case. 
+
+If I was to carry out this task again I would make sure to commit more as I noticed I did not commit when producing the code. Next time I would set checkpoints and commit at every milestone.


lib/airport.rb
```diff
@@ -0,0 +1,75 @@
...
+  def airport_error
+    fail "This plane is not stationed at this Airport" if airport_check
+  end 
+
+  def airport_check
+    checker = @planes_available.last
+    checker[:airport] != @airport
+  end 
+
+  def plane_already_landed?(plane)
```
PatrickBarsch, line 62: I think methods should end on "?" if they return "true"/"false".
The guard condition achieves something similar, but I am not sure wether it is in line with the convention.

lib/airport.rb
```diff
@@ -0,0 +1,75 @@
...
+    @planes_available.each do |airport_entry| 
+      fail "Plane has already landed" if airport_entry[:plane] == plane 
+    end 
+  end 
+
+  def skytraffic_checker
```
PatrickBarsch, line 68: Why does **skytraffic_checker** not take an argument, when **plane_already_landed** does?

spec/weather_spec.rb
```diff
@@ -0,0 +1,16 @@
+require_relative '../lib/weather'
+
+describe Weather do 
+  describe '#random_selector' do 
+    it 'checks random_selector return a number' do 
+      expect(subject.random_selector).to be_a_kind_of(Integer)
+    end 
+  end 
+  describe '#weather' do 
+    it 'checks weather function works correctly' do
```
PatrickBarsch, line 10: Maybe a bit more specific? Like, "returns 'CLEAR_SKIES' if rand returns a 1" ?

PR #2310, opened 2021-10-18
Extract weather

jjegarcia:


lib/plane.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  private def confirm_take_off
+    'confirmed take off'
+  end
+
+  private def update_state(state)
+    @state = state
+  end
+
```
NenehB, line 18: You can just write one private, you put it before the method and anything after the private method will be private.

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+    @spaces << plane
+  end
+
+  def take_off
+    fail 'no plane available' unless spaces.positive?
+    fail 'stormy weather' unless stormy? != WEATHER_STATES[:stormy]
+    plane = @spaces.pop
+    plane.take_off(id)
+  end
+end
```
NenehB, line 40: Really love the naming of the methods easy to understand what each method is meant to do

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+WEATHER_STATES = { stormy: true, not_stormy: false }.freeze
+STORMY_ODDS = [WEATHER_STATES[:not_stormy], WEATHER_STATES[:not_stormy],
+               WEATHER_STATES[:not_stormy], WEATHER_STATES[:stormy]].freeze
```
NenehB, line 3: Another way easier way would be to use rand(100) < 90 sunny else stormy

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+  it 'take off should prevent to take off in stormy weather' do
+    subject.land(Plane.new)
+    allow(subject).to receive(:stormy?).and_return WEATHER_STATES[:stormy]
+    expect { subject.take_off }.to raise_error('stormy weather')
+  end
+  it 'take off should prevent to take off if empty spaces' do
+    expect { subject.take_off }.to raise_error('no plane available')
+  end
+end
```
NenehB, line 54: Really nice, different test cases

PR #2311, opened 2021-10-18
Raul de la Hera Airport challenge release 1

rauldelahera:
4 user stories and 78% coverage for now. Still working on it.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,19 @@
+
+class Airport
+  attr_accessor :planes_landed
+  attr_accessor :capacity
+
+  def initialize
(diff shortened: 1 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,19 @@
...
+class Airport
+  attr_accessor :planes_landed
+  attr_accessor :capacity
+
+  def initialize
+    @planes_landed = 0
+    @capacity = 3
+  end
+ 
+  def full?
```
JackStorror, line 11: this method could also appear like
 def full?
      planes_landed >= capacity
  end

PR #2312, opened 2021-10-18
Airport Challenge

antlerhead:
Airport Challenge pull request

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,16 +1,12 @@
 
 class Airport
 
-    Default_capacity = 15
-
-    def initialize(capacity = Default_capacity)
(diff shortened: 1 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def take_off(plane)
+    fail 'cannot take off if weather stormy' if stormy?
+  end
+
+  def full?
+    @planes.length >= @capacity
+  end
+
+  def stormy?
```
MiaGordon91, line 25: Would you consider creating a random weather generator that is applied to an array of weather conditions? That way it might be a bit clearer when checking for stormy weather

PR #2314, opened 2021-10-18
Airport challenge: Starting methods and tests

MickStein:
Added methods and tests for landing plane, take off, ensure safe by not allowing landing when full.


spec/spec_traffic_controller_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'traffic_controller'
```
MarcBrosch, line 1: I would suggest to use "airport" as a class name

PR #2315, opened 2021-10-18
Mia Gordon: Attempt at Airport Challenge

MiaGordon91:
- I have attempted to complete the first 4 user stories.
- Creating the weather test was tricky as I wasn't sure how to test that the random weather generator worked. In the end I found myself writing a test to try and pass my code which isn't good. 
- There are two commented out tests which were additional attempts to test the stormy weather user story but didn't quite work.
-  I have 3 ruby files for Airport, Weather and Plane class. I created the plane class to pass the first user story however as I progressed through the stories I questioned the method inside the Plane class and have commented that out for now.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,27 +1,38 @@
 require_relative 'plane'
+require_relative 'weather'
 
 class Airport
 
-  def initialize
+  attr_accessor :capacity 
+
+  DEFAULT_CAPACITY = 2
+
+  def initialize( cap=DEFAULT_CAPACITY )
     @landed_planes = []
+    @capacity = cap
     @taken_off_planes = []
-    @capacity = 2
+    @forecast = []
   end
 
   def plane_land(plane) 
     # # "Plane needs to land"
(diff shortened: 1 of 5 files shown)


lib/weather.rb
```diff
@@ -0,0 +1,18 @@
...
+class Weather
+
+  attr_reader :todays_weather
+
+  def initialize
+    @todays_weather = []
+  end 
+
+  def todays_weather
+    weather = ["stormy", "light drizzle", "sunshine",
```
antlerhead, line 11: This is a really nice addition.

lib/airport.rb
```diff
@@ -1,39 +1,60 @@
...
 
 # airport = Airport.new
-# puts airport.plane_land(Plane.new)
-# puts airport.plane_land(Plane.new)
+# plane = Plane.new
+# puts airport.plane_land(plane)
+# # # puts airport.stormy
 # puts airport.plane_take_off
 
+
```
antlerhead, line 59: Good readability! very clear naming etc.

PR #2317, opened 2021-10-18
airport challenge

RJ21BL:
First PR for airport challenge.

### DIFF
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,0 +1,12 @@
+# class Weather
+
+#   attr_reader :sunny
+
+#   def initialize
+#     rand(10) == 0 ? @sunny = false : @sunny = true
+#   end
+
+#   def sunny?
+#     sunny
+#   end
+# end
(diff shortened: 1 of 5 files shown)


lib/weather.rb
```diff
@@ -0,0 +1,12 @@
+# class Weather
```
Yasmin-jones, line 1: Good use of SRP by creating a weather class

lib/weather.rb
```diff
@@ -0,0 +1,12 @@
...
+
+#   attr_reader :sunny
+
+#   def initialize
+#     rand(10) == 0 ? @sunny = false : @sunny = true
```
Yasmin-jones, line 6: for this method put this in your 'sunny?' method below and try rand(1..6) > 4 and see if that maybe works? 

if your sunny? method is true then it is stormy 
if your sunny? method is false then it is NOT stormy 
(maybe change the name of the method to 'stormy?' to avoid confusion)

PR #2346, opened 2021-11-19
airport_challenge

sherwingp:
Todo:
- [x] Write tests for edge cases
- [ ] Refactor

User stories:

- [x] As an air traffic controller 

- [x] As an air traffic controller 

- [x] As an air traffic controller 
        To ensure safety 

- [x] As the system designer

- [x] As an air traffic controller 
            To ensure safety 

- [x] As an air traffic controller 
        To ensure safety 


lib/airport.rb
```diff
@@ -0,0 +1,19 @@
+class Airport
+  attr_reader :planes, :capacity, :weather
+  DEFAULT_CAPACITY = 10
+  WEATHER_SCALE = 10
+
+  def initialize(capacity = 10)
```
tc1316, line 6: Could use capacity = DEFAULT_CAPACITY

lib/airport.rb
```diff
@@ -0,0 +1,19 @@
...
+    @planes = []
+    @capacity = capacity
+    @weather = rand(WEATHER_SCALE)
+  end
+
+  def full?
+    return @planes.length >= DEFAULT_CAPACITY
```
tc1316, line 13: Dont need @planes since its been declared in attr_reader

lib/airport.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def stormy?
+    @weather >= WEATHER_SCALE
```
tc1316, line 17: Likewise above

lib/plane.rb
```diff
@@ -0,0 +1,30 @@
+class Plane
+  attr_reader :flying
+  def initialize
+    @flying = true
+  end
+
+  def land(airport)
+    fail "airport is full" if airport.full?
+    fail "weather is stormy" if airport.stormy?    
+    fail "plane has already landed" if flying? == false
```
tc1316, line 10: Good use of guard conditions

spec/airport_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+describe Airport do
+  subject(:airport) { Airport.new }
+
+  it 'default capacity can be overridden' do
+    airport = Airport.new(3)
+    expect(airport.capacity).to be(3)
+  end
+
+  it 'can check for stormy weather' do
+    expect(subject.stormy?).to be_truthy.or be_falsy
```
tc1316, line 13: Cool!

spec/plane_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+describe Plane do
+  subject(:plane) { Plane.new }
+
+  context 'landing' do
+    it 'lands at airport' do
+      airport = Airport.new
+      allow(airport).to receive(:stormy?) { false }
+      subject.land(airport)
+      expect(airport.planes.length).to be(1)
```
tc1316, line 12: Alternative suggestion: expect(airport.planes).to include(plane)

spec/plane_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    it 'cannot land if airport is full' do
+      airport = Airport.new
+      Airport::DEFAULT_CAPACITY.times { airport.planes.push("plane") }
```
tc1316, line 17: Good use of calling DEFAULT_CAPACITY here to show that the airport will be full

spec/plane_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'plane'
+require 'airport'
+
+describe Plane do
+  subject(:plane) { Plane.new }
+
+  context 'landing' do
```
tc1316, line 7: Context use makes tests clearer to read

spec/plane_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'plane'
+require 'airport'
+
+describe Plane do
+  subject(:plane) { Plane.new }
+
```
tc1316, line 6: You could create the airport up here instead of in each test

tc1316 (changes requested):
Code review from Titus

lib/airport.rb
sherwingp: Thanks Titus! I didn't know that!

lib/plane.rb
sherwingp: Thanks Titus!

PR #2347, opened 2021-11-20
First pull request

chris-clement:


lib/airport.rb
```diff
@@ -4,35 +4,36 @@ class Airport
 
   attr_reader :planes
   attr_accessor :capacity
+  DEFAULT_CAPACITY = 25
 
   def initialize
     @planes = []
-    @capacity = 25
+    @capacity = DEFAULT_CAPACITY
   end
```
christopherxwebb, line 12: Everything is really elegant and neat. The code is very clear, well-formatted, and easy to read.

lib/airport.rb
```diff
@@ -4,35 +4,36 @@ class Airport
...
 
   def land(plane)
-    fail "Airport is full" if @planes.count >= @capacity
```
christopherxwebb: Great use of guard clauses, consistent throughout.

lib/airport.rb
```diff
@@ -4,35 +4,36 @@ class Airport
...
 
   def take_off
     fail "Stormy weather, can't take off" if stormy?
     fail "Airport is empty, no planes can take off" if empty?
-    @planes.pop
+    planes.pop
   end
 
   def stormy?
     rand(0..10).zero?
```
christopherxwebb, line 28: This is great, too, using a simple method ('zero?'). I often forget to call 'zero?' and user the '==' operator to point to '0'.

User_Stories.md
```diff
@@ -0,0 +1,58 @@
...
+|airport|land, take_off, planes, capacity, stormy?|
+
+Create stormy? method that returns true 1 in x times.
+If it is stormy, the take_off method should raise an error.
+
+As an air traffic controller 
+To ensure safety 
+I want to prevent landing when weather is stormy 
+
+Same as above but land should raise an error if stormy
```
christopherxwebb, line 58: A really clear User Stories.md (something I forgot to add myself!). Only possible suggestion for incorporation is: could there be mention of some of the edge cases? (Then again, if you've sorted those out already, it probably isn't necessary!)

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def initialize
+    @planes = []
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def land(plane)
+    fail "Airport is full" if planes.count >= capacity
+    fail "Stormy weather, can't land" if stormy?
+    fail "This plane has already landed" if landed?(plane)
```
christopherxwebb, line 17: This is a really smart method to use here.

spec/airport_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    context "when not stormy" do
+
+      before(:each) do
+        allow(@airport).to receive(:stormy?).and_return(false)
+      end
+
+      it { is_expected.to respond_to :take_off }
+
+      it "After instructing a plane to take off, I expect that plane to not be in the airport if the weather is not stormy" do
```
christopherxwebb, line 62: I wonder if this particular test could be written in a slightly less confusing way? I found myself struggling with the double-negative. Or perhaps just the last part ('if the weather is not stormy') is not entirely necessary (in spite of it helpfully specifying things further).

christopherxwebb (commented):
Well done! It's really neat and elegant. The Rspec testing is impressive, too. I'm sorry I don't have much more to add!

spec/airport_spec.rb
chris-clement: Agreed. This does not read well so I will look to rewrite this test to make it clearer.

User_Stories.md
chris-clement: Good point, I will look to add the edge cases to the User Stories.

PR #2348, opened 2021-11-21
airport challenge

bryn9696:


lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+        raise 'cannot land plane as airport full' if full?
+        raise 'cannot land plane: its stormy' if stormy?
+        @planes_landed << plane
+    end
+
+    def takeoff(plane)
+        raise 'cannot land plane: its stormy' if stormy?
+    end
+    
+    private
```
Mayo-Theodore, line 18: Why did you decide to make these methods private?

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def full?
+        @planes_landed.length >= @capacity
+    end
+
+    def stormy?
```
Mayo-Theodore, line 24: Great implementation of stormy conditions

spec/plane_spsec.rb
```diff
@@ -0,0 +1,4 @@
+require plane
```
Mayo-Theodore, line 1: Should this be require 'plane'? Did this work without the quotation marks

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    def land(plane)
+        raise 'cannot land plane as airport full' if full?
+        raise 'cannot land plane: its stormy' if stormy?
+        @planes_landed << plane
+    end
+
+    def takeoff(plane)
+        raise 'cannot land plane: its stormy' if stormy?
```
Mayo-Theodore, line 15: Could add some logic here, for when the plane takes off (when conditions are clear)

Mayo-Theodore (commented):
Well written, methods all had single responsibilities and were also concise.

PR #2349, opened 2021-11-21
airport_challenge

giorgigutsaevi:


lib/airport.rb
```diff
@@ -11,18 +11,22 @@ def initialize(capacity = DEFAULT_CAPACITY)
...
 	def land(plane)
-		fail "Unable to land during the storm" if stormy_weather?
+		fail "Unable to land during the storm" if stormy_weather? == true
 		raise "Airport full" if @hangar.size == @capacity
 		@hangar.push(plane)
 	end
 
 	def take_off
-		raise "Can't take off during the storm." if stormy_weather?
+		raise "Can't take off during the storm." if stormy_weather? == true
```
TomasGarciaDev, line 20: Maybe you don't need the "== true", the value of stormy_weather? already returns true or false

lib/airport.rb
```diff
@@ -11,18 +11,22 @@ def initialize(capacity = DEFAULT_CAPACITY)
 	end
 
 	def land(plane)
-		fail "Unable to land during the storm" if stormy_weather?
+		fail "Unable to land during the storm" if stormy_weather? == true
```
TomasGarciaDev, line 14: Maybe same thing here, no need of the "==true"

lib/airport.rb
```diff
@@ -11,18 +11,22 @@ def initialize(capacity = DEFAULT_CAPACITY)
...
 		@hangar.push(plane)
 	end
 
 	def take_off
-		raise "Can't take off during the storm." if stormy_weather?
+		raise "Can't take off during the storm." if stormy_weather? == true
 		@hangar.pop
 	end
 
+	def verify
```
TomasGarciaDev, line 24: Maybe the name could be a bit more descriptive and makes what is "verify"ing

TomasGarciaDev (commented):
Great use of "context" for your rspec. Great job!

PR #2350, opened 2021-11-21
Airport Challenge Week 1

EviePalaiochorinou:
Met all user story requests, only without a default weather double, so I had to put a stub in every relevant 'it' block.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,16 +1,30 @@
 class Airport
     attr_reader :planes
+    attr_reader :capacity
     def initialize(capacity = 1)
         @planes = []
         @capacity = capacity
     end
 
     def land(plane)
+        raise "Plane already landed once" if @planes.include? plane
+        raise "Cannot land when stormy" if get_weather == "stormy"
         raise "Airport full" if @planes.length == @capacity
         @planes << plane
     end
 
     def takeoff(plane)
+        raise "Plane already took off" if !@planes.include? plane
+        raise "Cannot take-off when stormy" if get_weather == "stormy"
         @planes.delete(plane)
+    end
+
+    def get_weather
+        weather = rand(1..10)
+        if weather >= 9
+            "stormy"
+        else
+            "sunny"
+        end
     end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,16 @@
+class Airport
+    attr_reader :planes
+    def initialize(capacity = 1)
+        @planes = []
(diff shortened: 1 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,30 @@
+class Airport
+    attr_reader :planes
```
msc49, line 2: this works and is good 
another way to write this could be 
attr_reader :planes, :capacity
that way its all on one line

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+        @planes << plane
+    end
+
+    def takeoff(plane)
+        raise "Plane already took off" if !@planes.include? plane
+        raise "Cannot take-off when stormy" if get_weather == "stormy"
+        @planes.delete(plane)
+    end
+
+    def get_weather
```
msc49, line 22: idk why but rubocop does not like methods starting with get

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
+class Airport
+    attr_reader :planes
+    attr_reader :capacity
+    def initialize(capacity = 1)
+        @planes = []
+        @capacity = capacity
+    end
+
+    def land(plane)
+        raise "Plane already landed once" if @planes.include? plane
```
msc49, line 10: this is a nice touch

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+        raise "Cannot land when stormy" if get_weather == "stormy"
+        raise "Airport full" if @planes.length == @capacity
+        @planes << plane
+    end
+
+    def takeoff(plane)
+        raise "Plane already took off" if !@planes.include? plane
```
msc49, line 17: quite look the use of the exclamation mark

spec/airport_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'airport'
+
+describe Airport do
+    
+    context 'land a plane' do
+        it 'checks if there is a land method' do
+            airport = Airport.new
+            expect(airport).to respond_to :land
```
msc49, line 8: instead of writing airport.new, you can use subject which creates an instance. Helps reduce a line

spec/plane_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'plane'
+
+describe Plane do
+
```
msc49, line 4: maybe check if plane is able to to create an instance of itself

PR #2353, opened 2021-11-21
Week 1 Airport Challenge

sabrinalord:


lib/plane.rb
```diff
@@ -0,0 +1,22 @@
+require_relative('./airport')
+
+class Plane
+  def land_at_airport(airport)
+    fail "The airport is full" if airport.full?
+    fail "It's too stormy to land" if airport.local_weather == "Stormy"
+    @in_airport = true
```
Inimesh, line 7: What is the value of @in_airport when Plane.new is run but not yet landed or taken off?

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
...
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @parked_planes = []
+    @local_weather = check_weather
+  end 
+
+  def full?
+    true if @parked_planes.count >= DEFAULT_CAPACITY
```
Inimesh, line 17: If I instantiated an airport as Airport.new(20), would this return the correct value?

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @parked_planes = []
+    @local_weather = check_weather
+  end 
+
+  def full?
```
Inimesh, line 16: Will this method ever return false?

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
+require_relative('./weather')
+
+class Airport
+  attr_accessor :capacity, :parked_planes, :local_weather
```
Inimesh, line 4: Do all of these instance variables need to be accessed (readable and writable) by other classes, or just be readable, or neither?

lib/plane.rb
```diff
@@ -0,0 +1,22 @@
+require_relative('./airport')
+
+class Plane
```
Inimesh, line 3: All round very easy to read and understand methods!

spec/airport_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+require 'weather'
+
+describe Airport do 
+  it 'has a default capacity' do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+
+  it 'keeps track of planes that have landed in the airport' do
+    parked_planes = 0
+    plane = Plane.new
```
Inimesh, line 12: Can see some repeated lines 'plane = Plane.new'. Maybe could do them in a 'before' block to maintain (DRY) (although I also do not know exactly how 'before' blocks work myself)

lib/airport.rb
sabrinalord: Good spot, updated to attr_reader as they need to be read by the plane class but not necessary to be writable

lib/airport.rb
sabrinalord: it will return nil so it would be falsey, but could be clearer

lib/plane.rb
sabrinalord: returns as nil :)

PR #2354, opened 2021-11-21
Week 1 Weekend Project Completed

jessgordon:


spec/airport_spec.rb
```diff
@@ -0,0 +1,103 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+  describe 'landing' do
+
+    it 'lands a plane at an airport' do
+      plane = Plane.new
+      airport = Airport.new("Heathrow")
```
Riky5, line 10: This line is repeated in most of your tests so it could be grouped at the beginning with the previous line. If in certain tests you need a different one you can override it in the single test.

jessgordon: brilliant spot, thank you Ricky

PR #2355, opened 2021-11-21
Finished airport_challenge

AlfonsoGhislieri:


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'weather'
+
+class Airport
+  attr_reader :planes, :weather
+  attr_accessor :capacity
+
```
Hereloss, line 6: Good use of the reader and accessor here to allow them to be used for tests!

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+    raise "Plane has not taken off" if plane_at_airport?(plane)
+    "#{plane} has taken off"
+  end
+  
+  private 
+
+  def full?
+    @planes.count >= @capacity
+  end
+end
```
Hereloss, line 41: Private method works well here!

lib/plane.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def takeoff(airport, weather = airport.weather)
+    raise "Plane already flying" if @flying
+    raise "Plane at different airport" unless airport.plane_at_airport?(self)
+    raise "Weather stormy cannot take off" if weather.stormy?
+    airport.remove_plane(self)
+    @flying = true
+  end
+
```
Hereloss, line 24: Short, succinct and clear for both of these! All errors make sense, and so does the order

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
+  def stormy?
+    rand(1..10).even?
```
Hereloss, line 3: The user stories did say the stormy should be rare - I think 50% is a bit high for this, maybe worth lowering this to 1/10 or 1/20?

spec/airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      subject.capacity.times { subject.add_plane(plane) }
+      expect { subject.add_plane(plane) }.to raise_error "Airport full"
+    end
+
+    it "Plane can land if airport not full" do
+      capacity = subject.capacity - 1
+      capacity.times { subject.add_plane(plane) } 
+      expect { subject.add_plane(plane) }.to_not raise_error 
+    end
+
```
Hereloss, line 23: Re-defining capacity here does make this test slightly harder to read - may be worth combining lines 19 and 20

spec/plane_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+  		end
+
+ 		 it "Cannot takeoff from airport if stored at different airport" do
+  			 airport = double("Airport")
+  			 allow(airport).to receive_messages(
+   				:planes => [],
+   				:plane_at_airport? => false,
+   				:weather => nil)
+  			 expect { subject.takeoff(airport) }.to raise_error "Plane at different airport"
+  		end
```
Hereloss, line 73: I feel a couple of take-off tests might be missing here - such just checking if a landed plane is able to takeoff? Not fully required (as is implied by others), but may be helpful to add

spec/plane_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+  			 it "Can takeoff from airport if weather is not stormy" do
+   				 weather = double(:Weather, :stormy? => false)
+   				 expect { subject.takeoff(@airport, weather) }.to_not raise_error 
+   			end
+  			
+  			 it "Cannot takeoff from airport if weather is stormy" do
+   				 weather = double(:Weather, :stormy? => true)
+   				 expect { subject.takeoff(@airport, weather) }.to raise_error "Weather stormy cannot take off"
+   			end
+  		end
```
Hereloss, line 109: Good use of doubles - works well!

spec/weather_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  it "Checks if responds to stormy method" do
+    expect(subject).to respond_to(:stormy?)
+  end
+
+  it "will return both true and false when running #stormy?" do
+    results = []
+    100.times { results << subject.stormy? }
+    expect(results.include?(true & false)).to eq true
+  end
```
Hereloss, line 13: Good way to check for randomness - a bigger number than 100 might be nice to add (just to be safe!) but should work either way

PR #2363, opened 2021-11-22
Nov 21 - Chris Thomas - Weekend Challenge

Hereloss:


lib/controller.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def pick_sky
+    return @planes_in_sky[0]
+  end
+
+  def pick_airport
+    return @airport.storage[0]
+  end
+
+end
```
AlfonsoGhislieri, line 80: How come you chose to use a controller class to control planes? Could the challenge maybe have been done with less?

AlfonsoGhislieri, line 80: Could some methods possibly be private, do the other classes need access to all these methods?

lib/plane.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def take_off
+    unless @in_air == false
+      raise "The plane is already in the air!"
+    end
+    unless @in_airport == true
+      raise "You can't take off from nowhere! You must have crashed!"
+    end
+    @in_air = true
+    @airport_at = "None"
```
AlfonsoGhislieri, line 38: could have maybe added an if, elsif, else block for better readability

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'airport'
+
+describe Airport do
+
```
AlfonsoGhislieri, line 4: could be using using a before condition here to create an airport since you do it on every test
```
Before do
  @airport = Airport.new
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  it "A airport knows it's own name, and can change this" do
+    airport = Airport.new
+    expect(airport.name).to eq("Heathrow")
+    expect(Airport.new(20,"JFK").name).to eq("JFK")
+  end
+
+  it "A plane can land at the airport" do
+    airport = Airport.new
+    plane = Plane.new
```
AlfonsoGhislieri, line 18: Could you have made a double with method stubs instead of having to use the an actual instance of plane in these tests?

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    airport = Airport.new
+    expect(airport.capacity).to eq 20
+  end
+
+  it "The airport's capacity can be changed upon creation" do
+    airport = Airport.new(30)
+    expect(airport.capacity).to eq 30
+  end
+
+end
```
AlfonsoGhislieri, line 55: Could take advantage of context blocks to break up tests for readability

spec/controller_spec.rb
```diff
@@ -0,0 +1,152 @@
+require "controller"
+
+describe Controller do
+
```
AlfonsoGhislieri, line 4: Could also use a before block for controller here

spec/controller_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+    controller.create_new_airport("JFK")
+    controller.change_airport_using("JFK")
+    plane3 = controller.pick_sky
+    controller.request_landing(plane3)
+    controller.stub(:weather?).and_return("Stormy")
+    expect(lambda { controller.request_landing(plane) }).to raise_error("It's too stormy to land!")
+    expect(lambda { controller.request_take_off(plane3) }).to raise_error("It's too stormy to take off!")
+  end
+
+end
```
AlfonsoGhislieri, line 152: Adding context blocks could be beneficial for easier readability and also take advantage of before blocks

spec/plane_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'plane'
+
+describe Plane do
+
```
AlfonsoGhislieri, line 4: use before block for plane

spec/weather_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'weather'
+
+describe Weather do
+
```
AlfonsoGhislieri, line 4: before block for controller

AlfonsoGhislieri: Also double of controller?

PR #2364, opened 2021-11-22
Partial solution of Airport Challenge

TomasGarciaDev:
I got stuck in user story 6, couldn't find a way to change the weather value stormy from true to false or vice versa.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,13 @@
+require_relative 'plane'
+
+class Airport
+  
+  def initialize
+    @airplanes = []
+  end
+
+  def land_plane(plane)
+    @airplanes << plane
+  end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,17 +1,25 @@
 require_relative 'plane'
 
 class Airport
-  attr_reader :airplanes
+  attr_reader :airplanes, :capacity
+
+  MAXIMUM_CAPACITY = 20
 
   def initialize
     @airplanes = []
+    @capacity = MAXIMUM_CAPACITY
   end
 
   def land_plane(plane)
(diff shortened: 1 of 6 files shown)


spec/airport_spec.rb
```diff
@@ -26,4 +26,8 @@
     Airport::MAXIMUM_CAPACITY.times { subject.land_plane(Plane.new) }
     expect { subject.land_plane(Plane.new) }.to raise_error("Airport full")
   end
+  
+  it "should have a default capacity of 20 when no armgument is given" do
```
giorgigutsaevi, line 30: A minor typo in  it definition `armgument` -> `argument`

lib/airport.rb
```diff
@@ -1,13 +1,17 @@
...
+
   attr_reader :airplanes, :capacity
+  attr_accessor :stormy
 
   MAXIMUM_CAPACITY = 20
 
   def initialize(capacity = MAXIMUM_CAPACITY)
     @airplanes = []
     @capacity = capacity
+    @condition = false
```
giorgigutsaevi, line 14: What is @condition instance variable's main purpose?

lib/weather.rb
```diff
@@ -0,0 +1,8 @@
+class Weather
+
+  attr_reader :stormy
+
+  def stormy?
+    stormy
```
giorgigutsaevi, line 6: Suggestion to change stormy to @stormy in the #stormy? method, so attr_reader can read it.

giorgigutsaevi (commented):
The code looks really elegant. There is nothing there, that should not be there. Very clean implementation and really enjoy the 1-liners as your unit-tests in spec.rb files! Keep up the good work!

PR #2365, opened 2021-11-22
Airport Challenge - Webb

christopherxwebb:
The user stories are complete but there are still a few edge cases that need ironing out (planes can currently takeoff an infinite amount of times, etc.).

The test coverage is >95% but there are certain tests (weather tests) that remain incomplete.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,8 @@
+require_relative 'plane'
+
+class Airport
+    
+    def land(plane)
+    end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,8 +1,35 @@
 require_relative 'plane'
 
 class Airport
-    
-    def land(plane)
+
+attr_reader :planes
+CAPACITY = 1
+
+  def initialize (capacity = CAPACITY)
+    @planes = []
+    @planes_taken_off = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    if @planes.count == 0
+        @planes.push(plane)
+    else fail "Airport full"
     end
+  end
+
+  def takeoff(plane)
+    @planes.pop
+    puts "#{plane} has now taken off! There are currently #{@planes.count} planes in the airport."
+    @planes_taken_off << plane
+  end
+
+  #def full?
+    #if @planes.count >= @capacity
+      #true
+    #else
+      #false
+    #end
+  #end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -3,12 +3,12 @@ require_relative 'plane'
 class Airport
 
 attr_reader :planes
-CAPACITY = 1
+DEFAULT_CAPACITY = 1
 
-  def initialize (capacity = CAPACITY)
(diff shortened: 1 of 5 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    airport.takeoff(plane)
+    expect(airport).to respond_to :takeoff
+  end
+
+  #it "raises error if pilot attempts to land and airport is full" do
+    #@planes == 1
+   # expect { Airport.new.land Plane.new }.to raise_error
+  #end
+
+  it "raise an error when you try to land a plane in a full airport" do
```
chris-clement, line 34: Well written test. I can see exactly what it is setting out to achieve.

lib/weather.rb
```diff
@@ -0,0 +1,16 @@
+require_relative 'plane'
+require_relative 'airport'
```
chris-clement, line 2: Is this require_relative necessary? Is anything from plane or airport being used in this weather file. I think you maybe only need to require_relative in the airport program.

lib/weather.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class Weather
+
+def stormy?
```
chris-clement, line 6: I'd suggest adding indentation as it is within your Weather class.

lib/weather.rb
```diff
@@ -0,0 +1,16 @@
+require_relative 'plane'
+require_relative 'airport'
+
+class Weather
```
chris-clement, line 4: I like that you made Weather a separate class.

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+    @capacity = desired_capacity
+    @weather = weather
+  end
+
+  def land(plane)
+    if @planes.count == 0 && @weather.stormy? == "false"
+        @planes.push(plane)
+    elsif @planes.count == DEFAULT_CAPACITY
+        fail "Airport full"
+    else fail "Weather is too stormy"
```
chris-clement, line 22: Should 'fail "Weather is too stormy"' be under the else statement? This reads more like a if or elsif condition

spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    airport = Airport.new
+    Airport::DEFAULT_CAPACITY.times { airport.land Plane.new }
```
chris-clement, line 36: good use of a constant here rather than a magic number.

chris-clement (commented):
Nice work! Lots of good code here, just a few suggestions which I have added comments for.

lib/weather.rb
christopherxwebb: A really good point. I think I did something which I'll try to avoid next time: i.e. just blanket put in 'require_relative' with the assumption that in the future these files will be interacting with each other. But they never do so is actually superfluous. Thanks!

lib/airport.rb
christopherxwebb: Yes, good point.

christopherxwebb:
Thanks, Chris! Really helpful comments, which I'll be sure to incorporate on my next commit.

PR #2366, opened 2021-11-22
Riky Moroni - Airport Challenge

Riky5:
Here's my solution to the Airport Challenge.
(Not exactly finished as I wrote in MYREADME.md.)
I still have a lot to improve. But good exercise to get used to TDD process.


spec/airport_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      it "should use the passed argument" do
+        expect(subjectNoDefaultCapacity.capacity).to eq(50)
+      end
+    end
+  end
+
+  describe "#take_off" do
+
+    it "should confirm the plane is no longer at the airport" do
+      allow(subject).to receive(:weather_stormy?).and_return(false)
```
jessgordon, line 30: I think this can also be written like:

allow(subject).to receive(:weather_stormy?) { false }

spec/airport_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      allow(subject).to receive(:include?).and_return(false)
```
jessgordon, line 31: Would this mean that the test below will always return false? i.e plane might be in the airport but because it's been told to return false to the method 'include?' it will return false whether the plane is in the airport or not? Sorry if I've misunderstood! :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    it "should not take off if weather is stormy" do
+      allow(subject).to receive(:weather_stormy?).and_return(true)
+      expect { subject.take_off(plane) }.to raise_error "Stormy weather: not possible to take off"
+    end
+
+    it "should take off only from the airport they are at" do
+      allow(subject).to receive(:weather_stormy?).and_return(false)
+      allow(subject).to receive(:include?).and_return(true)
+      expect(subject).to include plane
```
jessgordon, line 43: please could you talk me through how this test works? :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      allow(subject).to receive(:weather_stormy?).and_return(false)
+      allow(subject).to receive(:include?).and_return(false)
+      expect { subject.take_off(plane) }.to raise_error "Plane is not at the airport"
+    end
+
+    context "#delete" do
+      it "should delete a plane from array planes" do
+        allow(subject).to receive(:weather_stormy?).and_return(false)
+        allow(subject).to receive(:include?).and_return(true)
+        expect(subject.planes.find { |p| p == plane }).to eq nil
```
jessgordon, line 56: if possible please explain this line to me, another line I haven't used/ seen before :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  describe "#land" do
+    it "should land a plane successfully" do
+      allow(subject).to receive(:land).and_return(plane)
+      expect(subject.land).to eq plane
+    end
+
+    it "shouldn't land if airport has full capacity " do
+      allow(subject).to receive(:weather_stormy?).and_return(false)
+      allow(subject).to receive(:full_capacity?).and_return(true)
```
jessgordon, line 69: this is so smart! saves the need to fill the airport to capacity manually! nice!

spec/airport_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+require 'weather'
+
+describe Airport do
+  let(:subject) { described_class.new }
+  let(:plane) { double(:plane) }
+
+  it { is_expected.to respond_to :land }
+  it { is_expected.to respond_to :take_off }
+
+  context "#initialize" do
```
jessgordon, line 12: good use of context blocks to keep things organised :)

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  attr_accessor :planes
+
+  def initialize(capacity = 5)
+    @planes = []
+    @capacity = capacity
+    @weather = Weather.new.now?
+  end
+
+  def land(plane)
+    raise "Landing not possible, airport is full" if full_capacity?
```
jessgordon, line 16: great use of one line if statements throughout - keeps the code nice and easy to read and understand

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class Airport
+
+  attr_reader :capacity
+  attr_accessor :planes
+
+  def initialize(capacity = 5)
+    @planes = []
+    @capacity = capacity
+    @weather = Weather.new.now?
```
jessgordon, line 12: does this mean that the weather remains the same once an airport has been created? If so, perhaps consider moving the weather assignment to landing/take off so that it changes over time

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def weather_stormy?
+    @weather == :stormy
+  end
+
+  def full_capacity?
+    @planes.count >= @capacity
+  end
+
+end
```
jessgordon, line 36: really good length of methods, very easy to read and understand. SRP seems to be met here!

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
+  def now?
+    rand(1..4) == 3 ? :stormy : :sunny
```
jessgordon, line 3: nice refactoring

spec/weather_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+      expect(subject.now?).to eq :stormy
+    end
+
+    it "calls rand and returns sunny" do
+      allow(subject).to receive(:rand).and_return(2)
+      expect(subject.now?).to eq :sunny
+    end
+  end
+
+end
```
jessgordon, line 18: good test coverage for all files/classes

MYREADME.md
```diff
@@ -0,0 +1,241 @@
...
+
+User stories complete but lots to refactor and fix and add.
+Rspec: 10 examples, 0 failed 
+Changed methods to land and take_off.
+
+---
+### to be continued..
+I added some checks but more is needed to be considered completed.
+Lots of repeatition in tests.
+And I know some methods are quite long. I have to work on that and adjust tests in rspec accordingly...
```
jessgordon, line 241: what can I say... WOW. 
Very detailed, you've inspired me! 

Good you have irb demos, documented the rspec results and test coverage. This must help with learning/ reflection too.
Fantastic :)

Also.- I don't think your methods are too long as you mentioned. You could move the raise error lines to separate methods as e.g. landing checks and take off checks, but I don't know if you really need to as your methods are clear and only a few lines long anyway! Refactoring this may make it more confusing. I think format and readability of code was great.

jessgordon (commented):
Great work Ricky - I think your code is well formatted and easy to read! 

I read this over the weekend and found it really helpful in terms of formatting the spec files :) https://thoughtbot.com/blog/four-phase-test

spec/airport_spec.rb
```diff
@@ -0,0 +1,104 @@
+require 'airport'
+require 'plane'
+require 'weather'
+
+describe Airport do
+  let(:subject) { described_class.new }
```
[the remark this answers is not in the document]
Riky5: In this case the subject is Airport.new. Usually subject is the class you're testing. To avoid having to repeat in each test "subject = Airport.new" . This line sets it for the whole rspec , in this case airport_spec.rb. 
If in some specific tests you need to change the subject. You can override it in the single test by setting let(subject or whatever you want to call it) { Airport.new(50) } for example in the specific test/context. If you need a better explanation let me know.

lib/airport.rb
Riky5: Unfortunately yes, I thought about that when I was almost done lol . Great spot and advice :)

spec/airport_spec.rb
Riky5: I searched online how to test when you delete something from an array. And I found that line. Which it basically means --> Check the array(planes) in the subject(airport) and find that plane. The result should be nil when you deleted that plane.

spec/airport_spec.rb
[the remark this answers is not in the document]
jessgordon: ah amazing! thanks so much,  that makes sense and deffo would tidy up code! 🥳

spec/airport_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      end
+    end
+  end
+
+  describe "#take_off" do
+
+    it "should confirm the plane is no longer at the airport" do
+      allow(subject).to receive(:weather_stormy?).and_return(false)
+      allow(subject).to receive(:include?).and_return(false)
+      expect(subject).not_to include(plane)
```
jessgordon, line 32: is this stubbing the method so it's testing the stub rather than your code? 
(discussed in person)

PR #2367, opened 2021-11-22
Submitting solution to airport challenge

kim-morgan:
My solution to airport challenge

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,18 +2,25 @@ require_relative "plane"
 
 class Airport
 
-  attr_reader :plane
+  attr_reader :planes, :capacity
+
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
 
   def land(plane)
     fail "Airport full" if full?
-    @plane = plane
+    @planes << plane
   end
 
-  def takeoff
-    @plane = nil
+  def takeoff(plane)
+    @planes.select! {|other_planes| other_planes != plane}
   end
 
   def full?
-    @plane
+    @planes.length == capacity
   end
 end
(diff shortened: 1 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  attr_reader :planes, :capacity, :stormy
+
+  DEFAULT_CAPACITY = 10
+  LIKELIHOOD_OF_STORMINESS = 0.2
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+    rand < LIKELIHOOD_OF_STORMINESS ? @stormy = true : @stormy = false
```
Swa567, line 13: smart 👍
could be:
@stormy = rand < LIKELIHOOD_OF_STORMINESS

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+    takeoff_check(plane)
+    plane.flying = true
+    @planes.reject! { |planes| planes == plane }
+  end
+
+  def landing_check(plane)
+    fail "Airport full" if full?
+    fail "Cannot land at stormy airport" if @stormy
+    fail "Only planes can land at an airport" if plane.class != Plane
+    fail "Only flying plane can land" if plane.flying == false
```
Swa567, line 32: smart 👍
could be:
fail "Only flying plane can land"  unless plane.flying

lib/airport.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def takeoff_check(plane)
+    fail "Cannot takeoff from stormy airport" if @stormy
+    fail "Specified plane not at this airport" if at_airport?(plane) == false
```
Swa567, line 37: smart 👍
could be:
fail "Specified plane not at this airport" unless at_airport?(plane)

lib/airport.rb
kim-morgan: Thanks, good idea! Will implement.

PR #2370, opened 2021-11-22
Airport- Challenge

Swa567:
fail to created stub to implement user story 5 and 6


lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'plane'
+ 
+class Airport
+  CAPACITY = 14
```
kim-morgan, line 4: Is there a reason why it is 14?

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def land(plane)
+    fail 'Airport is full' if full?
+    # fail 'Weather is stormy! do not land' unless safe?
+    @hangar << plane
+    plane.landed
+  end
+
+  def full?
+    @hangar.size >= @capacity
```
kim-morgan, line 21: Nice and clear!

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  attr_reader :capacity, :hangar, :weather
+
+  def initialize(capacity = CAPACITY)
+    @capacity = capacity
+    @hangar = []
+    #@weather = rand(10)
+  end
+
+  def land(plane)
+    fail 'Airport is full' if full?
```
kim-morgan, line 14: Could the fail conditions maybe be moved to another method? (SRP)

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+    @hangar.pop
+    plane.taken_off
+  end
+
+  def in_hangar?(plane)
+    @hangar.include?(plane)
+  end
+
+  # def safe?
+  #   raise 'Weather is stormy, please stand by' unless @weather < 7
```
kim-morgan, line 36: Relates to the comment above, but maybe refactoring the @weather condition to be more descriptive would help it make more sense what 7 means

lib/plane.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'airport'
+
+class Plane
+  attr_reader :in_airport
+# assume plane is create at the airport
+  def initialize
+    @in_airport = false
+  end
+
+  def landed
```
kim-morgan, line 10: I wonder if landed and taken_off could be combined into one method? Or maybe they could be removed and the Airport takeoff and land methods could directly edit in_airport?

spec/airport_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'airport'
+
+describe Airport do
+  # let variable do not set the anonymous subject or allow expectations to be called on it implicitly
+
+  describe "#initialize" do
+    it "can be initialized with a default capacity" do
+      airport = Airport.new
+      expect(airport).to be_an_instance_of(Airport)
+      allow(airport).to receive_messages([])
```
kim-morgan, line 10: What is this doing? (Genuine question)

spec/airport_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      expect(airport.capacity).to eq 14
+    end
+
+    it "can be initialized with a custom capacity" do
+      expect(Airport.new(11).capacity).to eq 11
+    end
+  end
+
+  context "landing a plane"
```
kim-morgan, line 19: Nice use of contexts

spec/airport_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      #allow(airport).to receive_messages(plane)
+      # populate that array @hangar
+      airport = Airport.new
+      Airport::CAPACITY.times { airport.land(Plane.new) }
+      expect { airport.land(plane) }.to raise_error 'Airport is full'
+    end
+    
+    # it "raises an error when the weather is not safe for landing" do
+    #   plane = double(:plane, landed: true, taken_off: false)
+    #   airport = instance_double(Airport, :initialize, :safe? => false, :full? => false, :in_hangar? => false, land: plane)
```
kim-morgan, line 34: Can use srand() to introduce deterministic behavior into rand() method

README.md
```diff
@@ -11,79 +12,367 @@ Airport Challenge
...
 
-1. Fork this repo, and clone to your local machine
-2. Run the command `gem install bundler` (if you don't have bundler already)
-3. When the installation completes, run `bundle`
-4. Complete the following task:
+---
 
-Task
------
+We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off. Here are the user stories that we worked out in collaboration with the client:
```
kim-morgan, line 21: Very thorough readme document!

README.md
```diff
@@ -11,79 +12,367 @@ Airport Challenge
...
-Please create separate files for every class, module and test suite.
+| #Object | #Message |
+| :-----: | :------: |
+|         |          |
+|         |          |
 
-In code review we'll be hoping to see:
+# **_ Notes and Thoughts _**
+
+I was really unsure on how to approach the user stories: does the plane have a method to land(airport) (like land at the airport) or is it the airport has a method to land planes...
```
kim-morgan, line 106: I was also unsure about this, but I think it's good to put land and takeoff on the Airport class (like with boris bikes) since the main thing happening is that we are adding or removing planes to the @hanger attribute of the Airport class. (Obviously I might be wrong but that was my thought!)

README.md
s-dousse: 👍 thank you !

lib/airport.rb
s-dousse: no reason, in particular, I just pick a number 🙈

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'plane'
+ 
+class Airport
+  CAPACITY = 14
+  attr_reader :capacity, :hangar, :weather
+
+  def initialize(capacity = CAPACITY)
+    @capacity = capacity
+    @hangar = []
+    #@weather = rand(10)
```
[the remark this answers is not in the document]
s-dousse: Oh, very interesting! thank you! 🙏🏻
I have changed this: now have the @stromy attribute for the Weather class, which is then used by the airport instance to see if it is safe to land/take off.

lib/airport.rb
s-dousse: 👍

lib/airport.rb
s-dousse: yes, I changed the Weather class to just have the attribute @stormy (defines the risk of storm)

lib/plane.rb
s-dousse: yes, after going through the users' stories again, I removed it as I already had an in_hangar method to define the plane's location

spec/airport_spec.rb
s-dousse: This was a useless test; I have now deleted it 👍

spec/airport_spec.rb
s-dousse: Thank you for showing me how to use srand() 🙏🏻

README.md
s-dousse: Yes, you must be right! this makes more sense to me

lib/airport.rb
s-dousse: Might I need to move the in _hangar? method to the plane so the plane can be responsible for knowing its location and in someway keeping the airport informed

PR #2371, opened 2021-11-22
Airport_Challenge

KKassim:
Finished but unsure about test coverage?!

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -23,7 +23,7 @@ class Airport
     end
 
     def weather
-        the_weather = Kernel.rand(1..40)
+        the_weather = Nums.rand(1..40)
         if the_weather == 1
             return "Stormy"
         else
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,6 @@
+class Airport
+    attr_reader :planes
+    def initialize
+      @planes = []
+      @weather = check_weather
+    end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,6 +1,33 @@
 class Airport
     attr_reader :planes
     def initialize
-      @planes = []
-      @weather = check_weather
+      @plane_list = []
+      @weather = weather
     end
+
+    def land_plane(plane)
+        @plane_list << plane
+    end
+    
+    def take_off
+        landed_planes = @plane_list.length
+        current_weather = weather
+        if current_weather == "Stormy"
+            raise "Sorry, planes cannot take off in stormy weather."
+        else
+            @plane_list.pop
+        end
+        if landed_planes.-(1) == @plane_list.length
+            return "Flight has departed."
+        end
+    end
+
+    def weather
+        the_weather = Kernel.rand(1..40)
+        if the_weather == 1
+            return "Stormy"
+        else
+            return "Sunny"
+        end
+    end
+end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,4 @@
+require './lib/airport.rb'
+
+class Plane
+end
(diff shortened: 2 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def weather
+        the_weather = Nums.rand(1..40)
+        if the_weather == 1
+            return "Stormy"
+        else
+            return "Sunny"
+        end
+    end
+end
```
IbrahimAC, line 33: indentations are 4 not 2

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+    def take_off
+        landed_planes = @plane_list.length
+        current_weather = weather
+        if current_weather == "Stormy"
+            raise "Sorry, planes cannot take off in stormy weather."
+        else
+            @plane_list.pop
+        end
+        if landed_planes.-(1) == @plane_list.length
+            return "Flight has departed."
```
IbrahimAC, line 21: no return needed, just the string alone will be returned(99%sure on that)

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+        end
+    end
+
+    def weather
+        the_weather = Nums.rand(1..40)
+        if the_weather == 1
+            return "Stormy"
```
IbrahimAC, line 28: same again no return needed, just the strings alone is fine syntax and code functionality wise.

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+        else
+            return "Sunny"
```
IbrahimAC, line 30: indentations 4 not 2, return syntax not needed

lib/plane.rb
```diff
@@ -0,0 +1,4 @@
+require './lib/airport.rb'
```
IbrahimAC, line 1: you can just write require 'file_name' whatever is the file name just write that instead and it will work. In this case it would be 
require 'airport'

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    def land_plane(plane)
+        @plane_list << plane
+    end
+    
+    def take_off
+        landed_planes = @plane_list.length
+        current_weather = weather
+        if current_weather == "Stormy"
```
IbrahimAC, line 15: 4 not 2 indentations

spec/plane_spec.rb
```diff
@@ -0,0 +1,9 @@
+require './lib/airport.rb'
```
IbrahimAC, line 1: could use require 'airport' instead

spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
+require './lib/airport.rb'
+require './lib/plane.rb'
+
+describe Airport do
+  it 'creates a new instance of the Airport class' do
+    airport = Airport.new
```
IbrahimAC, line 6: no need to manually create instances , can use 'subject' as the alternative.
e.g. expect(subject).to be_a(Airport)

Instead of
expect(airport).to be_a(Airport)

spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    expect(airport).to be_a(Airport)
+  end
+
+  it 'calls planes to land at an airport' do
+    airport = Airport.new
+    plane = Plane.new
+    airport.land_plane(plane)
+    expect(airport.plane_list.length).to eq 1
+
```
IbrahimAC, line 15: needless line (no intent to sound so harsh)

IbrahimAC (commented):
To my knowledge most if not all of the code is perfectly functional. The changes I suggested/ things I came across were mostly syntax things/ saving times by writing shorter versions of the same code e.g. not using 'return' or not writing the whole path in the 'require' statement. These things are easily fixable and will be non issues with more practice

PR #2372, opened 2021-12-12
Airport Challenge Submission

cris-ch:


spec/feature_spec.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/plane'
```
BenCoyne, line 1: Having a separate feature_spec is a great way to add more context to the project and help efficiency when building. Well done!

BenCoyne (commented):
Your Commit Messages are detailed and clear. Easy to understand - very good!

BenCoyne (commented):
Have you considered (in airport_spec.rb) using a 'subject(:airport) { described_class.new }'? This would allow you to refer to 'airport' instead of 'subject'. However subject is perfectly acceptable!

PR #2373, opened 2021-12-12
airport-challenge

mikitaro-27:


spec/airport_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+  it { is_expected.to respond_to(:check_weather) }
+  
+  it 'has a capacity attribute, set to 3 by default' do
+    expect(airport.capacity).to eq(3)
+  end
+
+  it 'has capacity 2 when specified as parameter' do
+    airport = Airport.new(2)
+    expect(airport.capacity).to eq(2)
+  end
```
izaowl, line 19: Is it possible that this test is redundant when you also have test above?

ideas.md
```diff
@@ -0,0 +1,32 @@
+# describe Weather do
```
izaowl, line 1: Maybe this file is redundant therefore should not be included in this push?

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
...
+class Airport
+  attr_reader :capacity
+  attr_reader :hangar
+
+  def initialize(capacity = 3)
+    @capacity = capacity
+    @hangar = []
+  end
+
+  def land_plane(plane)
```
izaowl, line 12: How do you keep track of the number of planes which your hangar stores at any time?

lib/airport.rb
```diff
@@ -0,0 +1,32 @@
+require './lib/plane'
+
+class Airport
+  attr_reader :capacity
+  attr_reader :hangar
+
+  def initialize(capacity = 3)
```
izaowl, line 7: Capacity should perhaps be a static value set on the top of this class as customer wanted to be able to change it on the fly. If it is coded in the constructor is set in stone forever.

izaowl (commented):
Well done, ver ycler code and easy to understand.

PR #2377, opened 2021-12-12
Airport Challenge

BenCoyne:


lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
```
cris-ch, line 1: Wouldn't have been better to put this as a method in the Airport class, instead of as class itself?? It's more a question than a review, I'm just not sure what's the best practice...

cris-ch (commented):
Great code, with my limited knowledge I can't find anything that could have been done better.

PR #2379, opened 2021-12-13
Airport Challenge

ravensears:
Not fully finished, but passing current tests.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,5 @@
+require 'plane.rb'
+
+class Airport
+    
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,5 +1,30 @@
-require 'plane.rb'
+require 'plane'
+require 'weather'
 
 class Airport
+    DEFAULT_CAPACITY = 30
+
+    attr_accessor :airport, :capacity, :weather, :planes
     
+    def initialize (capacity = DEFAULT_CAPACITY, weather = Weather.new)
+        @airport = []
+        @capacity = capacity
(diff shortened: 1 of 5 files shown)


OrganicAC (commented):
I like that you specified weather = Weather.new in the code file!

lib/airport.rb
```diff
@@ -1,5 +1,30 @@
...
+    end
+
+    def land(plane)
+        raise 'Airport is full' if full
+        raise 'More Bad weather' if weather.bad?
+        airport << plane
+    end
+
+    def full
+        airport.length >= capacity
```
OrganicAC, line 28: This is a really simple and efficient way to check for this!

lib/airport.rb
```diff
@@ -1,5 +1,30 @@
-require 'plane.rb'
+require 'plane'
+require 'weather'
 
 class Airport
+    DEFAULT_CAPACITY = 30
+
+    attr_accessor :airport, :capacity, :weather, :planes
```
OrganicAC, line 7: Would being able to both read and write all four of these potentially cause an issue? Could some of them be read/write only?

OrganicAC (commented):
You have really descriptive RSpec tests, and the information in quotations is simple but lets you know instantly what it does without being confusing

spec/weather_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'weather'
```
OrganicAC, line 1: I really like this test. It's simple to read and easy to understand and well laid out

PR #2380, opened 2021-12-13
weekend1 airport challenge

robinucar:


lib/airport.rb
```diff
@@ -0,0 +1,53 @@
+require './lib/plane'
+# Airport class
+class Airport
+  DEFAULT_CAPACITY = 5
+  attr_accessor :planes, :capacity, :error_messages
+  def initialize(capacity = DEFAULT_CAPACITY)
+    # create a planes array
+    @planes = []
+    @capacity = capacity # Default is 5 but now it can be updated
+    @error_messages = [
```
eoinmakers, line 10: I don't think this is the best way to store these error messages - I can see why collating them in this way would feel like a + for readability, but it's slightly impractical for a dev that needs to make a change to one of them to first find where the error message is / find out which element in the array needs to be changed.

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+# Describe Airport class land method
+context Airport do
+  # subject = Airport
+  # land method
+  let(:plane) { Plane.new } # everytime when we call plane a new plane will be created.
+  describe '#land' do
+    it 'should land a plane' do
+      allow(subject).to receive(:stormy?) { "Sunny" }
+      subject.land(plane)
+      expect(subject.planes).to include(plane)
```
eoinmakers, line 12: Good use of stubs!

lib/airport.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def land(plane)
+    # error when airport is full
+    fail @error_messages[0] if full?
+    # error when weather is stormy
+    fail @error_messages[1] if stormy? == 'Stormy'
+    # Same plane should not land
+    fail @error_messages[2] if @planes.include?(plane)
+    # and plane will be pushed to planes array
```
eoinmakers, line 26: If the error messages were written directly in the method, it would make this method more readable, and you could avoid the use of comments. Generally speaking we try to write our code in a way that is readable enough that we don't also have to use comments.

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  
+  end
+  # take_off method
+  describe '#take_off' do
+    it 'should take off a plane' do
+      allow(subject).to receive(:stormy?) { "Sunny" }
+      subject.land(plane)
+      subject.take_off(plane)
+      expect(subject.planes).not_to include(plane)
+    end
```
eoinmakers, line 44: It's good to leave a newline under each test for readability

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    it 'should not let take off when weather is stormy' do
+      allow(subject).to receive(:stormy?) { 'Stormy' }
+      expect { subject.take_off(plane) }.to raise_error(subject.error_messages[3])
```
eoinmakers, line 47: I wouldn't use the real object for the 'expected' behaviour that you're comparing the 'actual' behaviour with. Thus I would write: 

expect { subject.take_off(plane) }.to raise_error("I would write the hardcoded error message here")

This would make your test a little more reliable.

eoinmakers: Otherwise, both sides of the test are using the real object, and you don't have a grounded / hardcoded expectation to compare the actual behaviour with.

eoinmakers (commented):
I thought this was a well written Airport Challenge, well done. You successfully implemented all of the user stories and demonstrated last week's learning objectives well. Nice work! 

My feedback is mostly around improving the readability for other devs, which saves time and makes code more easily digestible for those devs you're working with. 

A challenge I would pose to you is to explore the responsibilities in each class - if you do some work this week on domain model diagramming and thinking about which objects you should have and why, do you spot any behaviour in here that might be beneficial to draw out?

PR #2381, opened 2021-12-13
Victoria Black airport challenge

icvotria:


lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def land(plane)
+    raise 'No clearance to land: Airport is full' if full?
+    raise 'No clearance to land: Stormy weather' if stormy?
+    raise 'No clearance to land: Plane isn\'t in flight' unless plane.flying?
+    raise 'No clearance to land: Object is not a plane' unless plane.class == Plane
+    plane.land
+    hangar << plane
+  end
+end
```
JohnForster, line 44: This class is well written, but currently all the methods are public. It would be worth thinking about which of these should be part of the public interface of the class, and which should be set to be private. This includes the attr_readers at the top.

lib/weather.rb
```diff
@@ -0,0 +1,22 @@
...
+    weather_array = []
+    
+    weather_weights.each do |weather, weight|
+      weight.times { weather_array.push(weather) }
+    end
+    weather_array
+  end
+end
+
+module Weather
```
JohnForster, line 13: Good use of modules 👍

spec/airport_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+    plane1 = Plane.new
+    plane2 = Plane.new
+    plane3 = Plane.new
+    clear_airport_new = Airport.new(Airport::DEFAULT_CAPACITY, :clear)
+    
+    it 'allows a number of planes to land' do
+      clear_airport_new.land(plane1)
+      clear_airport_new.land(plane2)
+      clear_airport_new.land(plane3)
+      expect(clear_airport_new.hangar).to include(plane1).and include(plane2).and include(plane3)
```
JohnForster, line 20: This line here is testing state, not behaviour. When writing unit tests, you only want to be testing the _inputs_ and _outputs_ of the class, not the internal state. This will become clearer once you define the public interface. Your tests should only ever use the public interface of the class, and you should never change the public interface for your tests.

[The piggy bank exercise](https://github.com/makersacademy/skills-workshops/blob/main/test_driven_development/TDD_process_piggy_bank.md) is a good example of this. You can't see the contents of a piggy bank without breaking it, so you shouldn't write a test that peeks inside it.

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def stormy?
+    weather == :stormy
+  end
+
+  def land(plane)
+    raise 'No clearance to land: Airport is full' if full?
+    raise 'No clearance to land: Stormy weather' if stormy?
+    raise 'No clearance to land: Plane isn\'t in flight' unless plane.flying?
+    raise 'No clearance to land: Object is not a plane' unless plane.class == Plane
```
JohnForster, line 40: We don't want to be directly checking the class of the plane object. This is because of an idea called "Duck Typing" where if an object walks like a duck and quacks like a duck, we should treat it as if it is a duck.

In this example, we could create a Helicopter class that can still fly and takeoff and land, and so our Airport should be flexible enough to handle it, a line like this prevents us from doing that.

spec/weather_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'weather'
+
+describe Weather do
+  it 'returns clear when weather is clear' do
+    expect(Weather.weather_forecast).to eq(:clear).or eq(:stormy)
+  end
```
JohnForster, line 6: This test is good, but it doesn't check whether both `:clear` _and_ `:stormy` are possible values, only that at least one of them is. Have a look into using `srand` to take the random element out of tests.

JohnForster (commented):
Overall some great work here, you have demonstrated a good understanding of a lot of the concepts so far. I'd recommend that you have another look at testing _behaviour_ over _state_.

Keep it up!

PR #2382, opened 2021-12-13
Alex Jamshidi - Project Complete

Alex-Jamshidi:


smm3000 (commented):
Great to see that you updated the README!

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'plane'
```
smm3000, line 1: This has a clear layout of the different methods and is very readable. Nice use of '.pop'!

spec/airport_spec.rb
```diff
@@ -0,0 +1,113 @@
+require 'airport'
```
smm3000, line 1: Your tests are well laid out. It's very easy as a reviewer to be able to look through and understand what you're testing for.

smm3000 (commented):
Great to see that you updated the README!

spec/airport_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+  it 'after a landed plane takes off, it is no longer at the airport' do
+    allow(plane).to receive(:taken_off) { true }
+    allow(plane).to receive(:landed) { false }
+    allow(plane).to receive(:airborne) { true }
+    allow(weather).to receive(:forecast) { 'clear' }
+    heathrow = Airport.new
+    heathrow.land(plane, weather)
+    heathrow.takeoff(weather)
+
+    expect(heathrow.hangar).to eq []
```
eoinmakers, line 53: Do these stubs work effectively? How do you think you could test this? Hint: p statements in your tests to make sure the stubbed behaviour is as expected.

eoinmakers (commented):
There are good demonstrations of the learning objectives in this implementation. 

There are opportunities to make your code a little more DRY - which stands for Don't Repeat Yourself. There are some methods that are not being used, and I think it's a little over engineered in places, for example the functionality around the plane object itself being updated to represent being airborne etc. It's important to keep your implementation as simple as possible by sticking as closely as you can to the user stories. 

Well done!

PR #2383, opened 2021-12-13
Airport Challenge 

smm3000:


Alex-Jamshidi (commented):
Good use for guard statements with 'fail' method. Really clear and concise.

Alex-Jamshidi (commented):
classes clearly seperated by file and easily viewable at start of each file

Alex-Jamshidi (commented):
airport.rb:
Would max_capacity benifit from becoming a constant instead of method?

Alex-Jamshidi (commented):
Can clearlly see that a plane will land unless there is already a plane and will only take off if there is a plane. In either case; otherwise returning an error. Great readability.

Alex-Jamshidi (commented):
tests on airport are really clear, easy to follow. Good descriptions - can easily see what they do and coverage looks comprehensive.

Alex-Jamshidi (commented):
is 'attr_reader :plane' in airport_spec.rb coded to read anything yet?

Alex-Jamshidi:
Would it be helpful for repo README to include instructions on running code?

PR #2385, opened 2021-12-13
Airport_challenge_13.12.21

TMWcodes:


lib/airport.rb
```diff
@@ -0,0 +1,10 @@
+
+
+class Airport
+  attr_accessor :planes, :capacity
```
KiranKukar, line 4: I like the use of attributes_accessor for these two, it will make the code efficient and readable as you go

PR #2386, opened 2021-12-13
airport_challenge

jmodiale:


spec/airport_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'airport'
+
+describe Airport do
+    let(:airport) { Airport.new }
```
ElliottZaki, line 4: I really like your use of 'let' here. It makes the code line very succinct and easy to read.

spec/plane_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+    let(:airport) { Airport.new }
+
+    it 'lands at airport' do
+        expect(plane). to respond_to('land')
+    end
+
+    it 'takes off from the airport' do
+        expect(plane). to respond_to('take_off')
+    end
+
```
ElliottZaki, line 14: Good clean use of code. very easy to follow!

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+    def initialize(capacity = DEFAULT_CAPACITY)
+        @capacity = capacity
+        @planes = []
+    end
+    
+    def prevent_landing_capacity(plane)
+        raise "airport is full" if @planes.count > @capacity 
+        @planes << plane
+    end
+
```
ElliottZaki, line 14: I really like the use of raise here, the counting method and if its less than the default capacity.

lib/plane.rb
```diff
@@ -1,13 +1,17 @@
...
+    # @in_airport = false
   end
 
   def land
-  
+    Plane.new
+    # @in_airport = true
   end
 
 end
```
ElliottZaki, line 17: Overall what you have produced seems excellent so far! I was able to read and follow everything you've done the whole way through. Your use of syntax and language is self-explanatory. The only recommendations I could identify is, have you considered creating a README file to update as you complete the challenge. In here you can update the reader or colleague with where you are currently at with the programme. I would also recommend committing more frequently to keep track of changes. This might help with future debugging / shared project work.

PR #2387, opened 2021-12-13
Airport challenge attempt

J-son1:


lib/airport.rb
```diff
@@ -0,0 +1,28 @@
+
+
+class Airport
+  DEFAULT_CAPACITY = 10
```
penguat, line 4: I really love that you've set this up at the top of the class, it's a great place to put it!

penguat (commented):
I'd love to see some instructions on how to run the code

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class Weather
+  attr_reader :condition
+  
+  def stormy?
+    true
+  end
+
+  def set
+    @condition = rand(2)
```
Haydnstokes92, line 11: I like that you have refactored the randomise weather condition, it looks very clean and easy to read

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  describe '#initialize' do
+    it 'has a default capacity' do
+      expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+    end
+  end
+
+  describe '#capacity' do
+    it 'has variable capacity' do
+      subject.capacity = 50
```
Haydnstokes92, line 18: Within the airport spec when refactoring, you can define the 'subject' as Airport and then throughout the spec you can refer to the airport rather than the subject, this might be not be needed when the subject is only one thing however, if there was more than one subject in a file it might be a handy tip.

spec/airport_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'airport'
+
+describe Airport do
```
Haydnstokes92, line 3: subject(:airport) - is the refactoring for my comment below to give example.

spec/weather_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'weather'
```
Haydnstokes92, line 1: This is nice, having a completely different file for the weather condition makes it easy to read and doesn't over complicate the airport class.

Haydnstokes92 (commented):
I have left some comments below, overall this is nice code, I found it easy to read and understand, nicely refactored.

spec/airport_spec.rb
J-son1: Thanks for this tip! I'll be sure to use it going forward!

spec/airport_spec.rb
J-son1: That's great! thanks Haydn.

PR #2388, opened 2021-12-13
Airport weekend Challenge 

ElliottZaki:
Unfinished!

### DIFF
--- a/USER_STORIES.md
+++ b/USER_STORIES.md
@@ -4,6 +4,9 @@ Airport Challenge
 # Motivation: 
 Attempt to complete the weekend challenge with an aim to reinforce learning outcomes from the last week. 
 
+# Notable sources & references
+https://ruby-doc.org/
+
 # Build status:
 Start
 README / USERSTORIES.md
--- a/USER_STORIES.md
+++ b/USER_STORIES.md
@@ -1,0 +1,67 @@
+# Project Title:
+Airport Challenge
+
+# Motivation: 
+Attempt to complete the weekend challenge with an aim to reinforce learning outcomes from the last week. 
+
+# Build status:
+Start
+README / USERSTORIES.md
+Testing stage
+
+# User story 1:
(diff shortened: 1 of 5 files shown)


USER_STORIES.md
```diff
@@ -52,6 +64,14 @@ To ensure safety
 I want to prevent landing when weather is stormy 
 ```
 
+# Summary:
+
+Unfortunately, I ended up having less time than hoped to complete this weekend assignment. I had so far split the 3x user stories and managed to complete most of the first user story with landing and taking_off. 
+
+My next goal was to work on the second user story and to prevent a planes from being landed if the airport is full. I believe this would be done by creating a limited capacity on the airport. And to create a fail procedure once the airport is full otherwise to hold that plane instance.
```
jmodiale, line 71: I like that you're keeping track of the code progress in this file.

spec/airport_spec.rb
```diff
@@ -1,7 +1,20 @@
 require 'airport'
 
 airport = Airport.new
```
jmodiale, line 3: Rather than set a variable outside the 'describe' block you can set one within the 'describe' block using the before method or the let method, so that it runs before each example within the block like so:

before do
  airport = Airport.new
end

OR

let (:airport) { Airport.new }

spec/plane_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'plane'
+
+
+describe Plane do
+    # Is the plane airborne?
+    it { is_expected.to respond_to (:airborne?)}
+    true
```
jmodiale: The boolean appears outside the 'it' block, is there a reason why?

jmodiale:
The code is easy to follow and understand, it is also reusable. Good job Elliott!

jmodiale (commented):
The code is easy to follow and understand, it is also reusable. Good job Elliott!

PR #2390, opened 2021-12-13
Kat's Airport challenge pull request submission

kat-mcbrooks:


lib/weather.rb
```diff
@@ -0,0 +1,6 @@
+class Weather # weather needs its own class to satisfy SRP i.e. so Airport isn't responsible for it
+  def stormy? 
+    rand(5) == 4 # as it's stormy in rare cases, stormy? will only return true if rand(5) generates 
+                  # 4 i.e. probability 4 in 5. 
+  end
+end
```
SoJGo, line 6: Nice separation of Weather into its own class.

README.md
```diff
@@ -12,78 +12,20 @@ Airport Challenge
...
-Note that is a practice 'tech test' of the kinds that employers use to screen developer applicants.  More detailed submission requirements/guidelines are in [CONTRIBUTING.md](CONTRIBUTING.md)
 
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first.
+Question: I wasn't sure of the best place in my code to be setting the weather. Surely the weather needs to be set every time we run the depart and land methods, but I also felt it needed its own class? Would be grateful of feedback/advice on this. In my current code, the weather is only being set/randomly generated when we instantiate a new Airport, but surely it should be generated everytime a plane lands or departs... how to account for this?
 
-* **Submit a pull request early.**
+Question: Support/advice on how to write feature_spec files would be appreciated.
 
-* Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+Question: Not sure I've "defended against edge cases" sufficiently? How to do this?
```
SoJGo, line 31: Nice adding details about your process, would you perhaps want to add install/usage instructions?

spec/airport_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    allow_any_instance_of(Weather).to receive(:stormy?).and_return(true)
+    expect { subject.land(Plane.new) }.to raise_error "It's stormy. Planes cannot land at this time."
+  end
+
+  it 'stormy weather means planes cannot depart' do
+    allow_any_instance_of(Weather).to receive(:stormy?).and_return(true)
+    expect { subject.depart(Plane.new) }.to raise_error "It's stormy. Planes cannot depart at this time."
+  end
+
+end
```
SoJGo, line 56: Tests look comprehensive, I bet you have really good coverage!

spec/feature_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+airport.land(plane_easyjet)
+airport.depart
+
+airport_small = Airport.new(10)
+airport_small.capacity # should return 10
+plane_a = Plane.new
+plane_b = Plane.new
+airport.land(plane_a)
+airport.land(plane_b)
+airport.depart
```
SoJGo, line 20: I wish I could be useful re: features, but I have very much assumed and tried with mine, but this runs through scenarios using all classes which seems bang on to me.

spec/weather_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'weather'
+
+describe Weather do
+  it { is_expected.to respond_to :stormy? }
+
+  it 'is either stormy or not' do # tests that stormy on any instance of weather will be either true or false
+    expect(subject.stormy?).to be(true).or be(false)
+  end
+
```
SoJGo, line 9: Great us of .or, had not found that and certainly could have used it!

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @planes << plane
+  end
+
+  private 
+
+  def full?
+    @planes.length >= capacity
+  end
+
+end
```
SoJGo, line 36: Nice use of private.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 30
+  
+  attr_accessor :capacity # accessor allows the system designer to override capacity as appropriate
+  attr_reader :weather, :planes 
+    
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new) 
+    @planes = [] # every new instance of airport will have a planes array that stores how many planes are currently landed
```
eoinmakers, line 11: We aim to write code that is easily readable and understandable for other devs without the need for comments - in this case, I think your code is well written and I don't think you need to include these comments explaining their usage.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    @capacity = capacity  # every new instance of airport will have a default capacity of 30, unless the system designer overrides this by setting a different capacity. 
+    @weather = weather
```
eoinmakers, line 13: Nice dependency injection of Weather!

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new) 
+    @planes = [] # every new instance of airport will have a planes array that stores how many planes are currently landed
+    @capacity = capacity  # every new instance of airport will have a default capacity of 30, unless the system designer overrides this by setting a different capacity. 
+    @weather = weather 
+    
+  end
+
+  def depart(plane)
+    raise "It's stormy. Planes cannot depart at this time." if weather.stormy?
+    raise "This plane is not in the airport!" unless @planes.include?(plane) # is this what instructions mean by "defending against edge cases"?
```
eoinmakers, line 19: It is indeed an edge case that you've taken into consideration.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new) 
+    @planes = [] # every new instance of airport will have a planes array that stores how many planes are currently landed
+    @capacity = capacity  # every new instance of airport will have a default capacity of 30, unless the system designer overrides this by setting a different capacity. 
+    @weather = weather 
+    
+  end
+
+  def depart(plane)
+    raise "It's stormy. Planes cannot depart at this time." if weather.stormy?
```
eoinmakers, line 18: For .stormy?, we try to name methods after actions or verbs, so something like .is_stormy?, for example.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+    raise "This plane is not in the airport!" unless @planes.include?(plane) # is this what instructions mean by "defending against edge cases"?
+    @planes.delete(plane)
+    @planes
```
eoinmakers, line 21: I don't believe you need to return @planes here - the method has achieved its task of deleting the plane from the hangar.

spec/feature_spec.rb
eoinmakers: Feature tests look much like unit tests, but carry out a full user journey. Thus you can write these tests, but in the context of a 'describe' block, with a test declaration and expectation like _it 'does something' do_, etc.

eoinmakers (commented):
I thought this was an excellent airport challenge implementation, great work. I think you have great separation of concerns here, well defined and concise classes and methods, with the use of private methods where possible. 

You successfully implemented each of the user stories, and took edge cases into account. You've also demonstrated stubbing in your tests, which is great. 

Nice work!

PR #2391, opened 2021-12-13
Airport Challenge

helloecho12345:
Challenge not yet completed.

Need to integrate the random number generator to determine whether the weather is stormy or sunny so that the planes can land or takeoff.

Edge cases also need to be worked in.


README-process.md
```diff
@@ -96,3 +96,31 @@ Setting Up
 -----------
 
 Create lib and spec directories with associated files
+
+
+Work Plan
```
floriov, line 101: I like that you added a 'process' walkthrough for readers to understand what the code does.

PR #2394, opened 2021-12-17
Week 3 - TDD - Airport Individual Challenge

Aadam-Ali:
Completed all 6 user stories.


spec/airport_spec.rb
```diff
@@ -0,0 +1,66 @@
+require "airport"
+
+describe Airport do
+  it "responds to land" do
+    expect(subject).to respond_to(:land).with(2).arguments
+  end
+
+  it "stores a plane when it lands" do
+    plane = double(:plane)
```
samhpickering, line 9: We could save space by moving the plane definition to the top of the spec with `let(:plane) { double :plane }` as it's a common feature of the tests

spec/weather_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    expect(subject).to respond_to :stormy?
+  end
+
+  it "responds to rand_num" do
+    expect(subject).to respond_to :rand_num
+  end
+
+  it "responds with true when rand num is divisible by 10" do
+    weather = Weather.new
+    weather.stub(:rand_num).and_return(10)
```
samhpickering, line 14: It might be better to use srand to force random outcomes - rather than testing implementation specifics

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+  attr_accessor :capacity
+
+  DEFAULT_CAPACITY = 100
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane, weather)
```
samhpickering, line 14: It might make sense to move the weather instance inside of the airport class, as the airport is only dealing with one type of weather at a time

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  private
+
+  def full?
+    @planes.count == capacity
+  end
+
+  def stormy?(weather)
+    weather.stormy?
+  end
```
samhpickering, line 34: I'm not sure this is necessary - it doesn't do anything that weather.stormy? does not already do

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
+require_relative "plane"
```
samhpickering, line 1: Airport code looks clean and well named, nice job

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+require_relative "weather"
+
+class Airport
+  attr_accessor :capacity
```
oceansize, line 5: Be careful with `attr_accessor` — it's the 'nuclear' option when it comes to getters and setters, and it's best to concentrate on controlling access specifically on a method-by-method basis.

oceansize: Also in your particular case you don't ever call `airport.capacity = <value>` so using `attr_accessor` is redundant. `attr_reader` would be more appropriate here.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+  def land(plane, weather)
+    fail "Can not land, weather is stormy" if stormy?(weather)
+    fail "Airport full" if full?
+    @planes << plane
+  end
+
+  def take_off(plane, weather)
+    fail "Can not take off, weather is stormy" if stormy?(weather)
+    fail "Plane not in airport" unless includes?(plane)
+    @planes.delete(plane)
```
oceansize, line 23: This method isn't really making a place 'take off', just removing it from plane storage in the airport.

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
```
oceansize, line 1: I would avoid having empty classes. If you need a class I would expect this class to have some behaviours and state...

lib/airport.rb
oceansize: This is fine, arguably. Wrapping the `weather.stormy?` method means that you can keep `stormy?` as a method for your `Airport` and if the `Weather` object interface changes (for example, another developer renames it to `bad_weather?` you only need to change your `Airport` class once, on line 33.

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+class Weather
+  def stormy?
+    (rand_num % 10).zero?
```
oceansize, line 3: Fun way of doing this ☺️

PR #2396, opened 2021-12-17
Late November Cohort - TDD Airport Challenge

michael-zwickler:


spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'airport'
+
+RSpec.describe Airport do
+  describe '#call_to_land' do
+    it 'requests a flying plane to land on the airport' do
+      allow(subject).to receive(:stormy?) { false }
```
siellsiell, line 6: Using stubbing is the right idea here however it's not great that you have to stop a method on the class you are testing (`stormy?` is a method on `Airport` and you are testing `Airport`). 

Is there something about your class design that you could change to allow you to mock away the randomness without having to do what I describe above? 

Could putting weather-related code into a separate class help with that?

spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      plane = Plane.new
+      expect(subject.call_to_land(plane)).to eq "#{plane} landed!"
+    end
+    it 'throws an error when airport is full' do
+      allow(subject).to receive(:stormy?) { false }
+      subject.capacity.times { subject.call_to_land(Plane.new) }
```
siellsiell, line 12: nice one!

spec/plane_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'plane'
```
siellsiell, line 1: Good tests! You're testing a lot of scenarios. You could consider testing some more for example: test that a newly initialised plane should be considered flying.

spec/plane_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+  describe '#take_off' do
+    it 'lets the plane fly when it was landed before' do
+      subject.land()
+      subject.take_off()
+      expect(subject.flying?).to be true
+    end
+    it 'throws an error, when it was already flying' do
+      expect { subject.take_off() }.to raise_error('I am flying already!')
+    end
```
siellsiell, line 22: I'd be careful with using a single instance of the class that all test cases operate on (as you are doing here using `subject`). It can make what is being tested harder to follow and verify as the test cases depend on each other. 

In this case the last test case depends on the setup of the previous test case, for example.

siellsiell (commented):
Hey it's Simo, a coach :)
Thanks for submitting your challenge code and apologies for the slow response. This was a great attempt at the challenge. Your code is very readable, well documented and well tested. I left a few minor comments for you to have a look at.

lib/airport.rb
```diff
@@ -0,0 +1,66 @@
...
+# -- Initialize: Airport.new(opt. capacity)
+# -- Interfacing Methods: 
+#           #call_to_land(plane obj.) => throws errors in this sequence 
+#                                         when stormy 
+#                                         when airport is full
+#                                         when plane is already landed (forwarded from plane obj.)  
+#           #call_to_take_off(plane obj.) => throws errors in this sequence
+#                                         when stormy
+#                                         when plane is not in airport
+#                                         when plane is already flying (forwared from plane obj.)
```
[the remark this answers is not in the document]
michael-zwickler: Yep, totally agree :) Also most of it should go into a class diagram or a separate kind of documentation but not directly into the code...

lib/airport.rb
```diff
@@ -0,0 +1,66 @@
...
+
+class Airport
+  MAX_DOCKS_DEFAULT = 50
+  BFT_SCALE = 0..12
+  MAX_BFT = 8
```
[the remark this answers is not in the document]
michael-zwickler: You make me afraid of flying... I hope anyone working at any airport should know the Beaufortscale... but you're right wind-speed would more descriptive, just thought with BFT it was at least clear already what is the unit of measure

spec/airport_spec.rb
michael-zwickler: I was thinking about a separate class, but it would not have any more functionality than just providing a rand-number so I thought it was easier to just keep it in the airport itself. I think instead of stubbing :stormy? I could stub :rand instead. 

Longterm this would probably be a good example to implement an observer pattern as a class design, where a Wetterstation class informs the airport, when it's getting too stormy.? And then I could split up my tests into 2 main branches, one where the Wetterstation informed the airport about storm, and one where the Wetterstation informed about nice weather.

spec/plane_spec.rb
michael-zwickler: I think it is implicitly tested in ln 20, because in there I am testing that I cannot take off a new plane, but you're right I nowhere explicetly test how a new instance is initialized...

spec/plane_spec.rb
michael-zwickler: I simply had to try that out again, to be sure about it. In every 'it' block the subject is newly created, so the setup is not depending on a previous test case. In the screenshot all tests would pass and the subjects having a new/unique object id.

Anyway I will try to be more descriptive with using :let to define more explanatory subject-names.

image

michael-zwickler:
> 

Hi Simo, thanks for taking a look at my code and giving me some tips & tricks. I replied to some of them. Probably one of the key take-aways for me would be trying to be more descriptive in my variable names in production code as well as in testing to make it easier to read and understand. Therefore I really appreciate you taking a look at it as I feel it's some kind of authors blindness and helps a lot when a second pair of eyes is taking a look...

Michael

PR #2397, opened 2021-12-17
Pull Request: Completed Airport challenge, Orla Dunlop

odunlop:


spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'airport'
```
simba-tawonezvi, line 1: Code is easy to read.

spec/plane_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'plane'
```
simba-tawonezvi, line 1: Is this file necessary?

README.md
```diff
@@ -60,30 +47,96 @@ To ensure safety
...
+Initially, it seemed the natural decision to give a Plane class the methods of #land and #take_off as they're things *planes* do not airports - I ended up giving these behaviours to the Airport class for two reasons:
 
-For overriding random weather behaviour, please read the documentation to learn how to use test doubles: https://www.relishapp.com/rspec/rspec-mocks/docs . There’s an example of using a test double to test a die that’s relevant to testing random weather in the test.
+1. From writing the unit tests for these methods it became clear that the output I needed to test to fulfill the requirements of the User Story needed to come from an instance of the Airport class 
+2. The User Stories specify traffic controllers and a system desginer, **not** pilots - meaning I was designing functionality for their roles inside the airport (e.g. managing if planes land and/or take off).
 
-Please create separate files for every class, module and test suite.
+In summary, whilst the methods #land and #take_off describe what we would typically view as actions belonging to planes, they actually describe the actions of the traffic controllers granting/denying a plane permission to land.
```
siellsiell, line 58: Very nice to see your reasoning written out like this.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def take_off
+    fail "It's too stormy for a plane to take off right now" if FORECAST.stormy? == true
+    @recent_departure = @planes.pop
+  end
+
+  def gone?
+    !@planes.include?(@recent_departure)
+  end
```
siellsiell, line 28: This function can check whether the last plane is gone but you can't check the status of an arbitrary plane as the user story requires without knowing from which airport it has recently taken off (or landed).

You had some good reasoning for placing the takeoff and landing functions into the Airport class but can you see how, in addition, having similar functions in the Plane class could help with this?

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def override_capacity(new_capacity)
```
siellsiell, line 30: It looks like you only added this method for testing purposes. This is often an indication of design that could be improved. Is there another way that you could adjust the capacity and tests? Maybe using the `initialize` function?

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+describe Airport do
+  
+  describe '#land' do
+    it 'lets an air traffic controller instruct a plane to land at an aiport' do
+      expect(subject).to respond_to(:land).with(1).arguments
+    end
+    it 'raises an error when a plane tries to land when the airport is full' do
+      allow(Airport::FORECAST).to receive(:stormy?) { false }
+      Airport::CAPACITY.times { subject.land(Plane.new) }
```
siellsiell, line 11: Nice one!

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'airport'
+
+describe Airport do
+  
+  describe '#land' do
+    it 'lets an air traffic controller instruct a plane to land at an aiport' do
+      expect(subject).to respond_to(:land).with(1).arguments
+    end
+    it 'raises an error when a plane tries to land when the airport is full' do
+      allow(Airport::FORECAST).to receive(:stormy?) { false }
```
siellsiell, line 10: Nice use of stubbing to mock away the randomness!

The way you can do this without having to mess with the internal variables of the Airport class (which we want to avoid doing to a class we're testing) is using dependency injection: https://github.com/makersacademy/skills-workshops/blob/main/practicals/object_oriented_design/dependency_injection.md

spec/weather_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'weather'
+
+describe Weather do
```
siellsiell, line 3: Is there a way you could use stubbing to test this class in more detail?

spec/airport_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'airport'
+
+describe Airport do
+  
+  describe '#land' do
+    it 'lets an air traffic controller instruct a plane to land at an aiport' do
+      expect(subject).to respond_to(:land).with(1).arguments
```
siellsiell, line 7: This test isn't as strong as it could be. What should be the outcome of landing the plane? Test that.

siellsiell (commented):
Hey it's Simo, a coach :)

Thanks for submitting your challenge code and apologies for the slow response. This was a good attempt at the challenge. Your code is very readable and well documented. You've also made good use of mocking. 

There are places where designing your classes differently would've helped with improving your tests and meeting the requirements of the user stories fully. I left some comments with more details :)

PR #2398, opened 2021-12-21
completed airport challange

rizinaa99:


spec/plane_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'plane'
+
+describe Plane do
+  
+  describe '#land' do
```
oceansize, line 5: Instead of nesting `describe` here, you could use `context` for the inner one. `context` is an alias of describe, so it won't make much of a difference in terms of your output, but it will make the tests more readable.

PR #2400, opened 2021-12-30
Completed

alicegray33:
Not sure if I'm doing this right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,9 @@ Finally, try and take off again and get the expected error as the plane has take
 
 ## Further thoughts
 
-To come
+I often wondered which object was the most appropriate for which method. I originally planned the airport object to take the methods `land` and `take_off` however the user stories implied it should be the plane objects that should take those methods. Although its true that 'the plane lands at the airport' it should really be the airport that controls if the plane can land or take-off. Without being able to further question the 'client' I stuck to the user story to allow a user 'to instruct a plane to land at an airport.' This did result in me calling `airport.full?` from the plane object when the `full?` method could have been kept as a private method within `airport`
+
+Similarly, checking the weather is done with a method within `airport` but is called from `plane` and seems counter-intuitive but as it stands - and without being able to further query the client or get a better understanding of the user needs - the tests pass and the script meets the need of controlloing the flow of planes at an airport. As a further thought, checking the weather may be more likely to be done with an API.
 
 ## Task
 
--- a/README.md
+++ b/README.md
@@ -1,36 +1,97 @@
-Airport Challenge
-=================
+# Airport Challenge
 
+## Approach
+
+I started with each individual user story and wrote a test, then wrote the code to pass the test.
+
+After all five user stories completed and coded, I looked at the "edge cases" and wrote tests for them, then wrote the simplest code to pass the test.
+
+I have refactored where I can - lines 41 and 48 in plane_spec.rb are repeated but I couldn't figoure out a way to avoid the repetition. I tried using `before` but created unexpected errors so I felt the simplest solution was to leave the repetition in (within current time frame, at least). I used `let` to avoid other repetition when setting up the tests and used `allow` to take account of random weather.
+
+Methods are short and do one thing only. I had to make certain assumptions as unable to ask further questions to the client. An airport is a big place and a plane can be in various different areas of the airport, eg the plane could be on the runway, at the terminal or in a hangar.
+
+All tests pass, and have 100% coverage and rubocop is happy.
+
+I have not wrote the bonus "feature test" but the below example shows how it might be used:
+
+## Example
+
+Running irb and requiring plane.rb (which itself requires airport.rb):
 ```
-        ______
-        _\____\___
-=  = ==(____MA____)
-          \_____\___________________,-~~~~~~~`-.._
-          /     o o o o o o o o o o o o o o o o  |\_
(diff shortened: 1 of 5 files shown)


alicegray33:
The code is all really readable, understandable and short.

I liked how you referred to the DEFAULT_CAPACITY constant in your tests when testing if the airport is full. So if you change the default capacity in future then it will still pass.

I also liked how you used Array.new to define new array rather than just [] as its more obvious what you're doing

JennyEMorrison:
I find your code very readable and easy to understand. It is all very consistent throughout all of the user stories. You managed to get further than me, and It's really interesting to see how you approached the weather section of the challenge. Really nicely written.

README.md
```diff
@@ -1,36 +1,99 @@
-Airport Challenge
-=================
+# Airport Challenge
 
+## Approach
```
EdwardAndress, line 3: nice reflection, Alice!

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
+class Airport
+
```
EdwardAndress, line 2: How could you refactor this class such that the default capacity could be overidden?

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+  CAPACITY = 5
+
+  attr_reader :hangar
+
+  def initialize
+    @hangar = []
+  end
+
```
EdwardAndress, line 10: I like that you've added in the complexity around where a plane might be in the airport.  Did you consider how you might know whether a plane is on the runway, or not?

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+  def taxi_to_runway(this_plane)
+    @hangar.delete_if { |plane| plane == this_plane }
+  end
+
+  def full?
+    return true if @hangar.count >= CAPACITY
+    false
+  end
+
+  def check_weather
```
EdwardAndress, line 25: Did you consider a separate Weather class?

lib/plane.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'airport'
+
+class Plane
+
+  def initialize
+    @landed = false
+  end
+
+  attr_reader :landed
+
```
EdwardAndress, line 10: Someone asked about using `fail` more than once in a single method – was it you?  Can you think of a way to avoid this pattern and surface all problems (when there's more than one)?

spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'airport'
+
+describe Airport do
+  let(:plane) { Plane.new }
+
+  it { is_expected.to respond_to(:hangar) }
+
+  describe '#send_to_hangar' do
+    it 'stores a plane in the hangar' do
+      expect(subject.send_to_hangar(plane)).to eq true
```
EdwardAndress, line 10: Here, you'd probably want to check that the plane actually ends up in the hangar.

spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+  end
+
+  describe '#taxi_to_runway' do
+    it 'taxis a plane to the runway' do
+      expect(subject.taxi_to_runway(plane)).to eq []
```
EdwardAndress, line 16: And I guess here the important thing is that the plane has left the hanger but you might also change some state on the plane or have another collection of planes that are on runways...

spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+  end
+
+  describe '#full?' do
+    it 'returns false when hanger is empty' do
+      expect(subject.full?).to eq false
+    end
+    it 'returns true when hangar is full' do
+      5.times { subject.send_to_hangar(Plane.new) }
```
EdwardAndress, line 25: nice test!

spec/plane_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'plane'
+
+describe Plane do
+  let(:airport) do
```
EdwardAndress, line 4: nice use of `let`

spec/plane_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      subject.land(airport)
+      expect { subject.land(airport) }.to raise_error 'Unable to land, plane is already landed'
+    end
+  end
+
+  describe '#take_off' do
+    it 'takes off from an airport' do
+      subject.land(airport)
+      subject.take_off(airport)
+      expect(airport.hangar).not_to include(subject)
```
EdwardAndress, line 37: OK so here you're testing that the plane is removed from the hangar, which is great

spec/airport_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      expect(subject.full?).to eq true
+    end
+  end
+
+  describe '#check_weather' do
+    it 'returns random weather' do
+      expect(subject.check_weather)
```
EdwardAndress, line 32: I love that you've broken this long line out onto multiple lines. It's great.  In terms of how the test works though... would it still pass if the weather was broken?  What if the weather was broken in a way that meant it was always "cloudy"?

EdwardAndress:
Hey Alice! nice work here – you've clearly demonstrated the TDD learning objectives.  I've left a few comments inline against your code. If you have questions please ask them by DM on Slack :)

PR #2401, opened 2022-01-01
airport_challenge

Tom-V1:


lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+  def take_off(plane)
+    fail "Plane is not at the airport" unless @planes.include? plane
+
+    fail "Poor weather!" if weather? == true
+  
+    plane.undock
+    @planes.delete(plane)
+  end
+
+  def weather?
```
tomfuller, line 33: Does it make sense for the Airport to own the weather behaviour? This might be better in its own class.

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
+require "airport"
+
+describe Airport do 
+  
+  describe "#basic function checks" do
+
+    it "weather gives true or false" do
+      expect(subject.weather?).to eq(true).or(eq(false))
```
tomfuller, line 8: Could you use a mock to test a case when the weather is stormy and when its not separately?

spec/airport_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    it "weather gives true or false" do
+      expect(subject.weather?).to eq(true).or(eq(false)) 
+    end 
+
+    it "has default capacity" do
+      expect(subject.capacity).to eq(20)
+    end 
+
+    it "sets capacity" do
+      subject.capacity = 150
```
tomfuller, line 16: If you are going to change the value of a class variable its best to use a setter class, in this case it would be perfectly acceptable to also only change the capacity of an airport when the airport is created. This would be done doing something like airport = Airport.new(30) for example.

tomfuller:
This is a nice solution to the problem, you clearly grasped the learning objective for the week. I would think about how you could use mocks to test your weather function and whether the weather behaviour should have its own class. I would also think about what you need to meet the requirements and not do anymore than that, for example your plane docking functionality is a nice level of detail but not required.

PR #2402, opened 2022-01-03
First RSPEC project

SusanaCodes:
First time practicing with rspec.
Creating the basic test for a software to control the flow of planes at an airport. 
Creating on the basic of the main points of diagram; the air_controller, the weather and the fleet(with the plane).
Diagram added to the code.


lib/air_controller.rb
```diff
@@ -0,0 +1,10 @@
+require_relative "weather"
+
+class Aircontroller
```
tomfuller, line 3: A good practice for OOP is to take names directly from the nouns in user stories, in which case this should probably be called Airport.

lib/air_controller.rb
```diff
@@ -0,0 +1,10 @@
...
+
+  FLEET_SIZE = 10
```
tomfuller, line 5: Following the above comment, this would be called CAPACITY.

lib/fleet.rb
```diff
@@ -0,0 +1,19 @@
+require_relative "plane"
+require_relative "air_controller"
+
+class Fleet
```
tomfuller, line 4: With the above suggestions, could this logic live within an Airport class?

lib/plane.rb
```diff
@@ -0,0 +1,4 @@
+class Plane
```
tomfuller, line 1: This class isn't used anywhere

lib/weather.rb
```diff
@@ -0,0 +1,6 @@
+class Weather
+ 
+  def weather_today 
+    Random.rand(1..10) < 7 ? true : "stormy"
```
tomfuller, line 4: What are you trying to return with this function? at the moment it could return a string or a boolean.

spec/weather_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "weather"
+
+describe Weather do
+  it "tells if the weather is sunny or stormy" do
+    weather = Weather.new
+    expect(weather).to respond_to(:weather_today)
+    allow(weather).to receive(:weather_today).with(true).and_return("sunny")
```
tomfuller, line 7: What were you trying to stub here? A stub needs to be before an expect in order to have an effect on the test. Also as mentioned in the implementation, currently the 'weather_today' function doesn't return 'sunny' it returns true.

tomfuller:
I have left few notes where the code could be improved. The solution you have arrived at doesn't fully meet the requirements given in the user stories. Judging from your diagram and your design decisions it seems like you might have made some assumptions early on about how this should work? It would be great for me to understand your thought process and I can explain my thoughts much better over a call, so let's schedule a chat.

PR #2403, opened 2022-01-04
Airport Challenge

simba-tawonezvi:
Not complete.


lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+
+class Airport
+  attr_accessor :planes
+  def initialize
+    @planes = []
+    # @capacity
+  end
+
+  def land(airplane)
+    @planes.push(airplane)
```
odunlop, line 11: Good method - follows OOP: Method only performs one behaviour

lib/airplane.rb
```diff
@@ -0,0 +1,3 @@
+class Airplane
```
tomfuller, line 1: I would probably rename this to 'Plane'. It's simpler, harder to confuse with Airport and also would make the naming of the planes array more accurate.

tomfuller:
This is a good solution for the first user story. It looks like you have added empty functions for all of the user stories in advance. The more pragmatic approach is to take one user story at a time and only write the code as and when you need it. This avoids unused and unnecessary code, which in a larger enterprise codebase is very important because they are already massive and adding anything that isn't needed will make it bigger and more confusing to work on.

PR #2404, opened 2022-01-04
Adam Hindmarsh 

adamhind:
Completed challenge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,77 +13,29 @@ Airport Challenge
 
 ```
 
-Instructions
----------
+#Adam Hindmarsh README
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+##Decription 
+This project is to replicate the aid a air traffic controller to be able to monitor planes taking off and landing at different places. The program also ensures safety when planes are taking off and landing. 
 
-Steps
--------
 
-1. Fork this repo, and clone to your local machine
-2. Run the command `gem install bundler` (if you don't have bundler already)
-3. When the installation completes, run `bundle`
-4. Complete the following task:
+##How to use
 
-Task
------
+This program needs to be ran in the terminal using command prompts. 
+You will need to clone the Repo from GitHub and run it using IRB requiring './lib/airport' to begin. 
 
-We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off.  Here are the user stories that we worked out in collaboration with the client:
+There are two classes of objects that are used in the programs: Planes and Airports
 
-```
-As an air traffic controller 
-So I can get passengers to a destination 
-I want to instruct a plane to land at an airport
+###Planes don't have any direct methods but their landed status can be read by using the message '.landed'.
 
-As an air traffic controller 
-So I can get passengers on the way to their destination 
-I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
+###Airports 
+'.new' or '.new(50)' the argument sets the capacity of the airport. The default value is 20. 
+'.land(plane)' lands a plane at that airport.
+'.take_off(plane)' departs a plane from that airport.
 
-As an air traffic controller 
-To ensure safety 
-I want to prevent landing when the airport is full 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -13,77 +13,29 @@ Airport Challenge
...
-* High [Test coverage](https://github.com/makersacademy/course/blob/main/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc. 
-
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this at this moment.
-
-**BONUS**
+###Planes don't have any direct methods but their landed status can be read by using the message '.landed'.
 
-* Write an RSpec **feature** test that lands and takes off a number of planes
+###Airports
```
EdwardAndress, line 31: It's great that you added this explanation of how to use the program.  You might be interested to know that you can denote instance methods using the `#` symbol.  So you could write `#land(plane)` to make it really clear that the method is called on an instance, unlike `.new` which is called on the class.

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+  DEFAULTCAPACITY = 20
+  attr_reader :planes
+
+  def initialize(capacity = DEFAULTCAPACITY)
+    @capacity = capacity
+    @planes = []
+
+  end 
+
+  def land(plane)
```
EdwardAndress, line 13: We had a question about multiple `fail`s in class – maybe it was you that asked it? Anyway, at the moment, even if all three things are in a fail state, only one error will be thrown.  Can you think of a way to refactor this which would allow all three problems to be surfaced?

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def landed?(plane)
+    plane.landed
+  end 
+
+  def full?
+    @planes.count >= @capacity
+  end 
+
+  def stormy?
```
EdwardAndress, line 41: Did you consider implementing a weather class?  If so, what would it's responsibilities be?

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+    x = rand(1..100)
+    return true if x == 13
+  end 
+
+  def empty?
+    @planes.count.zero?
```
EdwardAndress, line 47: There's an array method called `#empty?` which you could use here

lib/airport.rb
adamhind: Looking at it now, it would make a lot of sense to have a weather class. We could call a method on it that determines the weather weather.now maybe? and, to be more adventurous define the likely hood of bad weather with when initialising so some airports have worse weather than others. 


DEFAULT_CHANCE = 100
 
class weather
  def initialize(chance = DEFAULT_CHANCE) 
    @chance = chance
   end

lib/airport.rb
adamhind: I'm not sure what you mean by 'multiple fails'.

README.md
adamhind: Thanks Eddie I'll make sure that I do that from now on.

lib/airport.rb
adamhind: Ah yes, thanks Eddie.

EdwardAndress:
Hey Adam, this is a really nice solution and you've clearly demonstrated the learning objectives of this module.  I've added a few comments in line against your code.  If you wanted to push things a bit further, you could create a weather class and, also, try some mocking (for the plane and weather, in airport_spec).

Drop me a line on Slack if you have any questions – I probably won't see any replies you make here.

spec/airport_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    it "prevents a plane from taking off if there is a storm" do 
+      plane = Plane.new
+      allow(subject).to receive(:stormy?) { false }
+      subject.land(plane)
+      allow(subject).to receive(:stormy?) { true }
+      expect { subject.take_off(plane) }.to raise_error "It is too stormy to take off"
+    end   
+  end 
+
+  describe '#land' do
```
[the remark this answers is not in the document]
adamhind: Someone has shown me this since I have done this task and it would made my tests cleaner.

lib/airport.rb
EdwardAndress: Very nice!! :)

lib/airport.rb
EdwardAndress: Sorry – I'm talking about the fact that you have several `fail` clauses in one method.

PR #2405, opened 2022-01-04
Sam Pickering

samhpickering:


spec/feature_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'airport'
+require 'plane'
+
+describe 'airport challenge' do
```
Aadam-Ali, line 4: This implementation of a feature test has been done very well and tests a range of possible situations to ensure maximum coverage of the codebase

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    plane.airborne = true
+    planes.delete(plane)
+  end
+
+  def landed?(plane)
+    planes.include?(plane)
+  end
+
+  def weather
```
Aadam-Ali, line 36: I would consider making 'Weather' it's own class, with a predicate method such as 'stormy?'

PR #2407, opened 2022-01-04
First 4 user stories

alexwood91:


lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'plane'
```
EdwardAndress, line 1: Is this needed? 🤔

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+
+class Airport
+
+  DEFAULT_CAPACITY = 15
+
+  def initialize
+    @planes = []
+  end
+
+  def capacity
```
EdwardAndress, line 11: How could you change this to allow the capacity to be overridden?

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
+class Plane
+
```
EdwardAndress, line 2: You could track state with a single instance var (either @airborne or @landed) and I think I'd prefer this as it means only one thing needs updating when the state changes.  Needing to update two things increases the chance that someone might forget.

Also, your `@landed` var is going to be `nil` when the plane is first created so, if you want to keep it, you'll need to assign a value in `initialize`.

If you wanted to trim it down to one var, you could choose one of those or you could go with something generic like `state`.  The problem with `state` is that it doesn't really convey much about what the states could be and it could be anything! So you'd then want to think about that and whether you could somehow prevent someone from setting a state of "OoO" or whatever :)

EdwardAndress:
Nice solution, Alex!  You've definitely demonstrated the learning objectives for this module, so well done there.  I've left a few comments in line against your code.  The next step would be to complete that Weather class :) ... But there's no obligation to continue.

PR #2408, opened 2022-01-04
Airport challenge commit

rosiewaite92:


airport_challenge_README.md
```diff
@@ -0,0 +1,30 @@
...
+
+As an air traffic controller 
+To ensure safety 
+I want to prevent takeoff when weather is stormy 
+
+As an air traffic controller 
+To ensure safety 
+I want to prevent landing when weather is stormy 
+
+I need to create a rspec test folder and relevant classes and methods for this.
```
tomfuller, line 30: Its really great to add the requirements to your README for context, but it also be good to have instructions on how to run your code and maybe brief description of your approach and/or a diagram.

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    @planes = []
+    @capacity = capacity 
+  end
+
+  def land_plane(plane)
+    fail "You cannot land" if full?
+    @planes << plane 
+  end 
+
+  def take_off
```
tomfuller, line 16: Is there a way we could take off a specific plane similarly to how land_plane does?

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    @planes.pop
+  end 
+
+  private 
+
+  def full?
+    @planes.count >= AIRPORT_CAPACITY
+  end 
+
+  def stormy?
```
tomfuller, line 27: Its great that you have separated the responsibility generating the random aspect of the weather and interpreting it.

Does it make sense that the airport interprets the weather as stormy or does this responsibility lie with the weather class?

spec/airport_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "airport"
+
+describe Airport do
+  it { is_expected.to respond_to :land_plane }
+  it { is_expected.to respond_to :take_off }
+
+  # it "expects planes to reduce by one when a plane takes off"
```
tomfuller, line 7: This feels like it would have been a useful test, Im wondering what stopped you from writing it? were you unsure on how to write it? did you run out of time? Either way its fine, but I think your test coverage would have been a lot stronger if you had written this and maybe one more test to cover the landing functionality.

spec/airport_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  it "raises an error when the airport is full" do 
+    subject.capacity.times { subject.land_plane(:plane) }
+    expect { subject.land_plane(:plane) }.to raise_error "You cannot land"
+  end 
+
+    it "raises an error when the weather is stormy to stop take off" do 
+      weather = Weather.new 
+      weather.todays_weather
+        expect { weather.stormy? }.to raise_error "You cannot take off"
```
tomfuller, line 22: This test is failing, it's really important not to commit failing tests. If you look at the comment in the airport class around the 'stormy?' function it might shine some light on why.

tomfuller:
This is a really great effort, you have clearly grasped the learning objectives for the week and achieved the stretch goal of the weather class. However you have committed a failing test which is a big no no in the software development community, you should always run your tests and ensure they are green before committing. 

I would also start thinking about why we write tests and what do they cover,  a test that tests a function's functionality is more useful than a test that only checks if a function can be called. For example this test 'it { is_expected.to respond_to :take_off }' would be covered by this test  '# it "expects planes to reduce by one when a plane takes off"' but you also then be testing the functionality of the function too.

airport_challenge_README.md
rosiewaite92: Thank you, I wasn't sure what to put on the read me but it has become clearer as the weeks have gone on :)

lib/airport.rb
rosiewaite92: so with pop I guess it would just take out the last plane to have landed. If I defined an argument for the method I could specify which plane to take off. Thanks I didn't even think of that!

lib/airport.rb
rosiewaite92: I had quite a lot of problems getting stormy to actually work (It was still not working when I committed this project) I did try it in the weather class too and still could not get it to work properly. But yes it would make more sense for that responsibility to fall within the weather class

spec/airport_spec.rb
rosiewaite92: If I am honest I cannot remember but I know at this point I was often struggling to write tests before the code so it may have been that I meant to come back to it and ran out of time.

spec/airport_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    airport = Airport.new(30)
+    expect(airport.capacity).to eq 30
+  end 
+
+  it "raises an error when the airport is full" do 
+    subject.capacity.times { subject.land_plane(:plane) }
+    expect { subject.land_plane(:plane) }.to raise_error "You cannot land"
+  end 
+
+    it "raises an error when the weather is stormy to stop take off" do
```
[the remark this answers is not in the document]
rosiewaite92: Thank you, I have updated this now

spec/airport_spec.rb
rosiewaite92: I did try to have a look at this but I could still not get it to work unfortunately. Would it have been better to comment the test out so it was not failing? Although I know that will then reduce my coverage.

PR #2409, opened 2022-01-04
completed up to 4th user story

jackiebenn:


lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+    fail 'Airport is full' if full?
+    @airport << plane
+  end
+
+  def take_off
+  end
+
+  private 
+
+  attr_reader :airport
```
michael-zwickler, line 22: attr_reader not needed, use @ airport in ln. 25

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
+class Airport
+
+  DEFAULT_CAPACITY = 20
+
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airport = []
```
EdwardAndress, line 8: Maybe a better name here would be something like `hangar` or `planes`?

spec/airport_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    expect { subject.land_plane Plane.new }.to raise_error 'Airport is full'
+  end 
+  it 'has a default capacity' do
+    expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+  end
+end
+
+describe 'initialization' do
+  it 'has a variable capacity' do
+    airport = Airport.new(50)
```
EdwardAndress, line 21: Nice that you can override the default capacity like this

PR #2410, opened 2022-01-04
Code Review

jyoung-gds:
User Stories: 
- As an air traffic controller, so I can get passengers to a destination, I want to instruct a plane to land at an airport
- As an air traffic controller, so I can get passengers on the way to their destination, I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
- As an air traffic controller, to ensure safety, I want to prevent landing when the airport is full 
- As the system designer, so that the software can be used for many different airports, I would like a default airport capacity that can be overridden as appropriate
- As an air traffic controller, to ensure safety, I want to prevent takeoff when weather is stormy 
- As an air traffic controller, to ensure safety, I want to prevent landing when weather is stormy 

Improvements:
- I didn't refactor the code at all, so it may be a bit messy.


README.md
```diff
@@ -87,3 +87,12 @@ Finally, don’t overcomplicate things. This task isn’t as hard as it may seem
 * **Submit a pull request early.**
 
 * Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+
+# Solution Design
+
+The solution I have designed uses three classes: Airport, Plane and Weather. The object interaction diagram below shows how these classes interact with the user and each other and what messages each can receive.
+
+### Object Interaction Diagram
```
tomfuller, line 95: The diagram is really great, its always good to be able to visualise your code like this. It would have been nice to have included instructions on how to run the code and the tests in the README.

lib/airport.rb
```diff
@@ -0,0 +1,15 @@
+class Airport
+  
+  attr_reader :docked_planes
+  DEFAULT_CAPACITY = 20
```
tomfuller, line 4: Is there a way you could have made the default capacity of the airport overridable as specified in the user stories?

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
+class Plane 
+  
+  attr_reader :location
+  def initialize
+    @location = Airport.new
```
tomfuller, line 5: Does it make sense semantically that a new Airport is created with every new Plane? It can really help when you're designing Object Oriented code to think about how those objects act and interact outside of your codebase.

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def takeoff(conditions)
```
tomfuller, line 8: Is there a better name for this param? Perhaps something weather specific?

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
...
+    fail "The plane is already in the air." if @location == "In-flight"
+    fail "The weather is too stormy to take off at the moment." if conditions.stormy? == true
+  
+    @location.docked_planes.delete_if { |plane| plane == self }
```
tomfuller, line 12: Location is also a little confusing as a variable, both in name and in practice. I think in name it can represent a state of the plane e.g. "in-flight" and also a physical location e.g. in the airport. In practice it can be a string or an Airport class that has its own functionality.

spec/plane_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+describe "landing tests" do
+
+  before(:each) do
+    @airport = Airport.new
+    @weather = Weather.new
+    @plane = Plane.new
+  end
+
+  it "should store a plane at a specified airport when landed" do
+    allow(@weather).to receive(:stormy?) { false }
```
tomfuller, line 27: There is a lot of repetition here in your test setups, is there a way to reduce this? Maybe looking at things like Before blocks.

PR #2412, opened 2022-01-21
Airport Challenge - Luke-Collins

Luke-Collins:


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'plane'
+
+class Airport
+  attr_accessor :airborne
+  attr_accessor :docked
+  attr_accessor :capacity
+  attr_accessor :forecast
+  
+  def initialize(capacity = 20)
```
benfielding74, line 9: Could you have split out capacity to a default variable so an argument doesn't need to be passed when initialising a new airport?

readme/README.md
```diff
@@ -0,0 +1,40 @@
...
+  def weather_error_takeoff
+    fail "Cannot take off in this weather" if @forecast == 'stormy'
+  end
+
+CODE COVERAGE
+I finished with a 98% (95.83% in the offending file) coverage rate, which I elected not to cover as it was a private line, and I understand that these need not be tested on?
+(May be old information)
+
+FEEDBACK
+I look forward to thie being reviewed and to tell me where I've gone wrong or can improve.
```
benfielding74, line 40: Great README, and clear and readable code with good use of Single Responsibility prinicples.

lib/airport.rb
Luke-Collins: The '= 20' means that if there is no argument, is uses 20, - so 20 is the default and no argument is required.

readme/README.md
Luke-Collins: Thank you ^-^

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+    @docked << plane unless full_error || weather_error_land
+  end
+
+  def takeoff(plane)
+    @airborne << @docked.delete(plane) unless weather_error_takeoff
+  end
+
+  private
+
+  def forecast_sum
```
siellsiell, line 26: This function doesn't output a sum so the naming seems off. I also don't see you calling this function anywhere. Best to delete things we don't use :)

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+    @forecast = @weather.sample
```
siellsiell, line 27: It's the intermediate`@forecast` variable really necessary here? It's best not to store state in the form of instance variables when we can avoid it as it increases the complexity of the system more than necessary. 

The `weather_error_takeoff` function is now no longer a self-contained function because its output depends on whether `forecast_sum` was run earlier or not.

readme/README.md
```diff
@@ -0,0 +1,40 @@
...
+  def full_error
+    fail "Airport Full" if @docked.length >= @capacity 
+  end
+
+  def weather_error_takeoff
+    fail "Cannot take off in this weather" if @forecast == 'stormy'
+  end
+
+CODE COVERAGE
+I finished with a 98% (95.83% in the offending file) coverage rate, which I elected not to cover as it was a private line, and I understand that these need not be tested on?
```
siellsiell, line 36: Yes typically we don't test private functions directly but that doesn't mean our tests can't _cover_ them (why is that? might be an interesting one to think about)

I'm guessing the line that wasn't covered was the one in the `forecast_sum` function. It's not covered because you're never calling it anywhere so the code can never run. Your implementation of the weather at the moment doesn't make it change randomly.

readme/README.md
```diff
@@ -0,0 +1,40 @@
+Airport Challenge
```
siellsiell, line 1: Thanks for writing this up! It can be very useful to see the reasoning when reviewing.

spec/airport_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+describe Airport do
+  plane = Plane.new
+  airport = Airport.new
+
+  it 'can land a plane' do
+    expect(airport.land(plane)).to eq([plane])
+  end
+
+  it 'a plane can take off' do
+    expect(airport.takeoff(plane)).to eq(airport.airborne)
```
siellsiell, line 13: This test would be stronger if you checked that the plane is now no longer in the airport. For example, calling `takeoff(plane)` twice in a row for the same plane should fail.

With the test as it is, I could rewrite `takeoff` like this:

```
 def takeoff(plane)
    airport.airborne
  end
```
And the test would still pass but the plane would still be in the list of docked planes in the airport.

spec/airport_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  plane = Plane.new
+  airport = Airport.new
+
+  it 'can land a plane' do
+    expect(airport.land(plane)).to eq([plane])
```
siellsiell, line 9: Can you apply a similar reasoning to your `can land a plane` test as outlined in the comment below? How could this test be made stronger?

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+
+class Airport
+  attr_accessor :airborne
+  attr_accessor :docked
+  attr_accessor :capacity
+  attr_accessor :forecast
+  
+  def initialize(capacity = 20)
+    @docked = []
+    @airborne = []
```
siellsiell, line 11: I'm not sure it's the airport that should keep track of whether a plane is airborne since that information would become incorrect as soon as the plane lands at a different airport. 

Which class might more naturally hold this information? The domain modelling exercise from this week will help here.

spec/airport_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  it 'a plane can take off' do
+    expect(airport.takeoff(plane)).to eq(airport.airborne)
+  end
+
+  describe 'docking 20 planes' do
+    it 'does not land a plane when airport is full (20)' do
+      20.times do airport.docked << plane end
+      expect { airport.land(plane) }.to raise_error('Airport Full')
```
siellsiell, line 19: Nice use of `raise_error`!

spec/airport_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  describe 'it lets us change default capacity' do
+    it 'lets us change the capacity' do
+      expect(airport = 50).to eq(50)
+    end
+  end
+
+  it 'prevents takeoff when weather is stormy' do
+    airport.forecast = "stormy"
```
siellsiell, line 34: In tests we want to avoid manipulating the internal state of the object directly -- we should only test the public methods of the class. When you find yourself adding  `attr_accessor` or `attr_reader` just for testing purposes, it often is a hint that your design could be improved to make testing possible without having to make things public just for the test.

In this case, it might indicate that the weather-related functionality should be in its own class so that you can replace it with a [double](https://github.com/makersacademy/course/blob/main/pills/doubles.md) when testing. 
The idea of [dependency injection](https://github.com/makersacademy/skills-workshops/blob/main/practicals/object_oriented_design/dependency_injection.md) might help with knowing how to separate the weather stuff out into its own class.

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+    @capacity = capacity
+  end
+
+  def land(plane)
+    @docked << plane unless full_error || weather_error_land
+  end
+
+  def takeoff(plane)
+    @airborne << @docked.delete(plane) unless weather_error_takeoff
+  end
```
siellsiell, line 22: Nice error handling!

siellsiell (commented):
Hey it's Simo, a coach :)

Thanks for submitting your challenge code and apologies for the slow response. This was a good attempt at the challenge. Your code is very readable and you've nicely hidden implementation details in private methods.

There are places where designing your classes differently would've helped with improving your tests and meeting the requirements of the user stories fully (for example right now the weather doesn't change randomly).

I left some comments with more details :) Do let me know if you have any questions!

PR #2413, opened 2022-01-21
f commit

laurenbethgray:


lib/airport.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'plane'
+class Airport
+
+  DEFAULT_CAPACITY = 10
```
MakersKyrstenStone, line 4: can a user change this on the fly? does the code need a variable on the airport class line 2?

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def land(plane)
+    fail 'Permission denied, airport at capacity.' if @planes.count >= DEFAULT_CAPACITY
+    @planes << plane
+  end
+
+  def take_off
+    @planes.pop
+    p "Confirmation, plane has left the airport."
```
tomfuller, line 20: You shouldn't put useful outputs in p/puts statements where they are lost to the terminal, returning the string so that it can be surfaced in a front end somewhere is much more useful.

lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
+  def planes
```
tomfuller, line 2: I don't think you need this function, I don't see it used anywhere and I don't know if its necessary.

read_me/READ_ME.md
```diff
@@ -0,0 +1,46 @@
+AIRPORT CHALLENGE
```
tomfuller, line 1: This is a great README, you have explained your approach and kept the requirements in there, the only thing I would add is instructions on how to run your code. Also just an FYI convention is to keep your README in a file called README.md at the top level of your repo, this repo came with one you could add to.

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'airport'
+
+describe Airport do
+  airport = Airport.new
+  it 'instructs plane to land at an airport' do
+    # arrange
+    plane = Plane.new
+    # assert
+    expect(airport.land(plane)).to eq [plane]
```
tomfuller, line 9: This is a good test, I would probably remove the comments once you have written it (if it helps you structure your tests then this is a good approach). I would also maybe keep `airport.land(plane)` in the set up part of your test and then assert that `airport.planes` is eq to `[plane]`.

lib/airport.rb
```diff
@@ -0,0 +1,23 @@
...
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail 'Permission denied, airport at capacity.' if @planes.count >= DEFAULT_CAPACITY
+    @planes << plane
+  end
+
+  def take_off
```
tomfuller, line 18: Is there a way your could make this function take off a specific plane rather than just the last one to land?

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  it 'instructs plane to take off from an airport' do
+    # arrange
+    plane = Plane.new
+    # assert
+    expect(airport.take_off).to eq "Confirmation, plane has left the airport."
```
tomfuller, line 16: This test will test that the take off function is run and reaches the final p statement but to check that it actually takes off a plane it would be better to put `airport.take_off` in the test setup and then assert that the plane is no longer in the plane array. For example, in this test at the moment you haven't actually created a plane/ put it in the airport to take off.

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  it 'prevents landing when airport is full' do
+    10.times { airport.land Plane.new }
```
tomfuller, line 20: Could you change the capacity to 1 and then try landing a second plane rather than land 10?

spec/airport_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    expect { airport.land Plane.new }.to raise_error 'Permission denied, airport at capacity.'
+  end
+
+    # Create a new airport with 30 and test
+  it 'Allow user to set capacity of airport' do
+    num = 30
+    custom_airport = Airport.new(num)
+    expect(custom_airport.capacity).to eq num
+
+    # Create a default airport using the DEFAULT_CAPACITY constant
```
tomfuller, line 30: This should be a separate test, as a general rule you want to avoid having more than one expect statement in a 
unit test.

tomfuller:
This is a nice solution and you clearly grasped the learning objectives for the week. I have left some comments where there are some improvements you could make. Its worth noting that unless you are building a command line application then you will likely never use p for a return value and its much better to use return instead.

PR #2414, opened 2022-01-21
Airport_challenge_wk3

CristinaGonzalezVaro:


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
+class Airport
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
```
shalini9, line 5: This code is very easy to read and understand.

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+    raise "Plane cannot take off. Weather is stormy." if stormy?
+  end
+
+private
+
+  def full?
+    @planes.count >= DEFAULT_CAPACITY
+  end
+
+  def stormy?
```
shalini9, line 25: To improve, there should be a Weather class for random weather conditions - you are already aware of this in your read.me file :)

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+    @planes = []
+  end
+
+  def land(plane)
+    raise "Plane cannot land. Airport is full." if full?
+    raise "Plane cannot land. Weather is stormy." if stormy?
+    @planes << plane
+  end
+
+  def take_off(plane)
```
shalini9, line 15: Perhaps consider using .delete when aeroplane departs (this will result in empty array)

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+private
+
+  def full?
+    @planes.count >= DEFAULT_CAPACITY
+  end
+
+  def stormy?
+    rand(1..7) > 5
```
shalini9, line 26: This is clever - any condition over over 5 is stormy.  Very nice :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  describe '#take_off' do
+    it 'tells plane to take off' do
+      expect(airport).to respond_to(:take_off).with(1).argument
+    end
+   
+    it 'raises an error if plane tries to take off when weather is stormy' do
+      allow(airport).to receive(:stormy?).and_return true
+      expect { airport.take_off(:plane) }.to raise_error "Plane cannot take off. Weather is stormy."
```
shalini9, line 35: Very nice match methods 'to raise_error' and with(1).arguement 👍

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    raise "Plane cannot land. Airport is full." if full?
+    raise "Plane cannot land. Weather is stormy." if stormy?
```
shalini9, line 11: Good short-hand use of 'if' statements

README/README1.md
```diff
@@ -0,0 +1,9 @@
+Airport Challenge 
+
+This project is the weekly challenge of my third week at Makers Bootcamp. The main goal is to put into practice TDD and ensure good practice.
+
+The aim of the challenge was to test drive the creation of a set of classes and methods to satisfy the user stories I was provided with.
+
+I broke down the user stories following black-box planning to understand what the input and output of each story would be. 
+ 
+Some of the challenges I faced relate to confirming when a plane is no longer at the airport, as well as wanting to move my stormy? method to a different class (Weather).
```
tomfuller, line 9: This is great, always useful to get context on the problem in the README! If you have time, instructions on how to run your code are also really useful and much appreciated.

lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
+  def initialize(plane)
```
tomfuller, line 2: I don't think you reference this plane variable within plane so you can probably leave the plane class empty.

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      allow(airport).to receive(:stormy?).and_return false
+      Airport::DEFAULT_CAPACITY.times do
+        airport.land(:plane)
+      end
+      expect { airport.land(:plane) }.to raise_error "Plane cannot land. Airport is full."
+    end
+
+    it 'tells plane to land' do
+      allow(airport).to receive(:stormy?).and_return false
+      expect(airport).to respond_to(:land).with(1).argument
```
tomfuller, line 24: I think a slightly more comprehensive test of the land function would be to call land on a plane and then check to see if that plane is in the airport. Testing that would also test that you can call land with one argument.

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  describe '#land' do
+    
+    it 'raises an error if plane tries to land when weather is stormy' do
+      allow(airport).to receive(:stormy?).and_return true
+      expect { airport.land(:plane) }.to raise_error "Plane cannot land. Weather is stormy."
+    end
+
+    it 'prevents plane from landing if airport is full' do
+      allow(airport).to receive(:stormy?).and_return false
+      Airport::DEFAULT_CAPACITY.times do
```
tomfuller, line 16: This is a nice creative solution to the problem of testing the capacity, however after adding the ability to set the capacity could this test be refactored to set the capacity at 0 and then try and land a plane?

Also a test around changing the capacity would be a nice touch!

tomfuller:
This is a really nice, simple solution. You clearly grasp the learning objectives for the week and reached the stretch goal. I have left some comments where there are some small improvements that can be made.

PR #2415, opened 2022-01-21
Please see my incomplete test challenge

abdulrm94:


README.md
```diff
@@ -16,29 +16,45 @@ Airport Challenge
...
+1. ~~Fork this repo, and clone to your local machine~~
+2. ~~Run the command `gem install bundler` (if you don't have bundler already)~~
+3. ~~When the installation completes, run `bundle`~~
 4. Complete the following task:
 
+Note by Abdul
+-----
+I struggled with this workshop, but gave it a good try. I only managed to get half way through with very basic tests. I would love to revisit it when I get a chance.
+
+## My flowchart
```
tomhenebury, line 36: This flowchart is **fantastic!** super useful for planning and reference.

lib/planes.rb
```diff
@@ -0,0 +1,29 @@
+class Planes
+  def planes_selector
+    all_planes = [
+      {:name=>"Driftscream", :status=>"in air"},
```
tomhenebury, line 4: great and appropriate use of symbols

lib/weather.rb
```diff
@@ -0,0 +1,14 @@
+class Weather
+  
+  def randomWeather 
+    probability = rand(1..26)
```
tomhenebury, line 4: possible ternary operator here?

spec/traffic_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'traffic_controller'
+
+describe TrafficController do
+
+  it "checks status of plane is grounded or in air" do
+    expect(subject.status).to eq("grounded").or eq("in air")
```
tomhenebury, line 6: great use of .or :)

tomhenebury (commented):
Great work!

README.md
```diff
@@ -16,29 +16,45 @@ Airport Challenge
...
+
+![Make a safe landing.](flowchart.png)
```
tomfuller, line 38: The flowchart is excellent, really useful for anyone looking at your code to understand what your thought process was!

lib/airports.rb
```diff
@@ -0,0 +1,10 @@
+airports = ["Camp Vernon Medical Center Heliport",
```
tomfuller, line 1: I like the attention to detail but I don't think is used anywhere? Its always better to only add things when you need them.

lib/planes.rb
```diff
@@ -0,0 +1,29 @@
...
+      {:name=>"Thunder", :status=>"grounded"},
+      {:name=>"Flightheat", :status=>"in air"},
+      {:name=>"Driftcore", :status=>"grounded"},
+      {:name=>"Greycry", :status=>"grounded"},
+      {:name=>"Dragonspark", :status=>"in air"},
+      {:name=>"Shy Duke", :status=>"grounded"},
+      {:name=>"False Falcon", :status=>"in air"}
+    ]
+
+  return all_planes.sample
```
tomfuller, line 14: With this approach you will always interact with a random plane. It would be more useful for an airport to be able to specify a specific plane to interact with.

lib/traffic_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  @@planes = Planes.new
+  @@status = @@planes.planes_selector.values[1]
+  @@plane_name = @@planes.planes_selector.values[0]
+  @@airport_capacity = "full"
+
+  def status
+ 
+    if @@status == "grounded"
+      print @@plane_name + " is " + @@status.to_s + "."
```
tomfuller, line 14: I would try to avoid putting useful outputs in print statements as this will be lost to the user. Its always better to put useful outputs in the return statement of a function.

lib/traffic_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      return "grounded"
+    else
+      print @@plane_name + " is " + @@status.to_s + "."
+      return "in air"
+    end
+    
+  end
+
+  def instruct
```
tomfuller, line 23: It doesn't look like this actually changes the state of a plane. I think it would also be better to have specific take off and landing functions rather than one generic one.

lib/traffic_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    if @@status == "grounded"
+      print " Time to take off!"
+      return "time to take off!"
+    else
+      print " Find an aiport to land at."
+      return "find an aiport to land at"
+    end
+  end
+
+  def check_capacity(capacity = @@airport_capacity)
```
tomfuller, line 37: I would rename the airport_capacity to default_capacity here as it is a default.

lib/traffic_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    if capacity == "full"
```
tomfuller, line 38: capacity might be better as a number that can then be checked if the number of planes currently at the airport would exceed the capacity if another one landed.

PR #2416, opened 2022-01-22
Wk3 Airport Challenge - T Henebury

tomhenebury:


lib/plane.rb
```diff
@@ -0,0 +1,18 @@
+class Plane
+
+  def initialize(in_flight = false)
+    @in_flight = in_flight
```
abdulrm94, line 4: I love how simple and easy this is to interpolate/understand, very good use of simplicity. The default in_flight value is also extremely useful incase you wish to change it

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def take_off
+    return raise ("This plane is already in flight - it cannot take off") unless @in_flight == false
```
abdulrm94, line 8: again, very simple but effective solution!

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    @in_flight = true
+  end
+
+  def land
+    return raise ("This plane is grounded, and doesn't need to land") unless @in_flight == true
+
+    @in_flight = false
```
abdulrm94, line 16: sorry for the repetition, but I am really enjoying reading your code, its so easy to follow, and the flight that you have updated your flight value when the plane has landed is great!!

spec/plane_spec.rb
```diff
@@ -0,0 +1,26 @@
+require './lib/plane'
+
+describe Plane do
+  let(:plane) { Plane.new }
```
abdulrm94, line 4: I haven't seen the use of this syntax before, looks very clean and minimal(again, your simplicity is extremely effective and the readability is perfect! Saves alot of time for someone analysing your code.

spec/plane_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  it '#can take off' do
+
+    expect(plane).to respond_to(:take_off)
```
abdulrm94, line 8: If i'm honest, my knowledge isn't that great, but we already discussed this morning, maybe this test could be more specific? From what we have learnt, respond_to just checks if a method exists? lets discuss in more detail!

spec/plane_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  it '#can land' do
+
+    expect(plane).to respond_to(:land)
+  end
+
+  it '#won\'t take off if in flight' do
+    plane.take_off
+
+    expect { plane.take_off }.to raise_error("This plane is already in flight - it cannot take off")
```
abdulrm94, line 19: again, Sir Henebury is using his knowledge to teach me! I didn't know about the raise error method. Simple and effective!

T-README.md
```diff
@@ -0,0 +1,100 @@
+# Airport Challenge - T's README
+
+## Plan
```
EdwardAndress, line 3: Nice job with the planning

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
+class Airport
+
+  STORM_ERROR = "There is a storm - no planes can take off or land".freeze
```
EdwardAndress, line 3: Nice use of a constant for this message

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def modify_hanger(new_capacity)
+    @hanger_capacity = new_capacity
+  end
+
+  def land_plane(plane)
+
+    return raise "Hanger is full" unless @hanger.length < @hanger_capacity
+    return raise STORM_ERROR unless check_weather != "stormy"
+
```
EdwardAndress, line 24: If the hangar is full and the weather is poor, at the same time, only one of the two errors will be thrown which means the other problem is masked.  How can you fix this?

lib/airport.rb
```diff
@@ -0,0 +1,48 @@
...
+  def take_off(plane)
+    return raise "#{plane} not in hanger" unless @hanger.include?(plane)
+    return raise STORM_ERROR unless check_weather != "stormy"
+
+    plane.take_off
+    @hanger.delete(plane)
+  end
+
+  def contains_plane?(plane)
```
EdwardAndress, line 37: by convention, a method ending in a `?` would return a boolean

EdwardAndress:
This is a good solution, which definitely demonstrate the learning objectives, Tom! I left a few small comments in line against the code.

PR #2417, opened 2022-01-22
Airport Challenge - Jonah Massam

JonahMassam:


MY_README.md
```diff
@@ -0,0 +1,14 @@
+Airport Challenge
```
DWatts15, line 1: I think this is a good README - it is giving insight into thought processes and challenges, and will likely help follow the commits as the 'random weather' code evolves

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
+class Airport
+
+  attr_reader :airport_storage
+  attr_reader :airport_capacity
+  attr_reader :weather
+  
+  DEFAULT_CAPACITY = 10
```
DWatts15, line 7: I think it was a good idea to make this clear as it has reduced ambiguity

lib/airport.rb
```diff
@@ -0,0 +1,33 @@
...
+  def initialize(airport_capacity = DEFAULT_CAPACITY)
+    @airport_storage = []
+    @airport_capacity = airport_capacity 
+    @weather = Weather.new
+  end
+
+  def land_plane(plane)
+    fail "Storage full" if full?
+    fail "Plane not flying" unless plane.flying?
+    fail "Too stormy" if weather.stormy?
```
DWatts15, line 18: These fail lines are useful and clear (I should have used fail in my code!). One small suggestion is to expand on the "Too stormy" line - maybe you could make it really explicit that the plane can't land in stormy conditions due to safety concerns (or something along those lines).

lib/plane.rb
```diff
@@ -0,0 +1,21 @@
...
+    airport.land_plane(self)
+    @flying = false 
+  end
+
+  def takeoff_from_airport(airport)
+    airport.send_plane(self)  
+    @flying = true
+  end
+
+end
```
DWatts15, line 21: Class and method names clear and self-explanatory

spec/airport_spec.rb
```diff
@@ -0,0 +1,73 @@
+require "airport"
+
+# srand(1) is clear weather, srand(2) is stormy weather.
+
+describe Airport do
+  context "Weather is clear" do
+    before(:each) do
```
DWatts15, line 7: Use of 'before' is a really good idea to reduce repetition of tests when the weather is Sunny or Stormy

spec/airport_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    it "Can't fly if stormy" do
+      new_plane = Plane.new(true)
+      srand(1)
+      new_plane.land_at_airport(subject)
+      srand(2)
+      expect { new_plane.takeoff_from_airport(subject) }.to raise_error "Too stormy"
+    end
+  end
+end
```
DWatts15, line 73: Tests seem to be solid and check edge cases. The test 'it' lines make sense and are clear. The tests seem to have been refactored and do not contain anything superfluous. Overall, easy to follow

spec/airport_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    it "Fails to send plane if not in storage" do
+      expect { subject.send_plane(Plane.new) }.to raise_error "Plane not in storage"
+    end
+
+    it "Stores landed plane" do
+      new_plane = Plane.new(true)
+      subject.land_plane(new_plane)
+      expect(subject.airport_storage).to include(new_plane)
```
siellsiell, line 26: One way you could refine these tests further: the idea of [testing behaviour rather than state](https://github.com/makersacademy/skills-workshops/blob/main/practicals/testing/behaviour_not_state.md). 

Here you had to make the `@airport_storage` publicly readable in order to test things -- that's often a sign that you're testing state. It's best to keep internal variables of a class private unless you have a strong reason to expose them.

Instead you could test behaviour: after landing a plane, you shouldn't be able to land it again, for example. Implicitly this tests the same thing as your test here but without accessing`@airport_storage` directly.

Similar reasoning applies to testing the `takeoff` method.

siellsiell (commented):
Hey it's Simo, a coach :)

Thanks for submitting your challenge code and apologies for the slow response. This was a great attempt at the challenge. Your code is very readable and well tested, and you've applied the idea of SRP well by splitting up areas of functionality into your three classes. 

I left a few minor comments for you to have a look at with ideas of how you could refine your code further.

Let me know if you have any questions :)

PR #2418, opened 2022-01-23
my_patch1

sapana2031:
I only manage to do 3 user story.

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,19 +1,28 @@
+require_relative 'plane'
 class Airport
-  
-  def land_plane
-  "Plane Landed"
+  attr_reader :planes_parking
+
+  def initialize 
+    @planes_parking = []
   end
 
-  def takeoff_plane
-    "Plane is flying"
+  def release_plane
+    my_plane = Plane.new
   end
-  def flying_plane
+
+  def land_plane(plane)
+    @planes_parking.push(plane)
+    "Plane has Landed"
+  end
+
+  def takeoff_plane(plane)
+    @planes_parking.delete(plane)
     "Plane is not in the airport"
   end
-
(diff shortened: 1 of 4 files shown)


spec/plane_spec.rb
```diff
@@ -0,0 +1,3 @@
+require 'plane'
+describe Plane do
```
Katerina-codes, line 2: If the class has no state (instance variables) or behaviour (methods) in it then it should be deleted.

spec/airport_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'airport'
+
+describe Airport do 
+  it "releases a plane" do
```
Katerina-codes, line 4: Nice, clear test names.

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+    @planes_parking = []
+  end
+
+  def release_plane
+    my_plane = Plane.new
+  end
+
+  def land_plane(plane)
+    @planes_parking.push(plane)
+    "Plane has Landed"
```
Katerina-codes, line 15: We don't need to display a string to confirm landing. As a programmer we need to know that this method works as we expect. We check this in our test, what could we test for to know the plane has landed?

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def takeoff_plane(plane)
+    @planes_parking.delete(plane)
+    "Plane is not in the airport"
```
Katerina-codes, line 20: We don't need to display a string to confirm take off. As a programmer we need to know that this method works as we expect. We check this in our method's test, what could we test for to know the plane has taken off?

PR #2419, opened 2022-01-23
Airport Challenge

benfielding74:
Challenge incomplete
I still need to correctly implement stormy weather and airport capacity methods to fulfill the user stories
I also need to refactor and properly write my tests for these functions


lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+  attr_accessor :capacity
+  attr_reader :planes
+  DEFAULT_CAPACITY = 10
+
+  def initialize
+    @capacity = DEFAULT_CAPACITY
+    @planes = []
+  end
+
+  def land_plane(plane)
```
Luke-Collins, line 14: SRP - definition lands (adds to array) and flags error, which is two processes.

Though I, myself, am not sure if this is acceptable for SRP or not.

Luke-Collins, line 14: SRP - definition lands (adds to array) and flags error, which is two processes.

Though I, myself, am not sure if this is acceptable for SRP or not.

Same applies for take_off

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+    fail 'Bad Weather' if stormy
+    @planes << plane
+    "Successful Landing"
+  end
+
+  def take_off(plane)
```
Luke-Collins, line 20: Quite sure this violates SRP, as there are multiple processes and it is quite long.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+        @planes.delete_at(i)
+      else  
+        "This plane is not at the airport"
+      end
+    end 
+  end
+
+  private
+
+  def stormy
```
Luke-Collins, line 34: Works to highlight only stormy, though other weathers would exist in a practical setting.
Addendum: Just saw readme comment and empty weather class, so this was clearly intended.

It suits purpose perfectly and is very neat, well done.

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+  end
+
+  describe '#take_off' do
+    it { is_expected.to respond_to :take_off }
+
+    it 'checks if there is a specific plane to take off' do
+      airport = Airport.new
+      panam = Plane.new
```
Luke-Collins, line 43: 'panam' is not a clear object for Plane.new.


If reading panam without the 'Plane.new' being right next to it, ie referenced elsewhere in code, it would not be clear what it is supposed to represent.

spec/plane_spec.rb
```diff
@@ -0,0 +1 @@
+require 'plane'
```
Luke-Collins, line 1: Unnecessary file?

README.md
```diff
@@ -1,89 +1,37 @@
-Airport Challenge
-=================
+** Airport challenge
```
Luke-Collins, line 1: Was a little confusing and messy to read your README within the default README..

Would advise having a separate file for clarity.

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
+require './lib/plane.rb'
```
Luke-Collins, line 1: Better not to use file directions, and instead just /require 'plane'/ as if the file is moved, it will not be found if the path is specified.

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'airport'
+
+describe Airport do
+
+  let(:plane) { double :plane }
```
Luke-Collins, line 5: What is this line for?

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  it { is_expected.to respond_to :capacity }
+
+  it 'allows the system designer to change capacity' do
+    airport = Airport.new
+    airport.capacity = 10
```
Luke-Collins, line 11: Your default capacity is also 10, so this doesn't test properly that it can change?

lib/airport.rb
```diff
@@ -0,0 +1,39 @@
...
+
+class Airport
+
+  attr_accessor :capacity
+  attr_reader :planes
+  DEFAULT_CAPACITY = 10
+
+  def initialize
+    @capacity = DEFAULT_CAPACITY
```
Luke-Collins, line 10: I think this would either overwrite any accessor capacity, or could otherwise be refactored into a parameter?

planning/land_plane method
```diff
@@ -0,0 +1 @@
...CB0qCOwsgWfrTs6mbSZfaopR/gZ7j5m9WTTE/H+TKYiaRC2ycVET61UxfE4xpVvL3a4g4djkuWbtPJ54WAyMwoG0afERPt5ub69uNcbGxn3D0YF4/fxS9PAEUiwl/ZdUEsC4Zfrx/BAKYgs4urGxzefow4BePhJhQMUF6yFGLKMGPBWBsqhcu9RErqqzIgAi5nr93uMuc7RBOLBcQehZRdr3nOonrUXgXAosrJkyNycRQIaFwfxIDtUCWeSTEPqTkwFREA10XxhDeOBfVpDStqgiZWzHrUsiA8Can6LkEnBUE5q/Afy7hL2EJbeaQosqwdrGDA4CuSukFwXFgQ/4ahdVKne8YhwLzB0jpQUQXiJkxaxbJ3slf9DKubFw4JhZ8YUJi/MyQDqQv6OnH7UQgAtR7qo+jc9TV5SRhfwOcoukN9tLIF+FiKPVwKVg57yIzs2jVMEoxjH+3kzZNO7UIj2x9GSo+NAl2zvz3Yi1oGKCE9xRxV6AE2wijmMJzYgm4U2lZm/MFQlo93oH6VbMl9xfCzmE2p+bdekvvNoZ07slxSlvUB6UUXUJ9y7bxkO8iG+BT4oHASVH+C5nUmacXEjHqJFLUKyRi9rl/VCfHIUZLb1YHJRFDgv1GHJT4SsmgbBqhW1feeQitMcIk7Vy3Dbi0S83Kk6W21U2zG+efmw3nP7aCake833w371em5v2Llt6/GtL5FyedP8BpamCLGimDN20Ze+B1dTk615b34P+5a6/aEvsJU/WUGphcRqYtpahvTcmkTc8JAYeKQshzA1Gl50aKoUjB5fnheSOL+Tp9VshG0GkeYiX5xBpgYHn0IPnGmHevRfomz36uht69yunnNQg+0GLdULwdu+xi3tXY15CmYypDLyHa5KisG4qSOEVfNEy36jnNOblDnNo2J6ENeopT5azEMf6JKCa+TPWjYyC9ye9HshM972Llc/I3CEPuGQRYj+yOOcBdJJZHeWQ2R3dkHoZaGFzk8Iu3PzPy6m/RXVq53NdqB1HrkJKKvwfGnVjS5C3RDwjtKD9Pcy5KD9cuLAMn+yKfiU59kp9ANDU/Rxan9uGiSea3tQtTYksgYAYhsh1c7G/jaIzng+YnMuMIvss+zweanGkKqsEALAuGtBoH/FH2ZwpxYh8Gk4wDdd5fHLBq+YtStuEt//PSr/4A</diagram></mxfile>
```
Luke-Collins, line 1: Purpose of file?

planning/take_off method
```diff
@@ -0,0 +1 @@
...fied="2022-01-21T17:24:51.118Z" agent="5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.71 Safari/537.36" etag="KnH94om06_N0DPl_UJQN" version="16.4.3" type="google"><diagram id="_87jV86NYZ9H6MGAtwBh" name="Page-1">5VjbcpswEP0aPyZjLsbkNY7bZiadSceZ5jGjwBrUCJYKYXC/vpIRYKzYdSd26SQvDDpa3c6eXS2MnFlSfeYki79iCGxkj8Nq5NyMbPvK9eRTAesamFhXNRBxGtaQ1QEL+gs0ONZoQUPIe4YCkQma9cEA0xQC0cMI51j2zZbI+qtmJAIDWASEmegjDUVco7497fAvQKO4Wdny9PkS0hjrk+QxCbHcgpz5yJlxRFG/JdUMmOKu4aUe92lPb7sxDqk4ZkCSe9W37xcTehuTe+DF8533cOHUs6wIK/SB9WbFumEAQkmIbiIXMUaYEjbv0GuORRqCWmYsW53NHWImQUuCP0CItfYuKQRKKBYJ0731mmqhvWfTUI4FD+DAgRqNEB6BOGBntx6QygVMQPC1HMeBEUFX/X0QraGotetoli+a6b9g3TJYv5XNksjDKnmr4WkueBGoNlFMMJJC0yfIi3rH5VIJmmOijFL1oDyT7BselFPhSytfu8+9lGWm7JIqUgF8uWRYBjHh4jKXJIonZVPGVMAiIxvmS2nW+mwFXEB12Gsmy3qA04SGzg2ObpZdoFljjcVbQeaNz+SXyXuLBvvIaHCHjAbbYF0p/Gmj7x32O273yXJb2YbsT6FZq69Za/qKaO1/KdrpexOte6RovSFFe/VRWfcHvTjNm/OD0G6Nh+TdNQuWXO06Bg79EmXHGYJTkkYb1v+Qrfup/Ty5292tN8zc7b2Suv1zpW7PoHVJqBo1sr2fhfowuH7QFFPFd4od0Y2BQfkbeV5SxmbIkG9mc8IJ+KHbWm71+Paz43mn8czE2/GMNfSt6huuWRRBAHne8869rstjkuvCPG0q873+eWMNM4h//P/NPc0GDoXOIxCVoJrgyQXyZH2+wDlFgtopLt0zFpey2f172PRt/cBx5r8B</diagram></mxfile>
```
Luke-Collins, line 1: Same here, what is the reason for this file?

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    expect(airport.capacity).to eq(10)
+  end
+
+  describe '#land_plane' do
+
+    it { is_expected.to respond_to :planes }
+    it 'recognises that a plane has landed' do
+      airport = Airport.new
```
Luke-Collins, line 20: Can instead put the airport = Airport.new line under the describe line, to allow it for the entire test desc, rather than each individual it statement.

Lucifer-Kett (commented):
Generally good, some things I didn't recognise and can learn from, keep up the great work!

Look forward to working with you, Lynsey and Barrington in Leeds! (4 of us now!)

lib/airport.rb
EdwardAndress: What would you suggest as a refactor, @Luke-Collins? :)

EdwardAndress:
Hey Ben! I see that @Luke-Collins has already given you some useful feedback around what could be improved here.  In which case, all I can really add is that it's a great effort which demonstrates the learning objectives for this module.  Well done picking up TDD, if it was new to you! :)

lib/airport.rb
EdwardAndress: Nice! I definitely agree about the weather.

Regarding land / takeoff, I can see good arguments for there being associated methods in both classes, since both Airport and Plane need to do something in those scenarios.  The Plane might change state from "flying" to "landed", for example, and the Airport needs to control who can / can't land / takeoff.  It's possible to get into a bit of a mess here though :)

Side note: I'd normally understand 'child' and 'parent' to infer inheritance, but I don't think that's how you're using the terms above.

Lucifer-Kett: Yes, I meant inheritance.

PR #2420, opened 2022-01-23
Airport Challenge, Fred Halliwell

FredHalliwell:


README.md
```diff
@@ -13,77 +13,65 @@ Airport Challenge
...
 
-Note that is a practice 'tech test' of the kinds that employers use to screen developer applicants.  More detailed submission requirements/guidelines are in [CONTRIBUTING.md](CONTRIBUTING.md)
+the argument will be whichever plane object you are trying to check for. The method will return true if the plane is at the airport, but false if not.
 
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first.
 
-* **Submit a pull request early.**
 
-* Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+# Thank you for reading, enjoy :)
```
tomfuller, line 77: Really great to see instructions on how to the run the code! I would personally leave in the user stories so that people looking at your code know what requirements you were trying to satisfy.

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+class Airport
+  attr_reader :landed_planes, :default_capacity, :capacity
+
+  attr_writer :weather, :landed_planes
+
+  DEFAULT_CAPACITY = 20
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @weather = Weather.new
+    @weather = @weather.random_weather
```
tomfuller, line 12: It's better not to overwrite variables if you can avoid it, maybe this could be called current_weather?

spec/airport_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  # has landing method
+  it { is_expected.to respond_to(:land) }
+
+  it 'plane can land' do
+
+    @airport.weather = :sunny
+
+    expect(@airport.land(@plane)).to eq (@airport.landed_planes)
```
tomfuller, line 24: It would be better to explicitly say what you would expect the new array to look like rather than reference the actual array. This is because if the plane isn't added to the array by the land function this test would still pass.

spec/airport_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+  # has take_off method
+  it { is_expected.to respond_to(:take_off) }
+
+  it 'plane can take off' do
+    
+    @airport.weather = :sunny
+
+    @airport.landed_planes << @plane
+
+    expect(@airport.take_off(@plane).include?(@plane)).to eq(false)
```
tomfuller, line 36: For readability here and a more robust test I would separate this out and in the it block call the take off function on a plane and then expect that landed_planes doesnt include it anymore.

tomfuller:
This is a really nice solution, you clearly grasp the learning objectives and reached the stretch goal. I have left some comments where there are small improvements that can be made. It looks like you have a decent understanding of RSpec and what needs to be tested, there are a few cases where your tests could be more robust. I would challenge you when you're writing tests to think about functionality you want to be covered by that test and whats the best possible way to achieve that. Some of that will come with experience though!

PR #2421, opened 2022-01-23
Sonali - Airport challenge 

sonalikr:


spec/plane_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    ap = Airport.new(1, "sunny")
+  
+    expect(plane.land(ap)).to eq("landed!")
+  end
+  
+  it 'cannot land due to more flights' do
+    plane = Plane.new
+    ap = Airport.new(2, "njjnnjn")
+    plane.land(ap)
+    plane.land(ap)
```
Edenlsw, line 23: DRY- I'm not sure if you need both line 22 and 23 in the test as they do the same thing.

spec/airport_spec.rb
```diff
@@ -0,0 +1,50 @@
+require './lib/airport.rb'
+
```
EdwardAndress, line 2: For some of the methods being tested here, the most important thing is what they do as "side effects" rather than what they return once complete.  So, for example, the important thing about `#add_flight` is that it increments @current by 1.

You effectively test for this below but think it's still worth reflecting on.

spec/plane_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    plane = Plane.new
+    ap = Airport.new(1, "sunny")
+    plane.land(ap)
+
+    expect(plane.take_off(ap)).to eq("took off! no longer in the airport")
+  end
+
+  it 'cannot take-off due to no flights' do
+    plane = Plane.new
+    ap = Airport.new
```
EdwardAndress, line 53: In a larger codebase, variable names like `ap` will be confusing. Make sure all your variable names mean something to the reader :)

PR #2423, opened 2022-01-23
Airport Challenge

JamesBuckle93:

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,5 +1,20 @@
+require 'plane'
+
 class Airport
+  attr_accessor :capacity
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
 
   def land(plane)
+    fail "Cannot land, airport full!" if @planes.length == @capacity
+    @planes << plane
   end
+
+  def take_off(plane)
+  end
+
 end  
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,5 @@
+class Airport
+
+  def land(plane)
+  end
+end  
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,15 @@
+require 'airport' 
+  
+describe Airport do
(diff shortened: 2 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,20 @@
+require 'plane'
+
+class Airport
```
MathewGrimward, line 3: Very Readable code, Can easily see what each method is doing and they are all named appropriately :)

spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'airport' 
+  
+describe Airport do
+  it 'allows a plane to land' do
+    airport = Airport.new
```
MathewGrimward, line 5: Instead of creating a new instance of Airport every time, you could do   subject { Airport.new } under describe Airport then every time you use subject it will automatically refer to this.

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+  attr_accessor :capacity
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    fail "Cannot land, airport full!" if @planes.length == @capacity
```
MathewGrimward, line 13: Nice use of .length, I wasn't aware that this could be used like this and I think it works great!

spec/airport_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'airport'
```
MathewGrimward, line 1: Great layout, spacing and indentation 👍

lib/airport.rb
```diff
@@ -0,0 +1,20 @@
...
+    @planes << plane
+  end
+
+  def take_off(plane)
```
MathewGrimward, line 17: Could use @planes.pop here to remove the last plane in the airport so that it confirms that the plane has left by removing the last plane of the array

MathewGrimward (commented):
This is a great easy to read code, which does exactly what it says it does. The indentation, spacing and capitalisation is all spot on. It follows the SRP well and DRY. This is great code!

PR #2424, opened 2022-01-24
Airport challenge

jconnollyweb:


spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
+require './lib/airport'
+
+describe Airport do
+  it 'responds to plane_take_off' do
+    airport = Airport.new(20)
```
Barrington-Brook, line 5: Since we are inside the Airport describe block you could use `subject` instead of creating a new Airport instance in each example.

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
...
+  def initialize(planes = [])
+    @planes = planes 
+    
+  end 
+
+  def plane_land(plane)
+    @planes.push(plane)
+  end
+
+  def plane_take_off(plane)
```
Barrington-Brook, line 16: This method is well-written and easy to read so you can tell what it does at a glance.

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
+class Airport
+
+  def initialize(capacity)
```
Barrington-Brook, line 3: The instance variables could be moved to one `initialize` method instead of two.

spec/airport_spec.rb
Katerina-codes: If the class has no state (instance variables) or behaviour (methods) in it then it should be deleted.

lib/plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
```
Katerina-codes, line 1: If the class has no state (instance variables) or behaviour (methods) in it then it should be deleted.

spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    expect(airport).to respond_to(:plane_take_off)
+  end
+
+  it 'responds to plane_lands' do
+    airport = Airport.new(20)
+    expect(airport).to respond_to(:plane_land)
```
Katerina-codes, line 11: At the moment, this only checks the method exists. I'd suggest testing that this method works as we expect. What could we test for to know the plane has landed?

spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  it 'does not allow planes to land when airport is full' do
+    airport = Airport.new(20)
+    plane = Plane.new
+    20.times do
+      airport.plane_land(plane)
+    end 
+    expect { airport.plane_land(plane) }.to raise_error 'Can not land plane as airport is full '
+
+  end
```
Katerina-codes, line 22: I think we're missing some tests here as we have more code in the Airport class.

PR #2425, opened 2022-01-24
Airport Challenge - Mathew Grimward

MathewGrimward:


spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+          expect { subject.land(plane) }.to raise_error 'It is stormy, you cannot land'
+        end
+
+        it 'allows system designer to set capacity' do
+          expect(subject.capacity).to eq Airport::DEFAULT_CAPACITY
+        end
+      end
+    end
+  end
+end
```
JamesBuckle93, line 46: As a whole file very easy to read

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    fail 'The Airport is full, you cannot land' if full?
+    fail 'It is stormy, you cannot land' if stormy?
+    @planes << plane
+  end
```
EdwardAndress, line 17: What if the airport was full and the weather was stormy at the same time? Would it be useful to surface both errors? Does your solution do that? :)

lib/airport.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def takeoff
+    fail 'It is stormy, you cannot take off' if stormy?
+    @planes.pop
+  end
```
EdwardAndress, line 22: Could you make a specific plane takeoff instead of the last one in the planes array?  Would that be an improvement?

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+describe Airport do
+  subject { Airport.new }
+
+  describe "#land" do
+    it 'can land a plane' do
+      plane = Plane.new
+      allow(subject).to receive(:stormy?) { false }
+      expect(subject.land(plane)).to eq [plane]
+    end
```
EdwardAndress, line 11: Nice use of a stub!

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+          allow(subject).to receive(:stormy?) { false }
+          Airport::DEFAULT_CAPACITY.times { subject.land(plane) }
+          expect { subject.land(plane) }.to raise_error 'The Airport is full, you cannot land'
+        end
+
+        it 'does not allow planes to land when stormy' do
+          plane = Plane.new
+          allow(subject).to receive(:stormy?) { true }
+          expect { subject.land(plane) }.to raise_error 'It is stormy, you cannot land'
+        end
```
EdwardAndress, line 38: Is it worth also verifying that the plane has not landed (does not now appear in the planes array)?

EdwardAndress: Same for the one above (where the airport is full).

spec/airport_spec.rb
```diff
@@ -0,0 +1,46 @@
+require './lib/airport'
```
EdwardAndress, line 1: You've clearly got a good grasp of TDD, including some of the more complex stuff, like stubbing.  Nice work.

PR #2426, opened 2022-01-24
Airport Challenge - Barrington Brook

barringtonbrook:


lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+      if @planes.include?(plane)
+        raise "plane already landed"
+      else
+        if full?
+          raise "no room"
+        else
+          @planes.push plane
+          plane.landed = true
+        end
+      end
```
jconnollyweb: This is all really clear and easy to understand

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def remove(plane)
+    plane.landed = false
+    @planes.delete_if { |p| p.equal? plane }
+  end
+
+  def full?
+    @planes.length == @capacity
+  end
+
```
jconnollyweb, line 43: This is all clean easy to read code, each section is in nice and unclustered

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather  
+  def stormy?
+    rand(6).zero?
+  end
+end
```
jconnollyweb, line 5: Small, simple and easy to read

spec/airport_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    allow(subject.weather).to receive(:stormy?) { true }
+    expect { subject.land(plane) }.to raise_error "Cannot land until weather conditions improve" 
+  end
+
+  it "raises an error if stormy and plane tried to takeoff" do
+    subject.land(plane)
+    allow(subject.weather).to receive(:stormy?) { true }
+    expect { subject.take_off(plane) }.to raise_error "Cannot take off until weather conditions improve"
+  end
+
```
jconnollyweb, line 43: Excellent use of functions running through all your code. You know your stuff.

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
+require 'plane'
+require 'weather'
+
+class Airport
+
+  AIRPORT_CAPACITY = 10
```
tomfuller, line 6: I would probably rename this DEFAULT_CAPACITY as it is only the default that can be overridden.

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+  private
+
+  def add(plane)
+    @planes.push plane
+    plane.landed = true
+  end
+
+  def remove(plane)
+    plane.landed = false
+    @planes.delete_if { |p| p.equal? plane }
```
tomfuller, line 37: This is cool and in other languages you would need to do something like this, but in ruby theres is an array delete() function that you might find nicer.

lib/plane.rb
```diff
@@ -0,0 +1,9 @@
+class Plane
+  
+  attr_accessor :landed
+  alias_method :landed?, :landed
+  
+  def initialize(landed = false)
+    @landed = landed
```
tomfuller, line 7: This value doesn't seem to be used anywhere, did you have something in mind for it?

tomfuller:
This is a really nice solution and you clearly grasp the learning objectives. I have left a couple comments where there are small improvements but overall it was good. Your tests are robust and appropriate. The only thing I would question is whether the landed value in the plane is necessary?

PR #2427, opened 2022-01-24

Will-Diable:
Unfinished airport challenge

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,3 +87,27 @@ Finally, don’t overcomplicate things. This task isn’t as hard as it may seem
 * **Submit a pull request early.**
 
 * Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+
+Planning Stage:
+
+controller
+plane - land in airport
+-----------
+controller
+plane - takeoff - confirm
+-----------
+controller
+plane - prevent landing when full
+-----------
+designer
+set max capacity
+fallback to default when 
+-----------
+controller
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -87,3 +87,31 @@ Finally, don’t overcomplicate things. This task isn’t as hard as it may seem
...
+fallback to default when 
+-----------
+controller
+prevent takeoff during stormy weather
+random weather array (in favour of clear/sunny)
+-----------
+as above, but reversed
+-----------
+use a stub to override random weather to ensure consistent test behaviour
+
```
davecarroll82, line 113: Very clear plan, nicely set out and easy to follow.

EdwardAndress:
Hey Will, this is a good attempt that shows you're picking up the basics of TDD. The main thing to reflect on in terms of improvements is your implementation of weather. I'd expect there to be some element of randomness there, perhaps using sample, like this...

```
fail "Too stormy" if WEATHER.sample == "stormy"
```
I'd also expect there to be just one method for landing and one method for taking off.

But, at the moment, it looks like user would choose a specific land or takeoff method taking into account what the weather is doing.  Maybe it's worth reflecting on your interpretation of the user stories..?

The tricky part would then be trying to test those methods!  For that, you can use stubs or you can use `srand()`.  If you'd like to learn more about what srand does, I recommend this short video on pseudorandom numbers: https://www.khanacademy.org/computing/computer-science/cryptography/crypt/v/random-vs-pseudorandom-number-generators.  Once you've watched it, you'll know what I mean when I say that `srand` sets the "seed" and fixes the pseudorandom number sequence to be the same every time.

PR #2428, opened 2022-01-24
airport challenge - Dave Carroll

davecarroll82:


Will-Diable:
Appreciate:
Very readable: Intuitive method/class names and perfectly spaced/indented.
Pleasure to read, very clean.

Suggest:
Perhaps add a message to the take_off method to explain the plane has taken off. I think this would meet the second user story better

spec/airport_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+plane = Aeroplane.new
+
+describe airport do
+ 
+  it 'can land the plane at the airport' do
+    expect(airport.land(plane)).to eq [plane]
+  end
+
+  it 'lets the plane take off and confirm' do
+    expect(airport.take_off).to eq plane
```
Katerina-codes, line 13: Does this test give your security that the **take_off** method works as you want it to? What value could you assert on to make you confident that the plane has indeed left the Airport?

spec/airport_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'Airport'
+
+airport = Airport.new
+plane = Aeroplane.new
+
+describe airport do
+ 
+  it 'can land the plane at the airport' do
+    expect(airport.land(plane)).to eq [plane]
```
Katerina-codes, line 9: Same feedback for this test as the test below :)

lib/Aeroplane.rb
```diff
@@ -0,0 +1,3 @@
+class Aeroplane
+
```
Katerina-codes, line 2: Don't create a class and leave it empty. If it doesn't have state (instance variables) or behaviour
(methods) the it shouldn't be here -.

lib/Airport.rb
```diff
@@ -0,0 +1,16 @@
...
+ 
+  def initialize
+    @planes = []
+  end
+ 
+  def land(plane)
+    @planes << plane
+  end
+
+  def take_off
```
Katerina-codes, line 13: Nice, clear method names

spec/airport_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'Airport'
+
+airport = Airport.new
+plane = Aeroplane.new
+
+describe airport do
+ 
+  it 'can land the plane at the airport' do
```
Katerina-codes, line 8: Good, clear test names.

Katerina-codes:
Nice work Dave, I would like to see the feedback I have given implemented when you get a chance. Don't forget to commit after each passing test.

spec/airport_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'Airport'
+
+airport = Airport.new
```
Katerina-codes, line 3: This line and line 4 should be inside each test and not up here.

PR #2430, opened 2022-01-24
Airport challenge now completed - Jaabir Yusuf

JaabirM:


lib/airport.rb
```diff
@@ -0,0 +1,36 @@
+require_relative "plane"
+
+class Airport
+  MAX_CAPACITY = 10
+
+  attr_reader :planes
+  attr_reader :capacity
```
OliverCutting, line 7: the two attr_readers could be one line i.e.

attr_reader :planes, :capacity

planning
```diff
@@ -0,0 +1,44 @@
...
+The methods are:
+
+Land
+Take off
+Is airport full
+Is weather stormy?
+
+Constant variable:
+
+Capacity - this can be changed as per the user story from the system designer
```
OliverCutting, line 44: Good clear and concise plan

spec/airport_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+  end
+
+  describe "#stormy weather" do
+    it "raises an error when landing in a stormy weather" do
+      allow(subject).to receive(:stormy?).and_return(true)
+      expect { subject.land_plane Plane.new }.to raise_error("Stormy weather")
+    end
+    it "raises an error when taking off in a stormy weather" do
+      allow(subject).to receive(:stormy?).and_return(true)
```
OliverCutting, line 36: DRY, were using the allow line to set the weather a few times here in our tests. Could we not have used a before - do block to set the weather for all the weather specific tests?

OliverCutting (commented):
The code is very well laid out, very easy to follow and sticks to SRP throughout. The plan is clear and concise. I think the main area where we can look for improvement would be in the tests as some refactoring can be done to break the larger test blocks down into smaller chunks which follow DRY principles better.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require_relative "plane"
+
+class Airport
+  MAX_CAPACITY = 10
```
tomfuller, line 4: As this can be set on creating I would rename MAX_CAPACITY to DEFAULT_CAPACITY for clarity.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  
+  attr_accessor :capacity
+
+  def initialize(capacity=MAX_CAPACITY)
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land_plane(plane)
+    fail "Airport full" if full?
```
tomfuller, line 16: It might be better to give a bit more detail in the errors, e.g. instead of ‘Airport full’ could be ‘Could not land plane as Airport is full’ or something like that.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    fail "Stormy weather" if stormy?
```
tomfuller, line 17: Same as above comment for “Stormy weather” error.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    planes.pop
+  end
+
+  private
+
+  def full?
+    planes.count >= capacity
+  end
+
+  def stormy?
```
tomfuller, line 32: I think  that ‘stormy?’ probably belongs in a weather class as generating weather doesn’t feel like behaviour that belongs to the Airport class.

planning
```diff
@@ -0,0 +1,44 @@
+THE AIRPORT CHALLENGE PLAN - This document is my plan of the Airport challenge before programming it.
```
tomfuller, line 1: The planning document is excellent, great to see your thoughts! A diagram showing how the classes and functions interact with each other would be an added bonus.

spec/airport_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    allow(subject).to receive(:stormy?).and_return(false)
+    plane = Plane.new
+    subject.land_plane(plane)
+    expect(subject.takeoff_plane).to eq(plane)
+  end
+
+  it "has a max capacity using default number" do
+    expect(subject.capacity).to eq(Airport::MAX_CAPACITY)
+  end
+
```
tomfuller, line 21: Would be good to see a test where you change the capacity of an airport.

spec/airport_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    subject { Airport.new(100) }
+    it "have initialized max capacity using non-default number" do
+      allow(subject).to receive(:stormy?).and_return(false)
+      subject.capacity.times do
+        subject.land_plane(Plane.new)
+      end
+      expect{ subject.land_plane(Plane.new)}.to raise_error "Airport full"
+    end
+  end
+  # describe "#randomness test = check if weather failure condition works" do
```
tomfuller, line 51: I know that the randomness test is commented but I don’t think you need it. Running something a large number of times to test randomness is inefficient and unnecessary. The rand function is provided by Ruby and so it’s safe to trust that it works.

tomfuller:
Overall this is good simple solution. You clearly grasp the learning objectives for this week. I have left some comments where there are some small improvements that can be made. Mostly I would think about how a separate weather class might work and how it would interact with the airport class.

PR #2432, opened 2022-01-24
Airport Challenge - Dan Watts

DWatts15:


lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  # initialize the Airport class with a capacity of 5 by default
+  def initialize(capacity = 5)
+    @capacity = capacity
+  end
+
+  # generate an instance of weather
+  def generate_weather
+    @weather = Weather.new
+    @weather = @weather.random_weather
+  end
```
JonahMassam, line 18: This function isn't really needed since you could create the Weather object in the initialize function to access random_weather whenever you want.

lib/weather.rb
```diff
@@ -0,0 +1,19 @@
+# create class Weather with attribute accessors for current_weather
+# and random_number
+class Weather
+  attr_accessor :current_weather, :random_number
```
JonahMassam, line 4: current_weather isn't used anywhere

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    expect { airport.take_off(plane) }.to raise_error
+  end
+
+  # edge case (?) check to stop a landed plane from accepting land method
+  it "prevents a landed plane from trying to land again" do
+    airport = Airport.new
+    plane = Plane.new("landed")
+    expect { airport.land(plane) }.to raise_error
+  end
+end
```
JonahMassam, line 75: Tests are clear, well commented and easy to follow.

lib/airport.rb
EdwardAndress: This is a fair comment but can you elaborate, @JonahMassam?  What's the disadvantage of using this separate method to generate the weather?

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  # land a plane in the airport and decrease available capacity by 1, 
+  # unless weather is stormy or capacity is zero (ie. airport full)
+  def land(plane)
+    raise if @weather == "Stormy"
+    raise if @capacity.zero?
+    raise if plane.status == "landed"
+    @capacity -= 1
+  end
```
EdwardAndress, line 27: Could you refactor this to surface all errors when there are multiple problems?  I.e. When the weather is stormy and the capacity is zero?  At the moment, only one of those two problems would be surfaced initially.

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  # instruct plane to take off and increase available capacity by 1, 
+  # unless weather is stormy
+  def take_off(plane)
+    raise if @weather == "Stormy"
+    raise if plane.status == "flying"
+    @capacity += 1
+  end
```
EdwardAndress, line 35: I like what you're doing with capacity here.  It's a bit different to what we normally see so that makes it interesting.  There's a bug though – you could take off 100 planes and that would increase the capacity by 100... perhaps that's beyond what the maximum should be? How can you retain this approach whilst enforcing a max capacity?

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    expect(subject.capacity).to eq(4)
+  end
+
+  # create Plane instance, call take_off method, and expect the capacity 
+  # to increase to 6 from default of 5.
+  it "allows ATC to instruct a plane to take off" do
+    plane = Plane.new("landed")
+    subject.take_off(plane)
+    expect(subject.capacity).to eq(6)
+  end
```
EdwardAndress, line 21: You can make this test more robust by checking that the capacity has changed by -1.

It's possible to do so in a very laborious way, where you first check the capacity, then execute your method, then check that the capacity has changed by the desired amount.

Alternatively, you can do this

```
expect {subject.takeoff(plane)}.to change {subject.capacity}.by -1
```

CAVEAT: I am prone to making small syntax errors when I write code like this, without being able to execute it (but hopefully you get the gist).

The exact implementation is RSpec specific but the broader point is more general and related to testing behaviour (that something changes by a certain amount) rather than state (that the capacity ends up being a specific number).

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  it "can receive a weather condition" do
+    expect(subject).to respond_to(:generate_weather)
+  end
+
+  # check that either Sunny or Stormy is generated
+  it "can generate random weather" do
+    airport = Airport.new
+    options = ["Sunny", "Stormy"]
+    expect(options).to include(airport.generate_weather)
+  end
```
EdwardAndress, line 41: What if generate_weather always returned "Sunny"?

Would the test pass? Would the program actually be broken?

spec/airport_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  # use stub to force the generate weather method to return "Stormy", 
+  # which should raise exception
+  it "prevents planes taking off when weather is stormy" do
+    airport = Airport.new
+    plane = Plane.new("landed")
+    # allow(airport).to receive(:generate_weather).and_return("Stormy")
+    airport.weather = "Stormy"
+    expect { airport.take_off(plane) }.to raise_error
+  end
```
EdwardAndress, line 51: nice use of the raise_error matcher

EdwardAndress:
Hi Daniel, this is a good solution that I think demonstrates the TDD learning objectives.  There are a few places where you might be able to improve things – I've left comments about those! :)

PR #2433, opened 2022-01-24
Airport challenge Shalini Nangia

shalini9:


lib/airport.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  attr_reader :planes_parking
+
+  def initialize 
+    @planes_parking = []
+    @weather1 = Weather.new
+  end 
+
+  def release_plane
+    plane1 = Plane.new
```
CristinaGonzalezVaro, line 14: Could plane1 on line 14 be @plane1 instead, so that on line 34 you wouldn't need to assign again plane1 = release_plane?

spec/airport_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "./lib/airport.rb"
```
CristinaGonzalezVaro, line 1: You could require the file by doing require 'airport' as you did in your airport file with require_relative

lib/airport.rb
```diff
@@ -0,0 +1,57 @@
...
+      land(plane1)
+    elsif @weather1.conditions(1)
+      land(plane1)    
+    elsif @weather1.conditions(2)
+      land(plane1)     
+    elsif @weather1.conditions(3) 
+      not_safe_to_land(plane1)
+    else
+      "weather unknown - do not land"
+      end
```
CristinaGonzalezVaro, line 45: Line 45 indentation 1 too deep

spec/weather_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  expect(weather.conditions(2)).to eq("drizzle")
+  end
+
+  it "generates random weather stormy" do
+    weather = Weather.new
+    allow(weather).to receive(:rand).and_return(3)
+    expect(weather.conditions(3)).to eq("stormy")
+    end
+
+end
```
CristinaGonzalezVaro, line 17: The file names are easy to understand and the overall structure is easy to follow.
SRP is also met in the methods.

Looking at the User Stories, I believe the below haven't been covered yet in the code/tests above:

> As an air traffic controller 
> To ensure safety 
> I want to prevent landing when the airport is full 

> As the system designer
> So that the software can be used for many different airports
> I would like a default airport capacity that can be overridden as appropriate

lib/airport.rb
```diff
@@ -0,0 +1,57 @@
+require_relative 'plane'
+require_relative "weather"
+
+class Airport
```
siellsiell, line 4: According to the requirements in the user stories, I would expect to have a single `land` method that prevents landing if the weather is bad or the airport is full, and a single `takeoff` method that removes the plane from parking. 

At the moment you have 3 methods that are in charge of landing planes in different ways. This is not great because a user of the class could still land a plane if they wanted to by callinng `land` when the weather is bad. Ideally, your class should prevent that. All methods that shouldn't be accessible to a user of the class should be private.

So the task is: how could you merge the following into a single method?
- checking the weather and returning an error if it's bad
- checking whether the airport is full and returning an error if it is
- adding a plane to the `@planes_parking` only if the two previous steps didn't return an error.

lib/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
+  def working?
```
siellsiell, line 2: What requirement does this method fulfill? Is it necessary?

lib/weather.rb
```diff
@@ -0,0 +1,16 @@
+class Weather
+
+  def conditions(n)
+    n = rand(4)
+      if n == 0
```
siellsiell, line 5: The number you are passing in to the `conditions` method doesn't have any effect because you then set `n` to the result of `rand(4)`, thus overwriting the value of the parameter `n`. This is also the reason why your weather checking code doesn't work currently.

siellsiell (commented):
Hey it's Simo, a coach :)

Thanks for submitting your challenge code and apologies for the slow response. Your code shows that you've thought about SRP -- it makes a lot of sense to split the weather-related functionality out into a separate class because that allows you to test it separately from the Airport functionality.

However, the design of your methods needs some rethinking, as currently it doesn't meet the requirements set out in the user stories. It would be a good idea to go over this code with a coach in more detail to dig into some of the issues you faced, for example with conditionals in the weather-related code, as well as with error handling.

I left some comments with more details. The best coach to reach out to for a follow up would be the coach currently coaching your cohort, as they're the most likely to have time, but you can reach out to someone else if you prefer. 

Feel free let me know if you have any questions  about any of my comments!

lib/weather.rb
```diff
@@ -0,0 +1,16 @@
...
+        p "sunny"
+      elsif n == 1
+        p "sunny_intervals"
+      elsif n == 2
+        p "drizzle"
+      else n == 3
+        p "stormy"   
+    end
```
siellsiell, line 13: The Array.sample method would allow you to avoid this if statement: https://ruby-doc.org/core-2.4.1/Array.html#method-i-sample

PR #2434, opened 2022-01-24
airport_challenge Kyrsten Stone

MakersKyrstenStone:
my files for this challenge

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -11,6 +11,7 @@ class Airport
   end
 
   def land_plane(plane)
+    fail "The Hanger is full you cannot land" unless @planes_in_hanger.length < 4
     # add a plane to the hanger
     @planes_in_hanger << plane  
     @planes_in_hanger.length
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -28,4 +28,16 @@ describe Airport do
     # ATC wants to know plane left airport
     expect(subject.planes_available).to eq (1)
   end
+
+  describe '#land_plane' do
+    # ATC wants to refuse landing if airport full
+    it 'refuse landing if airport full' do
+      subject.land_plane("747")
+      subject.land_plane("concorde")
+      subject.land_plane("xwing")
+      subject.land_plane("falcon")
+      expect { subject.land_plane("bigboy") }.to raise_error ("The Hanger is full you cannot land")
+    end
+  end
+
 end


spec/airport_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  it 'can let a plane take off' do
+    plane = Plane.new
+    @planes_in_hanger = 2
+    subject.plane_takeoff(plane)
+    # ATC wants plane to take off / leave airport
+    expect(subject.plane_takeoff(plane)).to eq (1)
+  end
+
+  it 'can confirm if planes at airport' do
+    plane =Plane.new
```
laurenbethgray: missing space after =

lib/plane.rb
```diff
@@ -0,0 +1,7 @@
+class Plane
+
+  def plane
+
+  end
+
+end
```
siellsiell, line 7: At the moment this class isn't very useful as the `plane` method doesn't do or return anything. You're not using or mocking it in your unit tests below either (you've used strings like "concorde" instead). As such, it doesn't serve much purpose and should be deleted. 

But this class might be useful for the second user story, which you haven't implemented fully. Can you see how?

```
As an air traffic controller 
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  it 'can let a plane take off' do
+    plane = Plane.new
+    subject.land_plane("747")
+    subject.land_plane("concorde")
+     # ATC wants plane to take off / leave airport
+    expect(subject.plane_takeoff("747")).to eq (1)
+  end
+
+  it 'can confirm if planes at airport' do
```
siellsiell, line 23: If you read the user story carefully it actually asks you to be able to confirm whether a **particular** plane is at the airport or not.

lib/airport.rb
```diff
@@ -0,0 +1,27 @@
...
+  def initialize
+    @planes_in_hanger = []
+  end
+
+  def planes_available
+    @no_of_planes = @planes_in_hanger.length
+  end
+
+  def land_plane(plane)
+    fail "The Hanger is full you cannot land" unless @planes_in_hanger.length < 4
```
siellsiell, line 14: It's good to put constant values like `4` into [constants](https://www.rubyguides.com/2017/07/ruby-constants/) as it allows the value to be changed more easily and also referred to in tests.

spec/plane_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'plane'
+
+describe Plane do
+  it { is_expected.to respond_to :plane }
```
siellsiell, line 4: Why is the `plane` method necessary?

spec/airport_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    expect(subject.land_plane(plane)).to eq (1)
+  end
+
+  it 'can let a plane take off' do
+    plane = Plane.new
+    subject.land_plane("747")
+    subject.land_plane("concorde")
+     # ATC wants plane to take off / leave airport
+    expect(subject.plane_takeoff("747")).to eq (1)
+  end
```
siellsiell, line 21: These are fair tests but I think they would be stronger if you checked that the plane is in the airport / is no longer in the airport. For example, if you tried to make the same plane land twice in a row, calling `land_plane` should fail. If you tried to make the same plane take off twice in a row, calling `plane_takeoff` should fail.

With the tests as they are, I could rewrite the methods like this:

```
 def land_plane(plane)
    1
  end

 def plane_takeoff(plane)
    1
  end
```

And the test would still pass.

siellsiell (commented):
Hey it's Simo, a coach :)

Thanks for submitting your challenge code and apologies for the slow response. This was a good attempt at the challenge. Your code is readable and you've got unit tests for all of your methods. 

I can see that you made a start at separating things into 2 classes, but didn't quite get far enough. Going further in that direction will help with improving your tests and meeting the requirements of the user stories fully.

I left some comments with more details :) Do let me know if you have any questions!

MakersKyrstenStone:
HI Simo
thanks for your feedbacki have responded inline with your comments belowKyrsten
    On Monday, 31 January 2022, 14:48:11 GMT, Simo Tchokni ***@***.***> wrote:  
 
 


@siellsiell commented on this pull request.

Hey it's Simo, a coach :)

Thanks for submitting your challenge code and apologies for the slow response. This was a good attempt at the challenge. Your code is readable and you've got unit tests for all of your methods.
I can see that you made a start at separating things into 2 classes, but didn't quite get far enough. Going further in that direction will help with improving your tests and meeting the requirements of the user stories fully.I ran out of time and that was all i could manage to do. also we hadnt done any diagramming at this point 
I left some comments with more details :) Do let me know if you have any questions!

In lib/plane.rb:
> +class Plane
+
+  def plane
+
+  end
+
+end
At the moment this class isn't very useful as the plane method doesn't do or return anything. You're not using or mocking it in your unit tests below either (you've used strings like "concorde" instead). As such, it doesn't serve much purpose and should be deleted.i have no idea what mocking is yet as i have not had the time to reach that area. i used "concorde" "747" etc because i wanted to check a specific plane rather than just "plane"But this class might be useful for the second user story, which you haven't implemented fully. Can you see how?no i am not sure i can, when my specific plane (i.e "concorde") takes off it is no longer in the airport when you list whats in the hangerAs an air traffic controller 

In spec/airport_spec.rb:
> +  
+  it 'can land a plane' do
+    plane = Plane.new
+       # ATC want to land the plane at the airport
+    expect(subject.land_plane(plane)).to eq (1)
+  end
+
+  it 'can let a plane take off' do
+    plane = Plane.new
+    subject.land_plane("747")
+    subject.land_plane("concorde")
+     # ATC wants plane to take off / leave airport
+    expect(subject.plane_takeoff("747")).to eq (1)
+  end
+
+  it 'can confirm if planes at airport' do
If you read the user story carefully it actually asks you to be able to confirm whether a particular plane is at the airport or not.thats why i used specific plane names
In lib/airport.rb:
> @@ -0,0 +1,27 @@
+# require 'plane'
+
+class Airport
+
+  def initialize
+    @planes_in_hanger = []
+  end
+
+  def planes_available
+    @no_of_planes = @planes_in_hanger.length
+  end
+
+  def land_plane(plane)
+    fail "The Hanger is full you cannot land" unless @planes_in_hanger.length < 4

It's good to put constant values like 4 into constants as it allows the value to be changed more easily and also referred to in tests.

In spec/plane_spec.rb:
> @@ -0,0 +1,6 @@
+require 'plane'
+
+describe Plane do
+  it { is_expected.to respond_to :plane }
Why is the plane method necessary?because its an object that the user story asked for, and i was still trying to implement the simplest soloution to the user story i could as we have been trained
In spec/airport_spec.rb:
> +  it 'can land a plane' do
+    plane = Plane.new
+       # ATC want to land the plane at the airport
+    expect(subject.land_plane(plane)).to eq (1)
+  end
+
+  it 'can let a plane take off' do
+    plane = Plane.new
+    subject.land_plane("747")
+    subject.land_plane("concorde")
+     # ATC wants plane to take off / leave airport
+    expect(subject.plane_takeoff("747")).to eq (1)
+  end

These are fair tests but I think they would be stronger if you checked that the plane is in the airport / is no longer in the airport. For example, if you tried to make the same plane land twice in a row, calling land_plane should fail. If you tried to make the same plane take off twice in a row, calling plane_takeoff should fail.

With the tests as they are, I could rewrite the methods like this:
 def land_plane(plane)
    1
  end

 def plane_takeoff(plane)
    1
  end
And the test would still pass.fair point but my level of current knowledge isn't quite up to getting there in the time i had
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

siellsiell:
Hey @MakersKyrstenStone 

Thanks for sharing your thoughts with me. I feel like my feedback might have been misinterpreted -- in general the feedback coaches give on code reviews is intended not so much to be issuing you with a grade or evaluating your performance, but rather to give you ideas of how your code could be made better in the future as you grow your coding skills. 

Ideally, I'd hope that my comments can serve as a resource to you for planning and evidencing your learning. Some of them might not make full sense right now and that's alright but they might begin to as you become more comfortable with TDD and come across some of the concepts you hadn't had the chance to learn yet. So I'd encourage you to revisit them later on.

PR #2435, opened 2022-01-24
pull request

KevinByrne:


lib/plane.rb
```diff
@@ -0,0 +1,40 @@
+class Plane
+  attr_reader :airport_capacity
+  attr_reader :current_capacity
+
+  def initialize
+    @current_capacity = 20
```
Katerina-codes, line 6: I think you only needed to check Airport capacity, not plane :)

spec/plane_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'plane'
+
+describe Plane do
+  it "instructs plane to land" do
```
Katerina-codes, line 4: Nice clear test names

lib/plane.rb
```diff
@@ -0,0 +1,40 @@
...
+    @airport_capacity = 21
+  end
+
+  def instruct_to_land
+    return "Plane has landed!"
```
Katerina-codes, line 11: For this method to be dynamic and useful, it needs to do something. Right now it just returns a string but the functionality we want to see is that the plane can land and that we have a way of keeping track of that. The same principle applies to the **instruct_to_takeoff** method.

PR #2436, opened 2022-01-25
completed 4 user stories, cannot do the last two (stormy weather)

siellsiell:
Hi Olivia, I'm Simo, a coach.

I've created for you so I can review your challenge. For the next times, please make a full request after you've completed the challenge so that we can find your code more easily :)

### DIFF
--- a/spec/weather_spec.rb
+++ b/spec/weather_spec.rb
@@ -1,0 +1,18 @@
+require 'weather'
+
+describe Weather do
+
+  subject(:weather) { described_class.new }
+
+  describe '#stormy?' do
+    it 'reports weather as true if it is stormy' do
+      allow(weather).to receive(:stormy?).and_return(true)
+      expect(weather).to be_stormy
+    end
+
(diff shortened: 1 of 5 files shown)


spec/weather_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'weather'
+
+describe Weather do
+
+  subject(:weather) { described_class.new }
+
+  describe '#stormy?' do
+    it 'reports weather as true if it is stormy' do
+      allow(weather).to receive(:stormy?).and_return(true)
```
siellsiell, line 9: Does this test really test the code in your class? It doesn't because you have mocked the method that you're testing (`stormy?`). Because you told me to return through Apple always return through and your test will pass even if you're cold is broken. Instead you can stub the `rand` method that's called from inside the `stormy?` method. This might help: https://stackoverflow.com/questions/45761380/how-to-stub-rand-in-rspec

PR #2439, opened 2022-01-29
Airport Challenge - Bromley Jones

nyahehnagi:
Initial pull request

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '3.0.2'
+ruby '3.1.0'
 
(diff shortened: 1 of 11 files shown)


eddiearnoldz (commented):
Passes most of tests (over 90%) but not full coverage. Bromley explained this was because of a mock configuration which i still can't quite grasp. Perhaps someone else could add more review to this.

README is thorough but should mention bundle install and ruby rvm requirements. could have added a feature test example as suggested in the rubric.

Didn't get time to talk through it all but very impressed and hope to learn a lot form reading and talking about it more

PR #2444, opened 2022-01-30
Airport Challenge - Francesco's attempt

now-what:
- Created Domain Model and Feature Tests.
- Implemented specs and code for the first three User Stories.
- Set a default capacity (4th story).
- Ran out of time to work on weather conditions.


lib/airport.rb
```diff
@@ -0,0 +1,26 @@
+require_relative "plane.rb"
+
+class Airport
+  DEFAULT_CAPACITY = 20
```
Rakshak81, line 4: Unused Constant - No tests seen for Default_Capacity

lib/airport.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def lift(plane)
+    "Plane has taken off and is no longer in airport"
+  end
+
+  private # as per walkthrough 15
+
+  def full?
+    true
```
Rakshak81, line 21: Method returns true, however we should have used 'count'

PR #2445, opened 2022-01-30
Airport Challenge: Stuck on an error message!

Chloeem:
I tried to figure out the way to resolve this error message but sadly couldn't quite get it. I'd really appreciate feedback on my approach and how to fix this error message. I've included a MYREADME.md which covers my approach/experience and outlines this error.

### DIFF
--- a/MYREADME.md
+++ b/MYREADME.md
@@ -90,3 +90,9 @@ Plane <-- flying? --> true/false
 Airport <-- take_off --> if !flying?
 
 ----------------------------------------
+
+## Issues
+- I need to learn how to access methods within other classes?
+- I need to better understand the use of .self within a class/it's use for accessing methods
+- I need to continue getting more familiar with rspec syntax 
+- I need to continue learning how to test 'behaviour > state'
--- a/MYREADME.md
+++ b/MYREADME.md
@@ -51,3 +51,42 @@ Plane <-- land --> Airport
 |------------|-----------|  
 | VARIABLES  |           |
 | METHODS    | stormy?() | 
+
+----------------------------------------
+
+## User Story 2
+```
+As an air traffic controller 
+So I can get passengers on the way to their destination 
+I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport.
+```
+
+### Nouns:
+- air traffic controller
+- passengers
+- destination
+- plane
+- airport
+
+### Verbs:
+- instruct
+- take off
+- confirm
+
+### Behaviour:
+"take off a plane"
+"confirm the plane is flying"
+
+### Functional representation:
+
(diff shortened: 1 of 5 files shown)


MYREADME.md
```diff
@@ -0,0 +1,122 @@
+# Airport Challenge
+We have a request from a client to write the software to control the flow of planes at an airport. The planes can land and take off provided that the weather is sunny. Occasionally it may be stormy, in which case no planes can land or take off. 
+
+## User Story 1
+```
+As an air traffic controller 
+So I can get passengers to a destination 
+I want to instruct a plane to land at an airport
+```
+
```
CTJMID, line 10: Really really like your README in general - definitely something that I would look at doing in the future

MYREADME.md
```diff
@@ -0,0 +1,122 @@
...
+       expected `Array#length` to have changed by -1, but was changed by 0
+     # ./spec/airport_spec.rb:18:in `block (3 levels) in <top (required)>'
+```
+
+----------------------------------------
+
+## Experience 
+I really wanted to be able to continue with the rest of this project. I was enjoying myself, but was unable to get past this error message (no1). I'm unsure whether its an incorrect use of the rspec "change" matcher syntax or I was over complicating things, or perhaps was misusing "arrange/act/assert". 
+
+With the methods inside the Plane class I felt it was important to be able to check that a plane was ready_to_land(), before it landed in the airport. However I am aware this wasn't mentioned in the user stories, so i'm unsure whether I should have included it?
```
JohnForster, line 117: This is a fair check, but what conditions would affect whether or not a plane was ready to land? Could you just throw an error if you tried to land a plane that was already landed, for example?

lib/airport.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def take_off 
+    "Plane has left the airport" if @planes.count.zero?
+    # Do I have a test for this?
+
+    @planes.pop if @planes.length >= 1
+
+    # fail "This plane is already in flight" unless !flying?
+    # ^^ How do i access a method inside another class?
```
JohnForster, line 24: Your @planes array contains many instances of the Plane class. To check if one of them is flying, you need to call `.flying?` on one of those instances. This might look like:

```ruby
if @planes.length > 0 
  plane = @planes.last
  raise "Can't take off a flying plane!" if plane.flying?
  @planes.pop()
end
```

JohnForster: Using `.pop` and `.last` does mean that you can only ever take off the plane that has landed most recently. Maybe you could add a `plane` argument to the `Airport#take_off` method so you can specify which plane you want to take off?

The `Array#delete` method might help with this.

PR #2446, opened 2022-01-30
Airport Challenge Submission

CTJMID:
I believe I have successfully completed this challenge, but, it would be good to see best in class examples and get any feedback for where I might have improved!

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -9,11 +9,21 @@ attr_reader :hanger
   end
 
   def land(plane)
-    @hanger << (plane)
+    if plane.flying == true
+       plane.landed_plane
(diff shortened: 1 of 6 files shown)


Chloeem (commented):
It all looks really good to me! 

I particularly like what you did with the hanger to store when planes are in the airport and using a randomiser for the weather.

Good use of allows over expects in the rspec files, guard clauses, 

On the rspec it just shows as one item missing coverage "95.83%  | lib/airport.rb | 24    | 1      | 35" but otherwise it runs fine. 

There is no domain model or personal README.md. I would have been good to see your approach and planning for the project. 

It would have been good to see the use of encapsulation.

PR #2447, opened 2022-01-30
Ben Partington - Airport Challenge 

benpartington:


spec/airport_spec.rb
```diff
@@ -0,0 +1,20 @@
+require_relative '../lib/airport.rb'
```
Jaroslaw-Roman-Cudzich, line 1: Does not pass rspec tests, yet.

spec/plane_spec.rb
```diff
@@ -0,0 +1,35 @@
+require_relative '../lib/plane.rb'
```
Jaroslaw-Roman-Cudzich, line 1: Does not pass rspec tests, yet.

PR #2449, opened 2022-01-30
Airport Challenge - Peter O

harmlessgoose:


lib/airport.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def land(plane, weather = weather())
+    fail 'Weather is stormy' if weather == 'stormy'
+    fail 'Airport is full' if @hanger.length >= @hanger_capacity
+    @hanger.push(plane)
+  end
+
+  def takeoff(plane, weather = weather())
+    fail 'Weather is stormy' if weather == 'stormy'
+    @hanger.delete(plane)
```
trushnarabadia, line 21: Love the use of delete here, would not have even thought to use delete.

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'plane'
+require 'airport'
+
+describe Airport do
+
+  it "lands a plane at an airport" do
+    plane = Plane.new
+    subject.land(plane, 'clear')
```
trushnarabadia, line 8: Love the use of subject as a shortcut

PR #2450, opened 2022-01-30
Airport_Challenge

Rakshak81:


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '3.0.2'
```
now-what, line 3: Might have been easier to upgrade to Ruby 3.0.2 rather than manually changing the version in the Gemfile? Or install 3.0.2 using RVM. I don't think it's good practice to manually edit Gemfiles.

Gemfile.lock
```diff
@@ -3,27 +3,27 @@ GEM
...
-    rspec-core (3.10.1)
+    rspec-core (3.10.2)
       rspec-support (~> 3.10.0)
-    rspec-expectations (3.10.1)
+    rspec-expectations (3.10.2)
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.10.0)
-    rspec-mocks (3.10.2)
+    rspec-mocks (3.10.3)
       diff-lcs (>= 1.2.0, < 2.0)
```
now-what, line 24: Not really sure about the manual edits in the gemfile.lock file because I can't replicate the initial issue.

lib/airport.rb
```diff
@@ -0,0 +1,21 @@
+ require_relative 'aeroplane'
+class Airport
+  def initialize
+    @plane = []
+    @planes_landed = 0
+  end
+  def land_plane(plane)
+    fail 'Airport full' if full?
+    @plane << plane
+    @planes_landed += 1
```
now-what, line 10: I don't think this counter is needed, because you already have a `@plane` array, and you can count in there how many planes have landed (like you do on line 19).

PR #2451, opened 2022-01-30
Airport challenge HF

Haroon2021:


READMEAIRPORTCHALLENGE.md
```diff
@@ -0,0 +1,11 @@
...
+
+-For the challenge we had to model the behaviour of an airport.
+
+-Planes were were allowed to take of and land
+
+-We tested each of the behaviour using both unit and feature testing.
+
+-We had a capacity limit on planes of 20
+
+-I looked through the boris bikes example to help guide my answers.
```
Ledders26, line 11: I find adding a readme quite a challenge as it's something very new to me.  But having more titles adds structure and I think helps with both readability and writeability.  I used https://meakaakka.medium.com/a-beginners-guide-to-writing-a-kickass-readme-7ac01da88ab3 for tips - for instance a header for purpose and one for how to run etc

I think also it has to be called just README.md, but worth checking with a coach

lib/Aeroplane.rb
```diff
@@ -0,0 +1,3 @@
+class Aeroplane 
+
+end
```
Ledders26, line 3: As there are no methods or code in this file beyond a declarative class, this file isn't needed (though appreciate as you'd gone through the challenge it would have been later on!)

spec/Aeroplane_Spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'Aeroplane'
+
+describe Aeroplane do
+
+end
```
Ledders26, line 5: As with the aeroplane file, as this file doesn't test anything it is not required at this stage

lib/Airport.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def initialize
+    @plane = []
+    @planes_landed = 0
+  end
+
+  def land_plane(plane)
+    fail 'Airport full' if @plane.count >= 20
+    @plane << plane
+    @planes_landed +=1
```
Ledders26, line 12: Instead of using the @planes_landed variable, you could instead call a query on the length of the @plane array, which would tell you how many items had been added.

spec/Airport_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'airport'
+describe Airport do
+  describe '#land_plane' do
+    it "instructs the plane to land" do
+        plane = Aeroplane.new
```
Ledders26, line 5: Really picky, but the indent on this is out - I know you said rubocop wasn't working, so to be fair this would normally have picked this up.

spec/Airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      expect(subject.land_plane(plane)).to eq plane
+    end
+  end
+  
+  describe '#land_plane' do
+    it 'Do not land if the Airport is full' do
+        20.times { subject.land_plane Aeroplane.new}
+        expect { subject.land_plane Aeroplane.new}. to raise_error 'Airport full'
```
Ledders26, line 13: Space required before } on these lines (again, working rubocop would have helped!)

spec/Airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  describe '#land_plane' do
+    it "instructs the plane to land" do
+        plane = Aeroplane.new
+      expect(subject.land_plane(plane)).to eq plane
+    end
+  end
+  
+  describe '#land_plane' do
+    it 'Do not land if the Airport is full' do
+        20.times { subject.land_plane Aeroplane.new}
```
Ledders26, line 12: Like this test!

lib/Airport.rb
```diff
@@ -0,0 +1,20 @@
+require_relative 'aeroplane'
+class Airport
+
+  def initialize
+    @plane = []
```
Ledders26, line 5: Like in the weekly work (Boris bikes) - I like the use of an array to store the details of new planes.  Makes future changes much easier (especially when you know where the code is going to go)

READMEAIRPORTCHALLENGE.md
Ledders26, line 11: Nice reference to something you had used to help!

PR #2454, opened 2022-01-30
Airport Challenge - attempt 1

meghanblyth:


spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      subject.stormy = false 
+      expect(subject.take_off(plane)).to eq plane 
+    end 
+
+    it "Should stop airplane from taking off if the weather is stormy" do 
+      subjct.stormy = true 
+      expect { subject.take_off(plane) }.to raise_error "The weather is stormy, the plane cannot take off" 
+    end 
+  end 
+end
```
SarahM55, line 62: This looks like a really good start - I won't comment on every single line but [this](https://github.com/sjmog/airport_challenge_exemplar/blob/master/spec/airport_spec.rb) is a 'finished' example to show where you could have taken the code next/how you could have refactored etc that might be useful to look over :)

docs/weather.rb
```diff
@@ -0,0 +1,12 @@
...
+  attr_reader :stormy 
+
+  def initialize
+    rand(10) == 10 ? @stormy = false : @stormy = true 
+  end 
+
+  def stormy?
+    stormy
+  end 
+end
```
SarahM55, line 12: I think this could have been incorporated into your 'airport.rb' file :)

SarahM55 (commented):
Hope this is a tiny bit helpful - apologies is it's not super insightful! I found this [video](https://www.youtube.com/watch?v=Vg0cFVLH_EM&ab_channel=SamMorgan) and this [completed example](https://github.com/sjmog/airport_challenge_exemplar) really helpful so might be worth taking a look through those too!

PR #2456, opened 2022-01-30
Airport Challenge CBuchan

CBuchan5:


lib/airport.rb
```diff
@@ -0,0 +1,4 @@
+class Airport
+def full
+end
```
EdCope, line 3: Indent methods to improve readability

spec/passenger_spec.rb
```diff
@@ -0,0 +1,14 @@
+require './lib/passenger.rb'
```
EdCope, line 1: Consider including the passengers later. As the user story focuses on just the plane landing.
Not boarding the plane itself.

lib/plane.rb
```diff
@@ -1,5 +1,11 @@
+
 class Plane
 def land_plane
```
EdCope, line 3: could rename to land. plane would be `self` in this instance. land(self).

PR #2457, opened 2022-01-31
Airport challenge

AliCo-Hash:
airport challenge attempt

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -5,10 +5,14 @@ class Airport
   end
 
   def land(airplane)
-    raise 'Cannot land airplane due to airport being full' if @airplanes.length >= @capacity
+    raise 'Cannot land airplane due to airport being full' if full?
     @airplanes << airplane
   end
 
   def take_off(airplane)
+  end
+  
+  def full?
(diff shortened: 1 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'airplane'
+
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :airplanes
+
+  def initialize(capacity = DEFAULT_CAPACITY)
```
nelsonclaire, line 7: How would you test for an airport that has something other than the default capacity?

lib/airport.rb
```diff
@@ -0,0 +1,29 @@
...
+    @airplanes = []
+  end
+
+  def land(airplane)
+    raise 'Cannot land due to airport being full' if full?
+    raise 'Cannot land due to stormy weather' if stormy?
+    @airplanes << airplane
+  end
+
+  def take_off(airplane)
```
nelsonclaire, line 18: You have added a plane to the array when it lands but I cannot see anywhere where you remove the plane from the array if it takes off?

PR #2458, opened 2022-01-31
My first pull request on the airport_challenge

jparsons-git:
Here is my effort to tackle the first weekend challenge - airport_challenge


lib/air_traffic_controller.rb
```diff
@@ -0,0 +1,27 @@
+# Describes the AirTrafficController class
```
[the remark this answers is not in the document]
winters145: However, when it comes to running the code maybe too many classes makes it a bit confusing

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+  def initialize(status)
+    @status = status
+  end  
+
+  # TODO - do we actually need these "shortcuts" on status when we have status?
+  def airbourne?
+    @status == "airbourne"
+  end  
+
+  def grounded?
```
winters145, line 15: I can't figure out where the actual method of 'grounded?' is referred to elsewhere

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
...
+
+class Plane
+  attr_accessor :status
+
+  def initialize(status)
+    @status = status
+  end  
+
+  # TODO - do we actually need these "shortcuts" on status when we have status?
+  def airbourne?
```
winters145, line 11: I like the use of airbourne? as it's own method to be called on

spec/air_traffic_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
+require_relative '../lib/air_traffic_controller'
+require_relative '../lib/airport'
+require 'air_traffic_controller'
+require 'find'
+
+describe AirTrafficController do
+  describe '#land' do
+    it "responds to land with an already grounded plane" do
+      plane = Plane.new("grounded")
```
winters145, line 9: The use of test doubles here could be good to ensure your test is only focussing on one class

winters145 (commented):
Over all very well written and neat code and tests. 
Very wide-coverage with the tests.
The addition of a README to explain how to use the script would have been helpful as there are a lot of methods & arguments involved

PR #2459, opened 2022-01-31
Airport Challenge 

lykathleen:


spec/weather_spec.rb
```diff
@@ -0,0 +1,13 @@
+require_relative '../lib/weather'
+
+describe Weather do
+  describe "#stormy?" do
+    it "weather is stormy" do
+      allow(subject).to receive(:conditions).and_return("stormy")
+      expect(subject.stormy?).to eq true
+    end 
+
```
felixtaljaard, line 10: Single responsibility principle: creating a spec for just weather is good!

lib/weather.rb
```diff
@@ -0,0 +1,11 @@
+class Weather
+
+  def conditions
+    conditions = ["stormy", "rainy", "sunny", "cloudy"]
+    conditions.sample
+  end 
+
```
felixtaljaard: Could use rand(), also may have to require another file to get the tests to pass

PR #2460, opened 2022-01-31
My attempt at the Airport Challenge 

trushnarabadia:


README.md
```diff
@@ -87,3 +87,12 @@ Finally, don’t overcomplicate things. This task isn’t as hard as it may seem
...
 * Finally, please submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am.
+
+
+My Approach to the Challenge
+---------
+
+The airport system is a system to allow airports to control the take offs and landings of planes. It does this by taking into consideration whether there is enough space to hold the plane and taking into consideration the weather. 
+
+
+I followed the TDD process that we have previously learned in the Boris Bikes challenge. I revised the code and process for that challenge and adapted it to this one. I have managed to complete the first 4 user stories but have struggled to incorporate the weather factor. I also had trouble grasping mocking behaviour on doubles.
```
harmlessgoose, line 98: Great to add your own part to the original README - I forgot on mine 😅

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
+require_relative 'plane'
+
+class Airport
+
+  DEFAULT_CAPACITY = 10
+
+  attr_accessor :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
```
harmlessgoose, line 9: By creating a `DEFAULT_CAPACITY` variable outside of the initialize method, I think this means you can't define the `DEFAULT_CAPACITY` when creating a new instance of the Airport class.

Instead, you could write  `def initialize(capacity = 10)` which will set the default to 10, but will also allow you to override it when making a new Airport, ie - `airport = Airport.new(20)`

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+  def land(plane)
+    fail 'The airport is full' if full?
+    @planes << plane
+  end
+
+  private
+
+  attr_reader :planes
+
+  def full?
```
harmlessgoose, line 29: Really great idea to make your methods with question marks, makes the code super simple and readable - wish I'd remembered that in mine! 😊

lib/airport.rb
```diff
@@ -0,0 +1,45 @@
...
+    @planes.count >= capacity
+  end
+
+  def empty?
+    @planes.empty?
+  end
+
+  def check_weather
```
harmlessgoose, line 37: Is your weather changing between Stormy and Not Stormy, or is this always making Stormy = true? 🤔

harmlessgoose (commented):
Gave me loads of ideas for how to improve mine! 😊

spec/airport_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      expect(airport.take_off).to eq plane
+    end
+
+    it 'does not allow a plane to take off if weather is stormy' do
+      plane = Plane.new
+      plane.check_weather
+      airport = Airport.new
+      airport.land plane
+      expect { airport.take_off}.to raise_error 'Cannot take off, weather is stormy'
+    end
```
harmlessgoose, line 27: Unfortunately this test failed when I ran it on my machine 😔

PR #2461, opened 2022-01-31
Airport attempt

breadmaking:


lib/airport.rb
```diff
@@ -0,0 +1,17 @@
...
+    @planes = []
+  end
+
+  def landing(plane)
+    raise 'No free spots to land - Aiport Full' if @planes.count == @capacity
+    @planes.push(plane)
+  end
+
+  def take_off(plane)
+    @planes.remove(plane)
```
heykathl, line 14: Did you mean @planes.pop?

spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
+require './lib/airport'
+
+describe Airport do
+  it 'responds to take off' do
```
heykathl, line 4: Adding in a describe '#takeoff' above line 4 to group them into a context block

spec/airport_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    airport = Airport.new(20)
+    expect(airport).to respond_to(:take_off)
```
heykathl, line 6: When I call airport.take_off(plane), it seems that the test isn't expecting an argument. Perhaps expect(airport).to respond_to(:take_off).with(1).argument?

lykathleen (commented):
Code was clear and easy to understand - coverage is 96.15%. 
README was not updated. As the challenge was not completed, there was no explanation on where and how the challenge would be completed. 
A suggestion is to run rubocop which improved the formatting of code.

PR #2464, opened 2022-02-17
Alex Sweeney

EdwardAndress:
# Your name

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    @capacity = capacity
+    @weather = weather
+    @planes = []
+
+
+  end
+
+
+  def land(plane)
+
```
EdwardAndress, line 19: Here, you'll only see the first of these two errors if both the airport is full and the weather is bad. You could consider separating the raising of errors from checking capacity and weather.

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+  def land(plane)
+    
+    raise 'airport full' if full?
+    raise 'weather is stormy' if stormy?
+
+    @planes << plane
+  end
+
+
+  def take_off(plane)
```
EdwardAndress, line 27: As above :)

lib/airport.rb
```diff
@@ -0,0 +1,47 @@
...
+    raise 'weather is stormy' if stormy?
+    raise 'plane not at this airport' unless at_airport?(plane)  
+    puts "#{@plane} has taken off"
+  end
```
EdwardAndress, line 31: Looks like the takeoff method doesn't remove the plane from the airport.

lib/weather.rb
```diff
@@ -0,0 +1,15 @@
+class Weather
+
+  def initialize
+    @weather_options = ['fine', 'stormy', 'fine', 'fine']
+  end
+
+  def stormy?
```
EdwardAndress, line 7: This could be simplified to

```ruby
def stormy?
  @weather_options.sample == 'stormy'
end
```

spec/airport_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+describe Airport do
+  subject(:airport) { described_class.new(20, weather) }
+  let(:plane) { double :plane }
+  let(:weather) { double :weather }
+
+  
+
+  describe '#take_off' do
+    context 'when not stormy' do
+      before do
```
EdwardAndress, line 12: Nice use of stubbing

spec/airport_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+
+      it 'raises an error if plane is not at this airport' do
+        another_airport = described_class.new(20, weather)
+        another_airport.land(plane)
+        expect { airport.take_off(plane) }.to raise_error 'plane not at this airport'
+      end
+    end
+
+  #   context 'when stormy' do
```
EdwardAndress, line 27: It's best to remove code that is no longer used.

spec/airport_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    context 'when not stormy' do
+      before do
+        allow(weather).to receive(:stormy?).and_return false
+      end
+
+      it 'tells a plane to land' do
+        expect(airport).to respond_to(:land).with(1).argument
+      end
+
+      context 'when full' do
```
EdwardAndress, line 57: Alternatively (to the `20.times do`), you cold override the default capacity here.  Your test does work though and you might prefer this approach!

EdwardAndress:
Nice work Alex. I think this is a pretty good solution. There are a few comments in line against your code.

PR #2465, opened 2022-02-26
Phil Airport Challenge

PhilBligh:
# Phil Bligh

# User stories 

- [X ] User story 1: "I want to instruct a plane to land at an airport"
- [X ] User story 2: "I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport"
- [X] User story 3: "I want to prevent landing when the airport is full"
- [X ] User story 4: "I would like a default airport capacity that can be overridden as appropriate"
- [X ] User story 5: "I want to prevent takeoff when weather is stormy"
- [X ] User story 6: "I want to prevent landing when weather is stormy"

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -13,7 +13,7 @@ class Airport
   end
 
   def stormy?
-    if rand(10) > 7
+    if Weather.new
       puts "It's stormy. You can not land or take off."
       return true
     else 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,24 @@
+class Airport
+
+  def initialize(capacity = 5)
+    @capacity = capacity
+    @landed_planes = []
+  end
+  attr_accessor  :landed_planes, :capacity
+
+  def full?
+    if @landed_planes.length == @capacity
+      fail "The airport is full"
+    end
+  end
+
+  def stormy?
+    if rand(10) > 7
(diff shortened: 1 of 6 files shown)


lib/plane.rb
```diff
@@ -0,0 +1,32 @@
...
+  
+  def initialize
+    @status = "Airborne"
+  end
+
+  attr_accessor :status
+
+  def land(airport)
+    airport.full?
+    fail "This plane has already landed" if @status.is_a?(Airport)
```
PhilBligh, line 14: Don't use fail, return a string instead

lib/airport.rb
```diff
@@ -0,0 +1,24 @@
+class Airport
+
+  def initialize(capacity = 5)
+    @capacity = capacity
+    @landed_planes = []
+  end
+  attr_accessor  :landed_planes, :capacity
+
+  def full?
+    if @landed_planes.length == @capacity
```
PhilBligh, line 10: Don't use == to avoid case where capacity is reduced below number of planes

spec/plane_spec.rb
```diff
@@ -0,0 +1,96 @@
+require 'plane'
+require 'airport'
+
```
PhilBligh, line 3: Find out how to use 'let' or 'before' to set up tests

PR #2466, opened 2022-02-26
Anusha's Implementation Of Airport Challenge

anushanesargi:
# Your name

Anusha Nesargi

# User stories 

Anusha Nesargi

# README checklist

- [x] how to install,
- [x] how to run,

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,16 @@ Airport Challenge
 
 ```
 
+
+Steps
+------
+
+1. Fork this repo, and clone to your local machine
+2. Run the command 'gem install bundler' (if you don't have bundler already)
+3. When the installation completes, run 'bundle'
+4. run 'rspec' to test the testcases
+5. 'ruby lib/airport.rb' runs just the code 
+
 Approach
 ---------
 
--- a/README.md
+++ b/README.md
@@ -13,77 +13,22 @@ Airport Challenge
 
 ```
 
-Instructions
+Approach
 ---------
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 10am Monday morning
+* I started with skimming through the complete challenge. Then written down what the challenge was all about.
 
-Steps
--------
+* Did not exactly follow the noun, verb approach but was able to breifly write down the tasks each user story was asking us to do.
 
-1. Fork this repo, and clone to your local machine
-2. Run the command `gem install bundler` (if you don't have bundler already)
-3. When the installation completes, run `bundle`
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -13,77 +13,32 @@ Airport Challenge
...
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/main/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc. 
+* I started with skimming through the complete challenge. Then written down what the challenge was all about.
 
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this at this moment.
+* Did not exactly follow the noun, verb approach but was able to breifly write down the tasks each user story was asking us to do.
 
-**BONUS**
+* When I started the TDD I faced a couple of issues with "simplecov" asking for ruby "3.0.2", it took me some time to figure out what was going on.
```
JohnForster, line 33: If you've installed rvm, you should be able to use it to change your local ruby version, so that you don't need to make changes to the gem file.

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+#   attr_reader :planes_at_airport
+
+  def initialize(capacity = 10)
+    @default_capacity = capacity
+    @planes_at_airport = []
+    
+  end
+
+  def land(*landing_plane)
+    landing_plane.each do |p|
```
JohnForster, line 11: Try to avoid single letter variables. Your meaning is clearer if you name this `plane`

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+      landing_exceptions(p)
+      puts "#{p} has landed"
+      @planes_at_airport << p
```
JohnForster, line 14: The class is the airport, so this could be renamed simply `@planes`

lib/airport.rb
```diff
@@ -0,0 +1,63 @@
...
+    @planes_at_airport.include?(plane)
+  end
+
+  def take_off_exceptions(departing_plane)
+    fail "No planes in the airport" if empty? == true
+    fail "Stormy weather, cannot take off" if weather == "Stormy"
+    fail "Plane not at the airport" if at_aiport?(departing_plane) == false
+  end
+
+  def landing_exceptions(landing_plane)
```
JohnForster, line 57: Where possible, your method names should describe actions. So it's better to use something like `#validate_landing` than `#landing_exceptions`

JohnForster (commented):
This is a generally execllent solution, well done. The only thing I'd suggest to improve is in your variable and method names.

lib/airport.rb
anushanesargi: Thank You so much. I will do a better job at naming going forward.

PR #2468, opened 2022-02-27
Airport Challenge

clovellbsc:
# Your name

Chris Lovell

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -12,7 +12,7 @@ class Airport
 
   def land_plane(plane)
     raise 'Airport at full capacity' if @planes.count >= @capacity
-    raise 'The plane is in the airport already' if plane.grounded?
+    raise 'The plane is in the airport already' if @planes.include?(plane)
     raise 'Weather conditions are not favourable for landing' if @weather.stormy?
     plane.grounded = true
     @planes << plane
@@ -24,7 +24,6 @@ class Airport
     raise "That plane is not in this airport" unless @planes.include?(plane)
     plane.grounded = false
     @planes.delete(plane)
-    'Plane left the airport'
   end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -12,7 +12,7 @@ class Airport
 
   def land_plane(plane)
     raise 'Airport at full capacity' if @planes.count >= @capacity
-    raise 'The plane is in the airport already' if plane.grounded?
+    raise 'The plane is in an airport already' if plane.grounded?
     raise 'Weather conditions are not favourable for landing' if @weather.stormy?
     plane.grounded = true
     @planes << plane
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,30 @@
+require './lib/plane'
+require './lib/weather'
+
+class Airport
+  attr_reader :planes, :capacity
+
+  def initialize(airport_capacity = 10)
+    @weather = Weather.new
+    @planes = []
+    @capacity = airport_capacity
+  end
+
+  def land_plane(plane)
+    raise 'Airport at full capacity' if @planes.count >= @capacity
+    raise 'The plane is in the airport already' if plane.grounded?
+    raise 'Weather conditions are not favourable for landing' if @weather.stormy?
+    plane.grounded = true
+    @planes << plane
+  end
+
+  def take_off(plane)
+    raise 'This plane is already in the air' unless plane.grounded?
+    raise 'Weather conditions are not favourable for taking off' if @weather.stormy?
+    raise "That plane is not in this airport" unless @planes.include?(plane)
+    plane.grounded = false
+    @planes.delete(plane)
+    'Plane left the airport'
+  end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,7 +2,7 @@ require './lib/plane'
 require './lib/weather'
 
 class Airport
-  attr_reader :planes
+  attr_reader :planes, :capacity
 
   def initialize(airport_capacity = 10)
     @weather = Weather.new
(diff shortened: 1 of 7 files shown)


joe-thornton (commented):
README is nice and clear

**Airport.rb**
Code is easily readable
Good, clear exception handling

Sensible to initialize weather with the airport - wish I'd done that

**airport_spec.rb**
See what you mean about using Object and rand for mocking - interested to discuss this approach. Perhaps could have mocked `stormy?` instead?

Airport capacity is hard-coded in test (10.times...) - may be issues here if the airport capacity changes.

**Plane**
Nice use of attr_accessor

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
+require './lib/plane'
+require './lib/weather'
+
+class Airport
+  attr_reader :planes, :capacity
+
```
leoht, line 6: You might not need to expose the attribute `capacity` to the outside with the accessor, as we probably don't want the caller/user of this class to access it directly?

You might ask: how do we test setting a custom capacity, then? Remember testing behaviour over state:
  * testing behaviour - given that we set a capacity of only 1, trying to land a plane *should raise an error*
  * testing state - given that we set a capacity of only 1, `airport.capacity` should return 1

Try to change your test below (in airport_spec, line 51) to assert on the class behaviour (it should raise an error if we go over capacity), rather than its state

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+    @weather = Weather.new
+    @planes = []
+    @capacity = airport_capacity
+  end
+
+  def land_plane(plane)
+    raise 'Airport at full capacity' if @planes.count >= @capacity
+    raise 'The plane is in the airport already' if plane.grounded?
+    raise 'Weather conditions are not favourable for landing' if @weather.stormy?
+    plane.grounded = true
```
leoht, line 17: You might not need the `grounded` attribute on the `Plane`  class - as you can already "know" whether a plane is "landed" or not (= is it included in the array or not?). Having to maintain this separate attribute `grounded` means that we know have two sources of truth for the same information, which can cause inconsistencies in the data, and might be error prone (for example, we might end up with a plane being in the array, but without its `grounded` attribute set to `true`)

lib/airport.rb
```diff
@@ -0,0 +1,30 @@
...
+    @planes << plane
+  end
+
+  def take_off(plane)
+    raise 'This plane is already in the air' unless plane.grounded?
+    raise 'Weather conditions are not favourable for taking off' if @weather.stormy?
+    raise "That plane is not in this airport" unless @planes.include?(plane)
+    plane.grounded = false
+    @planes.delete(plane)
+    'Plane left the airport'
```
leoht: Rather than returning a human-readable string here, I'd suggest returning a "meaningful" value the caller of the class can use, for two reasons:
 * returning a "simpler" value (such as the current value of `@planes`, a boolean, of even the plane being removed) is easier to handle than a human-readable message, which you could re-type elsewhere making typos in it
 * A human-readable string would a concern of presentation / displaying messages to the user of the program - so it should probably belong in another class responsible for everything dealing with presentation (if we were to add human-readable messages in our program)

lib/weather.rb
```diff
@@ -0,0 +1,18 @@
+class Weather
+
+  def initialize
+    @weather = 'sunny'
+  end
+
+  def weather_randomiser
+    case rand(1..10)
```
leoht, line 8: This can all be simplified further by keeping only the method `stormy?` - you would both chose a random value with `rand` *and* decide to return either 'sunny' or 'stormy' depending on this value, all in `stormy?`

spec/airport_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    allow_any_instance_of(Object).to receive(:rand).and_return(2)
+    subject.land_plane(new_plane)
+    expect(subject.take_off(new_plane)).to eql('Plane left the airport')
+    expect(subject.planes).to eq([])
+  end 
+
+  it 'plane unable to land when airport capacity is full' do
+    allow_any_instance_of(Object).to receive(:rand).and_return(1)
+    10.times { subject.land_plane(Plane.new) }
+    expect { subject.land_plane(Plane.new) }.to raise_error 'Airport at full capacity'
```
leoht: This is a nice example here of testing the behaviour re: overfilling capacity, rather than the state (see my comment above)

spec/weather_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'weather'
+
+describe Weather do
+  it 'weather can respond to stormy?' do
+    expect(subject).to respond_to(:stormy?)
+  end
+
+  describe Weather do
```
leoht: You probably don't need this nested `describe Weather` block here - you could have all 3 tests "flat" in the same top-level `describe`

spec/airport_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  it 'Landed plane gets stored at the airport' do
+    allow_any_instance_of(Object).to receive(:rand).and_return(8)
+    new_plane = Plane.new
+    expect(subject.land_plane(new_plane)).to eql([new_plane])
+  end
+
+  it 'plane to take off from an airport and confirm that it is no longer in the airport' do
+    new_plane = Plane.new
+    allow_any_instance_of(Object).to receive(:rand).and_return(2)
```
leoht: Instead of stubbing directly `rand` here - a better approach could be to stub the behaviour of `Weather` (like we've seen in this morning's workshop), something like:
```ruby
allow_any_instance_of(Weather).to receive(:stormy?).and_return(true)
```

that would stub *all* instances of `Weather`, which is probably fine for our case here, but an even better way would be to dependency-inject a double, or an instance of `Weather` that is specifically stubbed, in our `Airport` class - again, using a similar process that we saw in the mocking workshop

leoht (commented):
Great job :) I have a few comments here and there to improve the code and the tests, but overall that's already really good progress. Let me know if you have any further questions on the feedback, or if you want to discuss it

PR #2470, opened 2022-02-27
Rob Oman - Airport Challenge

r94o:
# Rob Oman

https://github.com/r94o/airport_challenge

# User stories 


HCVF:
Positives

- frequent commits with clear, succinct descriptions
- IRB transcript in README shows how program is used
- 100% test coverage exceeds expectations
-  DRY and SRP followed well
- Gaurd clauses used well and set a good example for me to follow in the future as I am trying to default to using them

Could Improve

- "Magic numbers" used in tests where constants may have been more appropriate
- I think you could get rid of the "expected to respond to" tests as your more advanced tests will cover the same cases

lib/plane.rb
```diff
@@ -0,0 +1,44 @@
...
+  def initialize
+    @flying = true
+  end
+
+  def land(airport)
+    fail "This plane is already at an airport" unless flying?
+    fail "Unable to land due to the weather" if stormy?
+    fail "Airport is full" if airport.full?
+    @flying = false
+    airport.hangar << self
```
leoht, line 15: That's an interesting approach you had here — the "ownership" of the data (which class owns the attribute?) for the planes array is on `Airport` (the instance variable is on this class), however it is being changed by `Plane`. This works, but is not ideal in terms of cohesion and maintainability — a way to improve this could be to move the methods `land` and `take_off` on the `Airport` class, and invert the dependency relationship of your two classes

It might sound counter-intuitive to think of "an airport landing planes", but in terms of data ownership and code cohesion, since the array belongs on the `Airport` class, it makes sense for the methods that alter the state (the program's behaviour) to *also* be on `Airport`

lib/plane.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def take_off(airport)
+    fail "This plane is not at an airport" if flying?
```
leoht, line 19: Nice use of guard conditions here!

lib/plane.rb
```diff
@@ -0,0 +1,44 @@
...
+    fail "Unable to take off due to the weather" if stormy?
+    fail "This plane is not at this airport" unless plane_at?(airport)
+    @flying = true
+    airport.hangar.delete(self)
+  end
+
+  def flying?
```
leoht, line 26: The two methods `flying?` and `check_weather` could probably be private as well, as we don't need to expose them to the outside

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+describe Airport do
+  
+  let(:plane) { double("plane") }
+  
+  it { is_expected.to respond_to(:full?) }
+
+  it "has a customisable capacity upon initialising instance" do
+    airport = Airport.new(50)
+    50.times { airport.hangar << plane }
+    expect(airport.full?).to eq true
```
leoht, line 12: Here, a better way to test the behaviour of the `Airport` class would be to assert that *an error is raised when trying to land another plane*, rather than asserting on `full?` (which should probably be a private method) - this is another example of "testing behaviour rather than state" - the expected behaviour should be that the landing raises an error

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  describe "#full?" do
+    it "returns true when the airport is full" do
+      # Default capacity is 30
+      30.times { subject.hangar << plane }
+      expect(subject.full?).to eq true
```
leoht, line 19: Same remark as above here - if we make the `full?` method private (it doesn't need to be accessed from the outside), we don't need to test it and we can simplify our tests

spec/plane_spec.rb
```diff
@@ -0,0 +1,76 @@
+require './lib/plane'
+
+describe Plane do
+  
+  let(:airport) { double("airport", :hangar => []) }
```
leoht, line 5: The comment I had above (about moving methods to the `Airport` class) would also simplify our tests - remember that, because either `Airport` depends on `Plane`, either the other way around (but not both), we'd need to use doubles only in one of these test files. If `Airport` depends on `Plane` (as I suggested), we'd need to use doubles only in Airport's tests, to mock the behaviour of `Plane`

leoht (commented):
Nice work :) The code is quite clear and readable, and the tests as well, my comments are mainly on code design, and how to simplify the model by moving ownership of data and behaviour (= the methods) to the `Airport` class, which will simplify a lot your code and tests as well - let me know if you have any questions around that, happy to discuss it further

PR #2471, opened 2022-02-27
Airport Challenge

gdrmvrm:
# Gayathri Dharmavaram

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,12 +1,15 @@
 require './lib/plane'
+require './lib/weather'
+
 
 class Airport
   attr_accessor :capacity
   DEFAULT_CAPACITY = 25
 
-  def initialize(capacity = DEFAULT_CAPACITY)
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
     @planes = []
     @capacity = capacity
+    @weather = weather
   end
 
   def return_planes
@@ -15,10 +18,12 @@ class Airport
 
   def land(plane)
     raise "Airport capacity is reached" if full?
+    raise "Stormy conditions - Plane cannot land" if @weather.get_conditions == "stormy"
     @planes << plane
   end
 
   def take_off(plane)
+    raise "Stormy conditions - Plane cannot take off" if @weather.get_conditions == "stormy"
     raise "No plane available to take off" if empty?
     raise "Plane not available to take off" if !@planes.include?(plane)
     @planes.delete(plane) if @planes.include?(plane)
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,20 @@
+require './lib/plane'
+
+class Airport
+
+  def initialize
+    @planes = []
+  end
+
+  def return_planes
+    return @planes
+  end
+
+  def land(plane)
+    @planes << plane
+  end
+
+  def take_off(plane)
+    
+  end
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,9 +1,12 @@
 require './lib/plane'
 
 class Airport
+  attr_accessor :capacity
+  DEFAULT_CAPACITY = 25
 
-  def initialize
+  def initialize(capacity = DEFAULT_CAPACITY)
     @planes = []
+    @capacity = capacity
   end
 
   def return_planes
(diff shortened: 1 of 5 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  attr_accessor :capacity
+  DEFAULT_CAPACITY = 25
+
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
+    @planes = []
+    @capacity = capacity
+    @weather = weather
+  end
+
+  def return_planes
```
leoht, line 15: I'd suggest renaming this method `planes`, as the fact it will "return" the list of planes is implied - this would look like this:
```ruby
def planes
   return @planes
end 
```

... which could be then further simplified by using Ruby's `attr_reader`

lib/weather.rb
```diff
@@ -0,0 +1,5 @@
+class Weather
+  def get_conditions
```
leoht, line 2: I'd suggest changing a bit the way other classes interact with `Weather` by simplifying the method to something like `def stormy?`, this way the other classes (like `Airport`) don't have to "know" which specific values will be returned - this mean you can simplify this:
```ruby
raise "Stormy conditions - Plane cannot land" if @weather.get_conditions == "stormy"
```
by this:
```ruby
raise "Stormy conditions - Plane cannot land" if @weather.stormy?
```

which has the advantage of being a bit simpler on the side of the caller, and we're not "leaking" implementation details (= the knowledge of the string values "stormy" and "sunny") outside of the `Weather` class

spec/airport_spec.rb
```diff
@@ -0,0 +1,83 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  describe ".land" do
+    it 'will not raise an error when a plane lands at the airport ' do
+      plane = Plane.new
+      weather = double("weather", :get_conditions => "sunny")
```
leoht, line 8: Nice use of doubles here to mock the weather's behaviour - good job :)

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+    raise "No plane available to take off" if empty?
+    raise "Plane not available to take off" if !@planes.include?(plane)
+    @planes.delete(plane) if @planes.include?(plane)
+  end
+
+
+  private
+
+  def empty?
+   return true if @planes.empty?
```
leoht, line 36: Since `@planes.empty?` returns the boolean value we want, this can be further simplified into 
```ruby
def empty?
  @planes.empty?
end
```

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def full?
```
leoht, line 39: Same remark as above here :)

lib/airport.rb
```diff
@@ -0,0 +1,42 @@
...
+    raise "Airport capacity is reached" if full?
+    raise "Stormy conditions - Plane cannot land" if @weather.get_conditions == "stormy"
+    @planes << plane
+  end
+
+  def take_off(plane)
+    raise "Stormy conditions - Plane cannot take off" if @weather.get_conditions == "stormy"
+    raise "No plane available to take off" if empty?
+    raise "Plane not available to take off" if !@planes.include?(plane)
+    @planes.delete(plane) if @planes.include?(plane)
```
leoht, line 29: Since the inclusion of the plane is already pre-checked by the guard condition just above, you might want to omit the condition here, as it won't reach this line if the plane is not in the array

spec/airport_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      airport = Airport.new(25, weather)
+      airport.land(plane)
+      expect(airport.return_planes).to match_array([plane])
+    end
+
+    it "will not allow to land plane when airport capcity is reached" do
+      weather = double("weather", :get_conditions => "sunny")
+      airport = Airport.new(25, weather)
+      airport.capacity.times { airport.land(Plane.new) }
```
leoht, line 24: It looks like you're exposing the `capacity` attribute in the `Airport` class only to be accessed here (I think - maybe I'm wrong), in this case I'd suggest removing it so you're not exposing the attribute - since it doesn't need to be - and you would directly use the value `25` in the test to iterate in the test. You're already using this value when calling `Airport.new` so it's fine to reference it here again explicitly - or even better, to first use a variable to store it:
```ruby
capacity = 25
airport = Airport.new(capacity, weather)
capacity.times { airport.land(Plane.new) }
```

leoht (commented):
Great work - I left a few comments here and there to improve the code, let me know if you have any questions :)

PR #2472, opened 2022-02-27
Makers Academy - Week 1 Weekend Challenge

PKilgarriff:
# Paul Kilgarriff

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


spec/airport_spec.rb
```diff
@@ -0,0 +1,81 @@
+require 'airport'
+
+describe Airport do
+  let(:serial_number) { 25_817 }
+  let(:airport) { Airport.new("Heathrow") }
+  let(:plane) { instance_double("Plane", :serial_number => serial_number) }
+  let(:weather) { instance_double(Weather, :weather_report => "Sunny") }
+
```
jonnyabrams, line 8: These let clauses look proper snazzy - gonna have to read up on this

lib/plane.rb
```diff
@@ -0,0 +1,18 @@
+class Plane
+  attr_reader :serial_number
+  attr_writer :location
+
+  def initialize(make = "Unknown", serial_number = "Unknown")
+    @location = "Unknown"
+    @make = make
+    @serial_number = serial_number
+  end
```
jonnyabrams, line 9: The creation of real planes with serial numbers is a great touch

jonnyabrams (commented):
Here y'go

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
+class Airport
+  DEFAULT_CAPACITY = 10
+  attr_reader :capacity
```
valentina-maggio, line 3: If you use the attr_reader, can the capacity be changed by the user?

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def initialize(name, capacity = DEFAULT_CAPACITY)
+    @name = name
+    @capacity = capacity
+    @planes = []
+    @weather = "Unknown"
+  end
+
+  def land(plane)
+    raise "Airport is full" if full?
```
valentina-maggio, line 13: I would move the 'raise' bits of code in a separate method to respect the SRP.

lib/airport.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def full?
+    @planes.count >= @capacity
+  end
+
+  def stormy?
+    @weather == "Stormy"
+  end
+end
```
valentina-maggio, line 54: Code very clear and readable!

spec/airport_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    it 'prevents planes from taking off when weather is stormy' do
+      airport.land(plane)
+      allow(weather).to receive(:weather_report).and_return("Stormy")
+      airport.update_weather(weather)
+      expect do
+        airport.take_off(serial_number)
+      end.to raise_error 'Cannot take off during stormy weather'
+    end
+  end
+end
```
valentina-maggio, line 85: I suggest to use 'context' based on the weather if it's stormy or not

spec/airport_spec.rb
PKilgarriff: Highly recommend - have a resource that I will link that is very overwhelming, but gives you places to jump off from.
https://relishapp.com/rspec/rspec-mocks/docs

lib/airport.rb
PKilgarriff: No not currently - it's only able to be overridden at the point of creation. Do you think I should add that in as a method?

lib/airport.rb
PKilgarriff: Absolutely, I will borrow your example!

spec/airport_spec.rb
PKilgarriff: Yes will do - need to read up on the usage

spec/airport_spec.rb
PKilgarriff: Above was not the deer in the headlights resource - this one is: https://github.com/rspec/rspec-mocks#method-stubs
Useful but needs chipping away at

lib/plane.rb
PKilgarriff: Thanks - definitely a distraction from writing the code, but kept me happier

spec/airport_spec.rb
PKilgarriff: Will follow your example and use contexts inside the instance methods

lib/airport.rb
valentina-maggio: I thought using the attr_accessor would best suit for the purpose of letting the the user change the capacity any time while running the program. I used the walkthrough example from the Boris Bike challenge as a guideline for this: 
https://github.com/makersacademy/course/blob/main/boris_bikes/walkthroughs/17.md

PKilgarriff: Thank you, I'll take a look. My only worry with that would be that a user could change it to anything, not just an integer

valentina-maggio: Good point! I didn't think about that!

PR #2478, opened 2022-02-28
First Pull Request

NicolaCortese:
# Nicola Cortese

# User stories 

# README checklist

- [ ] how to install,
- [x] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,4 +1,6 @@
 class Airport
+  
+  attr_reader :hangar
 
   def initialize
     @hangar = []
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -17,7 +17,13 @@ class Airport
   end
 
   def take_off
+    if empty?
+      raise "hangar is empty"
+    elsif stormy? 
+      raise "grounded due to storm"
+    else
     @hangar.pop
+    end
   end
 
   private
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  def full?
+    @hangar.length == @capacity
+  end
+
+  def empty?
+    @hangar.empty?
+  end
+
+  def stormy?
+    rand(100) >= 0
```
HarryClenyg, line 43: If you're doing rand you don't need to specify >= 0.

>= used for non-zero examples e.g. rand(100) >= 95.

HarryClenyg, line 43: If you're doing rand you don't need to specify >= 0.

>= used for non-zero examples e.g. rand(100) >= 95.

spec/airport_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  
+  # As an air traffic controller 
+  # So I can get passengers on the way to their destination 
+  # I want to instruct a plane to take off from an 
+  # airport and confirm that it is no longer in the airport
+  it 'instructs a plane to take off' do
+    allow(subject).to receive(:stormy?){false}
+    subject.land(plane)
+    expect(subject).to respond_to ('take_off')
+    expect(subject.take_off).to_not eq nil
```
HarryClenyg, line 23: could you use, 'to exist' here? 

also possible to use 'to_not be_nil'

PR #2479, opened 2022-02-28
Harry Dawson airport challenge

HarryClenyg:
# Harry Dawson

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,18 @@
+require_relative 'plane'
+
+class Airport
+    attr_reader :name
+
+  def initialize(name)
+    @name = name
+    @planes = []
+  end
+
+  def land 
+    "plane has landed at #{@name}" 
+  end
+
+  def take_off
+    "plane has taken off from #{@name}"
+  end
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,18 +1,31 @@
 require_relative 'plane'
 
 class Airport
-    attr_reader :name
+  DEFAULT_CAPACITY = 3
+    
+  attr_reader :name
+  attr_reader :capacity
 
-  def initialize(name)
+  def initialize(name, capacity = DEFAULT_CAPACITY)
     @name = name
     @planes = []
+    @capacity = capacity
   end
 
-  def land 
-    "plane has landed at #{@name}" 
+  def land(plane)
+    fail "Can't land in stormy weather." if stormy? == true
+    fail "Airport full. Capacity is #{@capacity}." if @planes.size == @capacity
+    fail "Plane already landed." if @planes.include? plane
+    @planes << plane
   end
 
-  def take_off
-    "plane has taken off from #{@name}"
+  def take_off(plane)
+    fail "Can't fly in stormy weather." if stormy? == true
+    fail "Plane already airborne." unless @planes.include? plane
+    @planes.pop
+  end
+
+  def stormy?
+    rand(5) > 4
   end
 end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,3 @@
+class Plane
+  
+end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,3 +1,3 @@
 class Plane
-  
+  @plane
 end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,15 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+  it 'allows plane to land at airport when instructed' do
+    airport = Airport.new("Heathrow")
+    expect(airport.land).to eq "plane has landed at Heathrow"
+  end
+
+  it 'allows plane to take off from airport when instructed' do
+    airport = Airport.new("Heathrow")
+    expect(airport.take_off).to eq "plane has taken off from Heathrow"
+  end
+
+end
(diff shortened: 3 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'plane'
```
NicolaCortese, line 1: Do you need to require plane in here?

lib/airport.rb
```diff
@@ -0,0 +1,18 @@
...
+
+class Airport
+    attr_reader :name
+
+  def initialize(name)
```
NicolaCortese: I like the way you init with a name

lib/airport.rb
```diff
@@ -0,0 +1,18 @@
...
+    @name = name
+    @planes = []
+  end
+
+  def land 
+    "plane has landed at #{@name}"
```
NicolaCortese: Lovely comment, the method needs an argument to be passed in for a plane to be landed, and probably added to the '@plane' array

lib/airport.rb
```diff
@@ -0,0 +1,18 @@
...
+    @name = name
+    @planes = []
+  end
+
+  def land 
+    "plane has landed at #{@name}" 
+  end
+
+  def take_off
+    "plane has taken off from #{@name}"
```
NicolaCortese: like the comment above for land, with the exception that you don't need a parameter here to take off, unless you want to specify what plane should take off

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
+
```
NicolaCortese: hot class

spec/airport_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'airport'
+require 'plane'
```
NicolaCortese: if you are testing the airport class in this spec, you should use a double to mimic the plane behaviour, ring-fencing the airport class tests from any changes made in the plane class

spec/airport_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+describe Airport do
+  it 'allows plane to land at airport when instructed' do
+    airport = Airport.new("Heathrow")
+    expect(airport.land).to eq "plane has landed at Heathrow"
+  end
+
+  it 'allows plane to take off from airport when instructed' do
+    airport = Airport.new("Heathrow")
+    expect(airport.take_off).to eq "plane has taken off from Heathrow"
+  end
```
NicolaCortese, line 35: Both tests are nicely written and I can easily understand what they are testing on my first read

NicolaCortese (commented):
this is awesome, JD

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+  attr_reader :capacity
+
+  def initialize(name, capacity = DEFAULT_CAPACITY)
+    @name = name
+    @planes = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    fail "Can't land in stormy weather." if stormy? == true
```
leoht, line 16: You can shorten `if stormy? == true` to simply `if stormy?`

lib/plane.rb
```diff
@@ -0,0 +1,3 @@
+class Plane
+  @plane
```
leoht, line 2: Looks like this instance variable is not needed/used here at all - you could probably remove it (it's fine to have an empty class)

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  it 'plane is added to @planes on landing' do
+    expect(landed_plane).to eq [plane]
+  end
+
+  it 'raises error if landed plane is asked to land again' do
+    airport.land(plane)
+    expect { airport.land(plane) }.to raise_error 'Plane already landed.'
+  end
+
+  it 'raises error if flying plane is asked to take off again' do
```
leoht, line 20: This test is valid, but the wording of it is slightly confusing -  `'raises error if flying plane is asked to take off again'` - it's not actually if asked to take off *again*, but to take off at all, as it never "landed" at the airport (when using a value defined by `let`, RSpec recreates a fresh value for each test, so the plane we're using here is new - the values are not shared by other tests)

(translated in terms of code, this plane was never added to the array, therefore trying to remove it ("take off") results in an error)

A maybe slightly better description of this test could be `'raises error if trying to take off plane that is not landed'`

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    new_airport = Airport.new("Gatwick", 2)
+    allow(new_airport).to receive(:stormy?).and_return(false)
+    new_airport.land("plane1")
+    new_airport.land("plane2")
+    expect { new_airport.land(plane) }.to raise_error 'Airport full. Capacity is 2.'
+  end
+
+  it 'prevents take off when weather is stormy' do
+    landed_plane
+    allow(airport).to receive(:stormy?).and_return(true)
```
leoht, line 47: Nice use of stubbing here to control the weather behaviour :)

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+require_relative 'plane'
+
+class Airport
+  DEFAULT_CAPACITY = 3
+    
+  attr_reader :name
+  attr_reader :capacity
```
leoht, line 7: Since we don't need to expose the value of `capacity` to the outside, we can probably remove this attribute reader?

leoht (commented):
Nice work! I've added a few comments here and there to slightly improve the code - a next step here (left as an exercise) could be to extract the weather logic to its own class :) 
Let me know if you have any questions on the feedback

PR #2480, opened 2022-02-28
Leo's airport challenge

leoht:
# Leo Hetsch

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


lib/airport.rb
```diff
@@ -0,0 +1,13 @@
+class Airport
+  def initialize
+    @planes = []
+  end
+
+  def add_plane(plane)
```
leoht, line 6: The code is readable with clear naming - good!

spec/airport_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+describe Airport do
+  it 'adds a plane' do
+    airport = Airport.new
+    airport.add_plane('AF7264')
+
+    expect(airport.planes).to(eq(['AF7264']))
+  end
+
+  # it 'adds plane?' do
```
leoht, line 11: This test is commented out - could we maybe remove it?

PR #2481, opened 2022-02-28
weekend challenge

dennihous:
Dennis Houston

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -5,7 +5,7 @@ class Airport
 
   def initialize(capacity = 10)
     plane = Plane.new
-    @hangar = [plane]
+    @hangar = []
     @capacity = capacity
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,31 @@
+require 'plane'
+
+class Airport
+  attr_accessor :hangar
+
+  def initialize(capacity = 10)
+    plane = Plane.new
+    @hangar = [plane]
+    @capacity = capacity
+  end
+
+  def land
+    raise IndexError.new("the hangar is full") if full?
+    hangar.push(plane)
+  end
+
+  def takeoff
+    raise IndexError.new("the hangar is empty") if empty?
+    @hangar.pop(plane)
+  end
+
+  def full?
+    @hangar.length >= @capacity
+  end
+
+  def empty?
+    @hangar.length == 0
+  end
+
+end
+ 
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,2 @@
+class Plane
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,28 @@
+require 'airport'
+require 'plane'
+
+
+describe Airport do
+
+  it 'should respond to land plane' do
+
+    expect(subject).to respond_to(:land)
+  end
+
+  it 'will instruct the plane to takeoff and not be in the airport anymore' do
+    num_planes = subject.hangar.length
+    subject.takeoff
+    expect(subject.hangar.length).to eq(num_planes - 1)
+  end
+
+  it 'should not let a plane land if the hangar is full' do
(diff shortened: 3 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,31 @@
+require 'plane'
+
+class Airport
+  attr_accessor :hangar
+
+  def initialize(capacity = 10)
```
arhussain1, line 6: Good use of setting the default variable to a maximum of 10

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    plane = Plane.new
+    @hangar = [plane]
```
arhussain1: might be better to leave this as an empty hangar. As when a new Airport instance is created it makes sense for it to be empty.

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    @capacity = capacity
+  end
+
+  def land
+    raise IndexError.new("the hangar is full") if full?
```
arhussain1, line 13: Good use of a guard condition, It makes this line really easy to read.

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    hangar.push(plane)
```
arhussain1, line 14: This line may cause an error, as plane is an instance variable in your initialize method above. It may be better to refactor and make your land method create a new Plane instance called 'plane' see below:

```
def land 
plane = Plane.new
raise IndexError.new("the hangar is full") if full?
@hangar.push(plane)
end
```

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def takeoff
+    raise IndexError.new("the hangar is empty") if empty?
+    @hangar.pop(plane)
+  end
+
+  def full?
```
arhussain1, line 22: Really good decision to pull this code out into its own method. This makes it easier to read when its called in other methods like in your :land method

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+    @hangar.length >= @capacity
```
arhussain1, line 23: Could have used ``==`` instead of ``>=``

lib/plane.rb
```diff
@@ -0,0 +1,2 @@
+class Plane
```
arhussain1, line 1: Really great choice of class name, really describes what this object is. Maybe try to go back through the user stories and see what is the role of the Plane object. Remember the single responsibility principle, each class should have one responsibility and its should focus on doing it well

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  it 'should respond to land plane' do
+
+    expect(subject).to respond_to(:land)
+  end
+
+  it 'will instruct the plane to takeoff and not be in the airport anymore' do
+    num_planes = subject.hangar.length
+    subject.takeoff
+    expect(subject.hangar.length).to eq(num_planes - 1)
```
arhussain1, line 14: Try not to access the instance variables when doing your tests. Try to test the methods rather than the instance variables.

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  it 'should not let a plane land if the hangar is full' do
```
arhussain1, line 17: Good descriptions of your tests, it makes it easier to understand what your testing

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    subject.capacity.times{ subject.hanger.push(Plane.new) }
+    expect{ subject.land(Plane.new) }.to raise_error(IndexError)
+  end
+
+  it 'should not let a plane takeoff if the hangar is empty' do
+    subject.hangar = []
+    expect{ subject.takeoff }.to raise_error(IndexError)
```
arhussain1, line 24: good use of { } to test behaviour of your code

PR #2482, opened 2022-02-28
Adam's airport challenge

adamwoodcock98:
# Adam Woodcock

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,3 @@
+class Airport
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,10 +1,17 @@
 class Airport
-  def land(plane)
+  attr_reader :planes
+  
+  def initialize
+    @planes = []
+  end
 
+  def land(plane)
+    @planes << plane
   end
 
   def take_off(plane)
-
+    raise 'Airport is empty' if @planes.empty?
+    @planes.pop
   end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -6,6 +6,7 @@ class Airport
   end
 
   def land(plane)
+    raise "Airport is full" if @planes.count >= 1
     @planes << plane
   end
 
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,6 +1,10 @@
 class Airport
   def land(plane)
-    
+
+  end
+
+  def take_off(plane)
+
   end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,7 +1,9 @@
 class Airport
   attr_reader :planes
+  attr_reader :capacity
   
-  def initialize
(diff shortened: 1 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def land(plane)
+    raise "Airport is full" if @planes.count >= capacity
+    @planes << plane
+  end
+
+  def take_off(plane)
+    raise 'Airport is empty' if @planes.empty?
+    @planes.pop
```
ChrisPinna, line 25: This might not work if the plane that has to take off is not the last one on the array index

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  it 'should allow specific planes to take off' do
+    expect(subject).to respond_to(:take_off).with(1).argument
+  end
+
+  it 'should ensure a plane is no longer at the airport when taking off' do
+    subject.land(plane)
+    subject.take_off(plane)
+    # the plane should've landed and taken off, so trying to take off another plane should not work.
```
ChrisPinna, line 23: really cool that you used the functionality to show how this error would come up, instead of just calling the take_off method on its own.

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  it 'should have a default capacity' do
+    # assuming a default capacity of 20
+    expect(subject.capacity).to eq 20
+  end
+
+  it 'should have a default capacity that can be overwritten' do
+    airport = Airport.new(50)
+    airport.capacity.times { airport.land(plane) }
```
ChrisPinna, line 34: maybe you could use 50.times { airport.land(plane) }
instead of airport.capacity.times { airport.land(plane) }
just so you don't have to reference back to the code to understand that you are putting 50 planes in the airport. But I'm not too sure that I'm right because it is very readable this way.

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'airport.rb'
+
+describe Airport do
+  let(:plane) { double :plane }
```
ChrisPinna, line 4: love that you have taken the plane class out of the picture, and that you are only testing the airport class

spec/airport_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  it 'should ensure a plane is no longer at the airport when taking off' do
+    subject.land(plane)
+    subject.take_off(plane)
+    # the plane should've landed and taken off, so trying to take off another plane should not work.
+    expect { subject.take_off(plane) }.to raise_error 'Airport is empty'
+  end
+
+  it 'should have a default capacity' do
+    # assuming a default capacity of 20
```
ChrisPinna, line 28: nice clarifying comment

PR #2483, opened 2022-02-28
Christian Pinna's pull request 

ChrisPinna:
# Christian Pinna

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -3,6 +3,7 @@ require_relative './airport.rb'
 class Plane
 
   def land(airport)
+    raise 'airport full' if airport.hanger.count == airport.DEFAULT_CAPACITY
     airport.hanger << self
   end
 
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -4,6 +4,7 @@ class Plane
 
   def land(airport)
     raise 'airport full' if airport.hanger.count == airport.DEFAULT_CAPACITY
+    raise 'landing denied, there is a storm!' if airport.storm == true
     airport.hanger << self
   end
 
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,15 +1,26 @@
-require_relative './airport.rb'
+require_relative './airport'
 
 class Plane
 
+  def initialize
+    @current_airport = nil
+    @flying = true
+  end
+
   def land(airport)
+    raise 'error, plane is not flying' if @flying == false
     raise 'airport full' if airport.hanger.count == airport.DEFAULT_CAPACITY
     raise 'landing denied, there is a storm!' if airport.storm == true
+    @current_airport = airport
+    @flying = false
     airport.hanger << self
   end
 
   def take_off(airport)
+    raise 'plane is not in the specified airport' if airport != @current_airport
     raise 'take-off denied, there is a storm!' if airport.storm == true
-    airport.hanger.select! { |plane| plane != self } 
+    @flying = true
+    airport.hanger.reject! { |plane| plane == self }
+
   end
 end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -8,6 +8,7 @@ class Plane
   end
 
   def take_off(airport)
+    raise 'take-off denied, there is a storm!' if airport.storm == true
     airport.hanger.select! { |plane| plane != self } 
   end
 end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,12 @@
+require_relative './airport.rb'
+
+class Plane
+
+  def land(airport)
+    airport.hanger << self
+  end
+
+  def take_off(airport)
+    airport.hanger.select! { |plane| plane != self } 
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/plane.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class Plane
+
+  def initialize
+    @current_airport = nil
+    @flying = true
+  end
+
+  def land(airport)
+    raise 'error, plane is not flying' if @flying == false
```
adamwoodcock98, line 11: Nice job bro - managed to include some of those edge cases, well done!

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
+require_relative './airport'
+
+class Plane
+
+  def initialize
+    @current_airport = nil
+    @flying = true
+  end
+
+  def land(airport)
```
adamwoodcock98, line 10:
```suggestion
  def land_at(airport)
```
Using `land_at(airport)` would make this read like a full sentence

spec/airport_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+describe Airport do
+	it 'can creat a new airport' do
+			expect { Airport.new }.to_not raise_error
+	end
+
+	it 'can creat a new plane' do 
+			expect { Plane.new }.to_not raise_error
+	end
+
+	it 'plane object responds to land method' do
```
adamwoodcock98, line 12: Maybe create a `plane_spec.rb` file for these kinds of tests.

spec/airport_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+			10.times { airport.hanger << Plane.new }
+			plane = Plane.new
+			expect { plane.land(airport) }.to raise_error 'airport full'
+	end
+
+	it 'method storm? should return true if stormy' do
+			airport = Airport.new
+			airport.storm = true
+			expect(airport.storm).to eq(true)
+	end
```
adamwoodcock98, line 55: We need to avoid vacuous tests (from the rubic). In this one, with `airport.storm = true` you are setting the conditions yourself, so in the `expect(airport.storm).to eq(true)` section, regardless of your code implementation in the `airport.rb` class file, this test will be successful. I hope I'm explaining this okay. This is where mocks and doubles come in.

What you would be looking for here would be code to test the actual _behaviour_ of the class. something like:
```ruby
airport = Airport.new # in real practice, this would be a double, because your logic to stop a plane taking off is in the plane class file, something like:
airport = double("airport", :storm? => true)
expect { plane.take_off(airport) }.to raise_error 'take-off denied, there is a storm!'
```
Now you're using the behaviour of your code to test the state of the storm, because if there is a storm, your code should fail. I'll try and explain this more if you need me to.

spec/airport_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+			airport.storm = true
+			expect { plane.take_off(airport) }.to raise_error('take-off denied, there is a storm!')
+	end
+
+	it 'a plane should not be able to land if the weather at the airport is stormy' do	
+			airport = Airport.new
+			plane = Plane.new
+			airport.storm = true
+			expect { plane.land(airport) }.to raise_error('landing denied, there is a storm!')
+	end
```
adamwoodcock98, line 70: I've now seen that you've done what I put in that previous long comment, so that would change my advise to the fact that you could just delete that test, as through these two you're by default testing that the `storm` predicate works.

spec/airport_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+	it 'a plane should not be able to take-off from an airport it is not in' do
+		airport = Airport.new
+		plane = Plane.new
+		expect { plane.take_off(airport) }.to raise_error('plane is not in the specified airport')
+	end
```
adamwoodcock98, line 76: Nice!!!!

lib/plane.rb
```diff
@@ -0,0 +1,26 @@
...
+    @current_airport = airport
+    @flying = false
+    airport.hanger << self
+  end
+
+  def take_off(airport)
+    raise 'plane is not in the specified airport' if airport != @current_airport
+    raise 'take-off denied, there is a storm!' if airport.storm == true
+    @flying = true
+    airport.hanger.reject! { |plane| plane == self }
```
adamwoodcock98, line 23:
```suggestion
    airport.hanger.delete(self) # this would return the deleted plane
    airport.hanger -= [self] # this would return the array, without the deleted plane
```
I'm not 100% sure these would work with your code, I'm only suggesting because this line took me a little while to figure out, so just trying to maybe suggest something that improves readability - maybe you tried these suggestions and they didn't work, not trying to teach you to suck eggs!

adamwoodcock98 (commented):
I think it's great how you've covered most of the edge cases, I was no where near that point so I'm impressed. My biggest takeaway was that the tests could definitely use some separation and use of mocks, but I know from the peer group you want a better understanding of this, so that makes sense. There's some better instructions in steps 19 & 20 of the Boris Bikes challenge.
But really well done Chris!

PR #2484, opened 2022-02-28
Valentina Maggio's Airport challenge - Week 1

valentina-maggio:
# Your name

Valentina Maggio

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,11 +140,11 @@ Software building process
 * Created weather_spec and started testing Weather class
 * Tested that take_off method raises error when weather is stormy
 * Tested that land method raises error when weather is stormy
-* Hedge case #1: added method to check if plane landed
-* Hedge case #2: added error message for takeoff when plane already taken off
-* Hedge case #3: added error message for landing when plane already landed at airport
-* Hedge case #4: added error message for takeoff when plane not at the airport
-* Hedge case#5: added error message for landing when plane already landed at other airport
+* Edge case #1: added method to check if plane landed
+* Edge case #2: added error message for takeoff when plane already taken off
+* Edge case #3: added error message for landing when plane already landed at airport
+* Edge case #4: added error message for takeoff when plane not at the airport
+* Edge case #5: added error message for landing when plane already landed at other airport
 * Created feature test to simulate flow of two planes to and from two different airports
 
 Technologies used
--- a/README.md
+++ b/README.md
@@ -22,25 +22,82 @@ Planes can take off and land when the weather is not stormy and when the aiport 
 Instructions to run the program
 -------
 
-1. Fork this repo, and clone to your local machine.
-2. Install Ruby version 3.0.2:
+1. The program can be run from the Terminal/Command Line
+2. Ruby version 3.0.2 required. Installation details:
 ```shell
 rvm install "ruby-3.0.2"
 ```
-3. If you don't have bundler already, run the following commands:
+3. To run the program, use the following commands:
 ```shell
-gem install bundler
+$ irb
+3.0.2 :001 > require './lib/airport.rb'
+ => true 
+3.0.2 :002 > require './lib/plane.rb'
+ => true 
+3.0.2 :003 > require './lib/weather.rb'
+ => true 
 ```
-4. When the installation completes, run:
+4. Enjoy!
+
+Output example with two planes and two airports
+-------
 ```shell
-bundle
-```
-5. Install RuboCop for code review:
-```shell
-gem install rubocop
+| => irb
+3.0. :001 > airport1 = Airport.new
+ => #<Airport:0x00007fbb4fa7bb80 @capacity=100, @hangar=[], @weather=#<Weather:0x00007fbb4fa7bab8>> 
+3.0.2 :002 > plane1 = Plane.new
+ => #<Plane:0x00007fbb5094bf70 @grounded=false> 
+3.0.2 :003 > plane2 = Plane.new
+ => #<Plane:0x00007fbb4f96d4f0 @grounded=false> 
+3.0.2 :004 > airport1.land(plane1)
+/Users/Valentina/Projects/makers_projects/airport_challenge/lib/airport.rb:51:in `land_check_list': Warning: stormy weather! Landing not allowed. (RuntimeError)
+	from /Users/Valentina/Projects/makers_projects/airport_challenge/lib/airport.rb:18:in `land'
+	from (irb):8:in `<main>'
+	from /Users/Valentina/.rvm/rubies/ruby-3.0.2/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
+	from /Users/Valentina/.rvm/rubies/ruby-3.0.2/bin/irb:23:in `load'
+	from /Users/Valentina/.rvm/rubies/ruby-3.0.2/bin/irb:23:in `<main>'
+3.0.2 :005 > airport1.land(plane1)
+ => [#<Plane:0x00007fbb5094bf70 @grounded=true>] 
+3.0.2 :006 > airport1
+ => 
+#<Airport:0x00007fbb4fa7bb80
+ @capacity=100,
+ @hangar=[#<Plane:0x00007fbb5094bf70 @grounded=true>],
+ @weather=#<Weather:0x00007fbb4fa7bab8>> 
+3.0.2 :007 > airport1.land(plane2)
+ => [#<Plane:0x00007fbb5094bf70 @grounded=true>, #<Plane:0x00007fbb4f96d4f0 @grounded=true>] 
+3.0.2 :008 > airport1.take_off(plane1)
+ => #<Plane:0x00007fbb5094bf70 @grounded=false> 
+ 3.0.2 :009 > airport1.take_off(plane1)
+/Users/Valentina/Projects/makers_projects/airport_challenge/lib/airport.rb:55:in `take_off_check_list': This plane has already taken off. (RuntimeError)
+	from /Users/Valentina/Projects/makers_projects/airport_challenge/lib/airport.rb:24:in `take_off'
+	from (irb):21:in `<main>'
+	from /Users/Valentina/.rvm/rubies/ruby-3.0.2/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
+	from /Users/Valentina/.rvm/rubies/ruby-3.0.2/bin/irb:23:in `load'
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -13,23 +13,91 @@ Airport Challenge
...
+3.0.2 :002 > require './lib/plane.rb'
+ => true 
+3.0.2 :003 > require './lib/weather.rb'
+ => true 
+```
+4. Enjoy!
+
+Output example with two planes and two airports
+-------
+```shell
```
PKilgarriff, line 44: This use of irb inputs and outputs is really helpful - removes any worry the user gets from running the program. Is it copy-pasted from terminal or did you have to type something out?

README.md
```diff
@@ -13,23 +13,91 @@ Airport Challenge
...
+| => irb
+3.0. :001 > airport1 = Airport.new
+ => #<Airport:0x00007fbb4fa7bb80 @capacity=100, @hangar=[], @weather=#<Weather:0x00007fbb4fa7bab8>> 
+3.0.2 :002 > plane1 = Plane.new
+ => #<Plane:0x00007fbb5094bf70 @grounded=false> 
+3.0.2 :003 > plane2 = Plane.new
+ => #<Plane:0x00007fbb4f96d4f0 @grounded=false> 
+3.0.2 :004 > airport1.land(plane1)
+/Users/Valentina/Projects/makers_projects/airport_challenge/lib/airport.rb:51:in `land_check_list': Warning: stormy weather! Landing not allowed. (RuntimeError)
```
PKilgarriff, line 53: Good explanatory error message

README.md
```diff
@@ -60,30 +128,34 @@ To ensure safety
...
+* Created plane_spec and started testing Plane class
+* Feature tested land method for Airport class, then proceeded with testing implementation in airport_spec
+* Feature tested take_off method for Airport class, then proceeded with testing implementation in airport_spec
+* Added test in plane_spec to check if plane left the airport
+* Added exception: error message raised when airport is full (tested for one plane)
+* Added default capacity (100) for airport and introduced ability to override it
+* Created weather_spec and started testing Weather class
+* Tested that take_off method raises error when weather is stormy
+* Tested that land method raises error when weather is stormy
+* Edge case #1: added method to check if plane landed
```
PKilgarriff: Useful to see the edge cases you have considered laid out like this

README.md
```diff
@@ -60,30 +128,34 @@ To ensure safety
...
+* Edge case #2: added error message for takeoff when plane already taken off
+* Edge case #3: added error message for landing when plane already landed at airport
+* Edge case #4: added error message for takeoff when plane not at the airport
+* Edge case #5: added error message for landing when plane already landed at other airport
+* Created feature test to simulate flow of two planes to and from two different airports
+
+Technologies used
```
PKilgarriff, line 150: Will be stealing this section for future readmes
Helpful both for acknowledging goals achieved and for CV building

README.md
```diff
@@ -60,30 +128,34 @@ To ensure safety
...
+-----
+* Ruby
+* Simplecov for testing coverage (test coverage: 99.29%)
+* Rubocop for code style check
 
-Finally, don’t overcomplicate things. This task isn’t as hard as it may seem at first.
+Future implementations to consider
+-----
+At the moment if the user changes the airport capacity for a number that is lower than the actual number of planes already at the airport, this doesn't raise any error. This needs to be fixed.
```
PKilgarriff, line 158: Good to have next steps for yourself as well - shows you know the current blind spots

lib/airport.rb
```diff
@@ -0,0 +1,59 @@
...
+require_relative 'weather'
+
+class Airport
+
+  DEFAULT_CAPACITY = 100
+
+  attr_reader :hangar, :weather
+  attr_accessor :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY, weather = Weather.new)
```
PKilgarriff, line 11: passing the weather to the airport when initialised like this is very neat

lib/airport.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def at_airport(plane)
+    @hangar.include?(plane)
+  end
+
+  def left_airport?(plane)
+    plane.grounded == false
+  end
+  
+  def land_check_list(plane)
```
PKilgarriff, line 47: This is such a great idea - used encapsulation and extracted all the guard clauses into a private method, same as for the take_off

lib/plane.rb
```diff
@@ -0,0 +1,16 @@
+class Plane
```
PKilgarriff, line 1: This is a good example of a very clear class that has a Single Responsibility (as far as my understanding goes)

lib/weather.rb
```diff
@@ -0,0 +1,8 @@
+class Weather
+  
+  def stormy?
```
PKilgarriff, line 3: This is interesting - how random is it? Do you have a third of a chance for each weather condition?
There might be an inbuilt way of calling rand, but I might just not understand Kernel here.

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      it 'prevents landing when airport is full' do
+        allow(airport).to receive(:at_airport).and_return(false)
+        allow(plane).to receive(:grounded).and_return(false)
+        Airport::DEFAULT_CAPACITY.times { airport.land(plane) }
+        expect { airport.land(plane) }.to raise_error 'Airport capacity reached. No more landing allowed.'
+      end
+    end
+
+    context 'when bad weather conditions: stormy weather' do
```
PKilgarriff, line 39: As discussed, really appreciate the splitting into context parts - makes for a very legible spec file

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+
+      it 'prevents takeoff if plane already took off' do
+        allow(airport).to receive(:stormy_weather?).and_return(false)
+        allow(airport).to receive(:left_airport?).and_return(true)
+        expect { airport.take_off(plane) }.to raise_error 'This plane has already taken off.' 
+      end
+
+      it 'prevents takeoff if plane not at airport' do
+        allow(airport).to receive(:stormy_weather?).and_return(false)
```
PKilgarriff: Is this line redundant after calling it during the before block in lines 53-55?

spec/weather_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'weather'
```
PKilgarriff, line 1: Also a very neat spec file

PKilgarriff (commented):
Really nicely done - haven't reviewed the previous commits, but the end result is very clear, obviously has good test coverage, and your classes adhere to the DRY principle. The Plane class in particular demonstrates a very clear single responsibility - and I prefer your descriptive method names instead of just making ground an attr_reader and writer

spec/airport_spec.rb
valentina-maggio: Thanks for highlighting this Paul! I totally forgot to remove those lines after adding the 'context' block. I've amended the spec code.

lib/weather.rb
valentina-maggio: Yes it's a third of a chance for each condition. I probably made it too confusing using Kernel. I'll have a look at a different solution.

lib/airport.rb
leoht: Indeed that's a great use of dependency injection - which will become necessary (as we'll see in the mocking workshop) when it comes to mock the dependency on `Weather` :)

However, you might not need to expose it using the attr_reader `:weather`, since we don't need to access the airport's weather instance from outside `Airport` (it is only used internally by Airport)

leoht: (This last remark probably holds for the `:hangar` and `:capacity` accessors too, if you don't need to expose it to the outside)

lib/airport.rb
leoht: Great idea indeed! I'd just rename it to something a bit more explicit (communicating it'll raise if one of the conditions is not met) - maybe something like `raise_if_cannot_land`?

lib/airport.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  private
+
+  def full?
+    @hangar.length >= @capacity
+  end
+
+  def stormy_weather?
+    @weather.stormy? == :stormy
```
leoht, line 36: Since the name of the `Weather`'s public method is `stormy?`, it might make sense for it to return a boolean (rather than a symbol or string), so it makes it easier to understand how to interact with it (also, by convention, in Ruby method names ending with a question mark return booleans)

lib/plane.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  attr_reader :grounded
+
+  def initialize
+    @grounded = false
+  end
+
+  def landed
```
leoht, line 9: I'd rename this to `set_landed` or `mark_landed` to avoid confusion with `landed` being a getter of a value (e.g when reading `landed`, I would instinctively expect it returns a boolean without altering the state)

spec/airport_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+require 'plane'
+require 'weather'
+
+describe Airport do
+
+  subject(:airport) { described_class.new }
+  let(:plane) { Plane.new }
+
+  describe '#land' do
+    context 'when good weather conditions' do
```
leoht, line 11: Great use of `context` here to setup the context for stormy weather :) 

One thing I'd suggest to improve how you've stubbed the weather would be to stub the behaviour on `Weather` rather than on `Airport`'s method `stormy_weather?`):

```ruby
let(:weather) { Weather.new } // creating weather
let(:airport) { Airport.new(100, weather) } // making sure it's injected into Airport

before do
  // stubbing weather's public method
  allow(weather).to receive(:stormy?).and_return(false)
end

```

In terms of mocking, this is a slightly better approach, as we're letting Airport call the Weather, so instead of altering Airport's private methods behaviour, we're mocking the behaviour of Weather. This might seem a bit abstract so don't hesitate to ask questions about this - the workshop might help seeing better how it all works :)

lib/airport.rb
valentina-maggio: Thanks for the suggestion Leo. I wasn't really convinced about the best way to write this method.

lib/plane.rb
valentina-maggio: Makes sense! I will remember this difference next time :)

spec/airport_spec.rb
valentina-maggio: Yeah I'm still trying to practice my abstract thinking, but I get confused sometimes. I see what you mean. Very clear explanation. Thanks! Looking forward to tomorrow's workshop!

PR #2490, opened 2022-03-27
Jordan Manu Pull Request

JordanManu:
# Jordan Manu

# User stories 

- [X] User story 1: "I want to instruct a plane to land at an airport"
- [X] User story 2: "I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport"
- [X] User story 3: "I want to prevent landing when the airport is full"
- [X] User story 4: "I would like a default airport capacity that can be overridden as appropriate"
- [X] User story 5: "I want to prevent takeoff when weather is stormy"
- [X] User story 6: "I want to prevent landing when weather is stormy"

# README checklist

- [X] how to install,
- [X] how to run,
- [X] and how to test your code?


lib/weather.rb
```diff
@@ -0,0 +1,17 @@
+class Weather
+
+  POTENTIAL_CONDITIONS = [:sunny, :sunny, :sunny, :sunny, :sunny, :sunny, :sunny, :sunny, :stormy,
```
NBenzineb, line 3: Is there potentially a better way of generating random weather conditions without as much repetition

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  private
+
+  def in_airport?(plane)
+    @planes.include?(plane)
+  end
+
+  def check_capacity
+    fail "Airport is full" if full?
```
NBenzineb, line 34: *maybe* Can this be in one line rather than creating a new method

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
+require_relative '../lib/airport'
+require_relative '../lib/plane'
+require_relative '../lib/weather'
+
```
NBenzineb, line 4: Really like your tests- they're very clear and easy to understand

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
+require_relative 'weather'
+require_relative 'plane'
+
+class Airport
+
+  DEFAULT_CAPACITY = 20
+
+  attr_reader :planes, :capacity
```
eoinmakers, line 8: Do planes and capacity need to be surfaced outside of this class?

lib/airport.rb
eoinmakers: I agree with the above - I think it's okay to leave this line in the land method, as it's one of the tasks that makes up the sole responsibility of allowing a plane to land.

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
+    check_capacity
+    check_weather(Weather.new)
```
eoinmakers, line 17: As you're passing in a default argument to the check_weather method, you could leave out Weather.new as an argument here.

lib/airport.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def in_airport?(plane)
+    @planes.include?(plane)
+  end
+
+  def check_capacity
+    fail "Airport is full" if full?
+  end
+
+  def check_weather(weather = Weather.new)
```
eoinmakers, line 37: If you wanted to further isolate your tests and pass a mock in place of the real weather class, you could carry this out with dependency injection. It could look something like:

```
    @weather = weather 
    . . . . 
end 
```

And then you can continue to call @weather.stormy? in this method. The benefit of this is that you could pass in a mock in place of weather in your tests, but the real program will use Weather.new, the default argument.

spec/airport_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  context "planes landing at airport" do
+    before(:each) do
+      allow_any_instance_of(Weather).to receive(:current_conditions).and_return :sunny
+    end
+
+    it "lands a plane at an airport" do
+      airport.land(plane)
+      expect(airport.planes).to include(plane)
+    end
```
eoinmakers, line 18: An empty line underneath each test improves readability

eoinmakers (commented):
Really great implementation of the Airport Challenge, well done.

PR #2492, opened 2022-03-27
Conor Hope - Completed Airport challenge

Conor-Developer:
# Conor Hope

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,41 @@
+require_relative './weather'
+
+class Airport
+
+  include Weather
+
+  attr_reader :hangar
+
+  DEFAULT_CAPACITY = 5
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+  end
+
+  def store_plane(plane)
+    return raise "You cannot land during stormy weather" unless weather == "sunny"
+
+    @hangar.push(plane) if full? == false
+  end
+
+  def remove_plane_from_hangar(plane)
+    return raise "You cannot take off during stormy weather" unless weather == "sunny"
+
+    @hangar.each_index do |index|
+      if @hangar[index] == plane
+        @hangar.delete_at(index)
+        break
+      else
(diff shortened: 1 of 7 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+
+class Airport
+
+  include Weather
+
+  attr_reader :hangar
+
+  DEFAULT_CAPACITY = 5
+
+  def initialize(capacity = DEFAULT_CAPACITY)
```
rupertpople, line 11: Is the init input(capacity = DEFAULT CAPACITY) needed? Below could you just put @capacity = DEFAULT CAPACITY?

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    @hangar.push(plane) if full? == false
+  end
+
+  def remove_plane_from_hangar(plane)
+    return raise "You cannot take off during stormy weather" unless weather == "sunny"
+
+    @hangar.each_index do |index|
+      if @hangar[index] == plane
+        @hangar.delete_at(index)
```
rupertpople, line 30: As well as deleting the plane from the hanagr should it not be returned somehow? As the plane will then be in transit again rather than being erased?

lib/weather.rb
```diff
@@ -0,0 +1,9 @@
+module Weather
+
+  WEATHER = ["sunny", "stormy"]
+
```
rupertpople, line 4: The weather is mostly sunny so perhaps the code could be changed to have sunny more often then 50/50.

rupertpople (commented):
I have put 3 comments in about the code that perhaps could be changed or improved to improve the methods on the airport.rb and weather.rb files.

lib/airport.rb
Conor-Developer: Upon creation of the airport, it will allow the user to assign the capacity.

e.g. heathrow = Airport.new(10)  - This allows a capacity of 10 planes to be stored at the airport.

PR #2493, opened 2022-03-27
Airport Challenge

monenavarro:
# Your name

Monica Navarro

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/airplane.rb
+++ b/lib/airplane.rb
@@ -1,9 +1,13 @@
 require "airport"
 
 class Airplane
+  attr_accessor :landed
 
-  def initialize
-    @airplane = []
+  def initialize(status = true)
+    @landed = status
   end
 
+  def landed?
+    landed
+  end
 end
--- a/lib/airplane.rb
+++ b/lib/airplane.rb
@@ -1,0 +1,9 @@
+require "airport"
+
+class Airplane
+
+  def initialize
+    @airplane = []
+  end
+
+end
--- a/lib/airplane_spec.rb
+++ b/lib/airplane_spec.rb
@@ -1,0 +1,4 @@
+require "airplane"
+
+describe Airplane do
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,7 @@
+require "airplane"
+
+class Airport
+
+  def land_airplane
+  end
+end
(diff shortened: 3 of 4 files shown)


lib/airplane.rb
```diff
@@ -0,0 +1,13 @@
+require "airport"
+
+class Airplane
+  attr_accessor :landed
+
+  def initialize(status = true)
+    @landed = status
+  end
+
+  def landed?
```
Josenewmano, line 10: This whole class is written very clearly and succinctly

lib/airplane_spec.rb
```diff
@@ -0,0 +1,4 @@
+require "airplane"
+
+describe Airplane do
```
Josenewmano, line 3: We discussed already, but this file should be in the spec folder

spec/airport_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+describe Airport do
+  it "enables an airplane to land when instructed to" do 
+    expect(subject).to respond_to :land_airplane
+  end
+
+  it "can instruct a plane to take off from an airport" do
+    expect(subject). to respond_to :take_off_airplane 
+  end
+
+  it "confirms that an airplane has taken off" do
```
Josenewmano, line 12: All of these 'it' statements are clear

lib/airport.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def land_airplane
+    fail "ATTENTION! Airport full!" if airport.size == capacity
+    fail "Airplane landed at airport" if airplane.landed
+    airplane.landed = true
+    airplanes << airplane
+  end
+
+  def check_airplane_status(airplane)
```
Josenewmano, line 22: I think this method could be a little clearer

Josenewmano: I'm not sure how though, as my code doesn't include the functionality of preventing planes from landing at multiple airports

lib/airplane.rb
Josenewmano: I've just noticed something though: Does your code make airplanes landed by default? That is how I am reading it, but I may be incorrect. 

Which airport are they in? It would be strange to have an airplane built in the sky, but I suppose that is actually how I did it also

Josenewmano (commented):
Well done, it seems to me that you've created good functionality ro prevent duplicate take off and landing events, and your code is mostly clear, succinct, and following the DRY mantra

PR #2494, opened 2022-03-27
Coral airport_challenge

Coral-Ann:
# Coral Baker

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,3 @@
+class Airport
+  
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,10 +1,17 @@
+require_relative '../lib/plane.rb'
+
 class Airport
+  attr_reader :hangar
+
+  def initialize
+    @hangar = []
+  end
 
   def land(plane)
-
+    @hangar << plane
   end
 
   def take_off(plane)
-  
+    
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -2,14 +2,15 @@ require_relative '../lib/plane.rb'
 
 class Airport
   attr_reader :hangar, :capacity
+  DEFAULT_CAPACITY = 10
 
-  def initialize
+  def initialize(capacity = DEFAULT_CAPACITY)
     @hangar = []
-    @capacity = 2
+    @capacity = capacity
   end
 
   def land(plane)
-    fail 'Hangar full.' if @hangar.count >= @capacity
+    raise 'Hangar full.' if @hangar.count >= @capacity
     fail 'Plane already grounded.' if @hangar.include? plane
     @hangar << plane
   end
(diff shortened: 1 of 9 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+  def land(plane)
+    land_checks(plane)
+    @hangar << plane
+  end
+
+  def take_off(plane)
+    take_off_checks(plane)
+    @hangar.delete(plane)
+  end
+
```
jmcnally17, line 23: The use of private methods really helps with isolating parts of the code that you don't want the user to access.

spec/airport_spec.rb
```diff
@@ -0,0 +1,97 @@
+require_relative '../lib/airport.rb'
+
+describe Airport do
+  let(:plane) { double :plane }
+  let(:plane2) { double :plane2 }
+  let(:airport) { Airport.new(5) }
+
```
jmcnally17, line 7: Tests have been ordered well with landing tests grouped together first before take off tests

spec/airport_spec.rb
```diff
@@ -0,0 +1,97 @@
+require_relative '../lib/airport.rb'
+
+describe Airport do
+  let(:plane) { double :plane }
+  let(:plane2) { double :plane2 }
```
jmcnally17, line 5: Use single line syntax for setting doubles with their own methods and return values in each unit test to make code more DRY

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+class Airport
+  attr_reader :hangar, :capacity
+  DEFAULT_CAPACITY = 10
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @hangar = []
+    @capacity = capacity
+    @weather = Weather.new
+  end
+
```
jmcnally17, line 13: Methods were kept very short which makes them easier to understand

lib/airport.rb
jmcnally17: This also greatly simplifies the code that the user will be using

spec/airport_spec.rb
jmcnally17: Can also take that a step further by using a "before do" bock of code to set the doubles and method results for all subsequent tests

spec/weather_spec.rb
```diff
@@ -0,0 +1,12 @@
+require_relative '../lib/weather.rb'
+
+describe Weather do
+
+  describe '#storm?' do
+    it { is_expected.to respond_to :storm? }
+  
+    it 'returns true if theres a storm.' do
+      expect(subject.storm?).to be(true).or be(false)
+    end
```
jmcnally17, line 10: Use "srand" to remove the random number generation and set the weather to what you want in the test

PR #2498, opened 2022-03-28
David Campbell - Airport_Challenge - 3rd User Story - 4th Commit

david-ewan-campbell:
# Your name

David Ewan Campbell (Ramble-ramble-70)

# User stories 

- 

# README checklist

- [x] how to install,
- [x] how to run,

### DIFF
--- a/README_David_Campbell_Airport_Challenge.md
+++ b/README_David_Campbell_Airport_Challenge.md
@@ -12,10 +12,16 @@ Makers Course WEEK 2 - AIRPORT CHALLENGE
 # gem bundler installed/updated
 # checked rspec up-to-date
 
-user story 1
+User story 1 - 3 covered
     # As an air traffic controller 
     # So I can get passengers to a destination 
     # I want to instruct a plane to land at an airport
+
+    # So I can get passengers on the way to their destination 
+    # I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
+
+    # To ensure safety 
+    # I want to prevent landing when the airport is full 
 
 -Objectives, turn user stories into objects & actions
 
@@ -31,3 +37,9 @@ README revised on third commit to make process taken clearer and concentrated on
 # With each step I ran a feature test first in irb then an rspec unit test, making sure they failed first and read result of each failed test. I created wrapper examples in my airport_spec then created each definition in my Airport class accordingly, making sure each step was failing then passing and ensuring 100 % coverage
 
 # then ran rubocop and tightened up code accordingly - no offences detected and added and commited after each method definition added passed unit tests
+
+-4th commit before pull request - added airport_capacity to limit number of planes allowed to land in airport
+- renamed and moved README file to root folder
+- ran rubocop to refactor/tidy code layout(one offence not corrected yet - unused method argument - will correct next session)
+
+- rspec unit tests 3 examples passing with 100% coverage
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,3 @@
+class Airport
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,3 +1,5 @@
 class Airport
 
+  def land_plane
+  end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,5 +1,8 @@
 class Airport
 
-  def land_plane
+  def land_plane(plane)
+  end
+
+  def plane_take_off(plane)
   end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,6 +1,12 @@
 class Airport
+  def initialize(airport_capacity)
+    @airport_capacity = airport_capacity
+    @number_of_planes = 0
+  end
 
   def land_plane(plane)
+    raise 'No space at airport for plane to land' if @number_of_planes == @airport_capacity
+    @number_of_planes += 1
   end
 
   def plane_take_off(plane)
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,4 @@
+require 'airport'
+
+describe Airport do
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,4 +1,5 @@
 require 'airport'
 
 describe Airport do
+  it { is_expected.to respond_to(:land_plane) }
 end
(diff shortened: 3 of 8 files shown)


README_David_Campbell_Airport_Challenge.md
```diff
@@ -0,0 +1,45 @@
...
+-Objectives, turn user stories into objects & actions
+
+object => airport(therefore destination) => plane(object)in air, action: tell plane to land, take off, cannot land when airport full...
+
+object => plane => action: landed at airport
+
+- rspec spec/airport_spec.rb to run tests from airport_challenge folder
+calling Class defined in airport.rb file
+
+README revised on third commit to make process taken clearer and concentrated on working with airport spec and airport object/class (deleted created plane class & spec file at this stage as over-complicating process)
```
david-ewan-campbell, line 35: Should have created plane object? Put A Plane Class in a separate file and require that file from other files...

README_David_Campbell_Airport_Challenge.md
```diff
@@ -0,0 +1,45 @@
...
+    # So I can get passengers to a destination 
+    # I want to instruct a plane to land at an airport
+
+    # So I can get passengers on the way to their destination 
+    # I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport
+
+    # To ensure safety 
+    # I want to prevent landing when the airport is full 
+
+-Objectives, turn user stories into objects & actions
```
david-ewan-campbell, line 26: I made a basic description of adapting the user story process up to the point I'd reached but maybe should have described the whole task and broken it down more clearly to help with the TDD process? Also to show how you take a client request and turn it from a user story into specific code that handles that request?

spec/airport_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    expect(airport).to respond_to(:land_plane)
+  end
+
+  it 'can instruct plane to take off' do
+    expect(airport).to respond_to(:plane_take_off)
+  end
+  
+  it 'does not allow planes to land when at capacity' do
+    10.times do
+      airport.land_plane(:plane)
```
david-ewan-campbell, line 15: Created a plane 'test-double' to test code without having created a plane object? What are the benefits/advantages at the test stage with creating a 'double' at the early test stage compared with creating another plane object in this case?

lib/airport.rb
```diff
@@ -0,0 +1,14 @@
...
+    @number_of_planes = 0
+  end
+
+  def land_plane(plane)
+    raise 'No space at airport for plane to land' if @number_of_planes == @airport_capacity
+    @number_of_planes += 1
+  end
+
+  def plane_take_off(plane)
+  end
```
david-ewan-campbell, line 13: Had not got to stage of using plane_take_off definition to start making space in the airport and le other planes land/take off and keep track of the state of airport_capacity in the counter?

spec/airport_spec.rb
almorcrette: What I did - and it may not be best practice - is to use actual objects of other classes in the tests initially. Then I was going to replace them later with doubles (but I ran out of time to do that).

At least initially, when the code is simple, using actual instances of other classes helped me to build the code across multiple classes that worked.

Maybe when you are immediately diving into a complex program, or improving on an existing complex program it's better to go straight for doubles.

lib/airport.rb
```diff
@@ -0,0 +1,14 @@
+class Airport
+  def initialize(airport_capacity)
+    @airport_capacity = airport_capacity
+    @number_of_planes = 0
+  end
+
+  def land_plane(plane)
+    raise 'No space at airport for plane to land' if @number_of_planes == @airport_capacity
```
almorcrette, line 8: Would a `>=` be better here to protect against edge cases?

spec/airport_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'airport'
+
+describe Airport do
+  subject(:airport) { airport = Airport.new(10) }
+  it 'can instruct plane to land' do
```
almorcrette, line 5: I don't think this test is sufficient to test your production code.

Your production code adds one to `@number_of_planes` but your test only looks to see if `airport` responds to `land_plane`.

spec/airport_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    expect(airport).to respond_to(:land_plane)
+  end
+
+  it 'can instruct plane to take off' do
+    expect(airport).to respond_to(:plane_take_off)
```
almorcrette, line 10: Again, assuming you want to build `plane_take_off` to remove `1` from your `number_of_planes`, then your test needs to be stronger.

lib/airport.rb
david-ewan-campbell: Yes, that's really good advice! The sort of thing spec/rubocop not throwing up and need to look out for myself.

david-ewan-campbell: Thank you!

spec/airport_spec.rb
david-ewan-campbell: No you're absolutely right Alexis, I removed a plane object I'd created in a separate file but needed to be working with a plane object and have that respond...

spec/airport_spec.rb
david-ewan-campbell: Yes definitely that was my next step but ran out of time...

spec/airport_spec.rb
david-ewan-campbell: Yes, I can see the advantages of doubles but as the remit definitely should have involved the creation of a plane object (which I did initially) though as you say that can come later if you are building a complex program.

Yes it makes sense to build instances of classes that can start working across your code...

spec/airport_spec.rb
david-ewan-campbell: I hop leaving comments unresolved means this relates to the issue you've raised with my code where I need to be building on the next step and making that test stronger for that next step... great advice Alexis

spec/airport_spec.rb
david-ewan-campbell: It was passing the most basic tests at this stage and maybe I was focusing on that '100 %' coverage passing each time and was just trying to get to grips with the feature/unit processes up to this stage BUT definitely not a strong test as far as the rest of the challenge and needed a lot more work. Will be fascinating to work on the process together this afternoon as I really want to vastly improve my process but still feel like I'm just getting to some sort of understanding of the basics...

david-ewan-campbell (commented):
Intend on following up unresolved comments as they indicate very helpful advice on ways I could work on and improve the strength of my tests.

PR #2499, opened 2022-03-28
Airport

rupertpople:
# Your name

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


lib/airport.rb
```diff
@@ -0,0 +1,18 @@
...
+class Airport
+attr_accessor :hangar, :dock
+    def initialize
+        @hangar = []
+    end
+
+    def instruct_land(plane)
+        @dock = {}
+        key = plane
+        dock[key] = "grounded"
```
Conor-Developer, line 12: Have you thought about assigning the "grounded" string as an instance variable within the Plane class because it might make it easier to check and change the status of the plane (e.g. flying or grounded).

lib/airport.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'plane'
+
+class Airport
+attr_accessor :hangar, :dock
```
Conor-Developer, line 4: Have you thought about removing :dock as an attr_accessor so it cannot be checked by the user?

lib/airport.rb
```diff
@@ -0,0 +1,18 @@
...
+    def initialize
+        @hangar = []
+    end
+
+    def instruct_land(plane)
```
Conor-Developer, line 9: Clearly describes the code procedure within this method

Conor-Developer (commented):
I was able to run the code with ease and clearly understood how it works.

PR #2500, opened 2022-04-23
Tim's Airport Challenge

timbo-graham:
# Your name

Tim Graham

# User stories 

- [+] User story 1: "I want to instruct a plane to land at an airport"
- [+] User story 2: "I want to instruct a plane to take off from an airport and confirm that it is no longer in the airport"
- [+] User story 3: "I want to prevent landing when the airport is full"
- [+] User story 4: "I would like a default airport capacity that can be overridden as appropriate"
- [+] User story 5: "I want to prevent takeoff when weather is stormy"
- [+] User story 6: "I want to prevent landing when weather is stormy"

# README checklist

- [+] how to install,
- [+] how to run,
- [+] and how to test your code?


lib/airport.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  end
+
+  def random_weather
+
+    rand(10)
+
+  end
+
+  def can_take_off?
```
Rmorbey, line 21: I like how you made two separate methods for 'can_take_off?' and 'can_land?' you've made it easy to understand.

lib/airport.rb
```diff
@@ -0,0 +1,65 @@
+require 'airport'
+require 'plane'
+
```
Rmorbey, line 3: You could add a Constant variable for capacity to reduce use of magic numbers.

PR #2503, opened 2022-04-24
Airport challenge (Erlantz)

ErlantzR:
# Erlantz Ramos

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,13 @@
+class Airport
+
+
+  def land(plane)
+    fail 'Airtport full' unless @plane == nil
+    @plane = plane
+  end
+
+  def take_off(plane)
+    "#{plane} no longer in #{self} airport"
+  end
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,6 +1,8 @@
 class Airport
 
-  def initialize(capacity = 1)
+  DEFAULT_CAPACITY = 1
+
+  def initialize(capacity = DEFAULT_CAPACITY)
     @planes = []
     @capacity = capacity
   end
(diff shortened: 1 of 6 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,41 @@
+class Airport
+
+  DEFAULT_CAPACITY = 1
```
jimmy-lyons, line 3: Is this a constant? Could you just set your initialized capacity to 1?

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def land(plane)
+    fail 'Airport full' if full?
+    fail 'Plane cannot land on stormy weather' if weather == "stormy"
+    planes << plane
+  end
+
+  def take_off(plane)
+    fail 'Plane cannot take off on stormy weather' if weather == "stormy"
+    fail 'Plane not in this airport' unless plane_in_airport?(plane)
```
jimmy-lyons, line 21: This is good. I should have written something similar.

lib/airport.rb
```diff
@@ -0,0 +1,41 @@
...
+    planes.delete(plane)
+    "#{plane} no longer in #{self} airport"
+  end
+
+  def weather
+    return "stormy" if rand(20) == 1
+    "sunny"
+  end
```
jimmy-lyons, line 29: I wrote weather as a new class. Yours is more succinct but since the weather is effectively a different entity to airport and plane, I think it should be it's own class.

spec/airport_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  describe "#take_off" do
+
+    it 'take_off method confirms plane is not longer in the airport' do
+      plane = Plane.new
+      allow(subject).to receive(:weather) { "sunny" }
+      subject.land(plane)
+      expect(subject.take_off(plane)).to eq ("#{plane} no longer in #{subject} airport")
+    end
```
jimmy-lyons, line 25: This is very clear - I like matching to the message.

PR #2504, opened 2022-04-24
Russell's Airport Challenge

Rmorbey:
# Russell Morbey

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -12,12 +12,12 @@ class Airport
   end
 
   def land(plane)
-    # fail 'Airport full' if full?
+    fail 'Airport full' if full?
     fail "Flight cannot land due to stormy weather" if bad_weather?
     airport << plane
   end
 
-  def take_off(plane)
+  def take_off(*)
     fail "Flight cancelled due to bad weather" if bad_weather?
     @airport.pop
   end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,40 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 1
+
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
+    @airport = []
+    @capacity = capacity
+  end
+
+  def land(plane)
+    # fail 'Airport full' if full?
+    fail "Flight cannot land due to stormy weather" if bad_weather?
+    airport << plane
+  end
+
+  def take_off(plane)
+    fail "Flight cancelled due to bad weather" if bad_weather?
+    @airport.pop
+  end
+
+  def bad_weather?
+    weather.stormy?
+  end
+
+  def weather
+    @weather ||= Weather.new
+  end
+
+  private
+
+  attr_reader :airport
+
+  def full?
+    airport.count >= capacity
+  end
+end
--- a/lib/weather.rb
+++ b/lib/weather.rb
@@ -1,0 +1,13 @@
+
+class Weather
+
+  def stormy?
+    forecast >= 6
+  end
+
+  private
+  
+  def forecast
+    rand(10)
+  end
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,114 @@
+require 'airport'
+require 'weather'
+
+describe Airport do
+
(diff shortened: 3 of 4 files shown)


lib/airport.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'plane'
+require_relative 'weather'
+
+class Airport
+  DEFAULT_CAPACITY = 1
+
+  attr_reader :capacity
+
+  def initialize(capacity = DEFAULT_CAPACITY)
```
timbo-graham, line 9: I think you can just give whatever value you want default to be here, like `capacity = 1`, without needing to set the DEFAULT_CAPACITY variable. As with all of my comments, apologies if I'm entirely wrong....

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def land(plane)
+    fail 'Airport full' if full?
+    fail "Flight cannot land due to stormy weather" if bad_weather?
+    airport << plane
+  end
+
+  def take_off(*)
+    fail "Flight cancelled due to bad weather" if bad_weather?
+    @airport.pop
```
timbo-graham, line 22: I guess .pop could be tricky if we had multiple planes in the airport, as it'll always just pop the most recent entry to the airport array. FWIW, I used .delete(plane)

lib/airport.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def bad_weather?
+    weather.stormy?
+  end
+
+  def weather
+    @weather ||= Weather.new
```
timbo-graham, line 30: Oo, nice - ||= is new to me

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
+
+class Weather
```
timbo-graham, line 2: Oh, nice - I didn't break weather into a separate class, but definitely should have. Good compartmentalisation

lib/weather.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  def stormy?
+    forecast >= 6
+  end
+
+  private
```
timbo-graham, line 8: As an open question, how come you used a private method here? I don't really understand private methods, so would be great to know

spec/airport_spec.rb
```diff
@@ -0,0 +1,114 @@
+require 'airport'
+require 'weather'
+
+describe Airport do
+
+  describe '#land' do
+
+    it 'instructs a plane to land at an airport' do
+
+      # User story 1
```
timbo-graham, line 10: This is a nice way to make clear what you're testing. Didn't even occur to me

spec/airport_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+    end
+
+    it 'prevents a plane from landing when airport is full' do
+
+      # User story 3
+      # As an air traffic controller 
+      # To ensure safety 
+      # I want to prevent landing when the airport is full 
+
+      subject.capacity.times { subject.land Plane.new }
```
timbo-graham, line 29: Like the use of the capacity variable here. Pretty sure I hardcoded a value

spec/airport_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+
+    it 'instructs a plane to land at an airport' do
+
+      # User story 1
+      # As an air traffic controller 
+      # So I can get passengers to a destination 
+      # I want to instruct a plane to land at an airport
+
+      # Arrange
+      airport = Airport.new
```
timbo-graham, line 16: FWIW, you can use `subject` instead of declaring airport = Airport.new each time, when the `describe` for the test gives the class name

spec/airport_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+
+      # User story 6
+      # As an air traffic controller
+      # To ensure safety
+      # I want to prevent landing when weather is stormy 
+
+      # Arrange
+      airport = Airport.new
+      stormy_weather = Weather.new
+      plane = Plane.new
```
timbo-graham, line 43: FWIW, you can do let(:plane) { Plane.new } underneath `describe Airport do`, then you don't have to do `plane = Plane.new` inside each test

Rmorbey: Super helpful cheers.

PR #2508, opened 2022-04-24
Gawain Hewitt's Airport Weekend Challenge, week 1

gawainhewitt:
# Gawain Hewitt

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


lib/airport.rb
```diff
@@ -0,0 +1,17 @@
+class Airport
+  attr_writer :capacity
+  DEFAULT_CAPACITY = 20
```
eoinmakers, line 3: I'd leave an empty line between attr_readers, constant variables and the first method definition for readability

lib/airport.rb
```diff
@@ -0,0 +1,17 @@
...
+    @capacity = DEFAULT_CAPACITY
+  end
+
+  def plane_landed
+    fail "airport full" unless @hangar_log < @capacity
+    @hangar_log += 1
+  end
+
+  def plane_take_off
+    @hangar_log -= 1
```
eoinmakers, line 15: What would happen if this method was run when no planes had ever landed?

lib/plane.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def initialize
+    @landed = false
+    @weather = :fair
+  end
+
+  def land(airport)
+    fail "plane already landed" if @landed
+    fail "landing is prevented in stormy weather" if @weather == :stormy
+    @landed = airport
```
eoinmakers, line 12: 'landed' as a variable description doesn't necessarily match up with airport as a value (in terms of its naming), when landed is a boolean value

spec/airport_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+describe Airport do
+
+  let(:planes) { [] }
+  let(:extra_plane) { Plane.new }
+
+  it 'prevents landing when full' do
+    20.times do
+      plane = Plane.new
+      planes << plane
```
eoinmakers, line 11: You don't use the planes variable again after this - is it needed?

spec/feature_spec.rb
```diff
@@ -0,0 +1,38 @@
+describe 'Feature Test' do
+  let(:heathrow) { Airport.new }
+  let(:ibiza) { Airport.new }
+  let(:new_york) { Airport.new }
+
+  let(:eoins_private_jet) { Plane.new }
```
eoinmakers, line 6: Thank you very much 😎

eoinmakers (commented):
A good interpretation of Airport Challenge, well done. 

You successfully met the specification of each user story, bar Weather, with some good examples of unit testing. 

There was separation of concerns between Plane and Airport as objects.

With weather, it appears that the only circumstance in which the weather will be stormy is in your tests - can you think of a way of randomising the weather conditions, so that sometimes it will be stormy, and sometimes sunny?

PR #2509, opened 2022-04-24
Airport Challenge completed pull request

lukestorey95:
# Your name

Luke Storey

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


lib/airport.rb
```diff
@@ -0,0 +1,59 @@
...
+    land_plane(plane)
+  end
+
+  def take_off(plane)
+    check_plane_at_airport(plane)
+    check_weather_condition
+    take_off_plane(plane)
+  end
+
+  def include?(plane)
```
leoht, line 22: If your method `include?` doesn't have to be part of the public methods (i.e the caller of the class is not going to need these methods), you could set it as a private one

lib/plane.rb
```diff
@@ -0,0 +1,27 @@
+class Plane
+  def initialize
+    @landed = nil
```
leoht, line 3: It might be good to set a default value that makes sense for `@landed` - since it's a boolean, setting it as an initial value to `nil` might cause unexpected behaviour down the line (e.g what if we'd call `.landed?` when this attribute is still `nil`)

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
+describe Airport do
+  let(:airport) { Airport.new }
+  let(:plane) { instance_double('Plane', land: true, take_off: false) }
+  before(:each) { allow(airport).to receive(:stormy?).and_return(false) }
```
leoht, line 4: Rather than stubbing the method `stormy?` on `Airport` (which is private), a slightly better solution would be to stub it on a `Weather` instance double which would be injected into `Airport` - so you can replace the default implementation of the Weather with this double, and stub its behaviour

spec/airport_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      it { should respond_to(:new).with(0) }
+    end
+
+    context 'should allow capacity to be set' do
+      it { should respond_to(:new).with(1) }
+    end
+  end
+
+  def land_plane
+    airport.land(plane)
```
leoht, line 19: For these helper methods, I'd suggest moving them all to the same place (maybe the beginning of the file), and having them receive arguments rather than making use of the surrounding context (which is implicit), e.g:
```ruby
def land_plane(airport, plane)
  airport.land(plane)
end
```

spec/plane_spec.rb
```diff
@@ -0,0 +1,36 @@
+describe Plane do
+  let(:plane) { described_class.new }
+
+  describe '#land' do
+    context 'when flying' do
+      it 'should land' do
+        expect(plane).to receive(:land).and_return(true)
+        plane.land
```
leoht, line 8: This specific test doesn't really test for something, as we're asking to verify that the method `plane.land` will be called, just before calling it (so our test asserts something that it does by itself)

To assert that the method worked correctly, we could think of a way to verify the `landed` attribute has been set to `true` — however this would mean exposing the internal state of our plane instance, rather than testing for its behaviour

In the end, since the only behaviour from this class is to raise an error in case the plane is already "landed", maybe this is what we should assert in this test — that an error would be raise if we try to land the plane a second time

leoht: (which seems to be done in the test just below!)

leoht (commented):
Well this is really great work — it's true you've pushed the SRP to its extremes here, so one method is really doing one thing, but it's good for you to experiment with this. In real situations there would likely be tradeoffs and other things to consider, so it's good to have a think about what is usually to be prioritised when it comes to splitting code into smaller chunks. When it comes to edge cases, it's very hard (sometimes impossible) to cover everything in more complex projects, so again it's a matter of prioritising - here most cases seem covered. Re: context in specs, there are tools the library can provide for us to write more concise setup code and tests, such as shared context, shared examples, etc - but the key is often to name things explicitly, and call helper methods, etc, explicitly as well, rather than relying too much on implicit context

Let me know if you have any further questions about this feedback — keep up the good work!

PR #2510, opened 2022-04-24
Airport Challenge Stevie Spiegl

S-Spiegl:
# Your name

Steven Spiegl

# User stories 

- [ x ] User story 1: "I want to instruct a plane to land at an airport"

# README checklist

- [x] how to install,
- [x] how to run,


S-Spiegl:
Line 15 should read: land and fly those planes e.g. heathrow.land_plane(boeing_747)

lib/airport.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def land_plane(plane)
+
+    fail 'airport full' if @planes.length == @capacity
+    fail 'weather too stormy for landing' unless @weather != "stormy"
+    @planes << plane
+  end
+
+  attr_accessor :planes
```
SamButton12, line 26: I think that the ruby convention is to use attr_reader over attr_accessor, and that they go before the methods in the class

SamButton12 (commented):
Test comment

PR #2511, opened 2022-04-25
airport_challenge

SamButton12:
# Sam Button

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


docs/airport_challenge.rb
```diff
@@ -0,0 +1,23 @@
...
+    @planes << (plane)
+  end
+  
+  def takeoff(plane)
+    raise 'cannot take off, weather is stormy' if stormy?
+  end
+
+  def capacity_full?
+    max_capacity = 10
+    @planes.size >= max_capacity
```
S-Spiegl, line 20: Can't say without running it, but should this be @planes.size > max_capacity? I think this would make max_capacity 9

docs/airport_challenge.rb
```diff
@@ -0,0 +1,23 @@
+class Airport
+  attr_reader :plane
+
+  def initialize(capacity)
+    @capacity = capacity
+    @planes = []
+  end
+
+  def land(plane)
```
S-Spiegl, line 9: Good, simple names. I feel like I overcomplicated mine with land_plane and release_plane.

spec/airport_spec.rb
```diff
@@ -0,0 +1,35 @@
+require './docs/airport_challenge'
+require './docs/weather'
+require './docs/plane'
+
+describe Airport do
+  it 'creates instance of airport' do
```
S-Spiegl, line 6: the rspec describes creating an instance of an airport but the arrange you run from within the rspec would create 10 airports, I think?

S-Spiegl (commented):
comment

docs/airport_challenge.rb
```diff
@@ -0,0 +1,23 @@
+class Airport
+  attr_reader :plane
```
eoinmakers, line 2: Do you need to expose the plane variable outside of this class?

docs/airport_challenge.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def initialize(capacity)
+    @capacity = capacity
```
eoinmakers, line 5: How would you implement a default capacity, in case the user didn't specify one? Maybe take a look at including a default argument: https://www.rubyguides.com/2018/06/rubys-method-arguments/

docs/airport_challenge.rb
```diff
@@ -0,0 +1,23 @@
...
+    @planes = []
+  end
+
+  def land(plane)
+    raise 'Airport is full, cannot land plane' if capacity_full?
+    @planes << (plane)
+  end
+  
+  def takeoff(plane)
+    raise 'cannot take off, weather is stormy' if stormy?
```
eoinmakers, line 15: How can having a plane take off be reflected? 

You're currently storing landed planes in @planes - they need to be decremented from this array once they take off, to show they're no longer in the airport.

docs/airport_challenge.rb
eoinmakers: Additionally, you should use the capacity you passed in in the initialize method, stored in @capacity. Otherwise, it's a duplication of code, and max_capacity will remain static at 10 regardless of the user input.

docs/plane.rb
```diff
@@ -0,0 +1,5 @@
+class Plane
+  def at_airport?
+    false
+  end
```
eoinmakers, line 4: Unused methods should be removed before opening a PR.

spec/airport_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    airport = Airport.new(10)
+    expect(airport).to be_kind_of(Airport)
+  end
+  
+  # it 'allows plane to land' do
+  #   airport = Airport.new
+  #   expect(airport).to respond_to :land
+  # end
+  # not needed as it is fully covered by a future test
```
eoinmakers, line 15: You can remove unneeded tests / commented out code / tests too.

docs/airport_challenge.rb
eoinmakers: In order for Weather's stormy? method to be called here, you could refactor your class with the following snippets (as an example):

```
def initialize(capacity, weather)
     @capacity = capacity
     @weather = weather
def
```

With the above, you would initialize a Weather instance with Weather.new, and pass that in as an argument when initialising Airport.new.

Then in your land/take_off methods, you could write: 

`raise 'cannot take off, weather is stormy' if @weather.stormy?`

The last remaining step will be to refactor your Weather class itself - currently, the result will always be the same. Can you think of a way you might randomise the result, so that it's sometimes sunny and sometimes stormy?

eoinmakers (commented):
Good progress with this, but there are some pockets of logic that are incomplete. Check out the snippets and suggestions I shared, and let me know if you'd like to take a look together!

PR #2512, opened 2022-04-25
Kieran's Airport Challenge

K-Carty:
# Kieran Carty

# User stories 

- [ ] I didn't manage to complete anything to do with the Weather user stories

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,35 @@
+require_relative 'plane'
+
+class Airport  
+  def initialize
+    @plane_inventory = []
+    @capacity = 10
+  end
+  def land(plane)
+    fail "Airport at capacity" if @plane_inventory.length >= @capacity
+    @plane_inventory << plane
+      "#{plane} landed"
+  end
+
+  def take_off(plane)
+    fail 'No planes on the ground' if @plane_inventory.empty?
+    @plane_inventory << plane
+    "#{plane} plane(s) taken off and no longer in the airport"
+  end
+  def plane_inventory
+    @plane_inventory.count
+  end
+  attr_reader :capacity
+end
+
+
+
+# need another test to see if land_plane and take off correctly edit the plane inv log
+# if @plane_inventory += planes <= @capacity
+#   else "UNSAFE TO LAND - Airport at capcity"
+# if (@plane_inventory += planes) > 10
+# @plane_inventory = @capacity
+# puts "UNSAFE TO LAND - Airport at capcity"
+# end 
+
+# plane = Plane.new
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,30 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+  it { is_expected.to be_kind_of(Airport)}
+  it { is_expected.to respond_to(:land).with(1).argument}
+  it 'can land planes' do
+    airport = Airport.new
+    plane = Plane.new
+    expect(airport.land(plane)).to eq("#{plane} landed")
+  end
+  it { is_expected.to respond_to(:take_off).with(1).argument}
+  it 'can limit the number of planes that can land' do
+    airport = Airport.new
+    expect(airport.plane_inventory).to be <= airport.capacity
+  end
+  describe '#land' do 
+    it 'raises an landing error when full' do 
+    airport = Airport.new
+    plane = Plane.new
+    @plane_inventory = [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 , 13, 14, 15, 16, 17, 18, 19, 20, 21] 
+    expect {airport.land(plane)}.to raise_error "Airport at capacity"
+    end
+  end  
+  it 'raises an take off error when empty' do 
+    airport = Airport.new
+    expect {airport.take_off(Plane.new)}.to raise_error "No planes on the ground"
+  end 
+end
--- a/lib/plane.rb
+++ b/lib/plane.rb
@@ -1,0 +1,3 @@
+class Plane
+   
+end
--- a/spec/plane_spec.rb
+++ b/spec/plane_spec.rb
@@ -1,0 +1,5 @@
+require 'plane'
+
+describe Plane do
+it { is_expected.to be_kind_of(Plane)}
+end


lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+    @capacity = 10
+  end
+  def land(plane)
+    fail "Airport at capacity" if @plane_inventory.length >= @capacity
+    @plane_inventory << plane
+      "#{plane} landed"
+  end
+
+  def take_off(plane)
+    fail 'No planes on the ground' if @plane_inventory.empty?
```
archiemartini, line 15: Used SRP nice

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require_relative 'plane'
```
archiemartini, line 1: SO THIS is how you split up files, gg

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+class Airport  
+  def initialize
+    @plane_inventory = []
+    @capacity = 10
+  end
+  def land(plane)
+    fail "Airport at capacity" if @plane_inventory.length >= @capacity
```
archiemartini, line 9: consider using SRP ie
@plane_inventory.length >= @capacity
could equal just 'full?' if you define below

lib/airport.rb
archiemartini: i didn't know

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def take_off(plane)
+    fail 'No planes on the ground' if @plane_inventory.empty?
+    @plane_inventory << plane
+    "#{plane} plane(s) taken off and no longer in the airport"
+  end
+  def plane_inventory
+    @plane_inventory.count
+  end
+  attr_reader :capacity
```
archiemartini, line 22: nice use of attr

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'airport'
+require 'plane'
+
+describe Airport do
+
+  it { is_expected.to be_kind_of(Airport)}
```
archiemartini, line 6: good, you're clearly letting RSpec guide you first

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  it { is_expected.to respond_to(:land).with(1).argument}
+  it 'can land planes' do
+    airport = Airport.new
+    plane = Plane.new
+    expect(airport.land(plane)).to eq("#{plane} landed")
+  end
+  it { is_expected.to respond_to(:take_off).with(1).argument}
```
archiemartini, line 13: I think argument specs are super useful, nice

lib/airport.rb
archiemartini: improves readability

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  it 'can limit the number of planes that can land' do
+    airport = Airport.new
+    expect(airport.plane_inventory).to be <= airport.capacity
+  end
+  describe '#land' do
```
archiemartini, line 18: clear outline of every function keeps things clarified, good practice

archiemartini: ie "decribe '#land' do"

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
+require_relative 'plane'
+
+class Airport  
+  def initialize
+    @plane_inventory = []
+    @capacity = 10
+  end
```
eoinmakers, line 7: It's good practice to leave an empty line underneath each method.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+  def land(plane)
+    fail "Airport at capacity" if @plane_inventory.length >= @capacity
+    @plane_inventory << plane
+      "#{plane} landed"
```
eoinmakers, line 11: You don't need the String included here, as we can assume the plane has indeed landed if no error has been raised. This is maybe more important later on when we start working towards specifications / practice tech tests, but for now it can just help to try to keep your implementation as simple as possible.

lib/airport.rb
eoinmakers: Do you need the capacity variable to be visible outside of this class? If not, and if this is only so you can test it, it's worth removing it - testing the capacity variable will come indirectly by testing its behaviour, ie. if you land 20 planes, landing a 21st will result in an error.

lib/airport.rb
```diff
@@ -0,0 +1,35 @@
...
+
+# need another test to see if land_plane and take off correctly edit the plane inv log
+# if @plane_inventory += planes <= @capacity
+#   else "UNSAFE TO LAND - Airport at capcity"
+# if (@plane_inventory += planes) > 10
+# @plane_inventory = @capacity
+# puts "UNSAFE TO LAND - Airport at capcity"
+# end 
+
+# plane = Plane.new
```
eoinmakers, line 35: It's good practice to remove any commented out code / notes before opening a PR - you can work any work in progress as such if you need to

spec/airport_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  it 'can limit the number of planes that can land' do
+    airport = Airport.new
+    expect(airport.plane_inventory).to be <= airport.capacity
+  end
+  describe '#land' do 
+    it 'raises an landing error when full' do 
+    airport = Airport.new
+    plane = Plane.new
+    @plane_inventory = [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 , 13, 14, 15, 16, 17, 18, 19, 20, 21] 
+    expect {airport.land(plane)}.to raise_error "Airport at capacity"
```
eoinmakers, line 23: Instead of altering the Airport object's state directly - the @plane_inventory variable, we could remove that direct access to better protect the internal state of the object, and you could land a plane 20 times manually.

Check out the .times method here: 
https://www.rubyguides.com/ruby-tutorial/loops/

For this test, you can land planes 20 times, and then keep your same expectation to check that the error is raised.

eoinmakers (commented):
Great work! You implemented the majority of the user stories, meeting the learning objectives of the week. 

Your implementation is concise, with a range of unit tests.

PR #2513, opened 2022-04-25
Alexandros Papagiannis.

Alexandros91:
# Your name

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '3.0.2'
+ruby '3.0.0'
 
 group :test do
   gem 'rspec'
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,0 +1,3 @@
+class Airport
+
+end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,3 +1,6 @@
 class Airport
 
+  def land
+  
+  end
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,6 +1,7 @@
 class Airport
 
   def land
-  
+    
   end
+
 end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,0 +1,4 @@
+require 'airport'
+describe Airport do
+
+end
--- a/spec/airport_spec.rb
+++ b/spec/airport_spec.rb
@@ -1,4 +1,11 @@
 require 'airport'
 describe Airport do
+  
+  describe '#land' do
+    it 'expects a new instance of the class to respond to the land method' do
+      expect { subject.to respond_to(:land)}
+    end
+  end
+
 
 end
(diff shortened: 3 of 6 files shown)


Alexandros91:
Alexandros Papagiannis.

Diagram for the challenge

As an air traffic controller 

As an air traffic controller 

As an air traffic controller 
To ensure safety 

As the system designer

As an air traffic controller 
To ensure safety 

As an air traffic controller 
To ensure safety 


Air traffic controller           get(to a destination)                              Air traffic controller    	get to destination.
Passengers                                                                 				Passengers			instruct
Destination                         instruct							Destination			land
Plane                                   land								Plane				take off
Airport													Airport				confirm (is no lon-
														weather                        ger at the airport)
												            system designer              ensure (safety)
													      software			      prevent landing
Air traffic controller             get (to a destination)				capacity			       prevent takeoff 
Passengers																      being used from 
Plane                                      take off											      differ. airports
Airport                                   confirm(is no longer								     can be overridden
                                                  at the airport)



Air traffic controller             ensure safety
                                                 prevent
Landing
When airport full


Air traffic controller               ensure safety
                                                  Prevent
Takeoff
When weather stormy

Air traffic controller               ensure safety
                                                  Prevent
landing
When weather stormy


System designer
Software                                 used for differ airports
Default capacity                    can be overridden

Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '3.0.2'
+ruby '3.0.0'
```
simon-2357, line 3: In general I think the best way to change this is using RVM from the command line to install the version of Ruby locally, rather than changing the Gemfile, in case there are other dependencies in the programme

spec/airport_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'airport'
+describe Airport do
+  
+  describe '#land' do
+    it 'expects to return a plane' do
+      expect(subject.land).to eq plane
```
simon-2357, line 6: Nice use of the implied subject in the test!

simon-2357, line 6: This test shouldn't check if an airport returns a plane when it's landed, but to check if it can land a plane. You can use a responds_to matcher to test this.

spec/airport_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'airport'
+describe Airport do
```
simon-2357, line 2: How could you write a test for a second user story? E.g. Airport won't allow a plane to land when it's full

spec/plane_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'plane'
+describe Plane do
+
```
simon-2357, line 3: If there are no tests to run, then no need to have a spec file here :)

lib/airport.rb
```diff
@@ -0,0 +1,7 @@
+class Airport
+
+  def land
```
simon-2357, line 3: What lands at the airport? I think this method needs an argument to tell the airport which plane is landing there.

PR #2515, opened 2022-04-25
Airport Challenge Complete

jimmy-lyons:
# Your name

Jimmy Lyons

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  def def_capacity(num)
+    @capacity = num
+  end
+
+  def check_weather
+    @current_weather = Weather.new.weather
+  end
+
+  def check_for_storm
+    fail unless @current_weather != 'stormy'
```
ErlantzR: What do you think about changing this to `fail if @current_weather == 'stormy'` ? It might be easy to read when this is raising the error

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  # make sure that planes can't take off when planes_at_airport == 0
+  def take_off(plane)
+    check_for_storm
+    @planes_in_airport.delete(plane)
+    plane
+  end
+
+  def def_capacity(num)
```
ErlantzR, line 26: This method might be redundant with the `attr_accessor :capacity` above

lib/airport.rb
```diff
@@ -0,0 +1,38 @@
+class Airport
```
ErlantzR, line 1: I think methods explain quite clearly what are they doing. I don't have problems to understand how the code works

spec/airport_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'airport'
```
ErlantzR, line 1: Tests seem clear and test appropriate features

PR #2519, opened 2022-04-25
Tom O'Neill airport challenge

tomoneill32:
# Tom O'Neill

# User stories 

# README checklist

### DIFF
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,18 +1,27 @@
 require_relative 'plane'
 
 class Airport
-  attr_reader :stored_planes
+  attr_reader :stored_planes, :capacity
+  CAPACITY = 5
 
-  def initialize
+  def initialize 
     @stored_planes = []
+    @capacity = CAPACITY
   end
 
   def land(plane)
+    fail 'Airport full!' if full?
     @stored_planes << plane
   end
 
   def depart(plane)
     @stored_planes.delete(plane)
+  end
+
+  private
+
+  def full?
+    @stored_planes.length >= @capacity
   end
 
 end
--- a/lib/airport.rb
+++ b/lib/airport.rb
@@ -1,25 +1,34 @@
 require_relative 'plane'
 
 class Airport
-  attr_reader :stored_planes
+  attr_reader :stored_planes, :weather
   CAPACITY = 5
 
   def initialize 
     @stored_planes = []
     @capacity = CAPACITY
+    @weather = self.set_weather
   end
 
   def land(plane)
     fail 'Airport full!' if full?
+    fail 'Cannot land - stormy weather!' if @weather == 'Stormy'
     @stored_planes << plane
   end
 
   def depart(plane)
+    fail 'No planes' if empty?
+    fail 'Plane not in airport' if !@stored_planes.include?(plane)
     @stored_planes.delete(plane)
   end
 
   def adjust_capacity(number)
     @capacity = number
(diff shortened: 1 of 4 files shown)


spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+describe Airport do
+  
+  describe '#land' do
+    it 'Can do so' do
+      plane = Plane.new
+      expect(subject.land(plane)).to eq([plane])
+    end
+
+    it 'Will not do so if the airport is full' do
+      Airport::CAPACITY.times { subject.land(Plane.new) }
```
lukestorey95, line 12: Good use of constant instead of magic number

lukestorey95, line 12: Good use of constant instead of magic number

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      expect { subject.land(Plane.new) }.to raise_error 'Airport full!'
+    end
+
+  end
+  
+  describe '#depart' do
+    it 'Can instruct a stored plane to take off from the airport' do
+      plane = Plane.new
```
lukestorey95, line 20: Since you are initialising a plane multiples times in the spec, you could consider using let to define a plane variable and then refer to that

`let(:plane) { Plane.new }`

spec/airport_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      subject.land(plane)
+      subject.depart(plane)
+      expect(subject.stored_planes).to eq []
+    end
+
+    it 'Will not do so if there are no planes in the airport' do
```
lukestorey95, line 26: Could reword the it statement to be less wordy and easier to read, perhaps by using should ( e.g. 'should not depart if there are no planes') or consider a context block to set the stage for multiple tests where the airport is empty.

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'plane'
```
lukestorey95, line 1: Believe if you require the files in the spec_helper you don't need to require them on each individual file as well

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def initialize 
+    @stored_planes = []
+    @capacity = CAPACITY
+    @weather = self.set_weather
+  end
+
+  def land(plane)
+    fail 'Airport full!' if full?
+    fail 'Cannot land - stormy weather!' if @weather == 'Stormy'
```
lukestorey95, line 15: Consider refactoring the condition into it's own method (SRP)

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+  private
+
+  attr_accessor :capacity
+
+  def full?
+    @stored_planes.length >= @capacity
+  end
+
+  def empty?
+    @stored_planes == []
```
lukestorey95, line 43: You could use .empty? which might be cleaner

lib/airport.rb
```diff
@@ -0,0 +1,46 @@
...
+    @stored_planes.delete(plane)
+  end
+
+  def adjust_capacity(number)
+    @capacity = number
+  end
+
+  def set_weather
+    die = rand(100)
+    die > 100 ? 'Stormy' : 'Sunny'
```
lukestorey95, line 31: Consider if there is a way to write this without the ternary

